PR #9, opened 2022-03-21
Add proto definitions for foxglove messages

defunctzombie:
foxglove.CompressedImage
foxglove.RawImage
foxglove.Log
foxglove.LocationFix

### DIFF
--- a/proto/foxglove/CompressedImage.proto
+++ b/proto/foxglove/CompressedImage.proto
@@ -1,0 +1,11 @@
+// https://foxglove.dev/docs/studio/messages/compressed-image
+
+syntax = "proto3";
+
+package foxglove;
+
+message CompressedImage {
+  uint64 timestamp = 1;
+  string format = 2;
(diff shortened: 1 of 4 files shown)


jtbandes:
If these files are hand-written can you update the glob in `.gitattributes` so github doesn't mark them as generated?

defunctzombie:
> If these files are hand-written can you update the glob in `.gitattributes` so github doesn't mark them as generated?

👍 

They are hand-written. I'm thinking thats sufficient for now? Maybe we can generate the json schemas from these hand-written protos.

proto/foxglove/CompressedImage.proto
```diff
@@ -0,0 +1,11 @@
+// https://foxglove.dev/docs/studio/messages/compressed-image
```
jtbandes, line 1: It would probably be ideal to have the full docs in here as comments, so the message files are really self-contained

PR #17, opened 2022-05-06
Add new SceneEntity and Primitive types

jtbandes:
**Public-Facing Changes**
Added types for drawing primitives in 3D.

**Description**
See https://github.com/foxglove/schemas/tree/jacob/add-markers/schemas#sceneentityupdate
Closes https://github.com/foxglove/schemas/issues/50


amacneil:
Re-sharing previous feedback that I don't like our inconsistent use of Marker vs Annotation. We should change one or the other to match.

I'm also not convinced we need either suffix. `foxglove.Cube` and `foxglove.Text` seem sufficiently unique to me?

src/schemas.ts
```diff
@@ -160,6 +160,522 @@ const foxglove_Pose: FoxgloveMessageSchema = {
...
+      description: "Embedded model. Either `url` or `mime_type` and `data` should be provided.",
+    },
+  ],
+};
+
+const foxglove_Markers: FoxgloveMessageSchema = {
+  type: "message",
+  name: "Markers",
+  description: "A list that can contain any number of any type of marker",
+  fields: [
```
amacneil, line 621: Why don't we remove `ImageAnnotations` and make this new `Markers` list support all markers?

defunctzombie: > Why don't we remove `ImageAnnotations` and make this new `Markers` list support all markers? I don't think it really matters whether the topic contains 2D or 3D markers or both?

I think it does. The ImageAnnotations unit values are in pixels and they are designed around annotating an image. They don't typically stand alone without an image under them. "Markers" are designed around scene building and the units are in scene space (world coordinates). Markers make sense on their own since they are the building blocks of a scene even without sensor data under them.

defunctzombie:
We should support this in the app prior to adding the documentation to the website. If we want to add to the message-schemas repo that seems sensible.

I think having a Marker suffix is useful (even tho I don't like Marker as a term I can't think of a better one). What separates `foxglove.CubeMarker` from `foxglove.Cube` is that CubeMarker will have fields to select color, and change properties that matter for rendering. If you only wanted a Cube in 3d space then you wouldn't need such fields on your message.

At their core - the "Marker" messages are for scene building. They are basically scene objects or renderables but beyond calling them "Marker" I've not better suggestion.

schemas/jsonschema/ArrowMarker.json
```diff
@@ -0,0 +1,167 @@
+{
+  "$comment": "Generated from ArrowMarker by @foxglove/message-schemas",
+  "title": "ArrowMarker",
+  "description": "A marker representing an arrow",
```
esthersweon, line 4: Add mention of "in 2D or 3D scene" for all marker messages?

schemas/jsonschema/KeyValuePair.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$comment": "Generated from KeyValuePair by @foxglove/message-schemas",
+  "title": "KeyValuePair",
+  "description": "An entry representing a key and its associated value",
```
esthersweon, line 4:
```suggestion
  "description": "A key with its associated value",
```

schemas/jsonschema/LineMarker.json
```diff
@@ -0,0 +1,240 @@
+{
+  "$comment": "Generated from LineMarker by @foxglove/message-schemas",
+  "title": "LineMarker",
+  "description": "A marker representing a series of points connected by lines",
```
esthersweon, line 4:
```suggestion
  "description": "A marker representing a series of connected points",
```

schemas/jsonschema/MarkerDeletion.json
```diff
@@ -0,0 +1,50 @@
+{
+  "$comment": "Generated from MarkerDeletion by @foxglove/message-schemas",
+  "title": "MarkerDeletion",
+  "description": "Deletion command to remove previously published markers",
```
esthersweon, line 4:
```suggestion
  "description": "Command to remove previously published markers",
```

schemas/jsonschema/Markers.json
```diff
@@ -0,0 +1,1593 @@
+{
+  "$comment": "Generated from Markers by @foxglove/message-schemas",
+  "title": "Markers",
+  "description": "A list that can contain any number of any type of marker",
```
esthersweon, line 4:
```suggestion
  "description": "A list of any number or type of markers",
```

esthersweon (commented):
Left comments on how to elaborate on certain message descriptions, as these will be used in docs.

jhurliman:
> I imagine this is going to require some significant changes to support triangle strip & triangle fan, but that can come later

My opinion is that we shouldn’t support these. 3js dropped support for them a while ago and just provides util methods to convert to triangle lists. There would be some bandwidth savings (not to the GPU, just disk or network) but compression would eliminate a lot of that savings.

schemas/ros1/TrianglesMarker.msg
```diff
@@ -0,0 +1,49 @@
...
+bool frame_locked
+
+# Additional user-provided metadata associated with the marker. Keys must be unique.
+foxglove_msgs/KeyValuePair[] metadata
+
+# 0-1-2, 3-4-5, ...
+uint8 TRIANGLE_LIST=0
+
+# 0-1-2, 1-2-3, 2-3-4, ...
+uint8 TRIANGLE_STRIP=1
```
jhurliman, line 28: -1

schemas/ros1/TrianglesMarker.msg
```diff
@@ -0,0 +1,49 @@
...
+
+# 0-1-2, 0-2-3, 0-3-4, ...
+uint8 TRIANGLE_FAN=2
```
jhurliman, line 31: -1

schemas/ros1/TrianglesMarker.msg
```diff
@@ -0,0 +1,49 @@
...
+
+# Drawing primitive to use for triangles
+uint8 type
```
jhurliman, line 34: Remove, always assume TRIANGLE_LIST

schemas/ros1/TrianglesMarker.msg
```diff
@@ -0,0 +1,49 @@
...
+# Vertices to use for triangles
+geometry_msgs/Point[] points
+
+# Solid color to use for the whole shape. One of `color` or `colors` must be provided.
+foxglove_msgs/Color color
+
+# Per-vertex colors (if specified, must have the same length as `points`). One of `color` or `colors` should be provided.
+foxglove_msgs/Color[] colors
+
+# Indexes into the `points` and `colors` attribute arrays, which can be used to avoid duplicating attribute data.
```
jhurliman, line 48: `# Indexes into the `points` and `colors` attribute arrays, which can be used to avoid duplicating attribute data. Leave empty for unindexed triangles.`

jhurliman (changes requested):

schemas/ros1/TrianglesMarker.msg
jtbandes: This comment is not helpful. Can you please provide some explanation or reasoning so that an intelligent response can be made? I think we previously discussed this briefly but I don't remember the reasoning for removing this.

jhurliman: https://github.com/foxglove/schemas/pull/17#issuecomment-1132255935

schemas/ros1/TrianglesMarker.msg
jtbandes: Adding 

> If omitted or empty, indexing will not be used. This default behavior is equivalent to specifying [0, 1, ..., N-1] for the indices (where N is the number of `points` provided).

schemas/jsonschema/LineMarker.json
jtbandes: Took the other suggestions, but I don't like this one as much

schemas/README.md
```diff
@@ -1593,6 +2172,548 @@ Raw image data
...
+</td>
+<td>
+
+Numeric identifier which must match if `kind` is `MATCHING_ID`.
+
+</td>
+</tr>
+</table>
+
+## SceneEntityUpdate
```
amacneil, line 2420: It feels like this schema should be called `SceneEntities`, in keeping with all our other array type schemas. 

I know it technically contains SceneEntityDeletion as well, but that is still a type of SceneEntity.

schemas/README.md
```diff
@@ -992,6 +1455,122 @@ Line number in the file
...
+
+</td>
+<td>
+
+Solid color to use for the whole model. If `use_embedded_materials` is true, this color is blended on top of the embedded material color.
+
+</td>
+</tr>
+<tr>
+<td><code>use_embedded_materials</code></td>
```
amacneil, line 1508:
```suggestion
<td><code>embedded_materials</code></td>
```

it's already a boolean

schemas/README.md
```diff
@@ -453,6 +593,160 @@ Image format
...
+
+</td>
+<td>
+
+Color of the arrow
+
+</td>
+</tr>
+<tr>
+<td><code>metadata</code></td>
```
amacneil, line 646: I don't really see the point of all this metadata in the primitives, nor do I understand how we would even show it to the user since I assume clicking a SceneEntity would select that entire SceneEntity

schemas/README.md
```diff
@@ -1593,6 +2172,548 @@ Raw image data
...
+
+Timestamp of the deletion. Only matching entities earlier than this timestamp will be deleted.
+
+</td>
+</tr>
+<tr>
+<td><code>type</code></td>
+<td>
+
+[enum SceneEntityDeletionType](#enum-sceneentitydeletiontype)
```
amacneil, line 2396: This could be a simple `delete_all` bool. Doesn't save many bytes, but it seems simpler to me when reading the docs, rather than clicking through into a whole enum with only two values.

schemas/README.md
```diff
@@ -1593,6 +2172,548 @@ Raw image data
...
+
+</td>
+<td>
+
+Type of deletion action to perform
+
+</td>
+</tr>
+<tr>
+<td><code>id</code></td>
```
amacneil, line 2406: Did you consider accepting an array of `ids` to delete? This would also make it simpler to publish `SceneEntityDeletion` directly on a topic without needing to wrap it in `SceneEntityUpdate`.

schemas/README.md
```diff
@@ -94,6 +120,120 @@ name | value | description
...
+
+</td>
+<td>
+
+Position of the arrow's tail and orientation of the arrow. Identity orientation means the arrow points in the +x direction.
+
+</td>
+</tr>
+<tr>
+<td><code>length</code></td>
```
amacneil, line 158: why not `shaft_length`, to match `shaft_diameter`, `head_length`, `head_diameter`?

schemas/README.md
```diff
@@ -825,6 +1157,137 @@ Intensity of detections
...
+<tr>
+<td><code>colors</code></td>
+<td>
+
+[Color](#color)[]
+
+</td>
+<td>
+
+Per-point colors (if specified, must have the same length as `points`). One of `color` or `colors` must be provided.
```
amacneil, line 1257: Seems like this complexity could be removed. If people want per-segment colors, they can already do that by adding multiple Lines to a SceneEntity.

schemas/README.md
```diff
@@ -825,6 +1157,137 @@ Intensity of detections
...
+<tr>
+<td><code>indices</code></td>
+<td>
+
+uint32[]
+
+</td>
+<td>
+
+Indices into the `points` and `colors` attribute arrays, which can be used to avoid duplicating attribute data.
```
amacneil, line 1270: ELI5?

schemas/README.md
```diff
@@ -992,6 +1455,122 @@ Line number in the file
...
+<tr>
+<td><code>mime_type</code></td>
+<td>
+
+string
+
+</td>
+<td>
+
+MIME type of embedded model (e.g. `model/gltf-binary`). Either `url` or `mime_type` and `data` should be provided.
```
amacneil, line 1542: We should document: If I set `mime_type` with `url`, would it override the http response `content-type` header? Or will we always ignore this field with `url`?

The former option might be useful in certain situations (if you can't control the http server).

schemas/README.md
```diff
@@ -1593,6 +2172,548 @@ Raw image data
...
+<tr>
+<td><code>colors</code></td>
+<td>
+
+[Color](#color)[]
+
+</td>
+<td>
+
+Per-vertex colors (if specified, must have the same length as `points`). One of `color` or `colors` must be provided.
```
amacneil, line 2683: Same comment as on LinePrimitive - seems like we could remove this complexity.

schemas/README.md
```diff
@@ -29,11 +37,29 @@ Generated by https://github.com/foxglove/schemas
...
 
+## enum LineType
+
+An enumeration indicating how input points should be interpreted to create lines
+
+name | value | description
+---- | ----- | -----------
+`LINE_STRIP` | 0 | 0-1, 1-2, ..., (n-1)-n
+`LINE_LOOP` | 1 | 0-1, 1-2, ..., (n-1)-n, n-0
+`LINE_LIST` | 2 | 0-1, 2-3, 4-5, ...
```
amacneil, line 59: `LINE_LIST` seems unnecessary, since you could just add multiple Lines to your SceneEntity.

That would mean you could remove this whole enum and just have a `loop` boolean on the Line schema.

schemas/README.md
```diff
@@ -1593,6 +2172,548 @@ Raw image data
...
+
+</td>
+<td>
+
+Length of time (relative to `timestamp`) after which the entity should be automatically removed. Zero value indicates the entity should remain visible until it is replaced or deleted.
+
+</td>
+</tr>
+<tr>
+<td><code>frame_locked</code></td>
```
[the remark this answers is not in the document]
defunctzombie: Entities always follow their frame_id. This setting indicates whether the entity should be interpolated in time using its timestamp with transforms rather than using the latest transform for the frame.

schemas/README.md
defunctzombie: LINE_LIST is a rendering primitive. The difference between rendering a LINE_LIST from a set of points in one message vs rendering 1000 lines (imagine lane lines) from 1000 messages is both a massive message overhead and rendering overhead.

Its better to think about these messages as giving you ways to access rendering primitives.

schemas/README.md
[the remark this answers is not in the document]
amacneil: That is not what the docs for this field says (or at least not how I interpreted it).

Can you give me an example of when I would want this feature enabled and disabled?

schemas/README.md
amacneil: I see - GL refers to this as `LINES`.

https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/drawElements

Should we stick to the standard naming, or do we think our name is less confusing? For me personally our name seems less confusing, but I'm not a 3d developer.

src/schemas.ts
```diff
@@ -141,26 +141,565 @@ const foxglove_Quaternion: FoxgloveMessageSchema = {
...
+      description: "Whether to use materials embedded in the model, or only the `color`",
+    },
+    {
+      name: "url",
+      type: { type: "primitive", name: "string" },
+      description:
+        "URL pointing to model file. Either `url` or `mime_type` and `data` should be provided.",
+    },
+    {
+      name: "mime_type",
```
amacneil, line 570: Apparently they are called "media types" now:

https://www.iana.org/assignments/media-types/media-types.xhtml

schemas/README.md
jtbandes: Without the verb I'm a bit confused about what this might mean. "Has some embedded materials"? "Did embed materials"?

schemas/README.md
jtbandes: Or what about SceneEntityUpdates?

schemas/README.md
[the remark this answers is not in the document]
jtbandes: Enabled: cylindrical piece of a robot arm, should move with the arm as /tf changes.
Disabled: lidar point cloud, should be interpreted as fixed in world/map frame even though the detection coordinates are in lidar frame.

schemas/README.md
amacneil: 🤷 we don't have verbs for any other bool flags. but maybe this one is justified to be different. I don't feel strongly about it but curious what others think.

schemas/README.md
jtbandes: >  I assume clicking a SceneEntity would select that entire SceneEntity

This would be a regression from the old3d panel behavior where you can select a single cube in a cube list, or a single point in a point cloud.

Note: in verifying this claim I realized there's a bug(?)missing feature(?) in old3d where the specific clicked cube in a cube list is not displayed. However, the index is available in the code and I guess is just being ignored. I think we may have changed this at some point.

image

schemas/README.md
jtbandes: This was leaving it open for a third option to delete all entities under the same `namespace`, so depends how confident we are about the omission of the `namespace` feature. (Also I don't think we should design our API just around the simplicity of the docs, since there are also things we could do to improve the docs experience here, such as embedding the enum values directly in the message field docs.)

schemas/README.md
jtbandes: I didn't think about it. Would you still combine this with the `type` field (or `delete_all` boolean if we rename it that)? Would we remove the `deletions` array from SceneEntityUpdate or would it still be useful enough to be able to delete+add in the same batch?

schemas/README.md
jtbandes: We could do that. The tradeoff is that the total length (`shaft_length+head_length`) is not included in the message anywhere.

schemas/README.md
jtbandes: > If people want per-segment colors, they can already do that by adding multiple Lines to a SceneEntity.

That kind of works, except that multiple Lines in a SceneEntity can't be connected in a polygon or line strip — they would just be like a line list (which means sending extra copies of points, if they need to be the last point of one line & the first point of the next).

schemas/README.md
jtbandes: The diagram here is pretty good: http://www.opengl-tutorial.org/intermediate-tutorials/tutorial-9-vbo-indexing/

schemas/README.md
jtbandes: As currently written it was intended to be that it's ignored if `url` is set. But I suppose we could allow it to be treated as an override.

schemas/README.md
jtbandes: Also, we would have to verify that each Line primitive has the same `pose` in order to render them efficiently in a batch. With this current design we are trying to avoid wasting CPU time on auto-batching, rather just allowing the user to provide data in batches if they need more efficient rendering. All the segments (with per-point colors) could be included in one batch with the current design.

schemas/README.md
jtbandes: I think all our other bools can be interpreted as "is it X" or "does it X" and it makes sense, while this one is a bit different which is why it sounds weird to me. I also don't have strong feelings here though.

schemas/README.md
jtbandes: Another idea: SceneUpdate(s)?

schemas/README.md
```diff
@@ -4,21 +4,29 @@ Generated by https://github.com/foxglove/schemas
...
 - [enum PointsAnnotationType](#enum-pointsannotationtype)
 - [enum PositionCovarianceType](#enum-positioncovariancetype)
+- [enum SceneEntityDeletionType](#enum-sceneentitydeletiontype)
+- [ArrowPrimitive](#arrowprimitive)
 - [CameraCalibration](#cameracalibration)
 - [CircleAnnotation](#circleannotation)
 - [Color](#color)
 - [CompressedImage](#compressedimage)
+- [CubePrimitive](#cubeprimitive)
+- [CylinderPrimitive](#cylinderprimitive)
```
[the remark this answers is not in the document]
jtbandes: I'll let others weigh in on this one. For one minor point, the shorter names are more easily confused in code for other types/variables that don't come from @foxglove/schemas.

schemas/README.md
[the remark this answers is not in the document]
amacneil: @jtbandes thanks. That helped me realize that roman and I were saying the same thing, I'm just getting tripped up on language.

Ok so going back to my original point - it feels to me like the default setting should be for entities to be "locked in time" to their own transform frame rather than the fixed frame. Do others agree? If so, we could invert this flag so that `true` means "lock to fixed frame".

schemas/README.md
amacneil: I like `SceneUpdate`

schemas/README.md
amacneil: Makes sense. Since neither of us feel strongly, @defunctzombie do you want to weigh in?

schemas/README.md
amacneil: Interesting. If we're going to make clicks select the individual primitives, then what is the point of having a SceneEntity? I thought the whole idea was so that I could build an "entity" out of primitives, and have it behave like one "object" in the scene.

I guess another option is it could behave like a "grouped object" in google slides, or we could let you hold a modifier key if you want to select individual primitives rather than the whole entity.

schemas/README.md
amacneil: 👍 don't feel strongly about this

schemas/README.md
amacneil: Yes I guess we could potentially remove `deletions` from the `SceneEntityUpdate` and just ask you to publish a deletion followed by addition. If we did that it would then make sense to rename `SceneEntityUpdate` to just `SceneEntities` since it would only include additions.

I'm not sure how common different workflows will be, but I could totally see `delete_all` being commonly used alongside additions (if you want to simply redraw the world). It might be less common to include specific ID deletions alongside additions, although still plausible (if you had some node publishing updates from a tracker).

Net/net I feel like I would lean towards keeping `SceneUpdate` (aka `SceneEntityUpdate`), which includes options to (a) clear scene (delete all), (b) delete one or more entities by IDs, and (c) add one or more entities. In addition, it feels cleaner to _require_ `SceneUpdate` and not support publishing individual scene entity / deletion messages directly on a topic, since that would result in multiple schemas being published on the same topic, which is technically legal but unnecessarily messy.

thoughts?

schemas/README.md
amacneil: 👍 yeah I did not fully grasp the rendering perf impacts here

schemas/README.md
amacneil: this is over my head so I'm just gonna assume you know what you are talking about 😅

schemas/README.md
amacneil: discussed in other thread

schemas/README.md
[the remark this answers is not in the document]
defunctzombie: The default _is_ that entities are locked in time to their own transform frame. That's why `frame_locked` is _false_ by default. A value of _false_ means _interpolate_ (i.e. locked in time to my frame) while a value of _true_ means: ignore my timestamp and display me at the latest position of my frame relative to the scene.

schemas/README.md
defunctzombie: The `use` feel superfluous. If an embedded_material is present it should be used. We can allow a topic setting to toggle it out-of-band of we want to provide a visualization option but the message doesn't need to convey that.

schemas/README.md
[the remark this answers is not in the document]
defunctzombie: I don't feel strongly since most code will likely access it via a namespaced `foxglove` type or package. Notably we use `CircleAnnotation` to denote that specific circle is unique to annotation. As long as we don't immediately see a value in some other `Cube` message then dropping Primitive seems ok. Keeping Primitive seems equally fine. In a more developed world I'd possibly argue for `foxglove.scene.Cube` 🤣  but beyond that Cube or CubePrimitive is fine.

schemas/README.md
jtbandes: > If an embedded_material is present it should be used.

well no, that's the whole point of this field. Gives the user a way to disable the usage of the embedded material. Hende `use_embedded_materials=false`.

schemas/README.md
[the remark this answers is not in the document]
jtbandes: > The default _is_ that entities **are locked** in time to their own transform frame. That's why <code>frame_<b>locked</b></code><b> is _false_</b> by default.

:philosoraptor:

I don't think we all agree on what locked means here. Maybe this field needs a new name.

schemas/README.md
jtbandes: > It might be less common to include specific ID deletions alongside additions, although still plausible (if you had some node publishing updates from a tracker).

Sounds common to me! At least in self driving...

> require SceneUpdate and not support publishing individual scene entity / deletion messages directly on a topic, since that would result in multiple schemas being published on the same topic

Why would that result in multiple schemas on the same topic? You could have zero or more topics that are `SceneUpdate`, and zero or more that are `SceneEntityDeletion` or `SceneEntity`, that doesn't mean a single topic is switching its schema mid-stream.

schemas/README.md
defunctzombie: Is this something you need to configure when sending the message? You added the mesh - why not add one with no material? We don't override other properties of a mesh.

Is this somehow a very common thing to do? Seems like this field could be dropped?

defunctzombie: If this field needs to remain and is about enabling/disabled a property of the mesh I would favor using the words `enable` or `disable`

schemas/README.md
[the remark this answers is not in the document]
amacneil: I guess since we are all totally talking past each other, maybe we just keep this as-is since it at least matches the ROS conventions?

I think the ros docs actually did a better job of explaining this concept: _"If this marker should be frame-locked, i.e. retransformed into its frame every timestep"_.

Now that I thought about this some more, I actually think the default of `false` (objects stick to their original point in the fixed frame) makes sense. For any observations (sensor returns, tracked objects) they should not move with the transform frame. A value of `true` (objects move along with the transform frame) only makes sense for objects that are physically attached to that frame (e.g. mesh markers/entities), which seems like the exception rather than the rule.

Motion to close this thread.

schemas/README.md
amacneil: Isn't the `id` field effectively namespaced per topic? If I published `SceneEntity` on topic A and `SceneEntityDeletion` on topic B, I would not expect it to delete any objects from topic A.

schemas/README.md
amacneil: I think always treating it as an override (if set) probably makes more sense. I can think of why it might be useful, and it seems to simplify the logic. People can always leave it empty.

schemas/README.md
jtbandes: > You added the mesh - why not add one with no material?

Perhaps you are not a 3D artist and someone just handed you a mesh to use. Or you want to use the same mesh in multiple ways, such as a regular-looking car and a "ghost car" at the same time. Just some ideas. I don't feel strongly that this is an important feature, just explaining my guess at why it exists.

schemas/README.md
jtbandes: Oh yeah that's right

defunctzombie (approved):

defunctzombie:
Not that the public facing changes are none - these schemas are experimental and not documented for use yet.

PR #22, opened 2022-05-12
Rename DataField to PackedElementField

jtbandes:
**Public-Facing Changes**
DataField type is renamed to PackedElementField


defunctzombie (approved):

amacneil:
Now that I understand what this actually does, DataField was a pretty good name for it. It definitely needed a better description though.

jtbandes:
🫠

defunctzombie:
> Now that I understand what this actually does, DataField was a pretty good name for it.

Disagree. It describes a single field in an element of an array. DataField is too generic.

PR #23, opened 2022-05-12
Clean up schema descriptions

jtbandes:
**Public-Facing Changes**
Updated descriptions for several schemas.

**Description**
Clean up from review of https://github.com/foxglove/message-schemas/pull/14

### DIFF
--- a/schemas/jsonschema/CircleAnnotation.json
+++ b/schemas/jsonschema/CircleAnnotation.json
@@ -1,7 +1,7 @@
 {
   "$comment": "Generated from CircleAnnotation by @foxglove/message-schemas",
   "title": "CircleAnnotation",
-  "description": "A circle annotation",
+  "description": "A circle annotation to be superimposed onto 2D images",
   "type": "object",
   "properties": {
     "timestamp": {
--- a/schemas/jsonschema/CompressedImage.json
+++ b/schemas/jsonschema/CompressedImage.json
@@ -1,7 +1,7 @@
 {
   "$comment": "Generated from CompressedImage by @foxglove/message-schemas",
   "title": "CompressedImage",
-  "description": "A compressed image",
+  "description": "A raw image",
   "type": "object",
   "properties": {
     "timestamp": {
--- a/schemas/jsonschema/DataField.json
+++ b/schemas/jsonschema/DataField.json
@@ -1,7 +1,7 @@
 {
   "$comment": "Generated from DataField by @foxglove/message-schemas",
   "title": "DataField",
-  "description": "List of fields included for every entity in an accompanying `data` field array",
+  "description": "List of fields for every element in an accompanying `data` field array",
   "type": "object",
   "properties": {
     "name": {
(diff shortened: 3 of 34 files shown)


schemas/jsonschema/CircleAnnotation.json
```diff
@@ -1,7 +1,7 @@
 {
   "$comment": "Generated from CircleAnnotation by @foxglove/message-schemas",
   "title": "CircleAnnotation",
-  "description": "A circle annotation",
+  "description": "A circle annotation to be superimposed onto 2D images",
```
defunctzombie, line 4: Something about `to be superimposed` feels "off". Either it's the passive voice or maybe something else.

`A circle annotation for a 2D image`?

schemas/jsonschema/CompressedImage.json
```diff
@@ -1,7 +1,7 @@
 {
   "$comment": "Generated from CompressedImage by @foxglove/message-schemas",
   "title": "CompressedImage",
-  "description": "A compressed image",
+  "description": "A raw image",
```
defunctzombie, line 4: Why does this say raw? This is a compressed image. "Raw image" to me means something else:

https://en.wikipedia.org/wiki/Raw_image_format

schemas/jsonschema/DataField.json
```diff
@@ -1,7 +1,7 @@
 {
   "$comment": "Generated from DataField by @foxglove/message-schemas",
   "title": "DataField",
-  "description": "List of fields included for every entity in an accompanying `data` field array",
+  "description": "List of fields for every element in an accompanying `data` field array",
```
defunctzombie, line 4: DataField is now PackedElementField?

schemas/jsonschema/ImageAnnotations.json
```diff
@@ -9,7 +9,7 @@
       "items": {
         "$comment": "Generated from CircleAnnotation by @foxglove/message-schemas",
         "title": "CircleAnnotation",
-        "description": "A circle annotation",
+        "description": "A circle annotation to be superimposed onto 2D images",
```
defunctzombie, line 12: Same feedback as on earlier CircleAnnotation message

schemas/jsonschema/DataField.json
defunctzombie: The `data` field part of this should be removed from this description cause thats a property of the message. The message doesn't have to use the `data` name for the array field.

defunctzombie (approved):
Some wording feedback. I didn't know the right file to comment on so I commented on the json schema files but the comments apply to all variants of the files.

schemas/jsonschema/CompressedImage.json
jtbandes: I guess I changed the wrong one 🤦  the other one also says compressed.

schemas/jsonschema/DataField.json
jtbandes: The rename is a separate PR which I requested your review on

jtbandes:
The source of truth is schemas.ts

schemas/jsonschema/CircleAnnotation.json
esthersweon: That revision sounds good ^

PR #26, opened 2022-05-17
Update description for PackedElementField

jtbandes:
**Public-Facing Changes**
Updated description for PackedElementField

### DIFF
--- a/schemas/jsonschema/Grid.json
+++ b/schemas/jsonschema/Grid.json
@@ -112,7 +112,7 @@
       "items": {
         "$comment": "Generated from PackedElementField by @foxglove/message-schemas",
         "title": "PackedElementField",
-        "description": "Defines a field within packed data. Commonly used to represent a field of a single element within a byte array containing multiple packed elements.",
+        "description": "A field present within each element in a byte array of packed elements.",
         "type": "object",
         "properties": {
           "name": {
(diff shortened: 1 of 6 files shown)


defunctzombie (approved):

schemas/jsonschema/Grid.json
```diff
@@ -112,7 +112,7 @@
       "items": {
         "$comment": "Generated from PackedElementField by @foxglove/message-schemas",
         "title": "PackedElementField",
-        "description": "List of fields for every element in a byte array",
+        "description": "Defines a field within packed data. Commonly used to represent a field of a single element within a byte array containing multiple packed elements.",
```
esthersweon, line 115: Can remove "Defines" from beginning of every version of this schema. Don't think the other schemas are formatted this way.

esthersweon:
```suggestion
  "description": "A field present within each element in a byte array of packed elements.",
```

esthersweon (changes requested):
Suggested changes to description.

schemas/jsonschema/Grid.json
esthersweon: Or maybe even "each packed element in a byte array".

esthersweon (approved):

PR #33, opened 2022-05-20
Add supported encoding formats to RawImage and CompressedImage

### LINKED ISSUE #32: Add supported format/encoding in description for RawImage and CompressedImage
Updated in website docs here: https://foxglove.dev/docs/studio/panels/image

esthersweon:
**Public-Facing Changes**
- Add supported encoding formats to RawImage and CompressedImage 

Resolves #32

### DIFF
--- a/schemas/jsonschema/CompressedImage.json
+++ b/schemas/jsonschema/CompressedImage.json
@@ -27,7 +27,7 @@
     },
     "format": {
       "type": "string",
-      "description": "Image format"
+      "description": "Image format (webp, jpeg, png)"
     }
   }
 }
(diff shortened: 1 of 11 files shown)


jtbandes (approved):

amacneil (approved):

amacneil:
ready to merge?

amacneil (changes requested):
On second thoughts, I do not think we should have two sources of truth on this in our docs. It seems inevitable that someone will add support for a new format, and update one of these lists but not the other (especially since they live in separate repos).

I would prefer something like:

```
[Image format](https://foxglove.dev/docs/studio/panels/image#supported-encoding-formats)
```

It won't look as great in generated schema source code, but is obvious enough, and will improve the website docs.

esthersweon:
Closing this PR, and addressing in https://github.com/foxglove/schemas/pull/60.

PR #40, opened 2022-06-14
Fix ROS packaging again

jtbandes:
Rework CMake setup again to avoid using `..` by copying generated files into the ros_foxglove_msgs dir. This is necessary because bloom-release will end up copying the ros_foxglove_msgs dir only, losing access to files from the parent dir during the final build.


amacneil:
😭

PR #46, opened 2022-07-13
Consolidate Transform and FrameTransform

amacneil:
**Public-Facing Changes**
- Move `translation` and `rotation` fields directly into `FrameTransform`
- Remove otherwise unused `Transform` message

**Justification**
- Simpler to create `FrameTransform` messages without unnecessary nesting
- `Transform` is otherwise unused in both our current schemas and those proposed by https://github.com/foxglove/schemas/pull/17
- If we ever did need to reintroduce `Transform` in the future, there is no need for it to be a nested field on `FrameTransform` and each type can have its own `translation` and `rotation` fields. But so far no one has been able to explain why we would need a dedicated `Transform` message so for now this is a non-issue.


jhurliman (approved):

defunctzombie (changes requested):
I'd like to hear the plan on how we will support this in a non-breaking manner for users that are already creating the foxglove.FrameTransform message.

amacneil:
Per discussion we will maintain backwards compatibility in Studio, but are ok breaking backwards compatibility with the proto schemas since these are not in wide use yet. Existing mcap files containing old schemas will continue to work.

PR #48, opened 2022-07-13
Add canonical name to schema comments

amacneil:
**Public-Facing Changes**
Update schema comments to add the canonical message name. For example
- `foxglove_msgs/Color` (ROS 1)
- `foxglove_msgs/msgs/Color` (ROS 2)
- `foxglove.Color` (proto)
- `foxglove.Color` (json schema - updated the `title` field)
- `Color` (typescript)

**Rationale**
At least in Foxglove Studio, these message names are significant. When linking someone to a schema, it is nice if the schema name is directly mentioned so that it is obvious how to use it (especially for ROS users).

Originally I considered keeping the `Generated by @foxglove/schemas` text, but I'm not sure that this adds anything. `@foxglove/schemas` is an npm syntax package name and not relevant to most of these languages. Also, the schemas weren't so much generated _by_ this project as generated _for_ it, and are part of the project. They do happen to be generated files, but I don't think that is relevant to the schema consumer.


src/generateJsonSchema.ts
```diff
@@ -48,7 +48,7 @@ export function generateJsonSchema(schema: FoxgloveMessageSchema): Record<string
         break;
       case "enum":
         fieldType = {
-          title: `${field.type.enum.name}: ${field.type.enum.description}`,
+          title: `foxglove.${field.type.enum.name}`,
```
jtbandes, line 51: Why remove the enum description?

jtbandes (commented):
> Originally I considered keeping the `Generated by @foxglove/schemas` text, but I'm not sure that this adds anything. `@foxglove/schemas` is an npm syntax package name and not relevant to most of these languages. Also, the schemas weren't so much generated _by_ this project as generated _for_ it, and are part of the project. They do happen to be generated files, but I don't think that is relevant to the schema consumer.

I like sticking to the principle that all generated files should include a note explaining that they were generated, and pointing the reader to the code that generated them. Otherwise someone who comes along wanting to change a file does not realize that it's a generated file, and wastes some time figuring that out and hunting for the code/project that generated it. I have personally been confused enough times this way (and appreciated when this info existed) that I started doing it too.

I don't see the advantage of putting `foxglove_msgs/PackedElementField` in the ROS .msgs, `foxglove.PackedElementField` in the .protos, or `// PackedElementField` in the .ts files. All of these are self-evident based on where the file came from (in ROS it'll be in the foxglove_msgs/msg dir with filename PackedElementField.msg) or its contents (in protobuf there's `package foxglove` and `message PackedElementField`, and in TS there's `class PackedElementField`). 🤷

src/generateJsonSchema.ts
amacneil: Because it's not a title. Also I checked every instance of these and even though they had slightly different text from the `description` field, they were still redundant in every case.

ros_foxglove_msgs/ros1/CameraCalibration.msg
```diff
@@ -1,4 +1,5 @@
-# Generated from CameraCalibration by @foxglove/schemas
+# foxglove_msgs/CameraCalibration
```
defunctzombie, line 1: I'd like to advocate for keeping the `generated from` text but having it reference he repository. `@foxglove/schemas` is an npm specific name and wouldn't help you figure out where the code was that generated these files.

I do not think having `foxglove_msgs/CameraCalibration` is adding anything as a comment.

schemas/jsonschema/CameraCalibration.json
```diff
@@ -1,6 +1,5 @@
 {
-  "$comment": "Generated from CameraCalibration by @foxglove/schemas",
-  "title": "CameraCalibration",
+  "title": "foxglove.CameraCalibration",
```
defunctzombie, line 2: This seems more correct as a title since the `foxglove.` namespace part is important.

Removing the comment on where these were generated is a downgrade.

schemas/proto/foxglove/CameraCalibration.proto
```diff
@@ -1,4 +1,4 @@
-// Generated from CameraCalibration by @foxglove/schemas
+// foxglove.CameraCalibration
```
defunctzombie, line 1: This comment is not useful. Its a proto file that has a `CameraCalibration` message in the `foxglove` namespace. Users of proto understand this equals `foxglove.CameraCalibration`.

The comment should mention what repo generated this and could also link to our MessageSchema docs which help you discover how to then use them in studio.

schemas/typescript/CameraCalibration.ts
```diff
@@ -1,4 +1,4 @@
-// Generated from CameraCalibration by @foxglove/schemas
+// CameraCalibration
```
defunctzombie, line 1: This comment is not useful.

Mentioning where the file was generated from and referencing the schema docs to discover how to leverage the schemas in Studio would be useful.

ros_foxglove_msgs/ros1/CameraCalibration.msg
defunctzombie: If you want to make it clear that it belongs to a ros package, just be explicit and say `Part of the foxglove_msgs package` - even tho I'm not sure that matters since a user is gonna need to go find the install instructions for foxglove_msgs package anyway - maybe the comment can link there but then its kind of silly since how did you install those in the first place?

amacneil: I link people to this repo, sometimes directly to the individual messages. I would like it to be clear to the reader what the name of this package and schema is, since it is not otherwise mentioned within the file.

schemas/jsonschema/CameraCalibration.json
amacneil: Restored comment

amacneil:
@defunctzombie @jtbandes updated with feedback

schemas/proto/foxglove/CameraCalibration.proto
amacneil: removed for proto files

schemas/typescript/CameraCalibration.ts
amacneil: removed for typescript

defunctzombie (approved):

jtbandes (approved):

PR #55, opened 2022-07-31
add Event schema

james-rms:
This commit adds a schema that allows a robot to publish an Event.
An Event is intended to annotate a range of time with some metadata,
to facilitate searching the dataset. The content and field names
for this message closely mirrors the Event API exposed by Foxglove
Data Platform.

**Public-Facing Changes**

**Description**


src/schemas.ts
```diff
@@ -800,6 +800,48 @@ const foxglove_LaserScan: FoxgloveMessageSchema = {
...
+      type: { type: "primitive", name: "string" },
+      description: "key for this metadata",
+    },
+    {
+      name: "value",
+      type: { type: "primitive", name: "string" },
+      description: "value of the metadata",
+    },
+  ],
+};
```
jtbandes, line 819: Adding an identical KeyValuePair in https://github.com/foxglove/schemas/pull/17. Want to remove Event from the name and "metadata" from the descriptions so we can reuse it in both places?

src/schemas.ts
```diff
@@ -800,6 +800,48 @@ const foxglove_LaserScan: FoxgloveMessageSchema = {
...
+
+const foxglove_Event: FoxgloveMessageSchema = {
+  type: "message",
+  name: "Event",
+  description: "a message used to annotate some event of interest in the data",
```
jtbandes, line 824: nit: I believe we are capitalizing sentences for descriptions on the other types/fields in this file

src/schemas.ts
amacneil: Seems ok for now. The Event schema might evolve more over time (e.g. to support other structured data like nested arrays / structs) but we can cross that bridge when we get to it.

schemas/typescript/Event.ts
```diff
@@ -0,0 +1,17 @@
...
+import { EventKeyValue } from "./EventKeyValue";
+import { Time } from "./Time";
+
+/** a message used to annotate some event of interest in the data */
+export type Event = {
+  /** time when the event occurred */
+  timestamp: Time;
+
+  /** duration of event, starting at `timestamp` */
+  duration: Duration;
```
defunctzombie, line 13: Should duration be optional to indicate an instantaneous event? Or do we require specifying 0 length duration?

amacneil: optional imo (for schema formats that support optionals)

james-rms:
Closing this PR for now, will reopen once the plan for Event ingestion is more solidified.

PR #61, opened 2022-08-31
add frame_id to RawImage, CompressedImage, and CameraCalibration

### LINKED ISSUE #59: Add `frame_id` to image types
Image and CompressedImage need `frame_id` if we are going to display them in the 3D panel. CameraCalibration may need it as well if we want to use `frame_id` to confirm that the calibration topic corresponds to the image topic.

jtbandes:
**Public-Facing Changes**
Added `frame_id` to RawImage, CompressedImage, and CameraCalibration

**Description**
Closes #59 
Unblocks https://github.com/foxglove/studio/issues/4323


ros_foxglove_msgs/ros1/CameraCalibration.msg
```diff
@@ -62,3 +62,6 @@ float64[9] R
 # 
 # This holds for both images of a stereo pair.
 float64[12] P
+
+# Frame of reference for the camera. The origin of the frame is the optical center of the camera. +x points to the right in the image, +y points down, and +z points into the plane of the image.
+string frame_id
```
defunctzombie, line 67: Can we make this field first in the message?

schemas/proto/foxglove/CameraCalibration.proto
```diff
@@ -67,4 +67,7 @@ message CameraCalibration {
   // 
   // This holds for both images of a stereo pair.
   repeated double P = 8; // length 12
+
+  // Frame of reference for the camera. The origin of the frame is the optical center of the camera. +x points to the right in the image, +y points down, and +z points into the plane of the image.
+  string frame_id = 9;
```
defunctzombie, line 72: Can we make this field first in the message (still id 9?) is it useful to do that?

src/schemas.ts
```diff
@@ -741,6 +741,12 @@ Given a 3D point [X Y Z]', the projection (x, y) of the point onto the rectified
 This holds for both images of a stereo pair.
 `,
     },
+    {
+      name: "frame_id",
+      type: { type: "primitive", name: "string" },
+      description:
+        "Frame of reference for the camera. The origin of the frame is the optical center of the camera. +x points to the right in the image, +y points down, and +z points into the plane of the image.",
```
defunctzombie, line 748: Is it worth calling out that CameraCalibration should be in the same frame as its corresponding image? What happens if the CameraCalibration is a different frame?

defunctzombie (approved):
Seems reasonable. Might be worth thinking through:

https://github.com/foxglove/studio/issues/4323#issuecomment-1233407797

And whether there are other useful or common ways of tying cameras to calibration data and/or frames.

ros_foxglove_msgs/ros1/CameraCalibration.msg
jtbandes: We would need to add special behavior for protobuf to keep binary compatibility. On the other hand, maybe we don't need to keep binary compatibility since the definitions will be embedded anyway?

defunctzombie: I don't feel strongly about it being first but it does seem like one of those nice things. Like if I was writing the message by hand I would want that.

If we are providing protobuf examples it seems nice to downstream consumers to attempt to be backwards compatible - tho we undid that with FrameTransform. We should be clear _somewhere_ very obvious of our goals with protobuf compatibility. Might be worth saying these schemas are provided as examples and that you should be mindful of field ids when making changes?

src/schemas.ts
jtbandes: No idea really. Probably bad things.

PR #64, opened 2022-09-20
Remove "experimental" note from SceneEntity types

jtbandes:
**Public-Facing Changes**
Removed "experimental" language from SceneEntity and related types.


defunctzombie (approved):
Guess these are always subject to change and a little less experimental now.

PR #65, opened 2022-09-29
Generate bazel build rules for proto schemas

alokpr:
Auto-generate bazel build rules for proto schemas using gazelle and buildozer.

Also added missing zero enum value for LogLevel


jtbandes:
Closing out this PR since we've had some offline discussions and the proto schemas on the `main` branch are in a good state now. Feel free to ping in Slack if there are any loose ends you need help with.

PR #70, opened 2022-11-08
Add Flatbuffer Schema Support

snosenzo:
- Add code to generate flatbuffer schema files: `.fbs`.
 - Added test for expected output

To test compilation: install `cmake` and `flatbuffer` (via homebrew or other means) then run this command from the repo root directory: `flatc --ts  -o ./schemas/flatbuffer/output ./schemas/flatbuffer/foxglove/**.fbs`
 and see that it generates the schema files in the `schemas/flatbuffer/output` directory. 

Few Open Questions about this implementation:
- ~~should I use the protobuf enums names?~~
  - Should enums by namespaced by their parent ?
- ~~just to be sure there aren’t any wellknown types for time and duration like protobuf/ros?~~
    - ~~currently using something similar to the way we handle typescript generation~~
    - none exist
- ~~Do we want defaults for primitive types?~~
    - added defaults directly to the schema that are reflected in protobuf
- ~~should we copy protobuf file structure and namespaces (foxglove.etc)~~?
  - all necessary schemas are under the foxglove namespace


jameskuszmaul-brt:
> * should I use the protobuf enums names?
> * just to be sure there aren’t any wellknown types for time and duration like protobuf/ros?

There are no well-known types.

>   
>   * currently using something similar to the way we handle typescript generation
> * Do we want defaults for primitive types?

Note that the "default defaults" are zero, which is typically reasonable. I'd suggest explicit defaults for enums.

>   
>   * incorporated into schema or defaults per primitive added at generation
> * should we copy protobuf file structure and namespaces (foxglove.etc)?

File structure is generally whatever. Namespaces should definitely be consistent.

>   
>   * should enums be separate files or with their message schema like protobuf?

It doesn't strictly matter if they are in separate files or not. My typical philosophy is to not put them into separate files unless they are logically separate or there is a particular need to (e.g., enums are shared across many different types). Since it looks like you are doing this based on codegen, I think your current approach is fine.

> 
> NOTE: the compiler does warn that some `CameraCalibration` fields are not lowercase snake_case compatible.
> 
> ```
> /Users/samnosenzo/dev/schemas/schemas/flatbuffer/foxglove/CameraCalibration.fbs:26: 3: warning: field names should be lowercase snake_case, got: D
> ```

I think that in order to maintain consistency with the schemas in your other languages, you want to leave it as-is. The fbs compiler will sometimes mess with the names for certain codegen, which is why it watches for this (e.g., for Java, a `snake_case` name in the fbs file gets turned into a (not)`SnakeCase` name in the generated java code).

jhurliman:
> I think that in order to maintain consistency with the schemas in your other languages

ROS 2 does not support naming a field `D`, so it would be consistent with at least our ROS 2 generator to rename this to `d`.

jameskuszmaul-brt:
> > I think that in order to maintain consistency with the schemas in your other languages
> 
> ROS 2 does not support naming a field `D`, so it would be consistent with at least our ROS 2 generator to rename this to `d`.

Are the panels that consume these fields on the Studio side case-insensitive?

jhurliman:
> Note that the "default defaults" are zero, which is typically reasonable

Some notable exceptions:
 - w component of quaternions (1 is better)
 - r,g,b,a fields of color (1 is better)
 - x,y,z fields of scale (1 is better)

jameskuszmaul-brt:
Also re: defaults: note that, like protobufs, flatbuffers do have a concept of a field not being populated. It is also the case that the serialization for flatbuffers by default is configured to not serialize fields if the field is set to its default (so `has_X()` may return false even if the serializer did `add_X(default)`), although that is configurable https://google.github.io/flatbuffers/classflatbuffers_1_1_flat_buffer_builder.html#a16a8fd46b34ad7727406c37b65b6b27a .

jtbandes:
Is there some compile step we can run in CI to validate these generated files? For example, we do this for protobuf:

https://github.com/foxglove/schemas/blob/b36014a3987ea81cf0184b7abb2d6ab6a3888688/.github/workflows/ci.yml#L39-L40

jtbandes:
> should I use the protobuf enums names?

Protobuf has special names because the types are nested inside message types:

>   name: "SceneEntityDeletionType",
>   protobufParentMessageName: "SceneEntityDeletion",
>   protobufEnumName: "Type",

The enum which is called `SceneEntityDeletionType` in other languages becomes `SceneEntityDeletion.Type` in protobuf.

This was done partly because the way protobuf c++ code is generated, the enum names are in the parent namespace, not namespaced under the enum type itself. So multiple top-level enums with `UNKNOWN` values would conflict with each other.

I don't know if flatbuffers would have the same problem or not?

jameskuszmaul-brt:
> The enum which is called `SceneEntityDeletionType` in other languages becomes `SceneEntityDeletion.Type` in protobuf.
> 
> This was done partly because the way protobuf c++ code is generated, the enum names pollute their parent namespace, rather than being namespaced under the enum type itself. So multiple top-level enums with `UNKNOWN` values would conflict with each other.
> 
> I don't know if flatbuffers would have the same problem or not?

It depends on where/how the types are used.

In C++, `flatc` provides a `--scoped-enums` flag to let you choose how you want your enums generated. We always turn that on in our projects, but I think in upstream flatbuffers it is off by default. Other languages have language-specific behavior.

snosenzo:
> r,g,b,a fields of color (1 is better) & x,y,z fields of scale (1 is better)

Is it fine to just add ` = 1.0` to all fields named this way that are primitives? Or should I hard-code certain ones that should apply to these?

> Enum Scoping and Namespaces

So as it stands right now all of the enums except for the pre-determined (ByteVectorForNesting, Time, Duration) ones all live in the foxglove namespace.
 1. should I put these into the foxglove namespace?
 2. It seems like it might also be desireable to have trailing namespaces for enums that belong to protobuf parents. I can add specific namespace definitions to their schema to not pollute the foxglove namespace. so the LineType enum would be `namespace foxglove.LinePrimitive`. Is this more desirable than having them all in the foxglove namespace?
 
 > flatc provides a --scoped-enums
 It's a little unclear to me what this uses to determine scoping. Is it just if it's in the file with a `root_type`?

snosenzo:
> Is there some compile step we can run in CI to validate these generated files? For example, we do this for protobuf:

looked into this and can't find a github action for it like there is for protobuf unfortunately. Is there somewhere else I should look or do you know how easy it would be to write one? I just don't know that much about making github actions

jameskuszmaul-brt:
> It's a little unclear to me what this uses to determine scoping. Is it just if it's in the file with a `root_type`?

In C++ it generates a [Scoped Enum](https://en.cppreference.com/w/cpp/language/enum)

I.e., with `--scoped-enums` if you have a fbs file like

```
namespace foo;

enum Abc : byte {
  A = 0,
  B = 1
}
```

Then the generated C++ code will be
```cpp
enum class Abc : int8_t {
  A = 0,
  B = 1,
  MIN = A,
  MAX = B
};
```

Without `--scoped-enums`, it looks like `flatc`'s default C++ code generates
```cpp
enum Abc : int8_t {
  Abc_A = 0,
  Abc_B = 1,
  Abc_MIN = A,
  Abc_MAX = B
};
```

They also seem to have a flag to strip those prefixes off, but I doubt most people use it.

Other languages I think tend to have some sort of scoping to their enums by default, so this is less of an issue outside of C++.

internal/generateFlatbuffer.ts
```diff
@@ -0,0 +1,137 @@
...
+
+// fields that would benefit from having a default of 1
+const defaultOneNumberFields = new Set(["x", "y", "z", "r", "g", "b", "a", "w"]);
+function isDefaultOneField(field: FoxgloveMessageField): boolean {
+  return (
+    field.type.type === "primitive" &&
+    field.type.name === "float64" &&
+    defaultOneNumberFields.has(field.name)
+  );
+}
```
jtbandes, line 32: This doesn't belong here because it's not really specific to flatbuffers. If we want to add defaults to some fields, we need to add the default values to `schemas.ts` and just read them here.

internal/generateFlatbuffer.ts
```diff
@@ -0,0 +1,137 @@
...
+
+function primitiveToFlatbuffer(type: Exclude<FoxglovePrimitive, "time" | "duration">) {
+  switch (type) {
+    case "uint32":
+      return "uint";
```
jtbandes, line 37: Can this be uint32?

internal/generateFlatbuffer.ts
```diff
@@ -0,0 +1,137 @@
...
+    case "bytes":
+      return "[byte]";
```
jtbandes, line 39: Looks like `[ubyte]` or `[uint8]` would be preferable. Sign is not needed.

internal/generateFlatbuffer.ts
```diff
@@ -0,0 +1,137 @@
...
+      const fields = schema.values.map(({ name, value, description }) => {
+        if (description != undefined) {
+          return `/// ${description}\n  ${name} = ${value},`;
+        } else {
+          return `${name} = ${value},`;
+        }
+      });
+
+      // `///` comments required to show up in compiled flatbuffer schemas
+      definition = `/// ${schema.description}\nenum ${schema.name} : byte {\n  ${fields.join(
```
jtbandes, line 63: Should we use ubyte for enums? Might want to add validation that the values all fit within the range of the type we choose. I assume the compiler would validate this, so if we can get the compiler to be tested in CI, then that would be sufficient.

jtbandes (commented):
How does someone go about merging flatbuffer schemas into one, in order to put them in an mcap/websocket schema? Is it acceptable that each file has `root_type` in it? I would guess not, and we might want to remove `root_type` from our generated fbs files, but maybe there is some convenient way of merging them (like protoc's `--descriptor_set_out`)?

jtbandes:
> looked into this and can't find a github action for it like there is for protobuf unfortunately. Is there somewhere else I should look or do you know how easy it would be to write one? I just don't know that much about making github actions

Looks like `flatc` pre-compiled binaries are provided here: https://github.com/google/flatbuffers/releases/tag/v22.10.26

So it would be possible to simply download and run them in the CI step. Take a look at the "before" side here for an example of downloading & running binaries as part of GitHub CI: https://github.com/foxglove/mcap/pull/604/files

(If you go this route, might also want to add a SHA check for security as was suggested in https://github.com/foxglove/studio/pull/4745#pullrequestreview-1166012825)

jtbandes:
> So as it stands right now all of the enums except for the pre-determined (ByteVectorForNesting, Time, Duration) ones all live in the foxglove namespace.
>  1. should I put these into the foxglove namespace?

Yes, at least I would say they probably shouldn't be floating around in the top level namespace. I don't know how name/import conflicts work in flatbuffers, but if someone is mixing their own custom schema with our schemas, we wouldn't want our Time/Duration to conflict with one they might already have. Maybe @jameskuszmaul-brt can give more insight on whether that would actually be a problem.

>  2. It seems like it might also be desireable to have trailing namespaces for enums that belong to protobuf parents. I can add specific namespace definitions to their schema to not pollute the foxglove namespace. so the LineType enum would be `namespace foxglove.LinePrimitive`. Is this more desirable than having them all in the foxglove namespace?

Not sure I fully understand the different options here – what would it look like to use the enums in the different scenarios you describe?

scripts/updateGeneratedFiles.ts
```diff
@@ -72,6 +78,28 @@ async function main({ outDir, rosOutDir }: { outDir: string; rosOutDir: string }
     }
   });
 
+  await logProgress("Generating FlatBuffer definitions", async () => {
+    await fs.mkdir(path.join(outDir, "flatbuffer", "foxglove"), { recursive: true });
+    await fs.writeFile(
+      path.join(outDir, "flatbuffer/foxglove", "ByteVectorForNesting.fbs"),
+      BYTE_VECTOR_FB,
+    );
+    await fs.writeFile(path.join(outDir, "flatbuffer/foxglove", "Time.fbs"), TIME_FB);
```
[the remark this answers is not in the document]
jameskuszmaul-brt: The github repo is https://github.com/google/flatbuffers and the C++ includes are in a `flatbuffers` folder.

jameskuszmaul-brt:
> How does someone go about merging flatbuffer schemas and their includes into one blob, in order to put them in an mcap/websocket schema? Is it acceptable that each file has `root_type` in it? I would guess not, and we might want to remove `root_type` from our generated fbs files, but maybe there is some convenient way of merging them (like protoc's `--descriptor_set_out`)?

Having root types for all the messages is fine (and probably what you want). Honestly not entirely sure why flatbuffers has the concept, since I'm pretty sure all it does is generate some extra code for that type, when it would be fine to just always generate the extra code.


> Yes, at least I would say they probably shouldn't be floating around in the top level namespace. I don't know how name/import conflicts work in flatbuffers, but if someone is mixing their own custom schema with our schemas, we wouldn't want our Time/Duration to conflict with one they might already have. Maybe @jameskuszmaul-brt can give more insight on whether that would actually be a problem.

I'd prefer they be in the `foxglove` namespace. Polluting the top-level namespace could make things inconvenient.

snosenzo:
> re: root types

From what I can tell the `root_type` is only really used for serializing/deserializing json like what's shown in[ this tutorial](https://google.github.io/flatbuffers/flatbuffers_guide_tutorial.html) under "JSON to binary" section. It also does add an extra export from the compiled typescript files so `ArrowPrimitive__generated.ts` will contain an `export {ArrowPrimitive} from './foxglove/arrow-primitive.ts'` which could be preferable.

> re: enums under parent protobuf foxglove namespace. 

after some testing, adding the parent proto types to the namespace doesn't work because the schemas referencing those enums don't have access to know what the parent type is 😅. So the options are just have the enums under the `foxglove` namespace or go back to my original, closer-to-protobuf generator that added the enums definitions to the parent flatbuffer schema files. They would also be under the foxglove namespace, but would be imported from their generated parent typescript file rather than their own file. This makes a difference in typescript, but I'm not sure what kind of impact it has on C++ given everything is still under the same namespace

scripts/updateGeneratedFiles.ts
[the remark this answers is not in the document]
snosenzo: There's no real prescribed folder structure for flatbuffers as far as I can tell. I thought I'd follow the protobuf pattern of making it follow the namespace, but that's probably not necessary. Removed.

internal/generateFlatbuffer.ts
```diff
@@ -0,0 +1,137 @@
...
+  sec:long;
+  nsec:int;
+}
+`;
+
+export const DURATION_FB = `struct Duration {
+  sec:long;
+  nsec:int;
+}
+`;
```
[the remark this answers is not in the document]
snosenzo: yeah good call, added the comment in line with the protobuf. 
What kind of consideration should be made to the ROS times schemas in this context? Is this difference going to cause issues at read time?

internal/generateFlatbuffer.ts
snosenzo: Added to the schema under `defaultValue?: string` if that seems appropriate.

jtbandes:
> the schemas referencing those enums don't have access to know what the parent type is

1. can't they look them up from the data in schemas.ts?
2. I thought the parent type was usually the same schema where the enum gets used...

.github/workflows/ci.yml
```diff
@@ -36,6 +36,13 @@ jobs:
             echo "Generated schemas are up to date!"
           fi
 
+      - name: Validate Flatbuffer definitions
+        run: |
+          curl -LO https://github.com/google/flatbuffers/releases/download/v22.10.26/Linux.flatc.binary.clang++-12.zip 
+          echo "0821af82a3a736b0ba9235c02219df24d1f042dd  Linux.flatc.binary.clang++-12.zip" | shasum -a 1 -c
+          unzip Linux.flatc.binary.clang++-12.zip
+          ./flatc --ts -o /dev/null schemas/flatbuffers/*.fbs
```
jtbandes, line 44: If there were warnings/errors would it actually affect the exit code of this flatc invocation?

snosenzo: if there are errors, it returns an exit code of 1, so I believe it should exit out. On warnings, however, it still returns 0. I can add some logic to this to fail out when there's a warning though

jtbandes: Yeah I think if that's not too hard, it would be nice to catch warnings as well. (Assuming the warnings are actually useful and not just noisy/pedantic)

snosenzo:
Just updated to have the enums included in their parent table files again. I think this makes a bit more sense going forward and prevents empty files being generated for the enums that were alone in their schema files.

On another note, I've been working on setting up an example file to test in studio and realized (as far as I know) that the schema has to be in binary format to be attached to the mcap schema. Do we want to also hold those binary schemas (.bfbs files) in this repo, since those are what will need to be passed to the mcap schemas when they are registered? Not sure what the ideals would be for consumption of these schemas

jameskuszmaul-brt:
> I've been working on setting up an example file to test in studio and realized (as far as I know) that the schema has to be in binary format to be attached to the mcap schema

Correct, the MCAP definition presumes the binary schemas. Whatever you do, it seems like it should be able to be consistent with however you do it for protobufs.

scripts/updateGeneratedFiles.ts
```diff
@@ -83,16 +83,15 @@ async function main({ outDir, rosOutDir }: { outDir: string; rosOutDir: string }
...
     await fs.writeFile(path.join(outDir, "flatbuffers", "Time.fbs"), TIME_FB);
     await fs.writeFile(path.join(outDir, "flatbuffers", "Duration.fbs"), DURATION_FB);
+
     for (const schema of Object.values(foxgloveMessageSchemas)) {
-      await fs.writeFile(
-        path.join(outDir, "flatbuffers", `${schema.name}.fbs`),
-        generateFlatbuffers(schema),
+      // want enums with their corresponding parent tables for usage
+      const enums = Object.values(foxgloveEnumSchemas).filter(
+        (enumSchema) => enumSchema.protobufParentMessageName === schema.name,
```
jtbandes, line 90: let's rename these `protobuf` things now that they are not exclusive to protobuf

jtbandes:
Do the binary schemas include all the dependencies so they are fully self-contained? How does one create a merged binary schema from the individual .fbs files?

With Protobuf we originally recommended people use `protoc --include_imports --descriptor_set_out` to create the merged binary schemas, but over time we have moved toward recommending they generate them at runtime, since it's simpler: https://github.com/foxglove/ws-protocol/pull/269 (This is helped by the fact that the FileDescriptorSet.proto is shipped with google's protobuf libraries.)  Is there some equivalent we can do with flatbuffers?

jtbandes:
> Just updated to have the enums included in their parent table files again. I think this makes a bit more sense going forward and prevents empty files being generated for the enums that were alone in their schema files.

Nice. Does this mean we do not support an enum being used in two different message schemas? For a moment I thought we already had that with NumericType, but I realized it is nested under the PackedElementField message and that gets used from both Grid and PointCloud. If it's not supported then maybe we should add an assertion in the unit tests to check that an enum only gets referenced by its parent message type and no others.

jameskuszmaul-brt:
> Do the binary schemas include all the dependencies so they are fully self-contained?

They are fully self-contained. 

> How does one create a merged binary schema from the individual .fbs files?

You don't merge them. If you, e.g., wanted to include to disparate types in the same binary you'd have to wrap the normal reflection object into a vector or something else custom to represent the list. Since MCAP only needs the individual schemas for each channel, you don't need to merge things.

> 
> With Protobuf we originally recommended people use `protoc --include_imports --descriptor_set_out` to create the merged binary schemas, but over time we have moved toward recommending they generate them at runtime, since it's simpler: [foxglove/ws-protocol#269](https://github.com/foxglove/ws-protocol/pull/269) (This is helped by the fact that the FileDescriptorSet.proto is shipped with google's protobuf libraries.) Is there some equivalent we can do with flatbuffers?

There isn't a great equivalent (that I'm aware of) for flatbuffers (for some reason flatbuffers has two different reflection mechanisms, one of which produces the nice easy-to-store object we use here; the other which is more useful at runtime with generated code).

jtbandes:
> Since MCAP only needs the individual schemas for each channel, you don't need to merge things.

I guess my question was, what processes the `include` directives? And I suppose the answer is `flatc` does. So if the generated code doesn't have a way to build that dynamically then we'll have to recommend people create the .bfbs themselves.

I guess storing the .bfbs in the repo could make sense. We never stored the protobuf binary FileDescriptorSets here because it seemed normal for people to copy/paste .proto files into their project and integrate with their own build system. And also because there's no standard file extension for a binary proto schema. But since the .bfbs extension is standard maybe it would make sense here. cc @defunctzombie @amacneil for thoughts.

jameskuszmaul-brt:
> > Since MCAP only needs the individual schemas for each channel, you don't need to merge things.
> 
> I guess my question was, what processes the `include` directives? And I suppose the answer is `flatc` does. So if the generated code doesn't have a way to build that dynamically then we'll have to recommend people create the .bfbs themselves.

Yeah, `flatc` processes the include directives. If people are creating topics that are just the type in question, then generating the `.bfbs` could be helpful, but if they are using these types as fields within their messages, it isn't helpful.

jtbandes:
Currently Studio would only support the former, we don't support pulling out visualizable types from a nested field (see https://github.com/foxglove/studio/issues/2410)

defunctzombie:
> But since the .bfbs extension is standard maybe it would make sense here. cc @defunctzombie @amacneil for thoughts.

My thoughts immediately go to a miss-match between the `bfbs` file and the sibling flatbuffer definition file and it not being obvious. This is avoided by generating these on-demand or at runtime like we do for protos.

snosenzo:
> at runtime like we do for protos.

The only thing there is I don't think there's really a way to generate these flatbuffer schema files in Typescript at runtime at least not in the same way that the[ bag2mcap example](https://github.com/foxglove/mcap/blob/main/typescript/examples/bag2mcap/scripts/bag2mcap.ts). On demand beforehand with `flatc`, certainly. But as far as I can see in the docs there's no way to compile the flatbuffer schemas to binary using typescript via the API. 

Aside from that if they're writing or reading these from these schemas, they'll also need the generated readers/writers for their specific language if they're doing something other than parsing to/from JSON. I was going to say I don't want to push the responsibility of keeping the `bfbs` files in sync onto the consumer, but they're most likely going to have to be doing other `flatc` cli commands to compile these schemas to their chosen language anyway. So It probably doesn't make sense to have the `bfbs` files in the repo.

defunctzombie:
> So It probably doesn't make sense to have the `bfbs` files in the repo.

Ok. Let's start without the bfbs files. We can always revisit this decision once we have more information from users.

snosenzo:
This should be ready for a final pass now.

internal/generateFlatbufferSchema.test.ts
```diff
@@ -0,0 +1,118 @@
+import { generateFlatbuffers } from "./generateFlatbufferSchema";
+import { exampleEnum, exampleMessage } from "./testFixtures";
+
+describe("generateFlatBuffer", () => {
```
jtbandes, line 4:
```suggestion
describe("generateFlatbuffers", () => {
```

internal/generateFlatbufferSchema.ts
```diff
@@ -0,0 +1,165 @@
...
+}
+root_type ByteVector;
+`;
+
+// Same as protobuf wellknown types
+export const TIME_FB = `
+namespace foxglove;
+
+struct Time {
+  /// 	Represents seconds of UTC time since Unix epoch 1970-01-01T00:00:00Z
```
jtbandes, line 19:
```suggestion
  /// Represents seconds of UTC time since Unix epoch 1970-01-01T00:00:00Z
```

internal/generateFlatbufferSchema.ts
```diff
@@ -0,0 +1,165 @@
...
+  sec:uint64;
+  /// Nano-second fractions from 0 to 999,999,999 inclusive
+  nsec:uint32;
```
jtbandes, line 22: I don't think we want uint64+uint32 – the other things I've seen are either a single uint64, or two (u)int32s

internal/types.ts
```diff
@@ -30,6 +30,7 @@ export type FoxgloveMessageField = {
   required?: true;
   description: string;
   protobufFieldNumber?: number;
+  defaultValue?: string;
```
jtbandes, line 33: I think the default value for numeric fields probably shouldn't be a string, it should be the actual type of the field (`number` in the cases you've added, although I suppose we could have defaults for string fields). Then it can be up to each generator to know how to translate a number into valid syntax in its own language.

scripts/updateGeneratedFiles.ts
jtbandes: seems like this wasn't addressed yet

internal/types.ts
snosenzo: yeah I honestly am not sure what would be the right way to type that aligned with the `type` type, without putting it inside of that, but maybe that makes more sense since we might not want defaults of nested schemas?

jtbandes (approved):

PR #72, opened 2022-11-14
Add documentation about supported field values for image-related topics

jtbandes:
**Public-Facing Changes**
Added information about Foxglove Studio's supported values for CameraCalibration.distortion_model, RawImage.encoding, and CompressedImage.format.

**Description**
Closes https://github.com/foxglove/schemas/issues/32

### DIFF
--- a/internal/schemas.ts
+++ b/internal/schemas.ts
@@ -679,7 +679,7 @@ const CameraCalibration: FoxgloveMessageSchema = {
       name: "distortion_model",
       type: { type: "primitive", name: "string" },
       description:
-        "Name of distortion model\n\nValues supported by Foxglove Studio: `plumb_bob` and `rational_polynomial`",
+        "Name of distortion model\n\nSupported values: `plumb_bob` and `rational_polynomial`",
     },
     {
       name: "D",
@@ -771,7 +771,7 @@ const CompressedImage: FoxgloveMessageSchema = {
     {
       name: "format",
       type: { type: "primitive", name: "string" },
-      description: "Image format\n\nValues supported by Foxglove Studio: `webp`, `jpeg`, `png`",
+      description: "Image format\n\nSupported values: `webp`, `jpeg`, `png`",
     },
   ],
 };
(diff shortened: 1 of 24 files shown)


jtbandes:
Personally not a huge fan of having this live in the schemas, but it is certainly much easier than the approach where it lives alongside the docs generation and has to be slotted in the right places.

defunctzombie (approved):
I was originally 👎 on this being in the schemas since this becomes yet another place for us to update the supported values from Studio but since the schemas are _for_ studio use then maybe it isn't so crazy. Overall I am 👎 on this approach but am approving because I cannot recommend a better alternative if the goal is to document this information with the schemas. I think the schemas should do a better job of saying you should consult the panel docs for supported values.

internal/schemas.ts
```diff
@@ -678,7 +678,8 @@ const CameraCalibration: FoxgloveMessageSchema = {
     {
       name: "distortion_model",
       type: { type: "primitive", name: "string" },
-      description: "Name of distortion model",
+      description:
+        "Name of distortion model\n\nValues supported by Foxglove Studio: `plumb_bob` and `rational_polynomial`",
```
amacneil, line 682: _"Values supported by Foxglove Studio"_ implies to me that these schemas are some general purpose framework that is incidentally used by Foxglove Studio. Imo that is backwards, these schemas are the primary API for rendering things in Foxglove Studio, so it's weird using language as if they are a separate product.

So I would have said something like: 

```suggestion
        "Name of distortion model\n\nSupported values: `plumb_bob`, `rational_polynomial`",
```

amacneil (commented):
Added a comment above if we are going with this method.

That said, I'll accept I am in the minority here and everyone else seems to want these to live in the panel docs, so @defunctzombie please make a final call on this.

jhurliman:
I'm with @amacneil on this; these schemas are hyper-specific to Studio by design and don't make sense outside of the context of Studio. Putting Studio-specific documentation into the schema docs seems natural to me, in the same way the Studio extension API will include Studio-specific documentation.

amacneil:
product review: ship it

amacneil:
@jtbandes I assume there is a follow up PR needed to regen the docs in our website repo?

jtbandes:
https://github.com/foxglove/website/pull/853

PR #80, opened 2023-01-09
Add docs links for certain schema field descriptions

esthersweon:
**Public-Facing Changes**

- Add docs links for certain schema field descriptions 
- Migrated from https://github.com/foxglove/schemas/pull/60


amacneil (changes requested):
We had a long discussion about the various options for documenting enum values, and settled on https://github.com/foxglove/schemas/pull/72 (putting the actual values in the schema docs, rather than linking to an external docs page).

PR #83, opened 2023-02-14
Update descriptions for PointsAnnotationType and LineType

jtbandes:
**Public-Facing Changes**
Better docs for PointsAnnotationType and LineType

**Description**
Closes FG-1377


internal/schemas.ts
```diff
@@ -357,9 +357,13 @@ const LineType: FoxgloveEnumSchema = {
   protobufEnumName: "Type",
   description: "An enumeration indicating how input points should be interpreted to create lines",
   values: [
-    { value: 0, name: "LINE_STRIP", description: "0-1, 1-2, ..., (n-1)-n" },
-    { value: 1, name: "LINE_LOOP", description: "0-1, 1-2, ..., (n-1)-n, n-0" },
-    { value: 2, name: "LINE_LIST", description: "0-1, 2-3, 4-5, ..." },
+    {
+      value: 0,
+      name: "LINE_STRIP",
+      description: "Connected line segments: 0-1, 1-2, ..., (n-1)-n",
```
esthersweon, line 363: It's somewhat confusing that the `-` in `(n-1)` means something different from the `-` in the `-n` suffix, but not sure how to edit..

esthersweon: Maybe just ending with the `...`, like in the `LINE_LIST` description?

internal/schemas.ts
```diff
@@ -357,9 +357,13 @@ const LineType: FoxgloveEnumSchema = {
...
+    },
+    { value: 1, name: "LINE_LOOP", description: "Closed polygon: 0-1, 1-2, ..., (n-1)-n, n-0" },
```
esthersweon, line 365: Maybe just remove the `(n-1)-n` one here?

esthersweon (approved):
Left some minor comments.

jtbandes:
I do think the n-1 provides some helpful context even though the formatting is a bit confusing. I think I'll lean towards keeping it.

PR #84, opened 2023-02-15
[Python] Add foxglove-schemas-flatbuffer package

jhurliman:
**Public-Facing Changes**
- [Python] New foxglove-schemas-flatbuffer package exporting Python classes for Flatbuffer types and a get_schema() method for retrieving binary .bfbs schemas

**Description**
This closely resembles the foxglove-schemas-protobuf package

### DIFF
--- a/python/foxglove-schemas-flatbuffer/README.md
+++ b/python/foxglove-schemas-flatbuffer/README.md
@@ -1,0 +1,23 @@
+# Foxglove Schemas (Flatbuffer)
+
+This package provides [Flatbuffer](https://google.github.io/flatbuffers/) classes for [Foxglove Schemas](https://foxglove.dev/docs/studio/messages/introduction).
+
+## Installation
+
+Install via [Pipenv](https://pipenv.pypa.io/en/latest/) by adding `foxglove-schemas-flatbuffer` to your `Pipfile` or via the command line:
+
+```bash
+pipenv install foxglove-schemas-flatbuffer
+```
+
+## Usage
+
+Import types from the `foxglove` module as follows:
+
+```py
+from foxglove_schemas_flatbuffer.CompressedImage import CompressedImage
+```
+
+## Stay in touch
+
+Join our [Slack channel](https://foxglove.dev/join-slack) to ask questions, share feedback, and stay up to date on what our team is working on.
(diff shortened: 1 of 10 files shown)


python/foxglove-schemas-flatbuffer/foxglove_schemas_flatbuffer/__init__.py
```diff
@@ -0,0 +1,5 @@
+import importlib.resources as pkg_resources
+
+
+def get_schema(name: str) -> bytes:
+    return pkg_resources.read_binary(__package__, name + ".bfbs")
```
jtbandes, line 5: Docs say this function is deprecated

python/foxglove-schemas-flatbuffer/README.md
```diff
@@ -0,0 +1,23 @@
...
+```bash
+pipenv install foxglove-schemas-flatbuffer
+```
+
+## Usage
+
+Import types from the `foxglove` module as follows:
+
+```py
+from foxglove_schemas_flatbuffer.CompressedImage import CompressedImage
```
jtbandes, line 18: One thing I've been wanting to fix about the protobuf schemas is that I think you should be allowed to do `from foxglove_schemas_X import CompressedImage` without the extra nested module. Maybe you can find a way to make that work here and we can use the same approach for protobuf?

jhurliman: That unfortunately won't work with flatbuffers since each generated file contains `def Start(builder):` and `def End(builder):` which would collide if everything was merged into the same module.

jtbandes: what do you mean by "merged"? I was thinking we would create an `__init__.py` which has a line like `from foxglove_schemas_flatbuffer.CompressedImage import CompressedImage` for each schema. Then importing from the package root would import from the set of names that were imported to this `__init__.py`...unless I'm misunderstanding how python imports work.

jhurliman: Ah I see, that would help for the reading side (which uses the `CompressedImage` class), just not the writing side (which relies on free functions exported from `CompressedImage.py`).

jtbandes: 👍 maybe it makes more sense for protobuf than for flatbuffers then.

PR #90, opened 2023-02-21
Add `zfp` as a supported CompressedImage format

jhurliman:
**Public-Facing Changes**
- `zfp` is documented as a valid format for CompressedImage messages. The schema docs describe the constraints (header required, data must be 2D)

### DIFF
--- a/internal/schemas.ts
+++ b/internal/schemas.ts
@@ -785,7 +785,8 @@ const CompressedImage: FoxgloveMessageSchema = {
     {
       name: "format",
       type: { type: "primitive", name: "string" },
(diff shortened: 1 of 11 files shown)


defunctzombie (changes requested):
Holding per comments in: https://github.com/foxglove/website/pull/997#issuecomment-1441214313

defunctzombie:
Closing this. Not currently planning to add zfp as a compressed image format sibling to png and jpeg. We might re-visit this in the future as a new field for generic compression of the image data or additional parts of a string on top of the format field that is meant to have well-known compressed image formats.

PR #92, opened 2023-03-02
Add units to PointsAnnotation and CircleAnnotation

defunctzombie:
I'm a little meh on this change since we already say `2D image space` and a `Point2d` has `x, y` fields.

@jtbandes I think I need to run some script to re-generate the schemas?

### DIFF
--- a/internal/schemas.ts
+++ b/internal/schemas.ts
@@ -1040,17 +1040,17 @@ const CircleAnnotation: FoxgloveMessageSchema = {
     {
       name: "position",
       type: { type: "nested", schema: Point2 },
-      description: "Center of the circle in 2D image coordinates",
+      description: "Center of the circle in 2D image coordinates ([0,w] x [0,h] pixels)",
     },
     {
       name: "diameter",
       type: { type: "primitive", name: "float64" },
-      description: "Circle diameter",
+      description: "Circle diameter in pixels",
     },
     {
       name: "thickness",
       type: { type: "primitive", name: "float64" },
-      description: "Line thickness",
+      description: "Line thickness in pixels",
     },
     {
       name: "fill_color",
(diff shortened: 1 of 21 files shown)


internal/schemas.ts
```diff
@@ -1040,17 +1040,17 @@ const CircleAnnotation: FoxgloveMessageSchema = {
     {
       name: "position",
       type: { type: "nested", schema: Point2 },
-      description: "Center of the circle in 2D image coordinates",
+      description: "Center of the circle in 2D image coordinates ([0,w] x [0,h] pixels)",
```
jtbandes, line 1043: maybe just `(pixels)`?

internal/schemas.ts
```diff
@@ -1102,7 +1102,7 @@ const PointsAnnotation: FoxgloveMessageSchema = {
     {
       name: "points",
       type: { type: "nested", schema: Point2 },
-      description: "Points in 2D image coordinates",
+      description: "Points in 2D image coordinates ([0,w] x [0,h] pixels)",
```
jtbandes, line 1105: maybe just `(pixels)`?

jtbandes (approved):

jhurliman:
Some notes:
- `2D image space` describes the space, not the units used in the space. This PR is a useful clarification.
- `[0, w]` is not correct for two reasons: 1. if you want to restrict the units to clip inside the image, it should be the half-open interval `[0, w)` and 2. neither the schema nor Studio place the constraint of clipping inside the image. `<-10, -10>` is a perfectly fine coordinate, for example.

defunctzombie:
> * `[0, w]` is not correct for two reasons: 1. if you want to restrict the units to clip inside the image, it should be the half-open interval `[0, w)` and 2. neither the schema nor Studio place the constraint of clipping inside the image. `<-10, -10>` is a perfectly fine coordinate, for example.

Good to know. I will amend the PR to `(pixels)` and we can roll with that. If there is further confusion in the future it is easy to revise with more information.

internal/schemas.ts
defunctzombie, line 1043:
```suggestion
      description: "Center of the circle in 2D image coordinates (pixels)",
```

internal/schemas.ts
defunctzombie, line 1105:
```suggestion
      description: "Points in 2D image coordinates (pixels)",
```

PR #95, opened 2023-03-08
Remove "all float64s" comment from PointCloud schema

defunctzombie:
These fields do not need to be float 64s, they can also be float 32. This confused a customer who thought they had to send float64.

@jtbandes do we support the other field types or only float32/64? If we don't support the other field types we should remove them.


snosenzo:
packed element field supports all of the numericEnum types technically. We don't do anything in studio to make sure that they are a specific numeric type from the looks of it

jtbandes (approved):

PR #98, opened 2023-03-24
add FrameTransforms Message

snosenzo:
FG-2048


internal/schemas.ts
```diff
@@ -869,6 +869,20 @@ const FrameTransform: FoxgloveMessageSchema = {
   ],
 };
 
+const FrameTransformList: FoxgloveMessageSchema = {
+  type: "message",
+  name: "FrameTransformList",
```
defunctzombie, line 874: Do we like the `List` suffix naming convention over the pluralized naming convention for this message? pro/cons?

snosenzo: I thought list would be a bit clearer? I saw `List` on `TriangleList`,  so that's where I went first, but I also now see `PosesInFrame` (and `PoseInFrame`) and `ImageAnnotations`, etc. If we already have that pluralized form in other messages, I think I prefer to change this to `FrameTransforms` as the name. I think the `TriangleList` has  bit more baggage since it kinda references rendering in an instanced way (LINE_LIST as drawMode)

jtbandes (approved):

PR #102, opened 2023-04-09
Add TextAnnotation

### LINKED ISSUE #101: Add `TextAnnotation`
**Description**

Foxglove Studio currently supports 2D text annotations via an unofficial extension to `visualization_msgs/ImageMarker`. Bringing support to foxglove schemas would allow us to drop this ROS message.

jhurliman:
### Public-Facing Changes
- Adds `TextAnnotation` schema for 2D text annotations and `labels` field to `ImageAnnotations`

### Description

Fixes #101


internal/schemas.ts
```diff
@@ -1144,6 +1144,45 @@ const PointsAnnotation: FoxgloveMessageSchema = {
...
+  fields: [
+    {
+      name: "timestamp",
+      type: { type: "primitive", name: "time" },
+      description: "Timestamp of annotation",
+    },
+    {
+      name: "position",
+      type: { type: "nested", schema: Point2 },
+      description: "Center of the circle in 2D image coordinates (pixels)",
```
snosenzo, line 1160: This needs to be updated to specify where on the text annotation this is positioning by ie what's the anchor used to position the text

internal/schemas.ts
```diff
@@ -1161,6 +1200,12 @@ const ImageAnnotations: FoxgloveMessageSchema = {
       description: "Points annotations",
       array: true,
     },
+    {
+      name: "labels",
```
jtbandes, line 1204:
```suggestion
      name: "texts",
```
Matches SceneEntity.

internal/schemas.ts
snosenzo: Also on second look, should this use normalized image coordinates ([0-1])so that it doesn't depend on the resolution of the image?

jhurliman: @snosenzo if you want to make a breaking change to all of the image annotations I'm ok with that. But it would be confusing to have one annotation use normalized coordinates and the rest use pixel coordinates.

jhurliman, line 1160:
```suggestion
      description: "Bottom-left origin of the text label in 2D image coordinates (pixels)",
```

snosenzo (approved):

jtbandes (approved):

snosenzo:
Going to test this in studio with implementation before merging in case we need to change anything

snosenzo:
tested in studio with implementation, everything seems to be good. PR here with changes: https://github.com/foxglove/studio/pull/5789

 will land and deploy this now

PR #107, opened 2023-05-05
Extending PointsAnnotation and CircleAnnotation with object_id

markoSyrmia:
I have added object_id in PointsAnnotation and CircleAnnotation.

### Description
I need this change because i want to have  possibility to send object id through protobuf message, in order  to be able to display it on ImageView. This is only the first part, the second is contained in foxglove/studio repo.

I will post here link of PR (the second part from foxglove/studio repo), for better understanding why i need this changes. 
https://github.com/foxglove/studio/pull/5921

### DIFF
--- a/internal/schemas.ts
+++ b/internal/schemas.ts
@@ -1076,6 +1076,11 @@ const CircleAnnotation: FoxgloveMessageSchema = {
       type: { type: "nested", schema: Color },
       description: "Outline color",
     },
+    {
+      name: "object_id",
+      type: { type: "primitive", name: "string" },
+      description: "Id of circle",
+    },
   ],
 };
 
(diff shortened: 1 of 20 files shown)


jtbandes:
Hi,
What is the intended use case for the object id? I noticed that your other PR #109 references the object id being a simple label like "car" or "truck". Is it just a visual label, or do you want to tie the identifier to a particular instance of data?

markoSyrmia:
To an extent the answer is Yes on bot counts. We wanted to implement the ability to "render" the object ID on screen, and also have it share object ID with other representations of the data in other views. So that the user can identify that the rectangle in Image view is the same object as a cube in 3D view.

jtbandes:
For rendering the object ID on screen, have you tried TextAnnotation?
For displaying additional data, would adding a `metadata` field (similar to SceneEntity) work for your use case?

jtbandes:
We believe that TextAnnotations are the best way to achieve this feature.

PR #109, opened 2023-05-16
Extending SceneEntity with object_id and pictogramUrl

markoSyrmia:
I have added  object_id and pictogramUrl in SceneEntity.

### Description
I need this change because i want to have possibility to send object id and pictogram URL through protobuf message, in order to be able to display it on 3DView.This pictogram will closer describe object in 3DView(e.g. one image of car for all type of cars in view). The same is for object id(e.g. text  car for all type of cars, text truck for all type of truck)  This is only the first part, the second (displaying on 3DView) is contained in foxglove/studio repo.

### DIFF
--- a/internal/schemas.ts
+++ b/internal/schemas.ts
@@ -573,6 +573,16 @@ const SceneEntity: FoxgloveMessageSchema = {
         "Identifier for the entity. A entity will replace any prior entity on the same topic with the same `id`.",
     },
     {
+      name: "object_id",
+      type: { type: "primitive", name: "string" },
+      description: "id of object ",
+    },
+    {
+      name: "pictogramUrl",
+      type: { type: "primitive", name: "string" },
+      description: "Url of pictogram",
+    },
+    {
       name: "lifetime",
       type: { type: "primitive", name: "duration" },
       description:
--- a/ros_foxglove_msgs/ros1/SceneEntity.msg
+++ b/ros_foxglove_msgs/ros1/SceneEntity.msg
@@ -12,6 +12,12 @@ string frame_id
 # Identifier for the entity. A entity will replace any prior entity on the same topic with the same `id`.
 string id
 
+# id of object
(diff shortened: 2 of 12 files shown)


jtbandes:
Thanks for your contribution.

Regarding `object_id`: Are you intending for this to act as an identifier for the object, or to be displayed as text? Please note that the [SceneEntity type](https://foxglove.dev/docs/studio/messages/scene-entity) already has an `id` field. Additionally, the SceneEntity can contain a [TextPrimitive](https://foxglove.dev/docs/studio/messages/text-primitive) to display text. Are these existing features not sufficient for your use case?

Regarding `pictogramUrl`: We have heard some similar requests for "icons" in the past. However, we are not sure that adding a field directly to SceneEntity is the way we want to approach this problem. For example, another approach might be to add it as a new type of "primitive" that can be part of the scene entity. We've made a note of this request in our product feedback and will continue to gather input from other users and customers, but we don't have any ETA to share for the feature at this time.

markoSyrmia:
Regarding object_id:Yes we saw late that this was updated. When we started our implementation this feature was missing. This part does feel redundant now.

Regarding pictogramUrl: I am not sure I completely understand what You mean. I could create a new primitive to hold this information (or implement this in any other way You'd suggest), but are You saying that I should drop this entirely? If You already have a lot of interest regarding the "icons" I figure You'd welcome that someone offers to implement the requested feature… I am really keen on contributing to this community and would like to implement this feature as it looks like an interesting challenge…

jtbandes:
We have heard some anecdotal requests for the feature, but don't have any detailed requirements or examples of how the feature could work. I'm glad you are interested in contributing, but I believe it would be better to start with a [discussion](https://github.com/orgs/foxglove/discussions) to gather input and ideas from others too. I'll close this PR because we are not ready to accept this `pictogramUrl` change as written.

PR #110, opened 2023-05-23
Feature/conan package proto files

kevswims:
### Public-Facing Changes

None

### Description

Adds a conanfile and cmake project to compile the protobuf code and package it. 

We are using this in a couple of ways:

- Dependent code such as (https://github.com/foxglove/ws-protocol/tree/main/cpp/examples) no longer has to compile the protobuf files. They can just consume the package with conan.
- The cmake project is being used standalone from a Yocto Bitbake recipe to package the proto files for consumption as a yocto package.

### TODO:

- [ ] Document the CMakeLists.txt
- [ ] Add correct license, author, etc into the conanfile
- [ ] Update the example (https://github.com/foxglove/ws-protocol/tree/main/cpp/examples) to consume this package instead of the copied proto files.

Questions:

- How does Foxglove typically contribute packages to conan center?


jtbandes:
Hi, thanks for your contribution!

We've had a good experience so far with publishing Python packages for the schemas ([foxglove-schemas-protobuf](https://pypi.org/project/foxglove-schemas-protobuf/) and [foxglove-schemas-flatbuffer](https://pypi.org/project/foxglove-schemas-flatbuffer/)), making it easy to get up and running with the schemas in a new Python project. So it makes sense to try and offer a good experience for C++ users in this way as well. We would probably want the name of the package to be protobuf-specific, since the schemas can also be used with other encodings like flatbuffers.

It's my understanding that publishing a package via Conan usually requires making a PR to [conan-io/conan-center-index](https://github.com/conan-io/conan-center-index). For example, here's our [`mcap` recipe](https://github.com/conan-io/conan-center-index/blob/master/recipes/mcap/all/conanfile.py) and our [`foxglove-websocket` recipe](https://github.com/conan-io/conan-center-index/blob/master/recipes/foxglove-websocket/all/conanfile.py). We've followed their [contributing guide](https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/README.md) in the past in order to set up these packages. Do you think that making a PR to conan-center-index would be a better first step to making this available? (Off the top of my head I'm not sure whether any changes in this repo would be required to get it working.)

We do sometimes use conan in our source repos in order to build and test the packages (for example: https://github.com/foxglove/mcap/blob/main/cpp/build.sh) but I don't think it's required, though of course there is value in setting up CI checks to ensure as best we can that the published conan package would continue to work as we make changes to the source repo.

jtbandes:
Makes sense – do you think the conanfile needs to be included in this PR if it's primarily going to live in the conan-center-index?

I do think we'd want some form of CI testing in order to accept the CMakeLists being added here.

I'm actually not totally convinced the CMakeLists should live here at all. Right now we have nothing in the repo specific to C++. We have some Python-specific stuff because the pypi package is actually published from this repo. But if the conan package lives in the conan-center-index repo, maybe the CMakeLists should live there too? What do you think? I don't have a lot of experience with other conan packages to know how they handle this.

jtbandes:
What do you think of naming it `foxglove-schemas-protobuf` to match the python package? And I think it's nice because the repo name is `foxglove/schemas`.

kevswims:
That works for me. I'll update the Conan center pr tomorrow with the new
name.

jtbandes:
Hey, what's the latest on this? Still waiting for review on the conan-center-index PR?

kevswims:
@jtbandes Yeah, that PR is still pending. I got everything on it building so I think it should be good to go. It just can take a while to get through the review queue.

jtbandes:
It looks like things are moving along on the conan center PR 🎉 – so I'm going to close this ticket for now to indicate that our team doesn't have any more work planned to support this.

Once that package is published, I think a good action item would be to update the [websocket protocol example c++ server](https://github.com/foxglove/ws-protocol/tree/main/cpp/examples) to use the new package instead of having the proto files and CMake stuff duplicated in that repo.

It might also make sense to update the [MCAP docs](https://mcap.dev/guides/cpp/protobuf) and/or [Studio docs](https://foxglove.dev/docs/studio/messages/introduction) to include some info about how to get set up with C++ & protobuf.

Feel free to let us know if you have any other desired outcomes beyond the publishing of the package to conan-center. Thanks for working on this!

PR #119, opened 2023-07-07
Add typescript schema export utility

foxymiles:
### Public-Facing Changes
None

### Description
Add typescript schema export utility. This will be used in studio to import schema definitions in order to make them available for use in user scripts.


internal/exportTypescriptSchema.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+  const allSchemaNames = [
+    ...Object.values(foxgloveMessageSchemas),
+    ...Object.values(foxgloveEnumSchemas),
+  ].sort((a, b) => a.name.localeCompare(b.name));
+  let indexTS = "";
+  for (const schema of allSchemaNames) {
+    indexTS += `export * from "./${schema.name}";\n`;
+  }
+  schemas["index"] = indexTS;
```
jtbandes, line 26: I think this also needs to include `TIME_TS` and `DURATION_TS`.

Could we also use this new function [here](https://github.com/foxglove/schemas/blob/main/scripts/updateGeneratedFiles.ts#L106) to reduce duplication?

internal/exportTypescriptSchema.test.ts
```diff
@@ -0,0 +1,8 @@
+import { exportTypescriptSchema } from "./exportTypescriptSchema";
+
+describe("exportTypescriptSchema", () => {
+  it("exports schemas", () => {
+    const schemas = exportTypescriptSchema();
+    expect(schemas["ArrowPrimitive"]).not.toBeNull();
```
jtbandes, line 6: Thanks for adding a test! What do you think of using `expect.toMatchSnapshot()` instead?

foxymiles: Would we have to update the snapshot every time we add new schema?

jtbandes: Yeah but we already update all the [generated files](https://github.com/foxglove/schemas/tree/main/schemas) so I don't think it's a big deal

jtbandes: On the other hand, if we use this function to generate those files, then maybe having a test is redundant.

foxymiles: I've updated this to use a snapshot.

internal/exportTypescriptSchema.ts
```diff
@@ -0,0 +1,34 @@
+import { DURATION_TS, TIME_TS, generateTypeScript } from "./generateTypeScript";
+import { foxgloveEnumSchemas, foxgloveMessageSchemas } from "./schemas";
+
+/**
+ * Export typescript schema as source, keyed by the schema name.
+ *
+ * @returns a record of schema name => schema source.
+ */
+export function exportTypescriptSchema(): Record<string, string> {
```
jtbandes, line 9:
```suggestion
/**
 * Export typescript schemas as source, keyed by the file name.
 *
 * @returns a record of schema name => schema source.
 */
export function exportTypeScriptSchemas(): Record<string, string> {
```

I would go with "file name" because `index` is not really a "schema name". This also makes me wonder if we should add `".ts"` suffix to the keys as well?

Maybe use a Map?

internal/exportTypescriptSchema.ts
jtbandes, line 1: - I think you'll need to export this from `internal/index.ts`
- also would rename file to `exportTypeScriptSchemas.ts`

jtbandes (approved):
Left some final nits and suggestions but the main logic looks good to me. Thanks!

internal/exportTypescriptSchema.ts
foxymiles: > * I think you'll need to export this from `internal/index.ts`

👍 

> * also would rename file to `exportTypeScriptSchemas.ts`

The other files in this directory use the singular so I stuck with that.

jtbandes: The other functions also tend to operate on just 1 schema at a time

jtbandes:
@foxymiles "schemata" is a cute spelling...but the name of this repo is foxglove/schemas, so I think we should probably stick with that plural form 😬 

image

PR #124, opened 2023-07-28
Add CompressedVideo schema

defunctzombie:
Alternative to https://github.com/foxglove/schemas/pull/114 for https://github.com/foxglove/studio/pull/6543


internal/schemas.ts
```diff
@@ -790,6 +790,36 @@ const CompressedImage: FoxgloveMessageSchema = {
...
 
+const CompressedVideo: FoxgloveMessageSchema = {
+  type: "message",
+  name: "CompressedVideo",
+  description: "A single frame of a compressed video bitstream",
+  fields: [
+    {
+      name: "timestamp",
+      type: { type: "primitive", name: "time" },
+      description: "Timestamp of image",
```
jtbandes, line 801: Why do we keep the "image" terminology here? Should this be "video frame" instead or something like that?

internal/schemas.ts
```diff
@@ -790,6 +790,36 @@ const CompressedImage: FoxgloveMessageSchema = {
...
+        "Compressed video frame data. For packet-based video codecs this data must begin and end on packet boundaries (no partial packets), and must contain enough video packets to decode exactly one image (either a keyframe or delta frame).",
+    },
+    {
+      name: "format",
+      type: { type: "primitive", name: "string" },
+      description: "Video format\n\nSupported values: `h264`",
+    },
+  ],
+};
+
```
pezy, line 822: Why are `keyframe` and `metadata` no longer needed here?

snosenzo: these are no longer included because people often come from already having compressed video formatted streams and it would be too much of a heavylift for them to parse out this format-specific information and write it back to the message in a more accessible way. All of this information is included in the formatted `data` byte array

PR #126, opened 2023-08-25
Generate .idl schemas

jtbandes:
Adds OMG IDL definitions for Foxglove schemas.


internal/testFixtures.ts
```diff
@@ -17,8 +17,8 @@ export const exampleEnum: FoxgloveEnumSchema = {
   parentSchemaName: "ExampleMessage",
   description: "An example enum",
   values: [
-    { name: "A", value: 1, description: "Value A" },
-    { name: "B", value: 2, description: "Value B" },
+    { name: "A", value: 0, description: "Value A" },
```
snosenzo, line 20: Is it worth putting a comment here that these need to start at 0 for omgidl enum support?

internal/generateOmgIdl.test.ts
```diff
@@ -0,0 +1,183 @@
...
+        boolean field_boolean;
+
+        // bytes field
+        sequence<uint8> field_bytes;
+
+        // float64 field
+        double field_float64;
+
+        // uint32 field
+        uint32 field_uint32;
```
snosenzo, line 56: You could support default values for primitives (except string) with `@default(value=<val>)`. not sure if this was an intentional omission.

internal/generateOmgIdl.ts
```diff
@@ -0,0 +1,111 @@
...
+    case "boolean":
+      return "boolean";
+    case "float64":
+      return "double";
+    case "uint32":
+      return "uint32";
+  }
+}
+
+export const TIME_IDL = `struct Time {
```
snosenzo, line 18: should we namespace Time and Duration to `foxglove::`? It looks like we do that for flatbuffers

amacneil:
Not sure if there is anything specific you would like me to review, but I took a look at a few of the generated schemas and they seem good.

internal/generateOmgIdl.test.ts
jtbandes: not intentional, I just forgot/didn't realize that is a thing 😄

internal/testFixtures.ts
jtbandes: I think it's fine without since the assertion will catch it if this were to change

snosenzo (approved):

snosenzo (approved):

PR #128, opened 2023-09-22
Add includeTypedArrays option to generateTypeScript

jtbandes:
### Public-Facing Changes

Added an `includeTypedArrays` option to `generateTypeScript` to allow generating type definitions that include TypedArrays in addition to plain `number[]` arrays.

### Description

Relates FG-4976
See also https://github.com/foxglove/studio/pull/6890


internal/generateTypeScript.ts
```diff
@@ -26,7 +41,15 @@ export const DURATION_TS = `export type Duration = {
 };
 `;
 
-export function generateTypeScript(schema: FoxgloveSchema): string {
+export type GenerateTypeScriptOptions = {
+  /** Include TypedArray types on numeric array fields, e.g. `number[] | Uint32Array` */
```
snosenzo, line 45: Is it worth mentioning when it might be important to use this?

jtbandes: updated comment

snosenzo (approved):

PR #132, opened 2023-10-10
Update compressedimage format comment

snosenzo:
Studio technically support more than only those 3 strings. So we should update the comment to reflect that

### DIFF
--- a/internal/schemas.ts
+++ b/internal/schemas.ts
@@ -785,7 +785,8 @@ const CompressedImage: FoxgloveMessageSchema = {
     {
       name: "format",
       type: { type: "primitive", name: "string" },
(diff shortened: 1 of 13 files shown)


internal/schemas.ts
```diff
@@ -785,7 +785,7 @@ const CompressedImage: FoxgloveMessageSchema = {
     {
       name: "format",
       type: { type: "primitive", name: "string" },
-      description: "Image format\n\nSupported values: `webp`, `jpeg`, `png`",
+      description: "Image format\n\nChrome supported media types, such as `webp`, `jpeg`, `png`",
```
jtbandes, line 788:
```suggestion
      description: "Image format\n\nSupported values: image media types supported by Chrome, such as `webp`, `jpeg`, `png`",
```

jtbandes (approved):

PR #133, opened 2023-10-24
Add mention of Annex-B format requirement for h264 encoded data

snosenzo:


banisadr (approved):
Ty

jtbandes (approved):

internal/__snapshots__/exportTypeScriptSchemas.test.ts.snap
```diff
@@ -199,7 +199,7 @@ export type CompressedVideo = {
   /**
    * Video format
    * 
-   * Supported values: \`h264\`
+   * Supported values: \`h264\` (Annex-B formatted data only)
```
defunctzombie, line 202: Seems worth putting this comment on the "data" field? Or both?

defunctzombie (approved):

internal/__snapshots__/exportTypeScriptSchemas.test.ts.snap
snosenzo: I didn't put it on the data field because I don't think h265 or other formats have Annex B. I could put it on both, but we don't really have a good way of differentiating requirements of the data field under other potential formats

snosenzo: I believe at least h265 also has B frames, but yeah I say we come back to this when we add more official docs support

PR #135, opened 2023-10-30
Version update for annex b docs

snosenzo:
- update CompressedVideo docs to mention annexb


jtbandes (approved):
I don’t really understand the benefit of removing the newlines here 🤷

PR #142, opened 2024-02-23
experiment: changing protobuf directory to match Python package name

indirectlylit:
### Public-Facing Changes

None

### Description

Simplify Python protobuf schema generation by matching directory structure to Python package name


indirectlylit:
I briefly tested the generated Python package locally, and this still works:

```python
from foxglove_schemas_protobuf.CompressedImage_pb2 import CompressedImage
```

However I'm not entirely confident this change maintains full backward compatibility with the existing package.

jtbandes:
I see that the imports are now `import "foxglove_schemas_protobuf/foo.proto";`  Is that sort of naming scheme common among third-party protobuf packages? The built-in ones look like `import "google/protobuf/duration.proto";`  so one might expect to see `import "foxglove/protobuf/foo.proto";` — but I'm wondering what is done most commonly in third-party packages?

indirectlylit:
No, I doubt that this is common, and is the reason I didn't make this PR originally: it ties our `*.proto` files to the somewhat arcane Python package naming conventions (to say nothing about the possibility of eventually adopting a global `foxglove` [implicit namespace](https://peps.python.org/pep-0420/)).

Another option might be to generate a secondary set of `*.proto` files specifically for the Python package?

indirectlylit:
Another alternative might be to use https://github.com/danielgtaylor/python-betterproto which I believe provides more flexibility in specifying package names, among other things. (ref: [FG-6783](https://linear.app/foxglove/issue/FG-6783/pythonprotobuf-schema-gen-consider-alternatives-to-protoc-utility))

indirectlylit:
closing experimental PR

PR #143, opened 2024-03-11
Update links and studio references

jtbandes:
Update to latest links and product names.

Includes a pytest downgrade to work around a CI failure https://github.com/ros2/launch/issues/765 (similar to https://github.com/scientific-python/pytest-doctestplus/issues/240)


snosenzo (approved):
instead of just using `Foxglove` should we use `the Foxglove app`? is it confusing distinguishing between Foxglove the organization and Foxglove the app?

PR #145, opened 2024-04-29
Proof of Concept generate at go-protobuf library

jbcpollak:
### Changelog

this sketches out how to build a Go+Protobuf library. To depend on it via `go get github.com/foxglove/schema/go` the .go code would need to be committed to the repo but there are other deployment techniques and I didn't want to be proscriptive. This PR should have enough details to help integrate the process with the build system. I'm happy to make changes if necessary


james-rms:
Thanks for the PoC, this is helpful! we'd need to integrate into our build script(s) and possibly do some refactoring at that point. It wouldn't be fair to drag you through that process, so closing this PR.

PR #147, opened 2024-05-22
Remove melodic ci 

snosenzo:
The ros CI action to use node v20 which can no longer run on this version of ubuntu.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,9 +131,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


defunctzombie:
Is melodic making anymore bloom releases? If it is not then there's little value in continuing to test it and make builds for it.

@achim-k where would we find this information?

achim-k:
It's EOL since about a year (https://wiki.ros.org/Distributions). So you can't make new releases for it.

achim-k (approved):

jtbandes (approved):

snosenzo:
🤦 This wasn't off of `main` it was off of the other PR. Gonna re-do this

PR #148, opened 2024-05-22
Add explicit field IDs to flatbuffer schemas

jameskuszmaul-brt:
### Changelog
Added explicit field IDs to flatbuffer schemas.

### Docs

None

### Description

This allows the schemas to be compiled with `--require-explicit-ids` turned on in `flatc`, which can be nice to help prevent people from accidentally making breaking changes to flatbuffer schemas if they choose to reorder fields (or insert new fields in the middle of the schemas). Because of how flatc works with includes, if downstream users want to use `--require-explicit-ids` on their own schemas that happen to include foxglove schema types, they must update the foxglove schemas that they use to include IDs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
 
       - name: Validate Flatbuffer definitions
         run: |
-          output=$(flatc --ts -o /dev/null ./schemas/flatbuffer/*.fbs)
+          output=$(flatc --require-explicit-ids --ts -o /dev/null ./schemas/flatbuffer/*.fbs)
           if [ -n "$output" ]; then
             echo "::error::Flatbuffer schema compilation had warnings or errors. Fix them to proceed:"
             echo "$output"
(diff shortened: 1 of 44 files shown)


jameskuszmaul-brt:
Waiting on #147 for CI....

jtbandes (approved):
Something worth noting is that we have a mechanism here for specifying field ids explicitly for protobuf (`protobufFieldNumber`). This was added in #61 because we wanted to add a field, but put it in a different order in the IDL, without changing the ids of existing fields.

This isn't currently an issue for flatbuffers because we created these `.fbs` files after making that change (#70) and these ids you've added should correspond with the existing field order today. But anyway, since the Foxglove app uses the binary schemas from the MCAP file, this probably won't matter for our purposes — only for external users who are associating this schema with some pre-existing data.

...and actually, while writing this comment I discovered we added a new field in a different order in https://github.com/foxglove/schemas/pull/108 after the `.fbs` files were first published (cc @defunctzombie). It was probably a miss that we didn't do this field numbering thing back then. (Amusingly, we discussed the issue in https://github.com/foxglove/schemas/pull/108#discussion_r1189254346 for protobuf but forgot about flatbuffers.) Anyway, what's done is done :)

PR #151, opened 2024-05-23
Fix python CI

snosenzo:
Protobuf doesn't seem to have had big enough updates since the last release, so I just incremented the patch version.

Flatbuffers had field numbers added and I felt like that warranted a minor version bump.


jtbandes (approved):

jtbandes:
Reminder to tag these releases after merging so they get pushed to PyPI 🚀

PR #155, opened 2024-11-22
Update CompressedVideo format strings

jtbandes:
Update CompressedVideo documentation for new formats, and describe format and data in more detail

### DIFF
--- a/internal/schemas.ts
+++ b/internal/schemas.ts
@@ -816,8 +816,24 @@ const CompressedVideo: FoxgloveMessageSchema = {
     {
       name: "format",
       type: { type: "primitive", name: "string" },
-      description:
-        "Video format.\n\nSupported values:\n- `h264` (Annex B formatted data only)\n- `h265` (HEVC; Annex B formatted data only)\n- `av1` (Low overhead bitstream format)",
+      description: `Video format.
+
+Supported formats:
+
+- \`h264\`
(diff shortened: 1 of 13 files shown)


jhurliman (approved):

internal/schemas.ts
```diff
@@ -810,15 +810,11 @@ const CompressedVideo: FoxgloveMessageSchema = {
...
       name: "data",
       type: { type: "primitive", name: "bytes" },
-      description:
-        "Compressed video frame data.\n\nFor packet-based video codecs this data must begin and end on packet boundaries (no partial packets), and must contain enough video packets to decode exactly one image (either a keyframe or delta frame). Note: Foxglove does not support video streams that include B frames because they require lookahead.",
-    },
-    {
-      name: "format",
-      type: { type: "primitive", name: "string" },
-      description: `Video format.
+      description: `Compressed video frame data.
```
snosenzo, line 813: Should we point here to the `format` field for more info on what this should look like?

snosenzo (approved):

internal/schemas.ts
jtbandes: I figured since they are right next to each other maybe it’s fine? Could add it for redundancy though.

defunctzombie (approved):

PR #158, opened 2024-12-11
Fix PyPI publishing

jtbandes:
Running a publish to TestPyPI immediately followed by a publish to the "real" PyPI triggered [an error](https://github.com/foxglove/schemas/actions/runs/12283736365/job/34277883747) due to a file already existing. The workaround recommended in https://github.com/pypa/gh-action-pypi-publish/issues/283#issuecomment-2499296440 is to set `attestations: false` on the TestPyPI publish.


indirectlylit (approved):
reading through the linked thread, this seems fine to me because we're only disabling on the test jobs.

(it does sound like the recommended fix is to split test and prod into separate flows, but this workaround seems reasonable)

PR #159, opened 2024-12-13
FG-2933: Add CompressedAudio schema

fgwt202412:
### Changelog

Add CompressedAudio schema

### Docs

None

### Description

Foxglove does not currently support audio messages of any kind. In order to support playback of audio in Foxglove Studio, this schema provides the wire format for audio frames that Foxglove can play.

Tested locally using the MCAP recording demo and a new Audio panel.

Used in:
- https://github.com/foxglove/mcap/pull/1292
- https://github.com/foxglove/app/pull/8221

FG-2933


fgwt202412:
The CI / python workflow is failing: https://github.com/foxglove/schemas/actions/runs/12341162294/job/34439471378?pr=159

It seems like Python 3.7 is not supported for Ubuntu 24.04 x64 - not sure if runners recently changed. According to [this issue](https://github.com/actions/setup-python/issues/962), we would need to fix to `ubuntu-22.04` instead of `ubuntu-latest` _or_ upgrade Python version to something supported in Ubuntu 24.04.

internal/schemas.ts
fgwt202412, line 1: This `CompressedAudio` schema was the only addition - the rest was auto-generated by `yarn update-generated-files`

.github/workflows/ci.yml
```diff
@@ -5,7 +5,7 @@ on:
     branches: [main]
     tags: ["releases/**"]
   pull_request:
-    branches: ["*"]
+    branches: ["**"]
```
fgwt202412, line 8: I added this so that CI runs on my branches with a slash in the branch name. I'm happy to remove this if you're not interested in this change.

jtbandes: 🤔 I've been using slashes in my branch names and didn't have an issue with CI running: https://github.com/foxglove/schemas/pull/158

fgwt202412: This PR is targeting a branch other than `main` (my other branch, `fgwt202412/ubuntu-22_04-python-3_7), which does have a slash in it.

[Docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore)

jtbandes: ah, makes sense, this is for the target branch rather than the head branch. In some other repos we use `pull_request: {}` which may do the same thing by default? Anyhow, this seems fine

internal/schemas.ts
```diff
@@ -843,6 +843,44 @@ Specifically, the requirements for different \`format\` values are:
...
+    {
+      name: "format",
+      type: { type: "primitive", name: "string" },
+      description: "Audio format.\n\nSupported values: `opus`.",
+    },
+    {
+      name: "type",
+      type: { type: "primitive", name: "string" },
+      description: "Frame type.\n\n Supported values: `key`, `delta`.",
+    },
```
jtbandes, line 870: We historically chose not to include this for CompressedVideo, although I don't think I know the whole reasoning behind the decision. It's certainly caused us a lot of pain having to detect key frames ourselves, but on the other hand it's nicer for users. cc @defunctzombie @snosenzo for any thoughts / additional historical context

fgwt202412: In practice with the MCAP recording demo as it is in https://github.com/foxglove/mcap/pull/1293, the frames are all keyframes, which would make this field somewhat unnecessary. That might not be the case for customers though.

It definitely pains me to include this extra field in every message (more data => more storage costs), but it does mean there is no logic of detecting keyframes like you said. A cursory search did not come up with how to detect audio keyframes, so I'd be worried about the complexity of the logic.

snosenzo: `VideoDecoder.configure` has the same concept: https://developer.mozilla.org/en-US/docs/Web/API/VideoDecoder/configure#description

We do not need to support it.

nidanin: It was my understanding that audio formats don't really have key vs delta frames the way video does. Each frame is completely decodable on its own. Is that incorrect?

nidanin: We might also need an `isInterleaved` field for PCM audio unless we feel like we can dictate this.

nidanin: `interleaved` vs `planar` here:

https://developer.mozilla.org/en-US/docs/Web/API/AudioData/format

defunctzombie: The main thing I will push on in whatever schema proposal follows this one is that each field we require the user to fill out meets the following:

- Feels intuitive and natural to turn data from an encoder/output to the fields required in our message (i.e. if the user has to manually figure out things like `isInterleaved` then we should do that for them.
- Cannot be inferred or figured out by us from the stream of data (i.e. sample_rate might not be part of the data stream and assumed to be in the container - or maybe not).

nidanin: That makes sense. I think what's inferable might depend on the codec.

nidanin: Looks like the w3codec spec mandates multi-channel PCM is interleaved:

https://www.w3.org/TR/webcodecs-pcm-codec-registration/

Might be a good thing for us to follow since it does make the data layout more straightforward.

nidanin: Maybe we should call this something other than `CompressedAudio` since it sounds like we also want to support uncompressed PCM?

nidanin: > `VideoDecoder.configure` has the same concept: https://developer.mozilla.org/en-US/docs/Web/API/VideoDecoder/configure#description
> 
> We do not need to support it.

It looks like it's required for flac audio:
https://www.w3.org/TR/webcodecs-flac-codec-registration/

```
[description](https://www.w3.org/TR/webcodecs/#dom-audiodecoderconfig-description) is required, and has to be the following:

The bytes 0x66 0x4C 0x61 0x43 ("fLaC" in ASCII)

A metadata block (called the STREAMINFO block) as described in section 7 of [[FLAC]](https://www.w3.org/TR/webcodecs-flac-codec-registration/#biblio-flac)

Optionaly other metadata blocks, that are not used by the specification
```

nidanin: Also for aac the `description` is meaningful:

https://www.w3.org/TR/webcodecs-aac-codec-registration/

```
If [description](https://www.w3.org/TR/webcodecs/#dom-audiodecoderconfig-description) is present, it is assumed to a AudioSpecificConfig as defined in [[iso14496-3]](https://www.w3.org/TR/webcodecs-aac-codec-registration/#biblio-iso14496-3) section 1.6.2.1, Table 1.15, and the bitstream is assumed to be in [aac](https://www.w3.org/TR/webcodecs-aac-codec-registration/#dom-aacbitstreamformat-aac).

If the [description](https://www.w3.org/TR/webcodecs/#dom-audiodecoderconfig-description) is not present, the bitstream is assumed to be in [adts](https://www.w3.org/TR/webcodecs-aac-codec-registration/#dom-aacbitstreamformat-adts) format.
```

PR #160, opened 2024-12-15
Fix CI errors due to ubuntu 24.04 image rollout

fgwt202412:
### Changelog

None

### Docs

None

### Description

CI [broke recently](https://github.com/foxglove/schemas/actions/runs/12341162294/job/34439471378?pr=159) because Python 3.7 is not supported for Ubuntu 24.04 x64. See https://github.com/actions/setup-python/issues/962. This fixes the `python` job to use Ubuntu 22.04.


jtbandes (approved):
👍 matches the approach in https://github.com/foxglove/mcap/pull/1290

PR #162, opened 2025-01-29
Upgrade to Yarn v4

amacneil:
### Changelog
None

### Docs
None

### Description
Upgrade to Yarn v4 for consistency with our other repos.


.yarnrc
```diff
@@ -1,2 +0,0 @@
-version-tag-prefix "releases/typescript/v"
-version-git-message "NPM package v%s"
```
amacneil, line 2: We don't appear to be using this feature of Yarn in recent package releases, so it is safe to remove.

E.g. https://github.com/foxglove/schemas/pull/156

jtbandes (approved):

PR #165, opened 2025-02-01
Prepare repo for incoming sdk

amacneil:
### Changelog
None

### Docs
None

### Description
A few repo configuration changes we can land before the SDK (https://github.com/foxglove/schemas/pull/164).


eloff (approved):
I don't know if any of these changes will be an issue for the existing code here, but the fact that it passes CI gives some confidence. At any rate, we can fix things after the fact. I'm going to approve it so we can move forward.

PR #170, opened 2025-02-02
Move typescript package into subdirectory

amacneil:
### Changelog
None

### Docs
None

### Description
Migrate `@foxglove/schemas` typescript package into `typescript` directory. This prepares for other typescript packages to live in this repo.

Also restrict to only necessary files included in the npm package. Currently lots of stuff unintentionally included:

image

New package contents:

image


typescript/schemas/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../tsconfig.json",
```
jtbandes, line 2: Maybe extend `@foxglove/tsconfig/base` unless there is a strong reason to make the package dependent on its parent dir?

typescript/schemas/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "name": "Foxglove",
+    "email": "support@foxglove.dev",
+    "url": "https://foxglove.dev/"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/foxglove/foxglove-sdk.git",
+    "directory": "typescript/schemas"
+  },
+  "main": "./dist/typescript/schemas/src/index.js",
```
jtbandes, line 16: I'm confused as to why this path includes `typescript/schemas`, why isn't it just e.g. `dist/index.js` or `dist/typescript/index.js`? And do we have a `typescript/schemas/typescript` folder? That seems weird

typescript/schemas/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "./jsonschema": "./dist/typescript/schemas/src/jsonschema.js"
+  },
+  "files": [
+    "dist"
+  ],
+  "packageManager": "yarn@4.6.0",
+  "scripts": {
+    "build": "tsc -b",
+    "clean": "rimraf dist",
+    "prepack": "yarn workspace foxglove-sdk update-generated-files && yarn build"
```
jtbandes, line 29: Feels like this update-generated-files should not be called from the inner package maybe?

typescript/schemas/README.md
```diff
@@ -0,0 +1,3 @@
+# @foxglove/schemas
+
+This package contains TypeScript type definitions for the Foxglove schemas.
```
jtbandes, line 3: this description is missing a mention of json schemas?

jtbandes (commented):
I don't understand why there's both `schemas/typescript` and `typescript/schemas`

defunctzombie:
Did you test your new package in the app repo to ensure that all the package.json entrypoints resolve as expected? You should be able to `npm package` it and then `yarn install <path to pkg>` (or maybe use a yarn resolution to the packaged file locally)

typescript/schemas/tsconfig.json
amacneil: done

typescript/schemas/package.json
amacneil: Also answering your other question at the same time:

> I don't understand why there's both `schemas/typescript` and `typescript/schemas`

- `typescript/schemas` is the typescript `@foxglove/schemas` package. I thought about nesting putting it in `typescript/@foxglove/schemas` but that seemed overkill unless we plan to have some non-foxglove-namespaced packages in this repo
- `schemas/typescript` are the generated type definitions for our schemas

Now one option would be to move the generated typescript types into `typescript/schemas/src/types`, that would _almost_ lead to the schemas package being self contained. It might be a bit weird that not all the generated schemas would live under top level `schemas` package, but that's not as much of a concern since anyone using typescript would be much more likely to import the package rather than copy/pasting files from the repo.

The wrinkle is that we also currently bundle the jsonschemas into the typescript package, and nesting them _only_ under the typescript package seems less of a clear win, since people using other languages might want to easily find and copy/paste them (like they do with proto/flatbuffer/etc). Side note: I looked in our app and we only use the jsonschemas in a single test, so maybe we could stop bundling them into `@foxglove/schemas`?

This brings me to your question why the `package.json` main path is `dist/typescript/schemas/src/index.js` not just `dist/index.js`. It's because we import stuff from above the package, so typescript infers our `baseUrl` as the repo root, meaning the `dist` output is super nested. We could solve that by putting all the typescript + jsonschemas under `typescript/schemas/src` so that `src` compiles cleanly to `dist`, but as mentioned above that is a bit weird especially for jsonschema.

Another route I could take is a `prepack` script in this package which copies the files into the package at build time, and we gitignore the duplicates. That seems like probably the best option which allows us to clean up the `dist` dir.

Thoughts?

typescript/schemas/package.json
amacneil: Yeah probably not. It was already like this, but I can see an argument that this should have already happened, especially if we tagged a new version and CI already checked that the generated files are up to date.

I'll remove it.

typescript/schemas/README.md
amacneil: updated + examples

typescript/schemas/package.json
amacneil: I tried the latter approach and quite like it. See the latest commit on this branch:

https://github.com/foxglove/foxglove-sdk/pull/170/commits/c72299d838e6cf35e87fe1716840856722bf90f8

amacneil:
@defunctzombie confirmed the current version of this PR is compatible with app ([#8633](https://github.com/foxglove/app/pull/8633))

scripts/updateGeneratedFiles.ts
```diff
@@ -3,19 +3,26 @@ import fs from "fs/promises";
 import path from "path";
```
defunctzombie, line 3: If you wanted you could put the scripts into a `packages` folder and also have them part of the yarn workspace. Could even expose them with `bin` fields in package.json

defunctzombie (approved):

scripts/updateGeneratedFiles.ts
amacneil: ok will have a look at that in another PR, this one is big enough

PR #171, opened 2025-02-02
Build python wheels

amacneil:
### Changelog
None

### Docs
None

### Description
Our upcoming python sdk uses [maturin](https://www.maturin.rs/) to depend on rust code. Most python users will want to install from wheels, containing the precompiled rust binaries.

This PR adds support for compiling wheels on 86 different operating systems and architectures.

Python jobs are moved to a separate workflow because `ci.yml` was getting out of hand. We could potentially separate the remaining ROS + Rust + TypeScript workflows into separate files in the future too.


python/foxglove-sdk/build.rs
```diff
@@ -30,6 +30,7 @@ fn compile_protos() {
         ));
     }
 
+    /*
```
gasmith, line 33: Is this broken under CI?

.github/workflows/python.yml
```diff
@@ -0,0 +1,258 @@
+# Originally generated by `maturin generate-ci`
```
gasmith, line 1: Maybe include a summary of changes that we can refer to, e.g. when the time comes to regenerate this file.

amacneil: It's modified quite heavily so I don't think we would want to regenerate this unless we are trying to add a new OS or something, or if maturin ship a new major version breaking CLI. For example stuff like installing protoc, modifying it to work with our repo directory structure, using our git tag patterns, adding lint, adding build for non-sdk packages.

Not sure if it makes sense to explain all that in the comment?

python/foxglove-sdk/build.rs
amacneil: TLDR yes - for example`arduino/setup-protoc` doesn't support windows and protoc binaries for windows are not published.

IMO we should remove protoc from the build step, but I didn't want to do that here. We can either commit the compiled protos to git (like we're doing with all other libraries in this repo), or even better we can remove proto serialization from the python sdk entirely and let rust consistently handle it.

.github/workflows/python.yml
gasmith: Yeah, I figured this is something that would happen once in a blue moon, but I don't know how often or how dramatically maturin changes their generate-ci output. If you think the changes are fairly self-evident, I'm fine with the minimal comment.

gasmith (approved):

.github/workflows/python.yml
amacneil: updated comment with more details

PR #173, opened 2025-02-03
rust: Replace HashSet<TopicByChannel> with a HashMap

gasmith:
While the `HashSet` might be slightly more memory-/cache-efficient, it's causing some clippy indigestion. This is not a high-cardinality set, and we're not accessing it on the hot path. We might as well just keep it simple with a HashMap.


eloff (approved):
LGTM. We discussed this already elsewhere, your motivation for the change is sound and it makes the code cleaner.

PR #174, opened 2025-02-04
rust: Move tests indoors

gasmith:
We've had some confusion around which examples to follow. To avoid that, let's bring our various internal tests and unstable examples inside the crate, so we aren't forced to publicly expose things that we don't intend to support.

Changes here:
- Moved test utilities to `crate::testutil`.
- Moved external tests to `crate::tests`.
- Removed unstable examples.
- Added docs to `FoxgloveError`, because this started out as an attempt to do `#![warn(missing_docs)]`.
- Fixed tests that were incorrectly relying on `std::thread::sleep()` inside a tokio runtime.
- Exposed `WebSocketServerHandle::publish_parameter_values()` under the `unstable` feature to avoid unused_code/unused_import lints.
- Removed pub visibility for `crate::websocket`.


eloff:
> Fixed tests that were incorrectly relying on std::thread::sleep() inside a tokio runtime.

Do the tests still pass? IIRC that was deliberate because we're waiting for flume messages to be processed, and we have no deterministic way of doing that except wait in real time, not an async wait.

gasmith:
> > Fixed tests that were incorrectly relying on std::thread::sleep() inside a tokio runtime.
> 
> Do the tests still pass? IIRC that was deliberate because we're waiting for flume messages to be processed, and we have no deterministic way of doing that except wait in real time, not an async wait.

They pass locally, anyway, using `cargo test --all-features`... I'd be more than a little surprised if tests somehow depended on blocking a tokio thread with a `std::thread::sleep`.

eloff (commented):
This seems like a big improvement.

Integration tests are supposed to only rely on the public interface, but I always find that too limiting, and it usually ends up making things public only to support the tests, which seems worse than relying on internals in the integration tests. I think philosophically, this is the better approach to bring them "indoors".

I'll take a final look once the unstable example is updated.

rust/foxglove/src/lib.rs
```diff
@@ -164,40 +164,52 @@ mod encode;
...
+pub use websocket::{Parameter, ParameterType, ParameterValue};
 pub use websocket_server::{WebSocketServer, WebSocketServerHandle};
 
+/// An error type for errors generated by this crate.
 #[derive(Error, Debug)]
 #[non_exhaustive]
 pub enum FoxgloveError {
+    /// A unspecified unrecoverable error.
     #[error("Fatal error: {0}")]
     Fatal(String),
```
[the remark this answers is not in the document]
gasmith: Right! That's been on my mind too. `Generic` or `Unspecified` might sound a bit better. I was also thinking we might use a `Box<dyn Error + Send + Sync + 'static>` instead of a `String`, to give users an extra degree of freedom in error handling.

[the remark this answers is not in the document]
eloff: Not sure about the Boxed Error, is that so the user can recover the original Error?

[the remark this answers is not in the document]
gasmith: For what it's worth, I've seen this pattern elsewhere: [sqlx](https://docs.rs/sqlx/latest/sqlx/enum.Error.html), [object_store](https://docs.rs/object_store/latest/object_store/enum.Error.html).

rust/foxglove/src/lib.rs
```diff
@@ -164,40 +164,52 @@ mod encode;
...
 #[doc(hidden)]
-pub use log_context::{GlobalContextTest, LogContext};
+pub use log_context::LogContext;
 pub use log_sink::LogSink;
 pub use mcap_writer::{McapWriter, McapWriterHandle};
 pub use metadata::{Metadata, PartialMetadata};
 pub(crate) use time::nanoseconds_since_epoch;
+#[doc(hidden)]
+#[cfg(feature = "unstable")]
+pub use websocket::{Capability, Parameter, ParameterType, ParameterValue};
```
gasmith, line 191: I'm not sure if we want to keep these at the top-level in the long run. The names are so generic, that I feel we should put them under a module namespace. TBD on what we want to call that: "websocket", "viz", etc. But I think we can postpone making that call until we're ready to stabilize/unhide them.

gasmith: I chose to put them here, instead of keeping `pub mod websocket`, because the websocket module is quite complex, and I don't want to accidentally expose stuff from there that we don't want users to find.

eloff: Yeah that makes sense to me as well

rust/foxglove/src/lib.rs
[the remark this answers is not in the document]
eloff: Yeah, the logic seems sound to me, we can do that

rust/foxglove/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "rust-analyzer.cargo.features": ["unstable"]
```
amacneil, line 2: never seen a vscode directory nested before, should this be at the repo root?

gasmith: Whoops, that wasn't supposed to be part of this commit.

PR #175, opened 2025-02-04
Add readme to rust crate

bryfox:
This adds a high-level overview to a README file, which links to docs.rs for more information (available on next release).

The existing readme with development documentation, is moved to README-development.

Sharing content between the readme and crate documentation is problematic as noted in feedback below.


rust/foxglove/README-development.md
```diff
@@ -0,0 +1,17 @@
+# Foxglove Rust SDK
```
amacneil, line 1: suggest renaming this file to `CONTRIBUTING.md` to match what we've done elsewhere.

or add it to the repo-level CONTRIBUTING.md

rust/foxglove/README.md
```diff
@@ -1,17 +1,155 @@
...
+events to local [MCAP] files or a local visualization server that communicates with the
+Foxglove app.
 
-### Generate Protobuf schemas
+[Foxglove]: https://docs.foxglove.dev/
+[MCAP]: https://mcap.dev/
 
-```bash
-cargo run --bin foxglove-proto-gen
+# Getting started
```
gasmith, line 10: I think we should keep the readme minimal. One big reason is that a lot of the rustdoc syntax & conventions are not supported by either Github or (presumably?) crates.io.

The README should be pure markdown, and I think we should exclude code examples and references entirely. We have docs for that.

For example, check out `object_store`'s [readme](https://crates.io/crates/object_store).

rust/foxglove/src/lib.rs
```diff
@@ -201,3 +201,7 @@ pub enum FoxgloveError {
     #[error(transparent)]
     JSONError(#[from] serde_json::Error),
 }
+
+#[doc = include_str!("../README.md")]
```
gasmith, line 205: One option for keeping the README in sync with the rustdoc preamble would be to move this to the very top of the file, so that it gets prefixed to the crate docs. Also allows you to avoid the `ReadmeTests` hack.

gasmith: BTW, I'm not necessarily making a strong endorsement for this trick. Over time, we might want to enhance the readme with crates/github niceties (e.g., images, banners, code of conduct, etc) that we don't want cluttering our rustdocs.

rust/foxglove/README.md
bryfox: I looked at tokio as a reference; they seem to largely duplicate content, and examples seem to render reasonably. I'm happy to keep this minimal and link to docs.rs instead, but I also see your other comment below about including the readme in lib. Do you have a favored approach?

bryfox: I updated per this suggestion. One thing that isn't served well with the same content is automatic linking to rust code.

eloff (approved):
This seems good to me, short and sweet

PR #183, opened 2025-02-04
Remove errant .vscode/settings.json

gasmith:
This was accidentally committed with #174.


eloff (approved):
Should we also add .vscode to .gitignore?

gasmith:
> Should we also add .vscode to .gitignore?

Maybe? Probably not? I was just messing around with alternatives to modifying the default feature set in Cargo.toml. I don't expect this to be a common occurrence.

amacneil:
> Should we also add .vscode to .gitignore?

no, those are the shared workspace settings. you can set your own user settings separately

PR #185, opened 2025-02-04
rust: Add missing docs, expand FoxgloveError

gasmith:
- Add the last missing docs, and slap on a `#![warn(missing_docs)]` attribute.
- Remove the mostly-useless `PartialMetadata::new()`.
- Expand the `FoxgloveError` somewhat:
  - Rename `Fatal` as `Unspecified` and retain the source error for downcasting.
  - Add variants for things that don't benefit from being opaque.

### DIFF
--- a/rust/foxglove/src/channel.rs
+++ b/rust/foxglove/src/channel.rs
@@ -44,6 +44,7 @@ pub struct Schema {
 }
 
 impl Schema {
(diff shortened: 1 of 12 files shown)


bryfox (approved):
👍 I think the additions to Error make sense. I wonder if we want to move these to something like an `errors` module now or soon.

PR #187, opened 2025-02-05
rust: Add nested struct to example

amacneil:
### Changelog
None

### Docs
None

### Description
Add example of nested struct to rust example.


eloff (approved):
Could we fix this at the foxglove side by adding support for $ref?

We could try transforming schemas on the Rust side to express them in a way foxglove understands, which is a bandaid solution.

But _just_ supporting a wider range of valid schemas in foxglove seems like it would be more of a root cause fix.

gasmith:
> Could we fix this at the foxglove side by adding support for $ref?

I think that's the suggested direction for FG-10012.

amacneil:
Ok fair enough, I'll close PR for now and we can revive once the UI supports it.

PR #188, opened 2025-02-05
rust: Add support for ws-protocol time capability

gasmith:
Add the capability and publishing for server `time`.

I also moved the websocket tests under the websocket module. Apart from feeling more natural, this also gives us the ability to access `crate::websocket::protocol::server` without making it more visible throughout the crate.


rust/foxglove/src/websocket_server.rs
```diff
@@ -138,6 +138,13 @@ impl WebSocketServerHandle {
             .await;
     }
 
+    /// Publishes the current server timestamp to all clients.
+    #[doc(hidden)]
+    #[cfg(feature = "unstable")]
+    pub async fn publish_time(&self, timestamp_nanos: u64) {
```
gasmith, line 144: I see that the [cpp](https://github.com/foxglove/ws-protocol/blob/main/cpp/foxglove-websocket/include/foxglove/websocket/server_interface.hpp#L113) & [typescript](https://github.com/foxglove/ws-protocol/blob/main/typescript/ws-protocol/src/FoxgloveServer.ts#L239) implementations call this `broadcastTime`. Will update to match.

rust/foxglove/src/websocket.rs
```diff
@@ -431,6 +433,28 @@ impl Server {
...
+
+        // https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#time
+        let mut buf = BytesMut::with_capacity(9);
+        buf.put_u8(protocol::server::BinaryOpcode::TimeData as u8);
+        buf.put_u64_le(timestamp_nanos);
+        let message = Message::binary(buf);
+
+        let clients = self.clients.get();
+        for client in clients.iter() {
+            if let Err(err) = client.control_plane_tx.send_async(message.clone()).await {
```
bryfox, line 452: `control_plane_tx` seems reasonable for a start, but this does make me wonder if these should be prioritized, and/or dealt with in some way such that we don't queue up multiple time messages.

bryfox (approved):
I left a couple of questions, which might spur follow-up issues, but this looks good to me.

@jtbandes and @achim-k might have thoughts or opinions here.

eloff (approved):
I guess this is just how the API works, but I'm surprised the end user had to call broadcast_time explicitly.

I would have expected the server itself maybe publishes the current time periodically.

I'm curious why it works this way.

What do you think of Bryan's recommendation for a name for the method?

The implementation looks good to me. I agree with the test reorganization.

rust/foxglove/src/websocket.rs
eloff: Maybe debounce these? That seems to be coming up a lot lately as a future improvement

rust/foxglove/src/websocket.rs
```diff
@@ -431,6 +433,28 @@ impl Server {
         }
     }
 
+    /// Publish the current timestamp to all clients.
+    #[cfg(feature = "unstable")]
+    pub async fn broadcast_time(&self, timestamp_nanos: u64) {
```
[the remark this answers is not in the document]
gasmith: Yeah. I don't feel strongly about it. Maybe I'll develop stronger opinions as we fill in more of the interface. Consistency is nice.

rust/foxglove/src/websocket.rs
gasmith: Debouncing sounds reasonable, and prioritizing time messages ahead of other message types also seems reasonable, since queueing delays will be manifested as clock skew on the client side. I'm tempted to defer this until we have a better sense of the requirements for other message types.

gasmith:
> I guess this is just how the API works, but I'm surprised the end user had to call broadcast_time explicitly.
>
> I would have expected the server itself maybe publishes the current time periodically.
>
> I'm curious why it works this way.

I thought about that too. The [ws-protocol spec](https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#time) mention that this can be used to obtain "accelerated, slowed, or stepped control over the progress of time." Which sound a bit funkier than plain old clock synchronization.

We could provide a batteries-included helper for periodically broadcasting the wallclock time, if that would be useful. But I guess I'd like to learn more about how this interface is typically used before we do that.

PR #189, opened 2025-02-05
Core implement ws server listener interface for subs and adverts

eloff:
Add new callbacks to ServerListener:

```
fn on_subscribe(&self, _channel_id: ChannelId) {}
fn on_unsubscribe(&self, _channel_id: ChannelId) {}
fn on_client_advertise(&self, _channel: &ClientChannel) {}
fn on_client_unadvertise(&self, _channel_id: ClientChannelId) {}
```

This mirrors how the callbacks are defined in the [Python implementation](https://github.com/foxglove/ws-protocol/blob/main/python/src/foxglove_websocket/server/__init__.py#L508), and they function in a similar way (e.g. not firing for duplicate or erroneous requests)

I added RecordingServerListener to testutil to allow more easily testing ServerListener.

I modified existing tests to verify these callbacks are called at the appropriate times with the expected arguments, and not called for duplicate requests.


bryfox (commented):
Looking good to me. As discussed, I think we can probably bring the server listener out from behind the feature now.

There's a `client-publish` example which uses part of this — it might be good to augment that with the advertisement handling.

rust/foxglove/src/cow_vec.rs
```diff
@@ -93,18 +93,16 @@ mod tests {
...
         vec.push(5);
-        let thread_state = handle.join().unwrap();
+        handle.join().unwrap();
         let final_state = vec.get().to_vec();
 
         // Old snapshot should still be valid and have original length
         assert_eq!(old_snapshot.len(), 3);
         // Both threads should see 5 items in their final state
         assert_eq!(final_state.len(), 5);
-        assert_eq!(thread_state.len(), 5);
```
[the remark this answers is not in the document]
eloff: I'll fix the comment, this was a race condition. It would be 4 items if the thread runs before the push(5) after starting the thread.

rust/foxglove/src/tests/websocket.rs
```diff
@@ -162,19 +168,38 @@ async fn test_advertise_to_client() {
...
+        .expect("Failed to send");
+    // Send a duplicate subscribe message (ignored)
     client_sender
         .send(Message::text(subscribe.to_string()))
         .await
         .expect("Failed to send");
 
     // Allow the server to process the subscription
     // FG-9723: replace this with an on_subscribe callback
+    // (whoops, that won't work either, unless we do something like polling the recording_listener)
```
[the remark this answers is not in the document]
eloff: There doesn't seem to be a way to wait for the queue to be drained, other than polling it, checking is_empty - but that still doesn't mean the data was acted on, just that it was removed from the queue. So I think it's the wrong thing to do.

I'm not sure what a sensible way to do this would look like. It's easy enough to come up with something just for tests where we poll somehow and wait for the messages to be processed.

Is that just for tests, or do we need something more polished that we could expose to the user?

rust/foxglove/src/tests/websocket.rs
```diff
@@ -280,6 +314,18 @@ async fn test_log_only_to_subscribers() {
     // FG-9723: replace this with an on_subscribe callback
```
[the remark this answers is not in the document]
eloff: I'll make a new ticket for this and reference it instead.

rust/foxglove/src/websocket.rs
```diff
@@ -128,11 +131,19 @@ impl Default for Server {
...
-    fn on_message_data(&self, channel_id: ClientChannelId, payload: &[u8]);
+    fn on_message_data(&self, _channel_id: ClientChannelId, _payload: &[u8]) {}
+    /// Callback invoked when a client subscribes to a channel.
+    fn on_subscribe(&self, _channel_id: ChannelId) {}
+    /// Callback invoked when a client unsubscribes from a channel.
+    fn on_unsubscribe(&self, _channel_id: ChannelId) {}
+    /// Callback invoked when a client advertises a client channel. Requires the "clientPublish" capability.
+    fn on_client_advertise(&self, _channel: &ClientChannel) {}
+    /// Callback invoked when a client unadvertises a client channel. Requires the "clientPublish" capability.
+    fn on_client_unadvertise(&self, _channel_id: ClientChannelId) {}
```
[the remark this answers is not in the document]
eloff: Yeah, I think that's a nice idea. I don't know exactly how that would look (or if we can do that.) Something to keep in mind.

rust/foxglove/src/tests/websocket.rs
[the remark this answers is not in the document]
bryfox: In my opinion, just for tests. I guess we don't necessarily need to poll the queue — we could have a test helper that polls, with a timeout, whatever future we want to wait on in the test (e.g. `client_receiver.next()`)

rust/foxglove/src/websocket.rs
```diff
@@ -128,11 +131,19 @@ impl Default for Server {
...
 }
 
-/// Provides a mechanism for registering callbacks for
-/// handling client message events.
+/// Provides a mechanism for registering callbacks for handling client message events.
+/// All methods are optional.
 pub trait ServerListener: Send + Sync {
     /// Callback invoked when a client message is received.
-    fn on_message_data(&self, channel_id: ClientChannelId, payload: &[u8]);
+    fn on_message_data(&self, _channel_id: ClientChannelId, _payload: &[u8]) {}
```
gasmith, line 138: Is the `ClientChannelId` globally unique, or locally-unique to a particular client? I presume the latter? Might IDs be reused, from the perspective of an implementer? We should probably update the rust doc for that type, and add some commentary to these methods to help guide user expectations.

We should add a reference to https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#client-message-data. Same goes for the other methods, I think.

rust/foxglove/src/websocket.rs
```diff
@@ -128,11 +131,19 @@ impl Default for Server {
...
+    /// Callback invoked when a client subscribes to a channel.
+    fn on_subscribe(&self, _channel_id: ChannelId) {}
```
gasmith, line 140: For server-advertised channels, we allocate IDs internally. The ID is recoverable, but we're asking trait implementation to do some extra legwork to map a channel ID back to something they understand (like a topic). Maybe we go a bit further.

Presumably we also need to pass the client identity, whatever that may be (perhaps a struct with a view over client metadata, so we can add getter methods later without breaking compatibility?). Actually, I think this comment applies to all of the methods in this trait.

rust/foxglove/src/websocket.rs
```diff
@@ -128,11 +131,19 @@ impl Default for Server {
     }
 }
 
-/// Provides a mechanism for registering callbacks for
-/// handling client message events.
+/// Provides a mechanism for registering callbacks for handling client message events.
+/// All methods are optional.
 pub trait ServerListener: Send + Sync {
```
gasmith, line 136: Seems like we might also want callbacks for client connect and disconnect, so that an implementer know when to flush client-specific state (about channel advertisements, subscriptions, etc.).

gasmith (commented):
A few early comments on the trait, still reading through the rest.

rust/foxglove/src/tests/websocket.rs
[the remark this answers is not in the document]
eloff: That future is buried at the bottom of `Server::handle_connection`, which is what makes it tricky. I'll create ticket to track it, but I'd have to experiment and think about it as to how we could implement it in a sensible way.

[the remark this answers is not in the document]
eloff: FG-10395

rust/foxglove/src/websocket.rs
eloff: It should only be unique per client, which suggests maybe we should pass in some kind of identifier for the client as well (what? SocketAddr? make up an id?). The Python implementation doesn't address that, and has these same signatures.

The payload is fine, the message only contains the channel id and the payload, and we pass those separately to the callback. I'll double-check the other methods.

rust/foxglove/src/websocket.rs
eloff: What if we map the ChannelId back to the Channel and pass that instead to these two particular callbacks? That seems like it should work.

rust/foxglove/src/websocket.rs
eloff: The existing implementations don't go that far, but it seems like a good idea to me

rust/foxglove/src/websocket.rs
```diff
@@ -171,60 +182,30 @@ impl ConnectedClient {
...
             }
             Ok(ClientMessage::Unsubscribe { subscription_ids }) => {
-                let mut subscriptions = self.subscriptions_by_channel.lock();
-                subscriptions
-                    .retain(|_, subscription_id| !subscription_ids.contains(subscription_id))
+                let mut subscriptions = self.subscriptions.lock();
+                for subscription_id in subscription_ids {
+                    if let Some((channel_id, _)) = subscriptions.remove_by_right(&subscription_id) {
+                        if let Some(handler) = self.server_listener.as_ref() {
+                            handler.on_unsubscribe(channel_id);
```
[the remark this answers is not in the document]
eloff: Yeah, that's a good point

rust/foxglove/src/websocket.rs
eloff: I added a global integer id for clients, so the subscriber can tell them apart

rust/foxglove/src/websocket.rs
eloff: I implemented this

rust/foxglove/src/testutil.rs
```diff
@@ -3,5 +3,82 @@
 mod log_context;
 mod log_sink;
 
+use crate::channel::ChannelId;
+use crate::websocket::{ClientChannel, ClientChannelId, ServerListener};
 pub use log_context::GlobalContextTest;
 pub use log_sink::{ErrorSink, MockSink, RecordingSink};
+use parking_lot::Mutex;
+
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
eloff: Turns out not needed anymore

rust/foxglove/src/websocket.rs
```diff
@@ -96,11 +103,23 @@ pub(crate) struct Server {
 /// handling client message events.
 pub trait ServerListener: Send + Sync {
     /// Callback invoked when a client message is received.
-    fn on_message_data(&self, channel_id: ClientChannelId, payload: &[u8]);
+    fn on_message_data(&self, _client_id: ClientId, _channel_id: ClientChannelId, _payload: &[u8]) {
+    }
+    /// Callback invoked when a client subscribes to a channel.
+    /// Only invoked if the channel is associated with the server and isn't already subscribed to by the client.
+    fn on_subscribe(&self, _client_id: ClientId, _channel_id: Arc<Channel>) {}
```
gasmith, line 110: Can rename these args as `channel` now, instead of `channel_id`.

Also, maybe we just pass a `&Channel` so that implementations aren't tempted to clone or hold on to the arc.

rust/foxglove/src/websocket.rs
```diff
@@ -96,11 +103,23 @@ pub(crate) struct Server {
 /// handling client message events.
 pub trait ServerListener: Send + Sync {
     /// Callback invoked when a client message is received.
-    fn on_message_data(&self, channel_id: ClientChannelId, payload: &[u8]);
+    fn on_message_data(&self, _client_id: ClientId, _channel_id: ClientChannelId, _payload: &[u8]) {
```
gasmith, line 106: Should we also pass `&ClientChannel` here (and `on_client_unadvertise`)? We're already doing the lookup before invoking the callback.

gasmith (approved):
LG, just some minor dangling comments.

rust/foxglove/src/websocket.rs
eloff: Based on the slack discussion, I went with wrapper view structs for all the argument types, exposing the bare minimum for now.

We'll need to provide some way to lookup a ClientChannel to get the full data, as that is now the one place where we provide less data than the existing callbacks in the original Python implementation.

gasmith (approved):
LG. I suspect we'll need to expand ClientChannelView with schema information for it to be useful, but we can cross that bridge when we get there.

PR #192, opened 2025-02-06
Add publish_status and remove_status methods to WebSocketServer

eloff:
Relatively simple PR to add support for publishing status messages to all connected clients, and removing them by list of ids.

[WS Spec](https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#remove-status)

[Current Python implementation](https://github.com/foxglove/ws-protocol/blob/main/python/src/foxglove_websocket/server/__init__.py#L457)

One thing I did here is treat status messages the same as log messages, in that when the queue is full it will drop older messages from the data plane queue.


rust/foxglove/src/websocket.rs
```diff
@@ -197,18 +198,33 @@ impl ConnectedClient {
...
+
+    /// Send a status message to the client.
+    fn send_status(&self, level: StatusLevel, message: String, id: Option<String>) {
+        let status = protocol::server::Status { level, message, id };
         let message = Message::text(serde_json::to_string(&status).unwrap());
-        // If the message can't be sent, or the outbox is full, log a warning and continue.
-        self.data_plane_tx.try_send(message).unwrap_or_else(|err| {
-            tracing::warn!("Failed to send status to client {}: {err}", self.addr)
-        });
+        self.send_data_lossy(message, MAX_SEND_RETRIES);
```
gasmith, line 227: I wonder whether these should be sent over the control plane channel instead, so that we don't drop them. I presume status messages are relatively low-frequency and high-importance.

rust/foxglove/src/websocket_server.rs
```diff
@@ -168,6 +168,22 @@ impl WebSocketServerHandle {
...
+        level: crate::websocket::StatusLevel,
+        message: impl Into<String>,
+        id: Option<impl Into<String>>,
+    ) {
+        self.0
+            .publish_status(level, message.into(), id.map(|id| id.into()));
+    }
+
+    /// Removes status messages by id from all clients.
+    pub fn remove_status(&self, status_ids: Vec<String>) {
```
gasmith, line 183: Maybe make it a bit more generic?

```suggestion
    pub fn remove_status<II, S>(&self, status_ids: II)
    where
        II: IntoIterator<Item = S>,
        S: Into<String>,
    {
```

Or maybe that's overkill.

rust/foxglove/src/websocket_server.rs
```diff
@@ -168,6 +168,22 @@ impl WebSocketServerHandle {
             .await;
     }
 
+    /// Publishes a status message to all clients.
```
gasmith, line 171: Maybe add links to the ws-protocol spec for these.

rust/foxglove/src/websocket_server.rs
```diff
@@ -168,6 +168,22 @@ impl WebSocketServerHandle {
...
+    pub fn publish_status(
+        &self,
+        level: crate::websocket::StatusLevel,
+        message: impl Into<String>,
+        id: Option<impl Into<String>>,
```
gasmith, line 176: Since there's an optional argument, we might consider having the client construct the `Status`, and perhaps with a builder (imagine, e.g., `Status::error("oh no").id("x")`).

gasmith (commented):
LG in general, though I wonder about the use of the data plane channel.

rust/foxglove/src/websocket.rs
gasmith: That seems pretty reasonable, sure.

rust/foxglove/src/websocket_server.rs
eloff: I did that initially, but we end up needing a Vec<String> anyway, and maybe I'm wrong but I think IntoIterator<String>.collect::<Vec<String>>() will create a new Vect, so then it's just worse in the common case where you already had a Vec.

Better to get the user to do the collect into a Vec if they have an iterator, or just pass it in if they have a Vec.

rust/foxglove/src/websocket_server.rs
eloff: Seems slightly better, yes

rust/foxglove/src/websocket_server.rs
```diff
@@ -175,6 +174,18 @@ impl WebSocketServerHandle {
             .await;
     }
 
+    /// Publishes a status message to all clients.
+    /// https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#status
```
gasmith, line 178: Needs a bit more formatting for rustdoc. How about:

```suggestion
    ///
    /// For more information, refer to the [Status][status] message specification.
    ///
    /// [status]: https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#status
```

rust/foxglove/src/websocket_server.rs
```diff
@@ -175,6 +174,18 @@ impl WebSocketServerHandle {
...
+    pub fn publish_status(&self, status: Status) {
+        self.0.publish_status(status);
+    }
+
+    /// Removes status messages by id from all clients.
+    /// https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#status
```
gasmith, line 184:
```suggestion
    ///
    /// For more information, refer to the [Remove Status][remove-status] message specification.
    ///
    /// [remove-status]: https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#remove-status
```

rust/foxglove/src/websocket_server.rs
```diff
@@ -175,6 +174,18 @@ impl WebSocketServerHandle {
             .await;
     }
 
+    /// Publishes a status message to all clients.
+    /// https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#status
+    pub fn publish_status(&self, status: Status) {
```
gasmith, line 179: Need to add these methods on the blocking handle too.

PR #193, opened 2025-02-06
Add release workflows

amacneil:
### Changelog
None

### Docs
None

### Description
New release workflows:
- Draft release workflow to bump version number across all SDK packages (rust + python) and draft release notes
- Post release workflow to merge release tag back into main branch

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: ${{ env.SDK_TAG }}
-          fetch-depth: 0
+          fetch-depth: 0 # needed to checkout main
 
       - name: Git merge into main
         run: |
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -1,0 +1,26 @@
+name: Post Release
+
+on:
+  release:
+    types: [released]
+
+env:
+  SDK_TAG: ${{ github.event.release.tag_name }}
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ env.SDK_TAG }}
+          fetch-depth: 0
+
+      - name: Git merge into main
+        run: |
+          echo ${{ secrets.FOXGLOVEBOT_GITHUB_TOKEN }} | gh auth login --with-token
+          git config user.name "Foxglove"
+          git config user.email "support@foxglove.dev"
+          git checkout -b main --track origin/main
+          git merge --no-ff ${{ github.ref }}
+          git push origin main
--- a/.npmignore
+++ b/.npmignore
@@ -1,1 +1,0 @@
-ros_foxglove_msgs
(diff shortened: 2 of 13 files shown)


.github/workflows/post-release.yml
```diff
@@ -0,0 +1,26 @@
...
+  SDK_TAG: ${{ github.event.release.tag_name }}
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ env.SDK_TAG }}
+          fetch-depth: 0
```
bryfox, line 17: Seems fine, but I'm not seeing why we need the full history

scripts/bumpSdkVersion.ts
```diff
@@ -0,0 +1,74 @@
...
+    console.log(`Checking ${cargoFile}...`);
+    const content = await fs.readFile(cargoFile, "utf8");
+
+    if (!versionRegex.test(content)) {
+      console.log(`  ℹ️ Skipped, does not contain version field`);
+      continue;
+    }
+
+    // Only update the main version field, not dependencies
+    const updatedContent = content.replace(versionRegex, `version = "${versionNumber}"`);
```
bryfox, line 38: Should we enforce that the version number increments here?

scripts/bumpSdkVersion.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  if (!success) {
+    console.error("\n❌ Some versions could not be updated");
+    process.exit(1);
+  }
+
+  // run cargo check
+  console.log("\nRunning cargo check...");
+  const cargoCheck = spawn("cargo", ["check"], {
```
bryfox, line 57: Why spawn here instead of adding a step to the draft job?

bryfox (approved):
Couple of questions, but looks good to me. I'm not sure what the release note generation does, but that's easy enough to adjust.

scripts/bumpSdkVersion.ts
amacneil: good idea

scripts/bumpSdkVersion.ts
amacneil: I guess no real reason? Only that it when testing the bump script locally it is helpful to check the output `Cargo.toml` is still valid.

I have updated this to run `cargo tree` instead which seems sufficient to validate the `Cargo.toml` files without building the rust code.

.npmignore
```diff
@@ -1 +0,0 @@
-ros_foxglove_msgs
```
amacneil, line 1: unrelated to PR but this file is no longer needed as of https://github.com/foxglove/foxglove-sdk/pull/170

python/foxglove-sdk/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "foxglove-sdk-python"
-version = "0.1.0"
+version = "0.1.1"
```
amacneil, line 3: bumping for consistency with rust package

PR #194, opened 2025-02-07
Fix post-release workflow

amacneil:
### Changelog
None

### Docs
None

### Description
Fix git permissions in post-release workflow.


bryfox (approved):

gasmith (approved):
Cool. TIL about `.on.release`. I like the branch-and-merge strategy for draft releases too. Good stuff.

amacneil:
Yeah branch + merge is nice because it means you can test the exact release before it goes out, without any PRs unintentionally slipping in. We use that for main app too.

PR #197, opened 2025-02-07
rust: Automatically advertise clock over websocket

gasmith:
This is a breaking change to automatically advertise the local clock over the websocket by default.

Based on [this slack thread](https://foxglove.slack.com/archives/C07RB3SSJE5/p1738945649661899).

This also takes the timestamp stuff out from under the `unstable` feature, which forces us to make a choice about how we're going to expose websocket-specific, but generically-named, symbols like `Capability`. I went with keeping them under the `websocket` module. I'm open to alternatives.

Related to FG-10151


amacneil:
👏

rust/foxglove/src/websocket.rs
```diff
@@ -96,6 +101,7 @@ type WebsocketSender = SplitSink<WebSocketStream<TcpStream>, Message>;
 // Queue up to 1024 messages per connected client before dropping messages
 const DEFAULT_MESSAGE_BACKLOG_SIZE: usize = 1024;
 const DEFAULT_CONTROL_PLANE_BACKLOG_SIZE: usize = 64;
+const DEFAULT_ADVERTISE_CLOCK_PERIOD: Duration = Duration::from_millis(1000 / 60);
```
amacneil: 🤔 I'm wondering if there is a better way to do this without publishing so many clock messages.

In a non-simulated environment, viz client can assume that time advances on the server at the normal pace. So we really only need to sync the time infrequently (maybe every minute? certainly no more than once per second).

Time can update on the robot (e.g. clock is updated from gps) but we could avoid publishing clock messages at 60hz.

gasmith: Yeah, 60Hz is a bit much. I noticed while testing the the app only updates the timestamp at the top of the page when it gets a `Time` advertisement over the websocket. Maybe we can extend ws-protocol with a capability to indicate that the timesource is a clock ticking at the standard rate, and that the `Time` messages are merely for synchronization.

gasmith:
:hocho: 

Publishing a timestamp at 60Hz just feels wrong, but there doesn't seem to be a better alternative today with the current ws-protocol. Filed FG-10421 to investigate protocol improvements.

I'll revive the websocket module reorg in a separate PR.

PR #199, opened 2025-02-09
Turn rust examples into packages

defunctzombie:
### Changelog
None

Not worth noting on the changelog

### Docs
None

### Description
This change re-works the rust examples. Instead of all the examples being sources as part of the "foxglove" package each example is now its own package within the `rust/examples` folder. This makes it more obvious which dependencies are needed for an example and isolates the examples from each other. A reader can "copy-paste" an example as a complete starting point (including the Cargo.toml file).

This change also does away with the "examples/common" folder. I found this layer of indirection confusing for understanding  the code I would need to write.

Precedent: https://github.com/tokio-rs/axum/tree/main/examples

---
I read about the `--example` concept in `cargo` but found that it suffered from what I call "dependency co-mingling". As a reader of an example - I want it to be clear and obvious what I would need to get that example working without introducing extra dependencies.


rust/examples/ws-demo/src/main.rs
```diff
@@ -97,3 +76,45 @@ pub fn log(counter: u32) {
...
+    let args = Cli::parse();
+
+    let server = foxglove::WebSocketServer::new()
+        .name("ws-demo")
+        .bind(&args.host, args.port)
+        .start()
+        .await
+        .expect("Server failed to start");
+
+    tokio::task::spawn(log_forever(args.fps));
```
amacneil, line 107: lets move these `log_forever()` and `log_blocking()` methods into closures - make it even more obvious where things are happening.

rust/examples/client-publish/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "foxglove-client-publish"
```
gasmith, line 2: Maybe we also follow axum's example for naming. Instead of `foxglove-client-publish`, call this `example-client-publish`. That way it's obvious that it's an example, and not an official library.

This also sounds better in the context of a workspace, where you can run stuff from the top-level of the workspace using just the package name: `cargo run -p example-client-publish`.

rust/examples/mcap-demo/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "foxglove-mcap-demo"
```
gasmith, line 2: While we're here, can we drop the "demo" from this name? Maybe `example-mcap` or `example-mcap-writer`?

Likewise for the other "demo" names.

rust/examples/mcap-demo/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+foxglove = { path = "../../foxglove" }
+tokio = { version = "1.0", features = ["full"] }
```
gasmith, line 9: I'm tempted to rewrite this one without the tokio dependency, because there's really nothing about the mcap example that depends on or benefits from async-ness.

rust/examples/client-publish/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+foxglove = { path = "../../foxglove", features = ["unstable"] }
```
gasmith, line 8: I think we might want to do a little more to signal that this example uses the unstable feature, because we don't want folks using that by accident. Could we put this under `rust/examples/unstable`?

rust/examples/ws-demo/src/main.rs
```diff
@@ -28,17 +18,6 @@ static_typed_channel!(pub BOX_CHANNEL, "/boxes", SceneUpdate);
...
-    let mut counter: u32 = 0;
-    let duration = Duration::from_millis(1000 / u64::from(fps));
-    while !stop.load(Ordering::Relaxed) {
-        log(counter);
-        std::thread::sleep(duration);
-        counter += 1;
-    }
-}
-
 #[allow(dead_code)]
```
gasmith, line 21: It's not dead anymore. We can also get rid of `pub`.

gasmith (commented):
I'm a little saddened that we don't get spinny cubes in all of our demos, but I agree that it's better to make the examples copy-pasteable with clear dependencies.

eloff (approved):
This looks like a good improvement.

I wondered if we should use workspace dependencies instead of each example having separately versioned dependencies. But then you'd lose the copy+paste ability. So I think it's better as-is.

We could improve on the examples over time (e.g. to bring back spinny cubes) that doesn't need to hold this up.

PR #200, opened 2025-02-10
rust: Re-export prost_types::Timestamp under foxglove::schemas

gasmith:


eloff (approved):
I checked how this shows up in the docs, and it works just fine, as if it were our type. If you click through to the source you end up in prost_types.

I think it works.

An alternative is wrap the prost_types::Duration and Timestamp with our own struct. But we'd have to adjust the code gen and I'm not sure what we gain for our trouble.

gasmith:
> An alternative is wrap the prost_types::Duration and Timestamp with our own struct. But we'd have to adjust the code gen and I'm not sure what we gain for our trouble.

This way we also get to keep some handy conversion implementations that I'd expect most integrators to use:
- [`impl TryFrom<std::time::Duration> for Duration`](https://docs.rs/prost-types/latest/prost_types/struct.Duration.html#impl-TryFrom%3CDuration%3E-for-Duration-1)
- [`impl From<SystemTime> for Timestamp`](https://docs.rs/prost-types/latest/prost_types/struct.Timestamp.html#impl-From%3CSystemTime%3E-for-Timestamp)

gasmith:
Oo, actually, maybe we want to look at prost_wkt_types which offers conversions from chrono timestamps...

amacneil:
I think its worth more strongly considering offering our own structs for these, so that we avoid a leaky abstraction - for typed logging methods, ideally we should be able to switch out the serialization entirely without a breaking API change.

Would be good to understand if there are any downsides to rolling our own.

defunctzombie:
I am 👍 on this change landing so we can start trending in the right direction and ticketing the ideas surfaced in this thread. I do like the notion of being more "agnostic" at our schema level to the underlying serialization we've picked by default.

gasmith:
> I think its worth more strongly considering offering our own structs for these, so that we avoid a leaky abstraction - for typed logging methods, ideally we should be able to switch out the serialization entirely without a breaking API change.
> 
> Would be good to understand if there are any downsides to rolling our own.

The interface that we're exposing is the struct name, its public fields & methods, and its trait implementations. If we were to roll our own, it would be _exactly the same thing_ as what `prost_types` (or `prost_wkt_types`) offers, with the exception that we'd probably omit the [public methods](https://docs.rs/prost-wkt-types/latest/prost_wkt_types/struct.Timestamp.html#implementations) that we don't expect anyone to use (because our users treat these as write-only).

Note that we're not really leaking protobuf details here. That's just a [trait](https://docs.rs/prost/latest/prost/trait.Message.html) that is implemented for these types, and it's one we'd need to implement anyway for any replacement we build ourselves.

The upside of rolling our own would be to offer more `From<T>` conversions so that it's easier for folks to work with the types they care about (`std::time::Duration`, `SystemTime`, `chrono::DateTime`, etc.), and just do a final `.into()` when constructing the message.

The downside of rolling our own is that it's something else to write, document, and maintain. Not a big deal, I suppose.

amacneil (approved):

gasmith:
>     * we document timestamp fields (and duration nanoseconds) as unsigned; prost fields are signed

We can't really work around that one without changing the types in the underlying protobuf files. It's the well-known types themselves that define these fields as signed:

https://protobuf.dev/reference/protobuf/google.protobuf/#duration
https://protobuf.dev/reference/protobuf/google.protobuf/#timestamp

It might be possible to make them unsigned in a way that doesn't break wire compatibility (since [int32, int64, uint32, and uint64 are all wire-compatible](https://protobuf.dev/programming-guides/proto2/#updating)). But we're still changing the interpretation of those fields, which arguably does break compatibility (though possibly only for edge cases we don't care about).

>     * we document using the field names "sec" and "nsec" everywhere except for protobuf

Yeah, that's a regrettable wart. It would be straightforward to fix this one with our own types.

bryfox:
> We can't really work around that one without changing the types in the underlying protobuf files

If we defined our own types, could we provide a conversion? Large values of nsec could be rolled into the seconds. I guess timestamp seconds would technically be fallible, but protobuf places tighter bounds than the type allows anyway.

gasmith:
> > We can't really work around that one without changing the types in the underlying protobuf files
> 
> If we defined our own types, could we provide a conversion? Large values of nsec could be rolled into the seconds. I guess timestamp seconds would technically be fallible, but protobuf places tighter bounds than the type allows anyway.

Oh good point. The conversion should be infallible. I was looking at it the wrong way around.

PR #201, opened 2025-02-10
Generate pyo3 schemas for SDK

bryfox:
This PR produces well-known Foxglove schemas as `pyclass`es which can be used by [pyo3](https://pyo3.rs/) in the SDK. For the default channel logging interface, we plan to remove protobuf serialization on the python side, and instead use the existing serialization in SDK core. We've demonstrated that this can be done with little relative overhead; see FG-10358.

Most of the changes here are in `generatePyclass.ts`; per FG-10347, we're going to simplify build steps between schemas and SDK. For now, this produces output in a `pyclass` directory, and the build script for the python SDK copies those files into the project. (Open to suggestions here.)

The generator produces three files:

- the schemas, expressed as `pyclass`-annotated Rust structs & enums,
- the .pyi stub interface for the above,
- and a pyo3 submodule for export

Currently, the last file is not part of the SDK build, and the submodule is registered with a function call instead. This mechanism is set to be deprecated in pyo3. Once we have the build process unified from FG-10347, it'll be a little easier to make the declarative submodule syntax work.

Follow-up PRs will:
- add getters or accessors to the schema fields
- generate schema channel types
- update the python examples, and re-export the private schemas under the public module


.prettierignore
```diff
@@ -1 +1 @@
-schemas
+./schemas
```
bryfox, line 1: With the recent repo merge, the typescript source now has a `schemas` directory as well

bryfox:
@amacneil I think you have the vision for unifying the codegen (FG-10347). This is an intermediate step, and it's a little messy since the generated source is coupled to SDK source — ideas welcome, or we can plan to clean it up as part of that issue.

scripts/updateGeneratedFiles.ts
```diff
@@ -142,6 +150,43 @@ async function main({ outDir, rosOutDir }: { outDir: string; rosOutDir: string }
       generateMarkdown(Object.values(foxgloveMessageSchemas), Object.values(foxgloveEnumSchemas)),
     );
   });
+
+  await logProgress("Generating Pyclass definitions", async () => {
+    const dir = await fs.mkdir(path.join(outDir, "pyclass"), { recursive: true });
```
amacneil, line 155: I suggest writing these directly to the python directory, like we're doing for the typescript types.

The top level `schemas` directory is really just to make it easy for people to find if they want to copy/paste the .proto or other files into their own project. I can't think of any other reason to save the generated python code here.

We should try to keep the generated code in its own directory though, so we can add a top level `yarn clean`. I'll make a PR with that.

eloff (commented):
These produce 'Vec<u8>' for bytes, are we going to punt on the performance issue for now and hope it gets fixed upstream?

Still looking this over.

schemas/pyclass/schemas.rs
```diff
@@ -0,0 +1,2077 @@
...
+            // todo: prost should likely be configured to use unsigned types
+            seconds: value.seconds.try_into().unwrap_or_default(),
+            nanos: value.nanos.try_into().unwrap_or_default(),
+        }
+    }
+}
+/// A primitive representing an arrow
+#[pyclass]
+#[derive(Clone)]
+pub(crate) struct ArrowPrimitive {
```
eloff, line 125: Why are we taking this approach of duplicating the fields for the struct instead of wrapping the foxglove::schemas type?

`pub(crate) struct ArrowPrimitive(foxglove::schemas::ArrowPrimitive);`

Are there problems with that approach we can't overcome?

scripts/updateGeneratedFiles.ts
amacneil: Update: https://github.com/foxglove/foxglove-sdk/pull/202

python/foxglove-sdk/src/schemas.rs
```diff
@@ -0,0 +1,1921 @@
...
+//! Generated by https://github.com/foxglove/schemas
+#![allow(clippy::too_many_arguments)]
+#![allow(clippy::enum_variant_names)]
+#![allow(non_snake_case)]
+use pyo3::prelude::*;
+
+/// An enumeration indicating how input points should be interpreted to create lines
+#[pyclass(eq, eq_int)]
+#[derive(PartialEq, Clone)]
+pub(crate) enum LinePrimitiveLineType {
```
amacneil, line 11: thoughts on generating each of these as separate files?

if not, at least alphabetically sorting them would be a good idea

amacneil: oh I guess maybe they are alpha-sorted after the enums? I can't tell.

python/foxglove-sdk/src/schemas.rs
```diff
@@ -0,0 +1,1921 @@
+//! Definitions for well-known Foxglove schemas
+//! Generated by https://github.com/foxglove/schemas
```
amacneil, line 2: It would be nice if possible to put all the generated files in a subdirectory, rather than mixed in with normal code, so that we can easily clean it prior to regenerate.

In https://github.com/foxglove/foxglove-sdk/pull/202 I've started consistently doing this, I also found a bug in https://github.com/foxglove/foxglove-sdk/pull/203 as a result of not fully cleaning generated output directories.

python/foxglove-sdk/pyproject.toml
```diff
@@ -51,4 +51,4 @@ warn_unreachable = true
 
 [tool.black]
 include = '\.pyi?$'
-extend-exclude = "python/foxglove/_protobuf"
+extend-exclude = "python/foxglove/(_protobuf|_foxglove_py/schemas.pyi)"
```
amacneil, line 54: we could run `black` after generating code

python/foxglove-sdk/build.rs
```diff
@@ -1,13 +1,12 @@
-// use std::io::{ErrorKind, Read};
-use std::path::Path;
-// use std::process::Command;
+use std::{fs, path::Path, process::Command};
 
+// Deprecated. PB serialization will happen on the Rust side.
```
amacneil, line 3: any reason not to delete all this in the same PR?

python/foxglove-sdk/build.rs
```diff
@@ -30,54 +29,47 @@ fn compile_protos() {
...
 
-    /*
-    // Compile protos
-    let mut command = Command::new("protoc");
+    // Write __init__.py
+    let init_path = out_dir.join("__init__.py");
+    fs::write(init_path, init_imports.join("\n") + "\n").unwrap();
+}
+
+fn rustfmt(file: &Path) {
```
amacneil, line 37: this doesn't seem to be called anywhere?

python/foxglove-sdk/build.rs
```diff
@@ -30,54 +29,47 @@ fn compile_protos() {
...
+    let dest = Path::join(rust_dir, "schemas.rs");
+    println!("cargo:rerun-if-changed={}", src.display());
+    fs::copy(&src, &dest).expect("schemas.rs should be copied by build script");
+    rustfmt(&dest);
+
+    // python stub interface
+    let src = Path::join(src_dir, "schemas.pyi");
+    let dest = Path::join(python_dir, "_foxglove_py/schemas.pyi");
+    println!("cargo:rerun-if-changed={}", src.display());
+    fs::copy(&src, &dest).expect("schemas.pyi should be copied by build script");
```
amacneil, line 69: seems simpler if we can move all this (including `rustfmt`) into the generate script

schemas/pyclass/schemas.rs
bryfox: No, I think we can do it that way; I'll update.

python/foxglove-sdk/build.rs
bryfox: The python examples still use the protobuf types; I plan to remove once those are updated

bryfox:
> These produce Vec u8 for bytes, are we going to punt on the performance issue for now and hope it gets fixed upstream?

I'm punting on it in this PR since there's nothing consuming the type and I don't want this to bleed into the Rust code; it's big enough as-is. But I expect we'll need to deal with this ourselves in a future PR.

python/foxglove-sdk/src/schemas.rs
bryfox: The generated rust files are now in a subdirectory. I'll clean this up a big once #202 merges.

I don't see an easy way to do the same for the .pyi stubs, since (I think) they need to fit into the expected module hierarchy, and that doesn't all come from codegen.

python/foxglove-sdk/pyproject.toml
bryfox: 👍 The typescript script now runs `black` and `rustfmt`. If either tool isn't installed, this step is skipped.

scripts/updateGeneratedFiles.ts
bryfox: This is done now. I wonder if this shouldn't even be part of the same `update` task; they feel spiritually different, and contributors to `schemas` don't need to worry about this one. This is especially true since (per comments below) the pyclass generation now has optional dependencies on cargo, poetry, and black.

bryfox: I cleaned this up with your changes above. For now, I put this behind a `--include-sdk` flag, which keeps the diff smaller. I do think a separate command might be worthwhile, but that could probably wait for FG-10347.

python/foxglove-sdk/pyproject.toml
amacneil: can you make sure the format step always runs in CI (e.g. fails if either tool is not installed), so that we don't mark the PR as green unless the code is correctly formatted.

e.g. from this step https://github.com/foxglove/foxglove-sdk/blob/main/.github/workflows/ci.yml#L76-L89

.flake8
```diff
@@ -9,5 +9,6 @@ exclude =
     python/foxglove-schemas-flatbuffer/foxglove_schemas_flatbuffer/*.py,
     python/foxglove-schemas-protobuf/foxglove_schemas_protobuf/*.py,
     python/foxglove-sdk/python/foxglove/_protobuf/*,
+    python/foxglove-sdk/python/foxglove/_foxglove_py/schemas.pyi,
```
amacneil, line 12: can we move this now? we format this file right?

scripts/generate.ts
```diff
@@ -31,8 +39,13 @@ async function logProgress(message: string, body: () => Promise<void>) {
...
+const repoRoot = path.resolve(__dirname, "..");
+
+const pythonSdkRoot = path.resolve(repoRoot, "python", "foxglove-sdk");
+const pythonSdkSourceRoot = path.join(pythonSdkRoot, "src", "generated");
+const pythonSdkStub = path.join(pythonSdkRoot, "python", "foxglove", "_foxglove_py", "schemas.pyi");
+
+async function main({ clean, includeSdk }: { clean: boolean; includeSdk: boolean }) {
   const outDir = path.join(repoRoot, "schemas");
   const rosOutDir = path.join(repoRoot, "ros_foxglove_msgs");
   const typescriptTypesDir = path.join(repoRoot, "typescript/schemas/src/types");
```
amacneil, line 51: looks like you could hoist these up above the main function too - keep them all in one place

python/foxglove-sdk/src/lib.rs
```diff
@@ -200,5 +203,11 @@ fn _foxglove_py(m: &Bound<'_, PyModule>) -> PyResult<()> {
     m.add_class::<BaseChannel>()?;
     m.add_class::<PyWebSocketServer>()?;
     m.add_class::<PyMcapWriter>()?;
+
+    // Register the schemas module
+    // A declarative submodule is generated in schemas/pyclass/module.rs, but this is currently
```
amacneil, line 208: I think this file reference comment is out of date

.flake8
bryfox: I'm currently running it through `black` and not flake8, but this is not required; I removed it.

scripts/generate.ts
bryfox: `--include-sdk` is now gone, so I moved everything back into `main`, and these are all in one place again

scripts/generate.ts
```diff
@@ -155,6 +175,60 @@ async function main({ clean }: { clean: boolean }) {
...
+      spawnSync("cargo", ["fmt", "--", path.resolve(pymoduleFile)], spawnOpts);
+    } catch (err) {
+      console.error("Failed to format rust output using `cargo fmt`");
+      console.error(err);
+    }
+
+    // Pyi stub file
+    await fs.writeFile(pythonSdkStub, generatePymoduleStub([...enumSchemas, ...messageSchemas]));
+    try {
+      spawnSync("poetry", ["run", "black", path.resolve(pythonSdkStub)], spawnOpts);
```
gasmith, line 223: Looks like we're not checking the result from `spawnSync`. I noticed because on my system, `yarn generate` doesn't format `schemas.pyi`. I think that's because when I do `poetry run` in the top-level directory, it fails with `Command not found: black`.

For fixing the poetry run failure, maybe: `spawnSync("poetry", ["--directory", path.resolve(pythonSdkRoot), "run", ...])`.

typescript/schemas/src/internal/generatePyclass.ts
```diff
@@ -0,0 +1,472 @@
...
+ */
+export function generatePyclass(schema: FoxgloveSchema): string {
+  return isMessageSchema(schema) ? generateMessageClass(schema) : generateEnumClass(schema);
+}
+
+/**
+ * Generate a .pyi stub for the given schemas.
+ */
+export function generatePymoduleStub(schemas: FoxgloveSchema[]): string {
+  const header = ["from enum import Enum", "from typing import List"].join("\n") + "\n";
```
gasmith, line 42: Let's a add a `# Generated by ...` line too.

typescript/schemas/src/internal/generatePyclass.ts
```diff
@@ -0,0 +1,472 @@
...
+  return [header, ...definitions].join("\n");
+}
+
+/**
+ * Generate a rust module which exports the given schemas annotated with `pymodule`.
+ */
+export function generatePymodule(schemas: FoxgloveSchema[]): string {
+  const header =
+    [
+      `use pyo3::prelude::*;\n`,
```
gasmith, line 95: `# Generated by ...`

typescript/schemas/src/internal/generatePyclass.ts
```diff
@@ -0,0 +1,472 @@
...
+    .sort((a, b) => a.name.localeCompare(b.name))
+    .map(({ source }) => source);
+
+  return [header, ...definitions].join("\n");
+}
+
+/**
+ * Generate a rust module which exports the given schemas annotated with `pymodule`.
+ */
+export function generatePymodule(schemas: FoxgloveSchema[]): string {
```
gasmith, line 92: > Currently, the last file is not part of the SDK build, and the submodule is registered with a function call instead. This mechanism is set to be deprecated in pyo3. Once we have the build process unified from [FG-10347](https://linear.app/foxglove/issue/FG-10347), it'll be a little easier to make the declarative submodule syntax work.

I feel like we should leave this out until we're actually using it so that it doesn't rot.

Why will does this need to wait for unified codegen?

gasmith (commented):
LG in general. A few comments.

scripts/generate.ts
amacneil: See https://github.com/foxglove/foxglove-sdk/pull/209 I added an exec() helper

typescript/schemas/src/internal/generatePyclass.ts
bryfox: It doesn't need to wait, and based on #209 that's not going to get us any closer. The problem now is that our exported python modules are a mix of generated and python-native code, and I can't (as far as I know) reference the generated module [declaratively](https://pyo3.rs/v0.23.4/module.html#declarative-modules) per https://github.com/rust-lang/rust/issues/54727.

I've removed it for now; tracked by FG-10457.

.github/workflows/python.yml
```diff
@@ -115,6 +115,7 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: "3.9"
+      - run: rustup component add rustfmt
```
amacneil, line 118: should we use `setup-rust` instead? this feels a bit out of place

rust/foxglove/CONTRIBUTING.md
```diff
@@ -2,16 +2,27 @@
...
 ### Generate Protobuf schemas
 
 ```bash
 cargo run --bin foxglove-proto-gen
 ```
 
+### Generate schemas for the Python SDK
+
+```bash
+yarn generate
```
amacneil, line 19: The changes to this file and comments about Python don't seem relevant to the rust `foxglove` package

scripts/generate.ts
```diff
@@ -62,12 +72,24 @@ async function main({ clean }: { clean: boolean }) {
...
 
+  const pythonSdkRoot = path.resolve(repoRoot, "python", "foxglove-sdk");
+  const pythonSdkGeneratedRoot = path.join(pythonSdkRoot, "src", "generated");
+  const pythonSdkStub = path.join(
+    pythonSdkRoot,
+    "python",
+    "foxglove",
+    "_foxglove_py",
+    "schemas.pyi",
+  );
```
amacneil, line 83: fyi with `path.join` you can just use `/` directory separators, it will work cross platform (also I doubt this generate script will work on windows anyway, with all the subcommands it calls)

amacneil:
```suggestion
  const pythonSdkRoot = path.resolve(repoRoot, "python/foxglove-sdk");
  const pythonSdkGeneratedRoot = path.join(pythonSdkRoot, "src/generated");
  const pythonSdkStub = path.join(pythonSdkRoot, "python/foxglove/_foxglove_py/schemas.pyi");
```

.github/workflows/python.yml
bryfox: Oh yeah, much better

scripts/generate.ts
bryfox: Thanks, I pushed that change after seeing your comment but before the suggestion here.

amacneil (approved):

gasmith (approved):

PR #208, opened 2025-02-12
rust: Encapsulate client message parsing

gasmith:
It seemed to me like too much protocol parsing work was happening in `websocket`, so I pushed it down into `websocket::protocol::client`. I also filled in the remaining message types and added unit tests with examples copied from the ws-protocol spec.

Side-conversation: Why did we choose to newtype our various IDs (`ChannelId`, `SubscriptionId`, etc.)? They're sort of annoying to work with. Could we use type aliases instead?


eloff:
> Side-conversation: Why did we choose to newtype our various IDs (`ChannelId`, `SubscriptionId`, etc.)? They're sort of annoying to work with. Could we use type aliases instead?

If you use type aliases, it's easy to pass an integer as a ChannelId, but if you use new type then you get more type safety around it and have to make conversions to and from integers explicit.

The annoying part is the feature, not a bug.

We could decide we value convenience more than the type safety in this specific case, in which case type aliases would be better. They started off that way as type aliases. We can discuss it.

rust/foxglove/src/websocket/protocol/client.rs
```diff
@@ -2,9 +2,128 @@
...
+            ClientMessage::SubscribeParameterUpdates(_) => "subscribeParameterUpdates",
+            ClientMessage::UnsubscribeParameterUpdates(_) => "unsubscribeParameterUpdates",
+            ClientMessage::ServiceCallRequest(_) => "serviceCallRequest",
+            ClientMessage::SubscribeConnectionGraph => "subscribeConnectionGraph",
+            ClientMessage::UnsubscribeConnectionGraph => "unsubscribeConnectionGraph",
+            ClientMessage::FetchAsset(_) => "fetchAsset",
+        }
+    }
+
+    pub fn parse_message(message: Message) -> Result<Option<Self>, ParseError> {
```
eloff, line 58: I think the original reason this was in websocket.rs was to keep the websocket Message type and logic segregated. I don't think that's particularly important, and I'm fine with doing it this way. I'm just offering that as context.

rust/foxglove/src/websocket/protocol/client.rs
```diff
@@ -96,30 +215,169 @@ pub struct ClientChannel {
...
+}
 
-    let channel_bytes = (&msg[1..=4]).try_into().unwrap();
-    let channel_id = ClientChannelId::new(u32::from_le_bytes(channel_bytes));
-    Ok((channel_id, &msg[5..]))
+// https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#set-parameters
+#[derive(Debug, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct SetParameters {
+    pub parameters: Vec<SetParameter>,
```
eloff, line 261: This was Vec<server::Parameter>, which is a different type. ParameterValue and serde_json::Value are different. We have to convert this to a server::Parameter in the work I'm currently doing, and having them share a type is actually helpful (no conversion needed.)

What's the motivation for changing this?

eloff (commented):
I have one question for you, otherwise this looks pretty good

rust/foxglove/src/websocket/protocol/client.rs
eloff: Actually it didn't exist before, I added it :rofl: 

You can merge it like this if you want, and I'll change it to `Vec<server::Parameter>`  and delete SetParameter when I merge my code, unless you know of a good reason to do it this way instead.

gasmith: The client message didn't exist before, AFAICT. I figured you'd be working in this area, so I was going for a "least effort" solution. I had totally forgotten about the `Parameter` type, and I completely agree about reusing it.

eloff: Yep, you're right

eloff (approved):
LGTM

PR #210, opened 2025-02-12
Python SDK: integrate new channels and schemas

bryfox:
This PR integrates the new generated Schema and typed Channel classes from #201 and #206 respectively, and updates the examples accordingly. 

As part of this, I've adjusted the Schema classes to make them easier to work with:
- fields are now `Optional`, with defaults. Callers can omit any properties during construction.
- constructors are keyword-only, which should be more future-proof.

The protobuf dependency is no longer needed in the Python SDK.

For discussion: I've also re-worked the high-level API a bit. It works with any given foxglove Schema now (though could use better typing). But it no longer works with arbitrary protobuf messages. We could support that, but I'm not sure how valuable it is. We could support arbitrarily serialized data, but then you'd need to provide your schema, and the channel interface seems much nicer.


python/foxglove-sdk/python/examples/geometry/euler_to_quaternion.py
```diff
@@ -1,6 +1,6 @@
 from math import cos, sin
 
-from foxglove.schemas import Quaternion
+from foxglove._foxglove_py.schemas import Quaternion
```
amacneil, line 3: can we re-export these, consumers shouldn't be importing anything from `_foxglove_py` directly

python/foxglove-sdk/python/examples/ws-demo.py
```diff
@@ -29,15 +30,15 @@ def main() -> None:
...
+    tf_chan = FrameTransformChannel("/tf")
+
+    # Log JSON messages using the JsonEncoder.
+    # A ProtobufEncoder is also available.
     sin_chan = foxglove.Channel(
         topic="/sine", encoder=foxglove.JsonEncoder(), schema=plot_schema
     )
-    box_chan = foxglove.Channel(
-        topic="/boxes", encoder=foxglove.ProtobufEncoder(), schema=SceneUpdate
-    )
```
amacneil, line 37: did we lose the 3d examples? plan to bring them back?

python/foxglove-sdk/python/examples/ws-demo.py
```diff
@@ -29,15 +30,15 @@ def main() -> None:
 
     server = foxglove.start_server(port=8765)
 
+    # Log messages having well-known Foxglove schemas using the appropriate channel type.
+    box_chan = SceneUpdateChannel("/boxes")
+    tf_chan = FrameTransformChannel("/tf")
+
+    # Log JSON messages using the JsonEncoder.
+    # A ProtobufEncoder is also available.
```
amacneil, line 38: your PR comment said that ProtobufEncoder is not available - did I misunderstand that?

bringing this back to our use cases: for logging arbitrary structs I don't think it matters to the user what the wire encoding is, we should just pick something under the hood.

for users who have some existing pubsub system, they should have existing serialization logic so can pass us preserialized messages and schemas directly.

python/foxglove-sdk/pyproject.toml
```diff
@@ -27,7 +27,6 @@ package-mode = false
 requires-poetry = ">=2.0"
 
 [tool.poetry.group.dev.dependencies]
-types-protobuf = "^5.28.3.20241030"
```
amacneil, line 30: might be some stuff you can clean up in gitignore now too

python/foxglove-sdk/python/examples/ws-demo.py
bryfox: Still here. I moved the Foxglove schema examples to the top; this one now looks like:

```py
box_chan = SceneUpdateChannel("/boxes")
```

python/foxglove-sdk/python/examples/ws-demo.py
bryfox: 👍 I will update.

amacneil (approved):

bryfox:
High-level API and examples will get an update in a future PR per discussion.

PR #212, opened 2025-02-13
rust: Support the rest of parameter protocol

eloff:
This fills out support for the remainder of the parameter protocol.

I tried to follow the behavior of the Python server, with the exception of the bug I uncovered, which is fixed here. on_parameters_subscribe and on_parameters_unsubscribe is only called when a parameter is first subscribed to by a client, or last unsubscribed to. This corresponds to when you first need to worry about it, and when you can stop worrying about it, without having to do that bookkeeping yourself.

Something I copied, but am unsure of, is how set_parameters will not update parameters if there's no listener installed for the server. Should that be the behavior?

Marked as a draft because I still need to add tests.


rust/foxglove/src/websocket.rs
```diff
@@ -175,6 +175,28 @@ pub trait ServerListener: Send + Sync {
...
+    /// Callback invoked when a client requests parameters. Requires the "parameters" capability.
+    fn on_get_parameters(
+        &self,
+        _client: Client,
+        _param_names: Vec<String>,
+        _request_id: Option<&str>,
+    ) -> Vec<Parameter> {
+        Vec::new()
+    }
+    /// Callback invoked when a client sets parameters. Requires the "parameters" capability.
```
gasmith, line 187: It looks like the server implementation automatically notifies all subscribers. That feels like an important thing to mention, so that someone implementing this method doesn't feel compelled to call `server.publish_parameter_values`.

rust/foxglove/src/websocket.rs
```diff
@@ -470,6 +507,85 @@ impl ConnectedClient {
...
+            let request_id = request_id.as_deref();
+            let updated_parameters =
+                handler.on_set_parameters(Client(self), parameters, request_id);
+            // Send all the updated_parameters back to the client if request_id is provided.
+            // This is the behavior of the reference Python server implementation.
+            if request_id.is_some() {
+                let message = protocol::server::parameters_json(&updated_parameters, request_id);
+                self.send_control_msg(Message::text(message));
+            }
+            // Send the subscribed parameters to the client if the client is subscribed to them.
```
gasmith, line 535:
```suggestion
            // Send parameter updates to all subscribed clients.
```

rust/foxglove/src/websocket.rs
```diff
@@ -470,6 +507,85 @@ impl ConnectedClient {
...
+        }
+        // Get the list of parameter names that previously had no subscribers
+        let new_param_subscriptions = server.parameters_without_subscription(param_names.clone());
+        {
+            let mut subscribed_parameters = self.parameter_subscriptions.lock();
+            subscribed_parameters.extend(param_names);
+        }
+        if let Some(handler) = self.server_listener.as_ref() {
+            handler.on_parameters_subscribe(Client(self), new_param_subscriptions);
+        }
```
gasmith, line 573: This is racy. Two clients that subscribe to the same parameter might both trigger `on_parameters_subscribe` for that same parameter.

rust/foxglove/src/websocket.rs
```diff
@@ -470,6 +507,85 @@ impl ConnectedClient {
...
+        // First filter param_names down to only the ones the client is subscribed to.
+        {
+            let mut subscribed_parameters = self.parameter_subscriptions.lock();
+            param_names.retain(|name| subscribed_parameters.remove(name));
+        }
+        // Get the list of parameter names that now have no subscribers
+        let unsubscribed_parameters = server.parameters_without_subscription(param_names);
+        if let Some(handler) = self.server_listener.as_ref() {
+            handler.on_parameters_unsubscribe(Client(self), unsubscribed_parameters);
+        }
```
gasmith, line 586: This is racy in the same way as subscribe: you might get multiple `on_parameters_unsubscribe` calls for the same parameter.

But it also races _with_ subscribe in a much more dangerous way. Suppose a client A subscribes to a parameter X, and a client B concurrently unsubscribes from that same parameter. Consider the following sequence:

- A calls `server.parameters_without_subscription([X])`, gets `[]`
- B removes `X` from its subscribed parameters
- B calls `server.parameters_without_subscription([X])`, gets `[X]`
- A adds `X` to its subscribed parameters
- A calls `on_parameters_subscribe([])`
- B calls `on_parameters_unsubscribe([X])`

Now the listener thinks that no one is subscribed to parameter X, when in fact A is subscribed to parameter X.

I think you're going to need a synchronized structure on the server to keep track of changes to the union of subscriptions.

rust/foxglove/src/websocket.rs
```diff
@@ -470,6 +507,85 @@ impl ConnectedClient {
...
+            param_names.retain(|name| !subscribed_parameters.contains(name));
+        }
+        // Get the list of parameter names that previously had no subscribers
+        let new_param_subscriptions = server.parameters_without_subscription(param_names.clone());
+        {
+            let mut subscribed_parameters = self.parameter_subscriptions.lock();
+            subscribed_parameters.extend(param_names);
+        }
+        if let Some(handler) = self.server_listener.as_ref() {
+            handler.on_parameters_subscribe(Client(self), new_param_subscriptions);
```
gasmith, line 572: If this is supposed to represent (to the listener) the union of all client subscriptions, we don't need to pass the client identity.

rust/foxglove/src/websocket.rs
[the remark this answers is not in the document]
eloff: There isn't an invalid value in that we don't validate it, if it deserializes from JSON, it's "valid"

If there is no server handler installed, set parameters specifically is silently ignored. 

I can document these things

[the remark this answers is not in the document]
gasmith: By invalid, I mean a value that the implementation of the trait wants to reject. Like setting an integer parameter to a string value, or setting a floating point parameter to a value that's out of acceptable range.

rust/foxglove/src/websocket.rs
eloff: Good catch. The solution is probably going to suck...

rust/foxglove/src/websocket.rs
eloff: Great catch! I'm going to have to rethink this, I think it probably will need a synchronized structure on the server :(

rust/foxglove/src/websocket.rs
eloff: The client is the one that triggered the notification, I'm not sure if it's ever useful to know that though. I did it more out of consistency with the other callbacks. "A foolish consistency is the hobgoblin of little minds"...

I think I'll remove it then.

rust/foxglove/src/websocket.rs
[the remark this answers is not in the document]
eloff: It should keep the old value and return the old value (or no value if there wasn't an old value)

There's no way to provide an error back to the client via this interface. If we did implement that it would probably be via a method on the passed in Client object that lets you publish a status to just that client.

rust/foxglove/src/websocket.rs
```diff
@@ -470,6 +513,144 @@ impl ConnectedClient {
...
+            // We didn't remove any subscriptions
+            return;
+        }
+        let Some(handler) = self.server_listener.as_ref() else {
+            return;
+        };
+
+        let unsubscribed_parameters = {
+            let mut subscribed_parameters = server.subscribed_parameters.lock();
+            // With the server-wide lock held, which prevents new subscriptions,
```
gasmith, line 643: An alternative would be to keep a `Mutex<HashMap<String, usize>>` on the server, where the value is the number of clients. That way you don't need to worry about the order in which we update the per-server and per-client sets.

rust/foxglove/src/websocket.rs
```diff
@@ -470,6 +513,144 @@ impl ConnectedClient {
...
+            // get the parameters without any subscibers
+            let mut unsubscribed_parameters = server.parameters_without_subscription(param_names);
+            // Remove the unsubscribed parameters from the server's list of subscribed parameters
+            unsubscribed_parameters.retain(|name| subscribed_parameters.remove(name));
+            unsubscribed_parameters
+        };
+
+        handler.on_parameters_unsubscribe(unsubscribed_parameters);
```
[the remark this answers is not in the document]
eloff: You're right, we do want to call both callbacks, but the order matters. Yuck.

I think I can just fix it by holding the server.subscribed_parameters lock for the whole duration in both subscribe and unsubscribe.

rust/foxglove/src/websocket.rs
```diff
@@ -776,6 +954,7 @@ impl Server {
             control_plane_rx: ctrl_rx,
             subscriptions: parking_lot::Mutex::new(BiHashMap::new()),
             advertised_channels: parking_lot::Mutex::new(HashMap::new()),
+            parameter_subscriptions: parking_lot::Mutex::new(HashSet::new()),
```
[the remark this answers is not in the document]
eloff: Nope, I missed I it. Thanks!

rust/foxglove/src/websocket.rs
eloff: Yeah I considered that, but it's more of a pain to manage

eloff:
@bryfox could you take over review of this PR while Greg's out, I addressed everything he called out, so hopefully it's a quickish review.

rust/foxglove/src/websocket.rs
```diff
@@ -470,6 +544,157 @@ impl ConnectedClient {
...
+        if self.server_listener.is_some() {
+            for name in param_names.iter() {
+                if all_subscriptions.insert(name.clone()) {
+                    new_param_subscriptions.push(name.clone());
+                }
+            }
+        }
+
+        // Track the client's own subscriptions
+        let mut client_subscriptions = self.parameter_subscriptions.lock();
```
bryfox, line 648: I see why the lock on server subscriptions is held across the callback, but why this one too?

rust/foxglove/src/lib.rs
```diff
@@ -225,7 +225,6 @@ pub enum FoxgloveError {
     McapError(#[from] mcap::McapError),
     /// An error related to JSON encoding.
     #[doc(hidden)]
```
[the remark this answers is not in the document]
eloff: Probably, I'll double check

rust/foxglove/src/websocket.rs
eloff: Yeah, good catch

rust/foxglove/src/lib.rs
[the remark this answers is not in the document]
eloff: We don't seem to actually use it, maybe it can just be removed.

bryfox (approved):
I assume the merge conflicts will be straightforward to resolve, but I can take another look after that if you'd like.

PR #213, opened 2025-02-13
rust: Add support for websocket services

gasmith:
This change adds support for hosting custom services on the websocket server. As far as I can tell, this is mostly intended to support the ros-foxglove-bridge.

This turned into a pretty large change, but I'm not sure that splitting it up into advertisement vs. call and response (as the linear tickets are arranged) would've made things any easier.

The basic ideas:
 - Add the `Service` type (id, name, schema, handler).
 - Store an <Id, Service> map on the websocket server.
 - Provision a buffered response channel for each client.
 - Advertise and unadvertise services as necessary.
 - Handle calls:
   - Allocate a response (or reject with "too many requests")
   - Lookup service by ID (or reject with "unknown service")
   - Invoke handler.

The handler itself is a flexible trait, so that it can be as simple as a `Fn` closure, or as complex as you like. At the moment, the handler runs synchronously from the context of the client poll loop, but you can always `tokio::spawn` your way to a less-constrained environment.

While writing tests (and the perhaps-overblown example), I realized that I really _really_ wanted a full-featured crate for round-tripping ws-protocol messages. I think we might want to pull `websocket::protocol` out into its own `foxglove-ws-protocol` crate. Any objections?

And because it's a bit buried, you can check out the service registration demo [here](https://github.com/foxglove/foxglove-sdk/pull/213/files#diff-2dc63a716bbefb10ba159933f2a1d3efec9caa0c5cde4bac7b39073ec726d81bR24-R72).

Fixes: FG-9672
Fixes: FG-9674


gasmith:
@jtbandes you might want to take a look at this one, since I hear you have some familiarity with the ros-foxglove-bridge, which is the target use case for this feature.

rust/foxglove/src/channel.rs
```diff
@@ -56,6 +56,16 @@ impl Schema {
...
         }
     }
+
+    /// Returns a JSON schema for the specified type.
+    pub fn json_schema<T: schemars::JsonSchema>() -> Self {
+        let json_schema = schemars::schema_for!(T);
+        Self::new(
+            std::any::type_name::<T>(),
+            "jsonschema",
+            serde_json::to_vec(&json_schema).expect("Failed to serialize schema"),
```
bryfox, line 66: I know this predates the PR, but is `expect` what we want here? Is it OK because of some guarantees schemars makes?

rust/foxglove/src/websocket.rs
```diff
@@ -470,6 +477,45 @@ impl ConnectedClient {
...
+        };
+
+        // Lookup the requested service handler.
+        let Some(service) = server.get_service(service_id) else {
+            self.send_service_call_failure(service_id, call_id, "Unknown service");
+            return;
+        };
+
+        // Prepare a response, or fail if this client has too many concurrent requests.
+        let Some(responder) = resp_chan.prepare_response(service_id, call_id, &req.encoding) else {
```
bryfox, line 500: The server has a set of `supported_encodings`. If `req.encoding` isn't supported, maybe we can bail on behalf of the handler.

bryfox (commented):
I haven't gotten through all of this yet, but it looks great so far; I left a couple of small comments. It would be great to document at a high-level how this handles services — not unlike your PR description; maybe added to the Server.

> I realized that I really really wanted a full-featured crate for round-tripping ws-protocol messages

Yeah, writing tests around clients has been time-consuming. I'm assuming it wouldn't need to be published?

rust/foxglove/src/websocket.rs
gasmith: Oh, nice. Yeah, I'll work that in.

rust/foxglove/src/websocket.rs
```diff
@@ -887,6 +977,132 @@ impl Server {
...
+            // Ensure that the new service names are not already registered.
+            let mut services = self.services.write();
+            for service in services.values() {
+                if new_names.contains_key(service.name()) {
+                    return Err(FoxgloveError::DuplicateService(service.name().to_string()));
+                }
+            }
+
+            // Update the service map.
+            services.extend(new_services.into_iter().map(|s| (s.id(), Arc::new(s))));
```
jtbandes, line 1034: probably many hundreds of pages of debate have happened already on this subject, but IMO this is less readable than a for loop that calls `insert()` 😅

jtbandes (commented):
Some initial thoughts - haven't gotten through all the ResponseChannel stuff yet

rust/foxglove/src/websocket/service.rs
```diff
@@ -0,0 +1,145 @@
...
+pub use response::Responder;
+pub(crate) use response::ResponseChannel;
+pub(crate) use schema::MessageSchema;
+pub use schema::ServiceSchema;
+
+/// A service ID, which uniquely identifies a service hosted by the server.
+pub type ServiceId = u32;
+
+/// A service call ID, which uniquely identifies an outstanding call for a particular client.
+pub type CallId = u32;
```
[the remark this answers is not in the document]
gasmith: I did think about it. I'm not a big fan of using newtypes for IDs. It feels overly pedantic, involves a ton of boilerplate (a million derives, plus custom impls for `Display` and `From`), and generally feels less ergonomic. The benefit is that you get type-checking on the ID type, which is nice when you're juggling lots of IDs.

Since we use newtypes for IDs elsewhere in the SDK, I can follow the same pattern here. Maybe we can alleviate some of the boilerplate pain with a macro.

rust/foxglove/src/websocket/protocol/server.rs
```diff
@@ -208,6 +215,85 @@ pub fn unadvertise(channel_id: ChannelId) -> String {
...
+            schema_encoding: &schema.encoding,
+            schema: &schema.data,
+        }
+    }
+}
+
+// https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#advertise-services
+pub(crate) fn advertise_services<'a, II>(services: II) -> String
+where
+    II: IntoIterator<Item = &'a Service>,
```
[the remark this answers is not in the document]
gasmith: The only downside of using `Iterator` here would be punting the responsibility of constructing the iterator to the caller:

```rust
let services: Vec<Service> = ...;

// With `IntoIterator`
let msg = advertise_services(&services);

// With `Iterator` (also trivially compatible with `IntoIterator`)
let msg = advertise_services(services.iter());
```

For public APIs, I prefer `IntoIterator` for the flexibility. This is a private function, but I happened to have one callsite with a `Vec<Service>` and another callsite with an iterator.

rust/foxglove/src/websocket/service.rs
```diff
@@ -0,0 +1,145 @@
...
+pub struct ServiceBuilder {
+    id: ServiceId,
+    name: String,
+    schema: ServiceSchema,
+}
+impl ServiceBuilder {
+    /// Creates a new builder for a websocket service.
+    fn new(name: impl Into<String>, schema: ServiceSchema) -> Self {
+        static ID: AtomicU32 = AtomicU32::new(1);
+        let id = ID.fetch_add(1, Ordering::Relaxed);
```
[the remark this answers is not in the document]
gasmith: Maybe a misread? The caller isn't providing the ID here, we're picking one internally with the static atomic. There's a `with_id` method, but that's only for deterministic testing (and it's under a `#[cfg(test)]` attribute).

rust/foxglove/src/websocket/service.rs
```diff
@@ -0,0 +1,145 @@
...
+            .field("id", &self.id)
+            .field("name", &self.name)
+            .field("schema", &self.schema)
+            .finish_non_exhaustive()
+    }
+}
+
+impl Service {
+    /// Creates a new builder for a websocket service.
+    pub fn builder(name: impl Into<String>, schema: ServiceSchema) -> ServiceBuilder {
```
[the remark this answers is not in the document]
gasmith: Caller ergonomics. The `impl Into<String>` signals our intention to convert this into an owned string, and so if the caller already has one they can provide that (for free with no conversion overhead). But if the caller has a reference (like a `&'static str`), I don't want to force them to type out `.into()` or `.to_string()` on their end.

```rust
// I'd rather write this:
let service = Service::builder("my-service", schema);

// Than this:
let service = Service::builder("my-service".to_string(), schema);
```

If this was being used in a performance sensitive path, I might be slightly more opinionated about encouraging callers to handle their own conversions/ownership, but this is definitely not on the hot path.

rust/foxglove/src/websocket/service/handler.rs
```diff
@@ -0,0 +1,76 @@
...
+
+/// A websocket service call handler.
+pub trait Handler: Send + Sync {
+    /// Handles a service call request from a client.
+    ///
+    /// The caller can choose whether to handle the call synchronously or asynchronously.
+    ///
+    /// This method is invoked from the client's main poll loop and must not block. If blocking or
+    /// long-running behavior is required, the implementation should use [`tokio::task::spawn`] (or
+    /// [`tokio::task::spawn_blocking`]) to handle the request asynchronously.
```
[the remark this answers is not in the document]
gasmith: This isn't really a rust thing, per se, but there are some rusty elements.

In abstract: we have a single poll loop for servicing the websocket connection. When we get a service call request, we have a choice about whether we want to handle it synchronously, or spawn a task to handle it asynchronously. I chose the former, because I don't want to force an extra context switch for situations where it isn't necessary (e.g., and very simple handler, or an implementation that simply queues the request on a channel). The tradeoff is that you have somebody else's code running in the connection poll loop, and so you have to warn the user about the consequences of blocking for too long.

The rusty part: Async methods in traits are a [new-ish thing](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html), and they come with [ various caveats](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html#--impl-trait-in-public-traits), and particularly [they cannot be used for dynamic dispatch](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html#dynamic-dispatch), which is a bit of a deal-breaker here.

But really, I don't think we want an async interface here in the first place, because we don't want to do `service.call(...).await` from the connection poll loop anyway: that blocks the poll loop (or gives us a ton of futures to `select!` on, which sounds equally unpleasant). If we know that the call implementation is going to block, we're probably going to spawn a task for it.

And that raises the question of what kind of task to spawn, and which runtime to use. If it's an async handler, we should use `tokio::spawn`. If it's a CPU-intensive handler, we should use [`tokio::task::spawn_blocking`](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html). The connection poll loop is running on a tokio runtime, but a program may have multiple runtimes for isolation, and maybe they want to spawn the service handler on a different (non-tokio?) runtime for some reason.

So to summarize, we have a bunch of options:
1. Synchronous callback
2. `tokio::spawn`
  a. On our runtime
  b. On a different runtime
3. `tokio::spawn_blocking`
  a. On our runtime
  b. On a different runtime

I opted for 1, because it provides the greatest flexibility and performance, for the least amount of API complexity. I don't want to provide a bunch of wrappers around spawning tasks, because (a) that might end up being too limiting anyway and (b) it might give users the wrong impression about how things work. Users would do better to read the tokio documentation directly, rather than our regurgitation of it.

rust/foxglove/src/channel.rs
eloff: Yeah, this is a good use case for a panic. This error would only ever happen if the schemars crate was completely bugged. Developer error, not a runtime error.

rust/foxglove/src/websocket/service.rs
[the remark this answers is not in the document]
eloff: We can discuss if we want to use new type everywhere for the ids or not, but once we've decided we stick with it.

I lean towards the extra type-safety, which I think is how rust itself leans.

rust/foxglove/src/websocket/protocol/server.rs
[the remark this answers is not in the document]
eloff: IntoIterator includes Iterator, because all iterators implement IntoIterator for themselves. So it's strictly more flexible.

Could we express this as services: `impl IntoIterator<Item = &'a Service>`, or is this a limitation of that syntax?

rust/foxglove/src/websocket/service/request.rs
```diff
@@ -0,0 +1,20 @@
...
+
+use super::{CallId, ServiceId};
+
+/// A service call request.
+#[derive(Debug, Clone)]
+pub struct Request {
+    /// Service ID.
+    pub service_id: ServiceId,
+    /// Service name.
+    pub service_name: String,
```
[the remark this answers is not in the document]
gasmith: I was thinking about the implementation of a shared service handler, such as you'd expect to find in the ros-foxglove-bridge, or [this example](https://github.com/foxglove/foxglove-sdk/pull/213/files#diff-2dc63a716bbefb10ba159933f2a1d3efec9caa0c5cde4bac7b39073ec726d81bR106-R123).

The shared implementation needs some way to demux, and it doesn't care about service IDs. It cares about service names. So we should provide the service name. I wanted the `Request` to be `Send` so that it could be handled asynchronously without wrapping it.

I suppose an alternative would be to just pass a clone of the `Arc<Service>` itself. Maybe that's a better way to go.

rust/foxglove/src/websocket/service/response.rs
```diff
@@ -0,0 +1,164 @@
...
+
+#[cfg(test)]
+mod tests;
+
+/// The default maximum number of concurrent requests from a particular client.
+const DEFAULT_RESPONSE_CHANNEL_CAPACITY: usize = 32;
+
+/// A bounded response channel.
+pub(crate) struct ResponseChannel {
+    tokens: Arc<Mutex<usize>>,
```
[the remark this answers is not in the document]
gasmith: Hah, whoops. Yes, that's a clear improvement. Nice catch.

rust/foxglove/src/websocket/service/handler.rs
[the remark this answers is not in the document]
eloff: We have the same problem in all of the ServerListener callbacks.

We made them sync so they work in sync code, and generally they are expected to complete quickly. But the implementer would need to use spawn/spawn_blocking for that code.

We should duplicate these docs for the ServerListener trait.

rust/foxglove/src/websocket/service.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+}
+
+/// A websocket service.
+#[must_use]
+pub struct Service {
+    id: ServiceId,
+    name: String,
+    schema: ServiceSchema,
+    handler: Arc<dyn Handler>,
```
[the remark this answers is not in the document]
gasmith: > I guess I would pose the question, if you were able to delete all the Handler related stuff, do you feel like that would be a win for simplicity?

I don't think it would be a significant win, no. There's a lot of ceremony in `handler.rs` related to setting up implementations of `Handler` for wrappers around `Fn`, but all of that complexity is completely self-contained and private to the crate. And while it's ugly for all of its trait-bound & lifetime madness, it's not really all that complex - just ceremony.

We'd also get rid of `ServiceBuilder` and just have `Service`... so in the public-facing interface you'd have:

```rust
// With flexible handlers
let service = Service::builder(name, schema).handler(foo);

// Without
let service = Service::new(name, schema);
```

The downside is that you're forcing all services to share the same handler. With flexible handlers, that's still a choice. You could even roll it into the same object you use to implement `ServerListener`, if you want:

```rust
struct Listener { ... }
impl foxglove::websocket::ServerListener for Listener { ... }
impl foxglove::websocket::service::Handler for Listener { ... }
let listener: Arc<Listener> = ...;

// With flexible handlers
let mut services = vec![];
for (name, schema) in service_definitions {
    let service = Service::builder(name, schema).handler(listener.clone());
    services.push(service);
}
ws_server.add_services(services);

// Without
let mut services = vec![];
for (name, schema) in service_definitions {
    let service = Service::new(name, schema);
    services.push(service);
}
ws_server.add_services(services);
```

rust/foxglove/src/websocket.rs
gasmith: Yeah, seems to be a hot topic around these parts. ;)

I'll pull it apart.

rust/foxglove/src/websocket/protocol/server.rs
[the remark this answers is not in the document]
gasmith: > Could we express this as services: impl IntoIterator<Item = &'a Service>, or is this a limitation of that syntax?

Oh, for sure. Not sure why I didn't do that.

rust/foxglove/src/websocket.rs
```diff
@@ -821,13 +892,31 @@ impl Server {
...
+                            new_client.drain_queues();
+                        }
+                    }
+                }
+            } else {
+                // No services capability, no problem.
+                std::future::pending().await
+            };
+        };
+
```
eloff, line 919: Can't we use the control plane for service messages via send_control_msg? I don't see why we need a third type of queue here. The behavior of try_send or disconnect is fine for service calls too.

rust/foxglove/src/websocket.rs
eloff: Rust vs Go in a nutshell. I'm old school, I like the loop.

rust/foxglove/src/websocket/service.rs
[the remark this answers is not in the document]
eloff: Yes, this is my main concern here as well.

This adds a LOT of code across many types to support easily adding multiple services with add_services. It's a lot for us to understand and maintain. This review was no joke.

However, the burden for the end-user seems small, and we do the dispatch for multiple services, instead of the user, as is currently the case. That is nice for the user.

rust/foxglove/src/websocket.rs
eloff, line 500: The response encoding is always just a copy of the request encoding. Having the request and responder as two different objects means we have to create a copy of it.

If we had one object (Request?), and the responder functionality lived on it, then you don't have the need to duplicate the service id, call id and encoding. Also one less type for the user to read about in the docs and understand. What do you think of that?

The fasthttp library in Go does something like this by combining the http request and response objects into one RequestCtx object.

rust/foxglove/src/websocket/service/response.rs
```diff
@@ -0,0 +1,164 @@
...
+/// Represents a reservation for the client's service response queue.
+///
+/// Increments the inner counter (a reference to [`ResponseChannel::tokens`]) when dropped.
+struct Token(Arc<Mutex<usize>>);
+
+impl Drop for Token {
+    fn drop(&mut self) {
+        let mut tokens = self.0.lock();
+        #[cfg(debug_assertions)]
+        assert_ne!(*tokens, usize::MAX);
```
eloff, line 96: debug_assert_ne!

but not required because += 1 is checked_add in a debug build anyway and will panic on overflow

rust/foxglove/src/websocket.rs
```diff
@@ -887,6 +977,132 @@ impl Server {
...
+                }
+            }
+        }
+    }
+
+    /// Adds new services, and advertises them to all clients.
+    ///
+    /// This method will fail if the services capability was not declared, or if a service name is
+    /// not unique.
+    pub async fn add_services(&self, new_services: Vec<Service>) -> Result<(), FoxgloveError> {
```
eloff, line 1004: This doesn't need to be async. The only thing async it calls is control_plane_tx.send_async, which only suspends when the queue is full. You can use send_control_msg and slow clients will get disconnected. Then both the blocking and non-blocking handles can call it directly.

We'll make the queue depths configurable in the future so the user can adjust that if need be.

eloff (changes requested):

rust/foxglove/src/websocket.rs
gasmith: I had a few things that I wanted:

- Limit the absolute service call concurrency on a per-client basis.
- Fail fast, instead of doing a bunch of throw-away work in the service call handler.
- Give clients have an opportunity to recover from "too many requests", instead of just disconnecting them. This is a different situation from a client that can't keep up with the rate of messages it's receiving from the server.

I also figured that we'll eventually rework some of our outbound queuing, so I didn't want to make changes to the existing queues. Maybe we can revisit this in the future, and find a way to merge some of these concepts?

gasmith: Also, I think we want to try _VERY_ hard to deliver service call responses.

Consider the implementation of the client. When the client makes a call, they're awaiting a response. If the server silently drops the response because its queues happen to be full, the client will block forever.

eloff: I think you still achieve all of those objectives - I'm not asking to change the ticket system for the request semaphore.

The difference is that if the control queue fills, regardless of what the source of those messages are, we disconnect the client. The message doesn't get dropped like with the data plane. I think this is fine, because the service call limit is so much lower than the control plane queue size that it would only happen for slow clients that are backed up.

The disconnect instead of drop is not implemented yet, but should next week with FG-10441. And we should increase the control queue size to a much larger size in the same PR (and maybe give a way to configure it in the server builder.)

gasmith: OK, that's fair. I'll drop the `ResponseChannel` and replace it with a simple non-blocking semaphore for admission control.

rust/foxglove/src/websocket.rs
jtbandes: > Should we support services that have different request and response encodings?

This seems like more of an @achim-k question, but I think it would be valid to do this, e.g. if the server is willing to accept json requests but always returns ros1-encoded responses or something like that... I have a feeling we may have done some behavior like this in the existing ros bridge, but can't remember off the top of my head

eloff: Is there any harm in unifying them and passing it into sync handler anyway? If the user calls respond AND returns non empty bytes then you ignore it and log it as an error.

Unless I'm missing something I think you could do that.

gasmith: I think it muddies the interface considerably for both the `SyncHandler` and the `Handler`, and the only thing it buys us is reducing the number of arguments for the more expressive `Handler` interface.

```rust
fn sync_handler(client: Client, request: Request) -> Result<Bytes, Error> {
    let payload = request.take_payload().unwrap(); // It sucks that this is an Option.
    request.complete(Ok(payload));
    Ok(Bytes::new()) // Construct a new buffer for no reason?
}
```

Versus:

```rust
fn sync_handler(client: Client, request: Request) -> Result<Bytes, Error> {
    Ok(request.into_payload())
}
```

eloff: Fair enough, leave it as is

rust/foxglove/src/websocket/service.rs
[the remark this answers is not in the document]
gasmith: I went ahead and newtyped.

We should review the newtype IDs we have for consistency, usability, and visibility. I noticed, for example, that we don't currently expose `ChannelId` in a way that clients can name, and that `ClientId` doesn't implement `Display`.

rust/foxglove/src/websocket.rs
gasmith: I've implemented this. But to be honest, I still mildly prefer the old solution, with the dedicated service call response channel. It's more loosely-coupled and modular, since the client doesn't need to care about the semaphore, and the responder doesn't need to care about the client. While it is "one more thing to poll" in the client, I think service calls are sufficiently different from other styles of client interaction to merit their own queue.

eloff (commented):
Just a minor detail to fix up, I think this is almost ready

rust/foxglove/src/websocket.rs
achim-k: > > Should we support services that have different request and response encodings?
> 
> This seems like more of an @achim-k question, but I think it would be valid to do this, e.g. if the server is willing to accept json requests but always returns ros1-encoded responses or something like that... I have a feeling we may have done some behavior like this in the existing ros bridge, but can't remember off the top of my head

The ros bridge currently only supports `ros1msg` for service request/response eoncdings (it additionally supports json for client channels).

eloff (approved):
I think this is ready now. If either @jtbandes or @bryfox could give this PR a final look, we can merge it.

bryfox (approved):
I'm not convinced that the disconnect-instead-of-drop described by FG-10441 should be shared for all of the cases now calling `send_control_msg`. I think there are cases where sending an error (and in turn reporting a 'problem' in the app) is preferable.

I still think some internal documentation would be useful to provide a high-level of how these pieces interact (i.e., the "ceremony" described in the thread above); it took a little while to get up to speed by reading the code alone.

I don't think either of these need to block merging this PR, though.

PR #216, opened 2025-02-13
Update Python example and remove explicit encoders

bryfox:
This updates the Python example and underlying module to better represent where we're headed.

The logging API serves the three use cases below.

1. Custom schemas and serialization.
2. Custom schemas, where the SDK serializes for you. 
3. Messages well-known Foxglove schemas.

This PR adds an example for use case (1), using JSON. I think we could add an example using protobuf, but I don't want to add protobuf again as a dev dependency. We could consider organizing examples like we've done with Rust — each getting its own package.

Previously, the examples made use of provided `Encoder`s. This sort-of addresses use case (2), but for JSON still required users to provide a JSON schema. We plan to address this in a better way (FG-10466).

Use case (2) is already represented by the examples, and is unchanged here.

However, for discussion: I've also kept it easy to work with JSON — particularly for exploratory use cases. You can pass schemas and messages as dicts, and these will be treated as JSON.


python/foxglove-sdk/python/foxglove/channel.py
```diff
@@ -1,43 +1,95 @@
...
+            dictionary. Otherwise, you are responsible for serializing the message.
+        """
+        if isinstance(msg, bytes):
+            self.base.log(msg)
+
+        if self.schema_encoding == "json":
+            self.base.log(json.dumps(msg).encode("utf-8"))
 
-        self.base.log(payload)
+        raise ValueError(f"Unsupported message type: {type(msg)}")
```
gasmith, line 86: I think we're missing some elif/else, or return statements.

python/foxglove-sdk/python/foxglove/channel.py
```diff
@@ -1,43 +1,95 @@
...
+        """
+        Log a message on the channel.
+
+        :param msg: the message to log. If the channel uses JSON encoding, you may pass a
+            dictionary. Otherwise, you are responsible for serializing the message.
+        """
+        if isinstance(msg, bytes):
+            self.base.log(msg)
+
+        if self.schema_encoding == "json":
```
gasmith, line 83: I think we care about the message encoding, not the schema encoding.

python/foxglove-sdk/python/foxglove/channel.py
```diff
@@ -55,3 +107,17 @@ def log(topic: str, message: object) -> None:
         pass
 
     channel.log(message)
+
+
+def _normalize_schema(schema: Union[JsonSchema, SchemaDefinition]) -> SchemaDefinition:
+    if isinstance(schema, SchemaDefinition):
+        return schema
+    elif isinstance(schema, dict):
```
gasmith, line 115: Should we do anything else to validate that this is a valid jsonschema, and not just a random dict?

python/foxglove-sdk/python/foxglove/channel.py
bryfox: Thanks 🤦. Fixed and introduced some basic unit tests, which should now run in CI.

python/foxglove-sdk/python/foxglove/channel.py
bryfox: I added one check, that the root "type" is "object", which the app also checks for. Beyond that, I'm not sure — the app does have additional restrictions, but mcap is basically schema-agnostic. I'd like to revisit this with FG-10466.

gasmith (approved):

PR #218, opened 2025-02-14
Optimize serialization of byte buffers in Python SDK

bryfox:
This PR modifies the generated schema structs to improve the serialization of byte arrays. When a struct `Vec` field is constructed, pyo3 will iterate the vec and copy each element. This change updates the `Vec<u8>` fields to accept a python-native type, from which we copy the bytes at once. This doesn't require changing the protobuf interface in the Rust SDK. There may be further optimizations we can make, but this is pretty low-hanging fruit.

This also adds a point cloud channel to the live viz example, taken from the linked blog post.

Earlier profiling showed that this approach (similar code) gave a speedup of ~2. Writing 1000-pt point clouds in a tight loop to an mcap file increases the publishing throughput 2-3x:

<table>
<tr><th>Before</th>
</tr>
<tr><td>

```sh
duration:  1.333925s
start:     2025-02-14T08:57:23.15778-05:00 (1739541443.157780000)
end:       2025-02-14T08:57:24.491705-05:00 (1739541444.491705000)
compression:
	zstd: [21/21 chunks] [15.37 MiB/1.85 MiB (87.94%)] [1.39 MiB/sec]
channels:
	(1) /pointcloud  1001 msgs (750.42 Hz)   : foxglove.PointCloud [protobuf]
channels: 1
```

</td>
</tr>
<tr><th>After</th></tr>
<tr>
<td>

```
duration:  462.919ms
start:     2025-02-14T08:56:45.933291-05:00 (1739541405.933291000)
end:       2025-02-14T08:56:46.39621-05:00 (1739541406.396210000)
compression:
	zstd: [21/21 chunks] [15.37 MiB/1.52 MiB (90.09%)] [3.29 MiB/sec]
channels:
	(1) /pointcloud  1001 msgs (2162.37 Hz)   : foxglove.PointCloud [protobuf]
```

</td></tr>
</table>

### DIFF
--- a/python/foxglove-sdk/python/examples/live_visualization.py
+++ b/python/foxglove-sdk/python/examples/live_visualization.py
@@ -1,15 +1,26 @@
+import math
+import struct
 import foxglove
 import numpy as np
 import time
 
 from examples.geometry import euler_to_quaternion
-from foxglove.channels import SceneUpdateChannel, FrameTransformChannel
+from foxglove.channels import (
+    FrameTransformsChannel,
+    PointCloudChannel,
+    SceneUpdateChannel,
+)
 from foxglove.schemas import (
     Color,
     CubePrimitive,
     Duration,
     FrameTransform,
+    FrameTransforms,
+    PackedElementField,
+    PackedElementFieldNumericType,
+    PointCloud,
     Pose,
+    Quaternion,
     RawImage,
     SceneEntity,
     SceneUpdate,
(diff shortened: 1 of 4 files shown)


gasmith (approved):
:ship:

amacneil:
Is there a script for running performance tests like that? Would be nice to keep in the repo if its not already.

bryfox:
> Is there a script for running performance tests like that? Would be nice to keep in the repo if its not already.

The point cloud generation was very close to the one included here, but used more points. I wrote the original scripts with cProfile; I'm not sure how much value there is in including that. Maybe worth adding benchmarks more formally, though?

amacneil:
Sounds good I added scope to FG-9514

PR #219, opened 2025-02-14
Update MCAP handling in Python SDK

bryfox:
This fixes the implementation of `PyMcapWriter` to agree with the declared interface. Notably, it exposes the `close` method and documents the auto-closing behavior if a reference to the writer is not held.

To make this easier to work with, this also adds a context-managed function `new_mcap_file`, which frees the user from needing to maintain a reference to the writer. I included 'mcap' in the name since (a) we've decided to expose that name in the Rust SDK, and (b) it makes clear with the output format will be, since we're asking the user to supply a file name.

This also updates the example to produce a basic MCAP file.

### DIFF
--- a/python/foxglove-sdk/python/foxglove/__init__.py
+++ b/python/foxglove-sdk/python/foxglove/__init__.py
@@ -1,5 +1,6 @@
 import atexit
-from typing import Union
+from contextlib import contextmanager
+from typing import Iterator, Union
 from ._foxglove_py import (
     record_file,
     enable_logging,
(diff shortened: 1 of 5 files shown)


python/foxglove-sdk/python/foxglove/__init__.py
```diff
@@ -54,6 +55,19 @@ def verbose_off() -> None:
     disable_logging()
 
 
+@contextmanager
+def new_mcap_file(fname: str) -> Iterator[None]:
```
gasmith, line 59: TIL that `Iterator[T]` is a recommended abbreviation for `Generator[T, None, None]`.

gasmith (approved):
LG! I like the context manager.

PR #220, opened 2025-02-14
Provide read access to schema attrs in Python SDK

bryfox:
This PR makes all schema attributes readable in the Python SDK.

The easiest way to make this work is to add `get_all` to the pyclass definition. I've replaced the wrapping tuple structs with named structs, and now perform the conversion to `foxglove::schemas` in the log method. It's possible users will log the same message multiple times, but I would not expect that for high-frequency data.


python/foxglove-sdk/src/generated/channels.rs
```diff
@@ -58,20 +58,16 @@ impl CameraCalibrationChannel {
         Ok(Self(base))
     }
 
-    fn log(&self, msg: &schemas::CameraCalibration) {
-        self.0.log(&msg.0);
+    fn log(&self, msg: schemas::CameraCalibration) {
```
gasmith, line 61: What does pyo3 do when you pass an object by value like this? Obviously the object is still reachable from the python interpreter after the call, so does that imply that there's an unconditional clone on each call?

gasmith, line 61: We might want to add some annotations to these methods for better documentation:

```
class Vector3Channel(builtins.object)
 |  Vector3Channel(topic)
 |
 |  Methods defined here:
 |
 |  __repr__(self, /)
 |      Return repr(self).
 |
 |  log(self, /, msg)
 |
 |  log_with_meta(self, /, msg, metadata)
 |
 |  ----------------------------------------------------------------------
 |  Static methods defined here:
 |
 |  __new__(*args, **kwargs)
 |      Create and return a new object.  See help(type) for accurate signature.
```

gasmith: Since we're in the business of duplicating all the schema definitions anyway, maybe we replace non-`Copy` types `T` with `&T` so that the clones are cheap, and then pay the cost of copying once when we convert into the underlying protobuf.

bryfox: So keep

```rs
fn log(&self, msg: &schemas::CameraCalibration)
```

but implement the conversion from `&schemas::CameraCalibration` instead?

gasmith: You can keep the argument as pass-by-value, but for all of the fields within `CameraCalibration` that are not `Copy`, make them references.

```rust
struct CameraCalibration<'a> {
    timestamp: Option<Timestamp>,
    frame_id: &'a str,
    width: u32,
    height: u32,
    distortion_model: &'a str,
    D: &'a [f64]
    ...
}
```

`Timestamp`, `Duration`, and all the enums can all derive `Copy`.

I don't know how well that will work in practice. You might have to carry `Bound<>` around?

bryfox: We can't have lifetimes on the pyclass structs: https://pyo3.rs/v0.23.4/class.html#no-lifetime-parameters

I'm realizing this also relates to the optimization in https://github.com/foxglove/foxglove-sdk/pull/218 (merged after this was opened). I'll need some strategy for getting access to the python token since we're no longer doing the conversion during construction.

eloff:
> This PR makes all schema attributes readable in the Python SDK.
> 
> 
> 
> The easiest way to make this work is to add `get_all` to the pyclass definition. I've replaced the wrapping tuple structs with named structs, and now perform the conversion to `foxglove::schemas` in the log method. It's possible users will log the same message multiple times, but I would not expect that for high-frequency data.

I think this is fundamentally the right approach. We can't optimize away copies or other conversion overhead unless we delay the conversion until the log method.

python/foxglove-sdk/src/generated/channels.rs
eloff: The Python channel class log methods are Python calls and should have access to the GIL. Might need some changing how we plumb things around

eloff: Are we able to still pass the object by reference here?

bryfox: I need to revisit this; I'll mark the PR as a draft for now.

I was hoping for a quick workaround here, but it might end up being better to address all of the schemas at once.

bryfox: Done as part of #223

eloff (approved):
Approving this, but I think it should be pass by reference instead of by value in the channel log methods.

PR #222, opened 2025-02-18
rust: inline definitions for derived schemas

bryfox:
This updates the auto-generated JSON schema for custom types to inline definitions, rather than including `$ref`s. Current versions of Foxglove do not support the latter.

This specifies the `draft07` version of JSONSchema, which is currently the default in schemars.

I added an enum to the example to confirm that this works (does not produce a 'problem') in the app.


eloff (approved):
Very nice!

PR #223, opened 2025-02-18
python: improve documentation and auto-generate api docs

bryfox:
This improves the documentation of native types, and adds the `sphinx.ext.autodoc` extension to auto-generate Python module documentation.

We're currently using Sphinx v7 as v8 requires a newer version of Python, and v7 is not able to generate docs from the stub files.

The docs are not yet deployed anywhere but can be built from the python SDK directory using the [documented command](https://github.com/foxglove/foxglove-sdk/blob/f8889ccb7c8a487a16019666c30f38d2ce7f84d2/python/foxglove-sdk/CONTRIBUTING.md?plain=1#L62).


python/foxglove-sdk/src/lib.rs
```diff
@@ -57,6 +66,10 @@ impl Drop for PyMcapWriter {
 
 #[pymethods]
 impl PyMcapWriter {
+    /// Close the MCAP writer.
+    ///
+    /// You must call this when you are done logging, unless you are using the context-managed
+    /// :py:func:`new_mcap_file`, which handles closing the writer for you.
```
eloff, line 72: This doesn't exist yet, should we change this to match the docs for PyMcapWriter until then (closed on going out of scope, but you can close it earlier to surface any errors)?

eloff (approved):
Looks good, one question

python/foxglove-sdk/src/lib.rs
bryfox: It does! From #219, and used in the example here: https://github.com/foxglove/foxglove-sdk/blob/f8889ccb7c8a487a16019666c30f38d2ce7f84d2/python/foxglove-sdk/python/examples/write_log.py#L15

You're right about matching PyMcapWriter — I'll fix this description.

PR #224, opened 2025-02-18
python: broadcast time to clients

bryfox:
This adds support for broadcasting time messages to live clients. The name and signature match Rust.

To support this, the package now exposes a `Capability` enum (currently only containing "time").

I (roughly) ported the mcap playback example from the Rust SDK. The new example is in a `foxglove-sdk-examples` directory outside of the SDK directory, which more closely resembles our Rust example structure. See #199 for motivation. If we like this structure, we can move the other examples in a future PR.

I also made the `start_server` parameters keyword-only.


python/foxglove-sdk-examples/ws-stream-mcap/main.py
```diff
@@ -0,0 +1,128 @@
...
+    _offset_ns: int
+
+    """
+    Helper for keep tracking of the relationship between a file timestamp and the wallclock.
+
+    :param offset_ns: The offset from epoch to treat as "now".
+    """
+
+    def __init__(self, *, offset_ns: int):
+        self._start_time = time.time()
```
eloff, line 106: Not used?

python/foxglove-sdk-examples/ws-stream-mcap/main.py
```diff
@@ -0,0 +1,128 @@
...
+    parser.add_argument("--port", type=int, default=8765)
+    parser.add_argument("--host", type=str, default="127.0.0.1")
+    args = parser.parse_args()
+
+    file_name = args.file
+
+    server = start_server(
+        name=file_name, port=args.port, host=args.host, capabilities=[Capability.Time]
+    )
+    done = False
```
eloff, line 27: You don't need this, keyboard interrupt will break out of the code anyway

eloff (approved):
LGTM, some minor comments

python/foxglove-sdk-examples/ws-stream-mcap/pyproject.toml
```diff
@@ -0,0 +1,18 @@
+[project]
+name = "ws-stream-mcap"
+version = "0.0.0"
+description = "Example of streaming an MCAP file over a WebSocket connection"
+authors = [{ name = "Foxglove", email = "support@foxglove.dev" }]
+readme = "README.md"
+requires-python = ">=3.9"
+dependencies = ["foxglove-sdk @ file:../../foxglove-sdk/", "mcap>=1.2"]
```
bryfox, line 8: I'd planned to list the SDK dependency as the first supported version (whatever our next release is). We can come back and update that if we want to; for now, I'm linking to the relative file — that seems to be what the rust examples do too.

PR #226, opened 2025-02-19
python: Add a server listener that handles client message callbacks

bryfox:
This adds a server listener interface to the Python SDK.

This is implemented as a Protocol in python; pyo3 does not support extending a Python class. To provide reasonable typing on `start_server`, that's now a wrapping function in the public package, to avoid the stub interface referencing the external type.

Since the ws-related code is starting to grow in python, I've moved related classes and functions into `websocket_server.rs` (from lib.rs). The exports are still in the top-level `foxglove` package for now.

To support the message data callback, this also adds support for the `ClientPublish` capability and `supported_encodings`.

I updated the existing live_visualization example to exercise this.

### DIFF
--- a/python/foxglove-sdk/src/websocket_server.rs
+++ b/python/foxglove-sdk/src/websocket_server.rs
@@ -5,7 +5,7 @@ use foxglove::{
 };
 use pyo3::{
     prelude::*,
-    types::{PyBytes, PyDict, PyString},
+    types::{PyBytes, PyString},
 };
 use std::sync::Arc;
 use std::time;
@@ -61,14 +61,11 @@ impl ServerListener for PyServerListener {
                 topic: PyString::new(py, channel.topic()).into(),
             };
 
-            let kwargs = PyDict::new(py);
-            kwargs.set_item("client", client_info)?;
-            kwargs.set_item("channel", channel_view)?;
-            kwargs.set_item("data", PyBytes::new(py, payload))?;
-
+            // client, channel, data
+            let args = (client_info, channel_view, PyBytes::new(py, payload));
             self.listener
                 .bind(py)
-                .call_method("on_message_data", (), Some(&kwargs))?;
+                .call_method("on_message_data", args, None)?;
 
             Ok(())
         });
--- a/python/foxglove-sdk/src/websocket_server.rs
+++ b/python/foxglove-sdk/src/websocket_server.rs
@@ -5,7 +5,7 @@ use foxglove::{
 };
 use pyo3::{
     prelude::*,
-    types::{PyBytes, PyDict},
+    types::{PyBytes, PyDict, PyString},
 };
 use std::sync::Arc;
 use std::time;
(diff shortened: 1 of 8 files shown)


python/foxglove-sdk/src/websocket_server.rs
```diff
@@ -0,0 +1,193 @@
...
+    id: u32,
+}
+
+/// Information about a client channel.
+#[pyclass(name = "ClientChannelView", module = "foxglove")]
+pub struct PyClientChannelView {
+    #[pyo3(get)]
+    id: u32,
+    #[pyo3(get)]
+    topic: String,
```
eloff, line 40: This will create a Python string on every access, and involve at least two copies. One from rust str to owned string and a second on access to python string.

I don't think it will matter much to performance, since these are tiny, but we could save one copy by storing this a Python string and doing the conversion upfront.

Thoughts? We don't have to do this now, it's not a breaking change.

I'm more concerned about setting a good example for future code that we write than this particular case.

python/foxglove-sdk/src/websocket_server.rs
```diff
@@ -0,0 +1,193 @@
...
+            id: client.id().into(),
+        };
+
+        let channel_view = PyClientChannelView {
+            id: channel.id().into(),
+            topic: channel.topic().to_string(),
+        };
+
+        let result: PyResult<()> = Python::with_gil(|py| {
+            let kwargs = PyDict::new(py);
```
eloff, line 64: We're creating this temporary dict here to call every callback. Why not use positional arguments?

The user can still choose to use kwargs if they want, but then at least they're aware of the overhead they're adding.

python/foxglove-sdk/src/websocket_server.rs
```diff
@@ -0,0 +1,193 @@
...
+/// WebSocket. The default connection string matches the defaults used by the SDK.
+#[pyfunction]
+#[pyo3(signature = (*, name = None, host="127.0.0.1", port=8765, capabilities=None, server_listener=None, supported_encodings=None))]
+pub fn start_server(
+    py: Python<'_>,
+    name: Option<String>,
+    host: &str,
+    port: u16,
+    capabilities: Option<Vec<PyCapability>>,
+    server_listener: Option<Py<PyAny>>,
```
eloff, line 102: If we're taking a PyAny here, why do we need to export PyServerListener to Python? I don't think we need it all?

eloff (commented):
A couple questions about this. Overall the approach looks great.

python/foxglove-sdk/src/websocket_server.rs
bryfox: 👍 I changed this to PyString and now construct that directly from the &str.

python/foxglove-sdk/src/websocket_server.rs
bryfox: 👍 updated

python/foxglove-sdk/src/websocket_server.rs
bryfox: Good catch, thanks. Leftover from before I realized it had to be implemented in Python directly.

eloff (approved):
LGTM

PR #227, opened 2025-02-20
python: support status publishing on server

bryfox:
This adds `publish_status` and `remove_status` to the Python server interface, to match the Rust SDK.

I've also updated the existing server methods to not raise exceptions, since the Rust SDK doesn't return resuts. Currently, the only way for these to fail is if the server has been shut down.

This is based on #226 to minimize conflicts (because that re-organized the structure of the server-related code).


python/foxglove-sdk/python/foxglove/tests/test_server.py
```diff
@@ -0,0 +1,17 @@
...
+from foxglove import start_server, Status, StatusLevel
+
+
+class TestServer(unittest.TestCase):
+    def test_server_interface(self) -> None:
+        """
+        Exercise the server interface; will also be checked with mypy.
+        """
+        server = start_server()
+        server.publish_status(Status(StatusLevel.Info, "test", "some-id"))
```
eloff, line 13: Maybe we can just make these function arguments, instead of using a class like rust does. Saves an allocation, which doesn't matter much here. But it also is one less type for the user to care about.

python/foxglove-sdk/src/websocket_server.rs
```diff
@@ -0,0 +1,254 @@
...
+    #[pyo3(signature = (session_id=None))]
+    pub fn clear_session(&self, session_id: Option<String>) {
+        let Some(server) = &self.0 else {
+            return;
+        };
+        server.clear_session(session_id);
+    }
+
+    /// Publishes the current server timestamp to all clients.
+    pub fn broadcast_time(&self, timestamp_nanos: u64) {
```
eloff, line 151: Is this from another PR that you used as a base? Not a big deal, I'll review that one next. You can figure out the logistics around merging them.

python/foxglove-sdk/src/websocket_server.rs
```diff
@@ -140,23 +140,84 @@ impl PyWebSocketServer {
...
+        server.publish_status(status.0.clone());
+    }
+
+    /// Remove status messages by id from all clients.
+    pub fn remove_status(&self, status_ids: Vec<String>) {
+        let Some(server) = &self.0 else {
+            return;
+        };
+        server.remove_status(status_ids);
+    }
```
eloff, line 172: A bigger question about the API. Is it better to silently fail after the server is closed, or to throw an exception?

Since the default is to only close the server at exit, this seems ok to me as-is.

eloff (approved):
I think I'd be better to drop the status class and use regular arguments. I'm going to approve this same leave it up to you.

python/foxglove-sdk/src/websocket_server.rs
gasmith: Either way, we should probably document that behavior.

It would be user-friendlier to not raise an exception, so that users aren't forced to litter their code with `with contextlib.suppress(...):` statements, but my rusty side says it's better to explicitly raise errors instead of trying to guess what the user was trying to do. Since we're trying to gear our python SDK towards user-friendliness, I suppose I agree with Dan.

python/foxglove-sdk/python/foxglove/tests/test_server.py
gasmith: Strongly agree. I had [similar feedback on the rust PR](https://github.com/foxglove/foxglove-sdk/pull/192#discussion_r1945624142) about making this a bit more ergonomic. Maybe:

```python
def publish_status(level: StatusLevel, message: str, id: str | None = None):
```

gasmith (approved):

python/foxglove-sdk/src/lib.rs
```diff
@@ -211,6 +213,8 @@ fn _foxglove_py(m: &Bound<'_, PyModule>) -> PyResult<()> {
     m.add_class::<PyCapability>()?;
     m.add_class::<PyClient>()?;
     m.add_class::<PyClientChannelView>()?;
+    m.add_class::<PyStatus>()?;
+    m.add_class::<PyStatusLevel>()?;
```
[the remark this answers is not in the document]
bryfox: Yeah, I'd certainly like to re-organize this. Since there are a few concurrent PRs in-flight, I'd like to take a pass at that in a follow-up PR. I think it's mildly unfortunate we're baking the name "websocket" into these interfaces, but I'll likely follow whatever rust does in the end.

python/foxglove-sdk/src/websocket_server.rs
bryfox: Sorry, this is a wonky diff because I changed the conditionals to early returns; was added in #224.

PR #228, opened 2025-02-20
python: publish docs to gh pages

sofuture:
### Changelog
Publish python docs to https://foxglove.github.io/foxglove-sdk/python/

### Docs
None

### Description
Publish python docs via github action.


.github/workflows/python-docs.yml
```diff
@@ -0,0 +1,27 @@
+name: documentation (python)
+
+on: [push, pull_request, workflow_dispatch]
```
amacneil, line 3: normally we don't run on push AND pull_request, since that results in the workflow running twice for each commit. so the `push` action is usually limited to `main` branch.

.github/workflows/python-docs.yml
```diff
@@ -0,0 +1,27 @@
+name: documentation (python)
```
amacneil, line 1: why a separate workflow? seems a bit cleaner to add a new job to existing python workflow

.github/workflows/python-docs.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+      - name: Install dependencies
+        run: |
+          pip install sphinx sphinx_rtd_theme myst_parser
```
amacneil, line 16: we should add these dev dependencies to the root pyproject.toml so they are versioned, and install with poetry (see the main python.yml workflow)

.github/workflows/python-docs.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Sphinx build
+        run: |
+          sphinx-build ./python/foxglove-sdk/python/docs _build
+      - name: Deploy to GitHub Pages
+        uses: peaceiris/actions-gh-pages@v3
+        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+        with:
+          publish_branch: gh-pages
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: _build/
```
amacneil, line 26: can we publish it under a `python` subdirectory? we will probably want to publish cpp/typescript and others too.

pyproject.toml
```diff
@@ -2,11 +2,12 @@
 [project]
 name = "foxglove-sdk-monorepo"
 version = "0.0.0"
-requires-python = ">=3.9"
+requires-python = ">=3.11"
```
eloff, line 5: Python 3.9 is end of support Oct 2025, 3.10 Oct 2026.

I think this fine to require 3.11 as a minimum.

.github/workflows/python.yml
```diff
@@ -259,3 +259,31 @@ jobs:
...
+  docs:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.9"
```
eloff, line 273: Should this then also be 3.11?

pyproject.toml
sofuture: I assume we had kept it older intentionally, but sphinx 8.x required at least 3.10 and 8.2 3.11 so 🤷

eloff: We didn't have a good reason for supporting 3.9 other than we could at the time without big sacrifices. 3.11 is 2.5 years old now. I think it's fine to require that.

sofuture: 8.x was latest/default when installing. I wasn't familiar with the poetry workspace setup, we can use the already specified one.

sofuture:
Just rebased and forced a deploy

bryfox:
> Just rebased and forced a deploy

Still missing those docs. It must be that we need to build the app (`poetry run maturin build`) or have it cached from a previous step

PR #229, opened 2025-02-20
Rust ws server asset protocol

eloff:
This introduces another callback on ServerListener:

```
fn on_fetch_asset(&self, _uri: String, _responder: AssetResponder) {}
```

Where the AssetResponder has two methods

```
pub fn send_error(&self, error: &str)
pub fn send_data(&self, asset: &[u8])
```

And a public Client property. Alternately this could be private and we could pass the client as the first argument like the other callbacks. I don't feel strongly about which is better.

I added an example showing how to implement an asset server, although it doesn't show how the client would work. Not sure if I should add a client as well?

I also added a way to send status messages to a specific client on Client, because callbacks in general lack a way to give feedback about errors or warnings.

```
pub fn send_status(&self, status: Status)
```

Which mirrors publish_status on the server handle.


rust/foxglove/src/websocket.rs
```diff
@@ -797,6 +845,30 @@ impl ConnectedClient {
...
+    /// Send a fetch asset response to the client. Called from AssetResponder.
+    fn send_asset_response(&self, response: &[u8], request_id: u32) {
+        // https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#fetch-asset-response
+        let mut buf = Vec::with_capacity(10 + response.len());
+        buf.put_u8(protocol::server::BinaryOpcode::FetchAssetResponse as u8);
+        buf.put_u32_le(request_id);
+        buf.put_u8(0); // 0 for success
+        buf.put(response);
+        let message = Message::binary(buf);
+        self.send_control_msg(message);
```
bryfox, line 870: Should this be considered a control plane message? I guess it seems similar to service responses, but it also seems like we're starting to overload this channel.

rust/foxglove/src/websocket/asset_responder.rs
```diff
@@ -0,0 +1,25 @@
...
+        Self { client, request_id }
+    }
+
+    /// Send an error response to the client.
+    pub fn send_error(&self, error: &str) {
+        self.client.send_asset_error(error, self.request_id);
+    }
+
+    /// Send a successful response to the client with the asset.
+    pub fn send_data(&self, asset: &[u8]) {
```
bryfox, line 22: If this is sync, how does it play with longer-running responses? Can I respond with assets downloaded from s3, for example?

rust/foxglove/src/websocket.rs
```diff
@@ -774,6 +810,18 @@ impl ConnectedClient {
         self.send_control_msg(msg);
     }
 
+    fn on_fetch_asset(&self, server: Arc<Server>, uri: String, request_id: u32) {
```
bryfox, line 813: Services have a `call_id`; should this be consistent? Or is this semantically different? (I like request_id better.)

bryfox (commented):
What's the advantage of the AssetResponder over the approach used by services to return Bytes? I like this approach, but it seems like a new pattern for a user to learn.

I'm also still trying to understand how this will work if the user's implementation needs to do some slow work to get the asset. With python specifically, I have a similar question with services (#230) — it seems like we want to have some sort of out-of-band delivery of responses (services or assets) that correlate by call_id/request_id. Maybe this isn't a use case we need to support yet, though.

eloff:
> What's the advantage of the AssetResponder over the approach used by services to return Bytes? I like this approach, but it seems like a new pattern for a user to learn.

You mean take a Result with a single respond method instead of one each for success and failure? Yeah I can align that.

> 
> I'm also still trying to understand how this will work if the user's implementation needs to do some slow work to get the asset. With python specifically, I have a similar question with services (#230) — it seems like we want to have some sort of out-of-band delivery of responses (services or assets) that correlate by call_id/request_id. Maybe this isn't a use case we need to support yet, though.

ServerListener has a note:

These methods are invoked from the client's main poll loop and must not block. If blocking or
long-running behavior is required, the implementation should use [`tokio::task::spawn`] (or
[`tokio::task::spawn_blocking`]).

The only callbacks that return a value are on_get_parameters and on_set_parameters which are expected to be quick.

Service calls and asset responders use this responder object pattern where you can launch the work in a thread pool, or an async runtime, and then call respond() when the result is ready.

For cheap handlers this works well, as it does for on_get_parameters/on_set_parameters where we need the result right away. For expensive handlers it means the user has to do some kind of async or threads themselves.

Is that better than doing it for them by calling the handlers from the blocking thread pool to begin with?

rust/foxglove/src/websocket/asset_responder.rs
eloff: You can download the asset in a background thread or in a new tokio task, and then call this when it's fully buffered.

rust/foxglove/src/websocket.rs
eloff: I'm copying it from the spec where it's called requestId: https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#fetch-asset

Maybe they should be named the same.

rust/foxglove/src/websocket.rs
bryfox: If we're planning in FG-10441 to drop clients which fill up the control plane queue, then I think dropping the message would be preferable (though I'm not necessarily advocating for dropping clients). This assumes we don't want a separate channel for services, assets, etc.

rust/foxglove/src/websocket.rs
bryfox: 👍 I like request_id; I do think we should consider renaming `call_id` for consistency, but will see what @gasmith thinks.

bryfox:
> The only callbacks that return a value are on_get_parameters and on_set_parameters which are expected to be quick.

The service handler has a `sync_handler_fn`, and there's no equivalent for assets. Maybe I'm making too many assumptions about Python consumers of the SDK, but I'd expect most to reach for the sync version (which is what I started with in #230). Maybe we can make the asset responder version feel natural there too.

> I added an example showing how to implement an asset server, although it doesn't show how the client would work. Not sure if I should add a client as well?

In my opinion, right now we should focus on examples targeting the Foxglove app as the client, rather than building out custom clients. e.g., how could this be used to load a URDF into the 3d panel for live viz?

rust/foxglove/src/websocket.rs
gasmith: I still feel rather strongly that client-initiated RPCs (assets, services) should be handled differently, as I initially proposed in #213. When the client expects a response, we _**really**_ shouldn't drop it. When the client makes too many requests, we should send an immediate failure response, which should be very nearly infallible (or otherwise result in drastic measures to fully reset client state, like disconnecting the client).

If the server is allowed to drop responses, then clients need to "time out" their requests in order to cope, and that's always a losing game.

I think Dan was opposed to a proliferation of pollables, but perhaps services and assets could share a common bounded channel and admission controller.

eloff:
> > The only callbacks that return a value are on_get_parameters and on_set_parameters which are expected to be quick.
> 
> The service handler has a `sync_handler_fn`, and there's no equivalent for assets. Maybe I'm making too many assumptions about Python consumers of the SDK, but I'd expect most to reach for the sync version (which is what I started with in #230). Maybe we can make the asset responder version feel natural there too.

That'd be an anti-pattern with the fetch asset API. You're not supposed to block for long, so if you returned it as a Result we'd need to call that via spawn_blocking and document that as an exception from the other handler methods. But maybe that's better than requiring the fetch asset handler to take responsibility for doing its work in the background.

For async you'd need a different interface (ideally an async handler). I can't think of a nice way to structure that code via the ServerListener. It works for services because they don't use that interface. We could do a similar dedicated API  for fetch asset with a similar interface.

I can rewrite it like that if it makes sense.

> 
> > I added an example showing how to implement an asset server, although it doesn't show how the client would work. Not sure if I should add a client as well?
> 
> In my opinion, right now we should focus on examples targeting the Foxglove app as the client, rather than building out custom clients. e.g., how could this be used to load a URDF into the 3d panel for live viz?

That makes sense. I don't know how to do that in the app, I'll look into it.

rust/foxglove/src/websocket/asset_responder.rs
```diff
@@ -0,0 +1,25 @@
+use super::Client;
+
+/// A helper for sending asset responses to a client.
+pub struct AssetResponder {
```
gasmith, line 4: Perhaps we should rename the service responder as `CallResponder` or `ServiceResponder` to disambiguate?

rust/foxglove/src/websocket/asset_responder.rs
```diff
@@ -0,0 +1,25 @@
...
+    /// The client requesting the asset.
+    pub client: Client,
```
gasmith, line 6: If we must, let's expose an accessor method instead of the field. But this feels a little redundant to the `Client` that's provided in the callback.

rust/foxglove/src/websocket/protocol/server.rs
```diff
@@ -19,6 +19,8 @@ pub enum BinaryOpcode {
     TimeData = 2,
     ServiceCallResponse = 3,
     // FetchAssetResponse = 4,
+    // ServiceCallResponse = 3,
```
gasmith, line 22:
```suggestion
```

gasmith:
> > What's the advantage of the AssetResponder over the approach used by services to return Bytes? I like this approach, but it seems like a new pattern for a user to learn.
> 
> You mean take a Result with a single respond method instead of one each for success and failure? Yeah I can align that.

Yeah, might be nice to use the same pattern as the service `Responder`'s `respond` method, which takes a `Result<Bytes, String>`.

> For async you'd need a different interface (ideally an async handler). I can't think of a nice way to structure that code via the ServerListener. It works for services because they don't use that interface. We could do a similar dedicated API for fetch asset with a similar interface.

Maybe? A big motivation for doing services differently was to accommodate multiple unrelated service implementations, without forcing the user to build a routing layer. We don't have that same motivation here, since there's at most one implementation for fetching assets.

If we were to offer a separate `FetchAsset` trait, the main benefit would be the sync implementation sugaring we added for service handlers, and the ability for clients to decouple their implementation from other `ServiceListener` callbacks. Maybe those are nice to have, but they don't feel like big wins...

The bigger issue is about blocking the client's poll loop, which is an awfully subtle footgun. Maybe `FetchAsset` and `service::Handler` should be async traits, and we should unconditionally spawn a new tokio task to invoke them. A nice benefit of doing things that way (with the spawn on our side of the fence), is that we can easily track and kill those tasks when the client disconnects (or we otherwise reset session state). The main downside is the latency we incur for spawning a task when we don't actually need one (e.g., because the implementation is sync, or needs `spawn_blocking`).

rust/foxglove/src/websocket.rs
eloff: For log data, I think the approach of dropping messages when the queue is full is reasonable. The client can't keep up, and we don't want that to spread to other clients or the rest of the system. I think we're on the same page about these.

For RPCs, advertisements, partial state updates where you can't miss one and still have an accurate picture, etc this doesn't work. I think we're also on the same page that dropping these is bad.

So we have a second type of queue, and when it fills, instead of dropping the messages, we need to do something else. I think up to here we agree.

We could block the caller when the queue fills, but whether we do that by awaiting, or by actually blocking the thread, in both cases we're blocking other clients. This is bad, because the slow client is now dictating progress in the system. We have actual customer complaints about this with existing ws-protocol servers. I do not think this approach is tenable. You could try to cheat this by blocking in a newly spawned task, but all that would do is invent an inefficient unbounded queue.

The other option, don't block means using an unbounded queue, which will cause resource exhaustion, swapping and slowing down the system, and eventually a panic that abruptly aborts the process. I also do not think this is tenable.

That leaves just one option that I can see, disconnect the slow client. We can send an error first, so the client knows what happened, but it's the same thing. It's not that I like the option, but I don't see a viable alternative. We can make the queue size configurable, so the user can adjust it if they have a very bursty workflow and they're getting disconnected.

Now I think Greg is advocating for having a third class of messages where we control how many are in flight with a semaphore up front, and therefore we never reach the limit of the queue, so never have to block. I don't think this is necessary. It behaves exactly the same if the control queue is has space, the only difference is if the client is backed up, one of these messages might be the message too many that causes the client to be disconnected. I don't think that it matters which message triggers it, it's about to be triggered anyway. This holds as long as the control queue is many multiples the size of the RPC semaphore. I was thinking of making the limit for the queue 1024, and the RPC limit is something like 32. If there are less than 992 messages in the queue, there's no difference in behavior, if there are more, then it might be an RCP message that's the last straw, but something was going to be that anyway. I don't think it justifies having a third type of queue.

rust/foxglove/src/websocket.rs
eloff: I like Greg's suggestion

rust/foxglove/src/websocket/asset_responder.rs
eloff: Yeah, that makes sense

rust/foxglove/src/websocket/asset_responder.rs
eloff: I didn't provide it in the callback, seemed odd to have two copies of it, one as an arg, and one inside the responder which is also an arg. But it also seems odd that you have to access the client from the responder, just for this call, when it's an arg for the others.

Which seems better to you?

eloff:
> > > What's the advantage of the AssetResponder over the approach used by services to return Bytes? I like this approach, but it seems like a new pattern for a user to learn.
> > 
> > 
> > You mean take a Result with a single respond method instead of one each for success and failure? Yeah I can align that.
> 
> Yeah, might be nice to use the same pattern as the service `Responder::respond` method, which takes a `Result<Bytes, String>`.
> 
> > For async you'd need a different interface (ideally an async handler). I can't think of a nice way to structure that code via the ServerListener. It works for services because they don't use that interface. We could do a similar dedicated API for fetch asset with a similar interface.
> 
> Maybe? A big motivation for doing services differently was to accommodate multiple unrelated service implementations, without forcing the user to build a routing layer. We don't have that same motivation here, since there's at most one implementation for fetching assets.
> 
> If we were to offer a separate `FetchAsset` trait, the main benefit would be the sync implementation sugaring we added for service handlers, and the ability for clients to decouple their implementation from other `ServiceListener` callbacks. Maybe those are nice to have, but they don't feel like big wins...
> 
> The bigger issue is about blocking the client's poll loop, which is an awfully subtle footgun. Maybe the `on_fetch_asset` and `service::Handler::call` callbacks should be async, and we should unconditionally spawn a new tokio task to invoke them. A nice benefit of doing things that way (with the spawn on our side of the fence), is that we can easily track and kill those tasks when the client disconnects (or we otherwise reset session state). The main downside is the latency we incur for spawning a task when we don't actually need one (e.g., because the implementation is sync, or needs `spawn_blocking`).

Yeah these are exactly the kinds of considerations I've been thinking about.

I think ideally we take a sync or async callback *somehow* and then invoke it ourselves via spawn_blocking or spawn, and document that in the API. So no footgun, and the user can write their code according to whatever works best (e.g. sync for filesystem or Python, or async for a rust network operation), and we can "cancel" it when the client disconnects.

I think that isn't easily done via the ServerListener, we'd need something like:

```rust
ServerBuilder::new()
    .fetch_asset_handler(sync_callback)`
   
// Or
    .fetch_asset_async_handler(async_callback)
```

Which can be implemented something like:
   
```rust

use std::future::Future;
use tokio::task::{spawn, spawn_blocking};

/// Define an enum to distinguish between sync and async callbacks
enum Callback<ResultType> {
    Sync(fn() -> Result<ResultType, String>),
    Async(fn() -> impl Future<Output = Result<ResultType, String>> + Send + 'static),
}

impl<ResultType: Send + 'static> Callback<ResultType> {
    /// Execute the callback correctly based on its type
    async fn execute(self) -> Result<ResultType, String> {
        match self {
            Self::Sync(sync_fn) => {
                // Run sync function in a blocking thread
                spawn_blocking(move || sync_fn()).await.unwrap_or_else(|e| Err(format!("Task join error: {}", e)))
            }
            Self::Async(async_fn) => {
                // Run async function in an async task
                spawn(async_fn()).await.unwrap_or_else(|e| Err(format!("Task join error: {}", e)))
            }
        }
    }
}

#[tokio::main]
async fn main() {
    // Define a synchronous callback
    let sync_callback = Callback::Sync(|| {
        println!("Sync callback running");
        Ok(24)
    });

    // Define an asynchronous callback
    let async_callback = Callback::Async(|| async {
        println!("Async callback running");
        Ok(42)
    });

    // Execute both callbacks
    let res1 = sync_callback.execute().await;
    println!("Sync callback result: {:?}", res1);

    let res2 = async_callback.execute().await;
    println!("Async callback result: {:?}", res2);
}
```

gasmith:
>> The bigger issue is about blocking the client's poll loop, which is an awfully subtle footgun. Maybe the `on_fetch_asset` and `service::Handler::call` callbacks should be async, and we should unconditionally spawn a new tokio task to invoke them.
>
> I think that isn't easily done via the ServerListener, we'd need something like:
> 
> ```rust
> ServerBuilder::new()
>     .fetch_asset_handler(sync_callback)`
>    
> // Or
>     .fetch_asset_async_handler(async_callback)
> ```


As we discussed this morning, I think we should try to keep it simple, and lean towards async, since that's what we expect rust devs to be doing anyway.

> ```rust
> enum Callback<ResultType> {
>     Sync(fn() -> Result<ResultType, String>),
>     Async(fn() -> impl Future<Output = Result<ResultType, String>> + Send + 'static),
> }

I don't think it's possible to describe fn pointers with an `impl` return. But even if it were, I think I'd like to keep the callback interface as a trait, for wider flexibility. Then we can provide wrapper structs for sugaring concrete use cases, like we did for service callbacks.

We could follow [the pattern that `tower::Service` uses](https://github.com/tower-rs/tower/blob/954e4c7e8d889b3b77e68886b2c78d5bb45b74fb/tower-service/src/lib.rs#L311), and describe the future as an associated type. That should be less finicky than `async-trait` (with its pin-box-ness), or native async traits (which are still not dyn-compatible). Something like:

```rust
trait Call {
    type Error;
    type Future: Future<Output = Result<(), Self::Error>>;

    fn call(&self) -> Self::Future;
}

// Sugaring for an async function or block.
struct CallFn<F, Fut>(F)
where
    F: Fn() -> Fut,
    Fut: Future<Output = anyhow::Result<()>>;

impl<F, Fut> Call for CallFn<F, Fut>
where
    F: Fn() -> Fut,
    Fut: Future<Output = anyhow::Result<()>>,
{
    type Error = anyhow::Error;
    type Future = Fut;

    fn call(&self) -> Self::Future {
        self.0()
    }
}

// It's dyn-compatible, so long as the function and future are 'static.
impl<F, Fut> CallFn<F, Fut>
where
    F: Fn() -> Fut + 'static,
    Fut: Future<Output = anyhow::Result<()>> + 'static,
{
    fn boxed(self) -> Box<dyn Call<Error = anyhow::Error, Future = Fut>> {
        Box::new(self)
    }
}

// Examples follow...
async fn handler() -> anyhow::Result<()> {
    println!("hello world");
    Ok(())
}

fn main() {
    let call1 = CallFn(|| async {
        println!("hello world");
        Ok(())
    });
    let call2 = CallFn(handler);
}
```

rust/foxglove/src/websocket_server.rs
```diff
@@ -69,6 +73,31 @@ impl WebSocketServer {
         self
     }
 
+    /// Configure the handler for fetching assets.
+    pub fn fetch_asset_handler(mut self, handler: Arc<dyn AssetHandler>) -> Self {
+        self.options.fetch_asset_handler = Some(handler);
+        self
+    }
+
+    /// Configure a synchronous function as a fetch asset handler.
```
gasmith, line 82: We should document that these setters are mutually exclusive - there can be only one fetch asset handler.

gasmith (commented):
This is looking great!

rust/foxglove/src/websocket/fetch_asset.rs
```diff
@@ -0,0 +1,77 @@
...
+/// The result of a fetch asset request.
+pub type FetchAssetResult = Result<Vec<u8>, String>;
+
+/// A handler to respond to fetch asset requests
+/// See: https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#fetch-asset
+pub trait AssetHandler: Send + Sync + 'static {
+    /// Fetch an asset with the given uri and return it via the responder.
+    /// Fetch should not block, it should call `runtime.spawn`
+    /// or `runtime.spawn_blocking` to do the actual work.
+    fn fetch(self: Arc<Self>, _runtime: &Handle, _uri: String, _responder: AssetResponder);
```
gasmith, line 15: Passing `Arc<Self>` is a little unusual. I wonder whether we might just bury another Arc inside the `Fn` structs instead.

rust/foxglove/src/websocket.rs
```diff
@@ -72,13 +80,40 @@ impl From<ClientId> for u32 {
...
 
-impl Client<'_> {
     /// Returns the client ID.
     pub fn id(&self) -> ClientId {
-        self.0.id
+        self.id
+    }
+
+    /// Send a status message to this client. Does nothing if client is disconnected.
+    pub fn send_status(&self, status: Status) {
```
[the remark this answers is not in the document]
eloff: Public actually, it gives the user a way to send an error back to the client from a ServerListener. There are ways to broadcast to all users, but this lets one notify a particular client that there was a problem with their request when we didn't have other ways to communicate that, like in on_set_parameters.

rust/foxglove/src/websocket/fetch_asset.rs
```diff
@@ -0,0 +1,77 @@
...
+    F: Fn(Client, String) -> Fut + Send + Sync + 'static,
+    Fut: Future<Output = FetchAssetResult> + Send + 'static,
+{
+    fn fetch(self: Arc<Self>, runtime: &Handle, uri: String, responder: AssetResponder) {
+        runtime.spawn(async move {
+            let result = (self.0)(responder.client(), uri).await;
+            responder.respond(result);
+        });
+    }
+}
```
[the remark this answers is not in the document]
eloff: I'm embarrassed how much time I spent on it :laughing: 

It was like my own personal learning rust with too many linked-lists journey. Nothing I tried worked out, but this finally did.

The nice part is it lets us easily set a sync handler for languages like Python where we know we're going to do that, but also have an async handler for the rust API.

rust/foxglove/src/websocket/fetch_asset.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// or `runtime.spawn_blocking` to do the actual work.
+    fn fetch(self: Arc<Self>, _runtime: &Handle, _uri: String, _responder: AssetResponder);
+}
+
+#[doc(hidden)]
+pub struct SyncAssetHandlerFn<F>(pub F);
+
+impl<F> AssetHandler for SyncAssetHandlerFn<F>
+where
+    F: Fn(Client, String) -> FetchAssetResult + Send + Sync + 'static,
```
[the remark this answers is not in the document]
eloff: Yeah, that's a good idea

rust/foxglove/src/websocket/fetch_asset.rs
```diff
@@ -0,0 +1,77 @@
...
+/// See: https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#fetch-asset
+pub trait AssetHandler: Send + Sync + 'static {
+    /// Fetch an asset with the given uri and return it via the responder.
+    /// Fetch should not block, it should call `runtime.spawn`
+    /// or `runtime.spawn_blocking` to do the actual work.
+    fn fetch(self: Arc<Self>, _runtime: &Handle, _uri: String, _responder: AssetResponder);
+}
+
+#[doc(hidden)]
+pub struct SyncAssetHandlerFn<F>(pub F);
```
gasmith, line 19: Let's call this `BlockingAssetHandlerFn`, to match the expectation that we're going to invoke it with `tokio::task::spawn_blocking`.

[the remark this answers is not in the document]
eloff: Hmm, I think so, I'll try it

rust/foxglove/src/websocket.rs
```diff
@@ -819,6 +881,30 @@ impl ConnectedClient {
...
     }
+
+    /// Send a fetch asset error to the client.
+    fn send_asset_error(&self, error: &str, request_id: u32) {
+        // https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#fetch-asset-response
+        let mut buf = Vec::with_capacity(10 + error.len());
+        buf.put_u8(protocol::server::BinaryOpcode::FetchAssetResponse as u8);
+        buf.put_u32_le(request_id);
+        buf.put_u8(1); // 1 for error
+        buf.put(error.as_bytes());
```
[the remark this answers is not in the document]
eloff: Good catch!

rust/foxglove/src/websocket/fetch_asset.rs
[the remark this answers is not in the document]
eloff: Actually I tried it and I didn't like this, because it forces the user to specify the error type in the success case, because it can't be inferred. I think it's fine to get the user to call to_string() themselves.

rust/foxglove/src/websocket/fetch_asset.rs
eloff: I think any implementation of the trait would also need to be in an Arc, and this is the only way to enforce that

rust/foxglove/src/websocket/fetch_asset.rs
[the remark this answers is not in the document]
gasmith: I feel like the case where you have an infallible handler implemented as a closure is going to be pretty rare. Most of the time you're either going to have a function (where the error type is explicitly declared) or the implementation will be fallible (in which case the error type is inferred). Plus, you'll often have more than one place in your handler where you need to return an error. I'd much rather do this:

```rust
.fetch_asset_handler_async_fn(|_, url| async move {
    reqwest::get(url).await?.bytes().await
})
```

than this:

```rust
.fetch_asset_handler_async_fn(|_, url| async move {
    reqwest::get(url)
        .await
        .map_err(|e| e.to_string())?
        .bytes()
        .await
        .map_err(|e| e.to_string())
})
```

rust/foxglove/src/websocket/fetch_asset.rs
gasmith: You only need the `Arc` if you spawn.

Consider a server that prefetches assets on startup, and serves them up from memory. That could be totally synchronous.

rust/examples/ws-server-fetch-asset/src/main.rs
```diff
@@ -20,11 +20,11 @@ impl AssetServer {
 }
 
 impl AssetHandler for AssetServer {
-    fn fetch(self: Arc<Self>, _runtime: &Handle, uri: String, responder: AssetResponder) {
+    fn fetch(self: Arc<Self>, uri: String, responder: AssetResponder) {
         match self.assets.get(&uri) {
             // A real implementation might use std::fs::read to read a file into a Vec<u8>
             // The ws-protocol doesn't currently support streaming for a single asset.
-            Some(asset) => responder.respond(Ok(asset.clone())),
+            Some(asset) => responder.respond(Ok(Bytes::copy_from_slice(asset))),
```
gasmith, line 27: Nit: I know it's just an example, but if you store `Bytes` in the `AssetServer`, this becomes a cheap clone instead of a copy.

gasmith (commented):
I still think the callback wrappers should be generic over the error type, and that we don't really need the Arc receiver in the trait. Other than that, LG.

eloff:
> I still think the callback wrappers should be generic over the error type, and that we don't really need the Arc receiver in the trait. Other than that, LG.

I got rid of the Arc receiver. The user is almost always going to need to use an Arc, because the trait requires them to spawn unless they've got the data already sitting in memory somehow. They're going to need to do a `Foo { inner: Arc<InnerFoo> }` instead, which is maybe not as obvious without the Arc<Self> in the trait, and a little ugly. We also have a double indirect Box<Arc<T>>, but I doubt that will ever matter. I'm not convinced this was an improvement for the user.

I'm not convinced about the generic Err either. I tried it again to see what the error was, but I don't see a nice way around it:

If I use Err: Display, then if one implements AssetHandler::fetch and calls respond, it has to look like this:

```rust
fn fetch(self: Arc<Self>, uri: String, responder: AssetResponder) {
        match self.assets.get(&uri) {
            Some(asset) => responder.respond(Ok::<_, String>(Bytes::copy_from_slice(asset))),
            None => responder.respond(Err(format!("Asset {} not found", uri))),
        }
    }
```

You have to specify the type of of the unused Error here, which is strange and not intuitive. What do you choose, there is no error! Didn't you run into this as well?

```text
Some(asset) => responder.respond(Ok(Bytes::copy_from_slice(asset))),
   |                                      ------- ^^ cannot infer type of the type parameter `E` declared on the enum `Result`
```

Without that you'd need to return `.map_err(|e|e.t_string())`, which isn't as convenient but at least it's clear from the compile error what to do.

We should make these consistent between the service handler and asset handler, so let's talk about it after you read this and settle on a direction.

gasmith:
> If I use Err: Display, then if one implements AssetHandler::fetch and calls respond, it has to look like this:

I think `AssetResponder::respond` should still take a concrete `Result<Bytes, String>`. I was only proposing the generic error for the function call wrappers. The service handler is written the same way.

eloff:
> > If I use Err: Display, then if one implements AssetHandler::fetch and calls respond, it has to look like this:
> 
> I think `AssetResponder::respond` should still take a concrete `Result<Bytes, String>`. I was only proposing the generic error for the function call wrappers. The service handler is written the same way.

Oh, that makes sense. I couldn't see it.

rust/examples/ws-server-fetch-asset/src/main.rs
gasmith: Even better!

```suggestion
            Some(asset) => responder.respond(Ok(asset.clone())),
```

gasmith (approved):
:ship:

PR #230, opened 2025-02-20
python: add support for services

bryfox:
This adds support for ws services to the Python SDK.

The types generally follow the Rust SDK. The user can provide ad hoc handlers, so (like Rust) this allows different handlers for each service, rather than tying it to the listener interface. However, it seemed more pythonic to receive a function with arguments rather than a larger nested object. This generally matches the existing ws-protocol implementation in python. I think we should arguably remove the `service_name` from the arguments here (to be discussed).

Currently, you can define services only when calling `start_server`; this doesn't support for adding & removing services after the fact; that is tracked by FG-10182.

I added a standalone example which shows a few ways to implement services with JSON encoding.

service


python/foxglove-sdk-examples/ws-services/main.py
```diff
@@ -0,0 +1,125 @@
...
+
+    The handler should return a bytes object which will be sent back to the client.
+
+    The handler must return quickly.
+    """
+    log_request(service_name, client, call_id, encoding, payload)
+    return b"{}"
+
+
+# A handler can also be defined as a lambda
```
eloff, line 43: You can omit this, lambda is not encouraged in Python anymore

python/foxglove-sdk-examples/ws-services/main.py
```diff
@@ -0,0 +1,125 @@
...
+    ServiceSchema,
+    start_server,
+    Capability,
+    Client,
+    ServiceHandler,
+)
+
+
+def logging_handler(
+    service_name: str, client: Client, call_id: int, encoding: str, payload: bytes
```
eloff, line 23: The problem with offering this API is it's not possible to use async Python or do work in another thread.

It's nice and simple for simple services, but it makes more complex ones impossible.

The responder object pattern we used in rust allows for all the above, while being a little more complex to use for simple cases.

Do we offer two distinct APIs or should we try to offer one way of doing it? I'm in favor of the latter (it's actually part of the zen of Python that there should ideally be one and only one obvious way to do it)

eloff (commented):
I'd like Greg to take a look at this too

python/foxglove-sdk/src/websocket_server.rs
```diff
@@ -68,6 +75,43 @@ impl ServerListener for PyServerListener {
...
+        let service_name = request.service_name();
+        let call_id: u32 = request.call_id().into();
+        let encoding = request.encoding();
+        let payload = request.payload();
+
+        let result: PyResult<Vec<u8>> = Python::with_gil(|py| {
+            let args = (service_name, client_info, call_id, encoding, payload);
+            let result = self.handler.bind(py).call(args, None)?;
+            result.extract::<Vec<u8>>()
+        });
```
gasmith, line 103: Hm, I don't think we should be making calls into python directly from the rust client's poll loop. We should use `tokio::task::spawn_blocking` (or `tokio::spawn` if we want to natively support polling python async as rust futures, though I don't know if there's any benefit to doing so).

That, in turn, means we should implement `Handler` instead of `SyncHandler`. We should be able to keep the `Responder` as a rust implementation detail, though.

This same comment applies to other `ServerListener` callbacks. I don't think we should assume the python implementation is fast enough to be on the hot path.

bryfox:
Thanks for the feedback. Per discussion, we'll implement services (and asset serving, when it comes) with the responder pattern, but hide this behind a `spawn_blocking` so the python interface appears similar to what's here now.

PR #231, opened 2025-02-21
python: implement parameters on server listener

bryfox:
This adds `on_get_parameters` and `on_set_parameters` to the python server listener, along with an example.

I was a bit surprised by the behavior of the latter, and only understood it by reading the [ws-protocol spec](https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#parameter-values); it works more like a patch interface. I updated the docs slightly to clarify this, but I wonder if we should consider renaming the callback for this. (Would we also consider implementing a parameter store internally in a future version?)

I think we'll eventually want a higher-level interface in Python for this too — dealing with 'type' classes is not ergonomic — but having a consistent interface seems valuable as well.


python/foxglove-sdk-examples/ws-param-server/main.py
```diff
@@ -0,0 +1,97 @@
...
+
+    # Foxglove server callback
+    def on_get_parameters(
+        self,
+        client: foxglove.Client,
+        param_names: list[str],
+        request_id: Optional[str] = None,
+    ) -> list[Parameter]:
+        logging.debug(f"on_get_parameters: {param_names}, {client.id}, {request_id}")
+        return self.parameters
```
eloff, line 30: This is ignoring the param names, that seems like the wrong thing to do in an example

python/foxglove-sdk-examples/ws-param-server/main.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def on_set_parameters(
+        self,
+        client: foxglove.Client,
+        parameters: list[foxglove.Parameter],
+        request_id: str | None = None,
+    ) -> list[foxglove.Parameter]:
+        logging.debug(f"on_set_parameters: {parameters}, {client.id}, {request_id}")
+        existing_names = [p.name for p in self.parameters]
```
eloff, line 39: I think self.parameters should be a dict. See the rust example.

If you wanted to keep this as is you could use a set or dict comprehension.

Using list everywhere is simple but inefficient. I know it's only an example, but I like setting a good example.

python/foxglove-sdk-examples/ws-param-server/main.py
```diff
@@ -0,0 +1,97 @@
...
+        return parameters
+
+
+def main() -> None:
+    foxglove.verbose_on()
+
+    initial_values: list[Parameter] = [
+        Parameter(
+            "param1",
+            value=ParameterValue.Dict(
```
eloff, line 66: I see what you mean. This would be a much nicer interface in Python if you could just use native lists, dicts, bytes, etc and we did the conversion in our code.

Although we'd have to throw an exception at runtime if you passed the wrong type.

I feel like we could provide the correct typings though. Also this current code will still throw an exception at runtime if you pass the wrong types. So it's only slightly safer.

I guess for now this works and we can make it more ergonomic in a future iteration.

eloff (commented):
It was surprising behavior to me too.

We can rename the callback to make it clearer, and maybe add some docs to clarify how it should be used.

python/foxglove-sdk-examples/ws-param-server/main.py
bryfox: I originally wrote it with a dict, but thought that it got in the way of understanding the interface. I suppose having a lot of parameters could be typical; I'll update.

python/foxglove-sdk-examples/ws-param-server/main.py
bryfox: oh yes, good catch

python/foxglove-sdk-examples/ws-param-server/main.py
bryfox: I think it's reasonable to support both, so I'll keep this and have also filed FG-10589 for the alternative.

bryfox:
> We can rename the callback to make it clearer, and maybe add some docs to clarify how it should be used.

I updated the docs here; I filed FG-10592 for possibly renaming.

eloff (approved):
LGTM

PR #232, opened 2025-02-21
python: Add metadata to log()

amacneil:
### Changelog
None

### Docs
None

### Description
- Allow custom timestamps in python:

```py
chan = fg.Channel(topic="/aranet", schema=schema)
chan.log(msg, log_time=123, publish_time=456, sequence=42)
```

- Added `log()` kwargs to accept `log_time`, `publish_time`, `sequence` (https://github.com/foxglove/foxglove-sdk/pull/234)

- Remove `log_with_meta()` from python API in favor of `log()`


python/foxglove-sdk/python/foxglove/channel.py
```diff
@@ -71,18 +71,26 @@ def __init__(
 
         _channels_by_topic[topic] = self
 
-    def log(self, msg: Union[JsonMessage, bytes]) -> None:
+    def log(
+        self,
+        msg: Union[JsonMessage, bytes],
+        log_time: Optional[int] = None,
+        publish_time: Optional[int] = None,
+        sequence: Optional[int] = None,
```
gasmith, line 79: For typed channels, we have `log(bytes)` and `log_with_meta(bytes, PartialMetadata)`. We should be consistent.

FWIW, I prefer the style in this PR, or passing in metadata as optional kwargs. We could get rid of `PartialMetadata` from the python API.

gasmith (approved):

python/foxglove-sdk/python/foxglove/channel.py
gasmith: https://github.com/foxglove/foxglove-sdk/pull/234

We can either merge it into your branch, or merge it to main separately, or even discard it, as folks prefer.

eloff: Passing a metadata struct works in Rust, because it's a zero-cost abstraction. In Python it's clunky and it involves an allocation.

PR #233, opened 2025-02-21
python: rename new_mcap_file() to open_file()

amacneil:
The existing name for this method feels clunky - longer than it needs to be, since we have no plans to support other file types.

Proposing to rename to `open_file` for consistency with python stdlib

### DIFF
--- a/python/foxglove-sdk/python/docs/index.rst
+++ b/python/foxglove-sdk/python/docs/index.rst
@@ -26,7 +26,7 @@ Overview
 To record messages, you need at least one sink and at least one channel.
 
 A "sink" is a destination for logged messages — either an MCAP file or a live visualization server.
-Use :python:`record_file` or :python:`with new_mcap_file("")` to register a new MCAP sink. Use
+Use :python:`record_file` or :python:`with open_file("")` to register a new MCAP sink. Use
 :python:`start_server` to create a new live visualization server.
 
 A "channel" gives a way to log related messages which have the same schema. Each channel is
--- a/python/foxglove-sdk/python/examples/write_log.py
+++ b/python/foxglove-sdk/python/examples/write_log.py
@@ -12,7 +12,7 @@ args = parser.parse_args()
 
 def main() -> None:
     # Create a new mcap file at the given path for recording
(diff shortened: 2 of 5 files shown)


gasmith (commented):
The term `open` is extremely generic. The stdlib's `open` recalls the `open(2)` system call and libc bindings, and so it obviously means "open a file." I'm not sure I'd have that same expectation while looking at any other python package. We only allow opening the file for _writing_, whereas I'd also expect a file-oriented `open` to also open files for _reading_. Maybe we want to support reading in the future. Is that the same interface, or should it be different?

The inclusion of "new" in the `new_mcap_file` name that we picked is supposed to reflect the fact that we're creating a new file, or failing with `EEXIST`. We had a discussion several weeks ago about default behavior, and agreed that overwriting existing mcap files is the exceptional case, not the default. On the rust side, if you really want to overwrite existing files, you can open the file yourself. On python, I'm not sure we need to provide that degree of flexibility, but we should at least allow overwriting existing files. I think this is easy to address by adding a `allow_overwrite: bool = False` kwarg. The term "new" shouldn't be in the name of the API call.

And then there's `record_file`. We really should not have two wildly-different names for the plain and context-managed versions of this API call. Instead, we should have a _single call_ that returns an `MCAPWriter`, and that type should implement the context manager interface itself (`__enter__` and `__exit__`) so that folks can use it that way if they want to, or not.

So to summarize, I'd like to see:
- One function to open an MCAP file.
- Create new by default, with an `allow_overwrite` kwarg.
- Implement context manager on `MCAPWriter`.

I still don't have a great proposal for the name of this function. Maybe `open_mcap_writer`, despite the verbosity? Perhaps `open_mcap`, even though we don't support reading? Or `write_mcap`?

gasmith:
> So to summarize, I'd like to see:
> 
> * One function to open an MCAP file.
> * Create new by default, with an `allow_overwrite` kwarg.
> * Implement context manager on `MCAPWriter`.

Here's a sketch of that: https://github.com/foxglove/foxglove-sdk/pull/235

amacneil:
Closing in favor of https://github.com/foxglove/foxglove-sdk/pull/235

PR #234, opened 2025-02-21
python: Use kwargs for metadata on typed channels

gasmith:
This is built on top of @amacneil's #232, and we can either merge it into that PR branch, or merge it to main separately.

The idea is to make typed channels consistent with raw channels by removing `log_with_meta` and adding keyword-only arguments to `log` for timestamps and sequence.

Fixes: FG-10468


python/foxglove-sdk/python/foxglove/_foxglove_py/channels.pyi
```diff
@@ -43,12 +42,10 @@ class CameraCalibrationChannel:
...
-    ) -> "CameraCalibrationChannel": ...
-    def log_with_meta(
-        self,
-        message: "CameraCalibration",
-        metadata: "PartialMetadata",
-    ) -> "CameraCalibrationChannel": ...
+        log_time: int | None = None,
+        publish_time: int | None = None,
+        sequence: int | None = None,
+    ) -> None: ...
```
amacneil, line 48: I like the idea of removing `log_with_meta()`

the PR title is "use kwargs" - could we move the metadata to kwargs? so like

```py
ch.log(msg, log_time=123, publish_time=456, sequence=42)
```

gasmith: > the PR title is "use kwargs" - could we move the metadata to kwargs? so like
> 
> ```py
> ch.log(msg, log_time=123, publish_time=456, sequence=42)
> ```

Yep, that's exactly what this does.

bryfox (approved):
👍 This closes FG-10468.

My original thinking was to also keep `log_with_meta` for consistency between languages, but I think reducing surface area is preferable.

bryfox:
Actually, one change I'd suggest is to make the trailing few args keyword-only.

python/foxglove-sdk/python/foxglove/_foxglove_py/channels.pyi
amacneil: derp, I don't understand how to read typed python

PR #235, opened 2025-02-22
python: Unified open_mcap call

gasmith:
See https://github.com/foxglove/foxglove-sdk/pull/233#pullrequestreview-2634589625 for context. The core of this change is:
- Only one function to open an MCAP file. Naming up for discussion.
- Create new by default, with an `allow_overwrite` kwarg.
- Implement context manager on `MCAPWriter`.

Other changes here:
- Use `PathBuf`, which pyo3 can construct from `str|pathlib.Path`.
- Fix up a bunch of sphinx cross references. Tested with #228.
- Add some test coverage.


python/foxglove-sdk/src/lib.rs
```diff
@@ -20,35 +21,40 @@ mod websocket_server;
...
 struct BaseChannel(Arc<Channel>);
 
-///  A writer for logging messages to an MCAP file.
-///
-/// Obtain an instance by calling :py:func:`record_file`, or the context-managed
-/// :py:func:`new_mcap_file`.
-///
-/// If you're using :py:func:`record_file`, you must maintain a reference to the returned writer
-/// until you are done logging. The writer will be closed automatically when it is garbage
-/// collected, but you may also :py:func:`MCAPWriter.close` it explicitly.
```
gasmith, line 30: It seems that these things get obliterated by the docstrings we put in the pyi. It probably makes sense to keep the documentation in one place.

python/foxglove-sdk/src/lib.rs
```diff
@@ -145,14 +147,17 @@ impl From<PartialMetadata> for foxglove::PartialMetadata {
...
 
 /// Open a new mcap file for recording.
-///
-/// :param path: The path to the MCAP file. This file will be created and must not already exist.
-/// :return: A new `MCAPWriter` object.
 #[pyfunction]
-#[pyo3(signature = (path))]
-fn record_file(path: &str) -> PyResult<PyMcapWriter> {
+#[pyo3(signature = (path, allow_overwrite = false))]
+fn open_mcap(path: &str, allow_overwrite: bool) -> PyResult<PyMcapWriter> {
```
gasmith, line 152: I suppose we might want to surface [mcap writer options](https://docs.rs/mcap/latest/mcap/write/struct.WriteOptions.html) here too. Not sure whether we want to flatten those all out into kwargs, or have some nesting to separate file-related options from mcap-related options.

python/foxglove-sdk/src/lib.rs
gasmith: I've restored the duplicate docs. I'm still weeping.

python/foxglove-sdk/python/foxglove/_foxglove_py/__init__.pyi
```diff
@@ -127,9 +137,13 @@ def shutdown() -> None:
     """
     ...
 
-def record_file(path: str) -> MCAPWriter:
+def open_mcap(path: str | Path, allow_overwrite: bool = False) -> MCAPWriter:
```
amacneil, line 140: Thinking more about your point that `open()` in python can mean read or write - I could imagine a world where we might want to expose mcap read interfaces (e.g. open an existing mcap file and append some additional foxglove-schema channels).

So one idea would be to name this `write_mcap()`.

I'm not sure I like it yet - just throwing the idea out there.

python/foxglove-sdk/src/lib.rs
bryfox: Yeah, this has been driving me crazy. I've also noticed that vscode, at least, doesn't do anything special with rst. I want to test other IDEs, but we may want to get rid of that format from the stub docs. I'd like to do a full pass over the docs after all these features merge, and we could leave this until then.

bryfox (commented):
Implementation looks good, and I'm 👍 for the unified interface.

I like the name, but happy to hear discussion...

I like "mcap". `open` alone is too generic (especially since we also have a socket connection to the app); `foxglove.mcap.open` would be fine, but that'd be the only reason to introduce the module. `open_file` seems needlessly vague since you really want to name your file with '.mcap'.

"open"/"write"/"read": I think "open" sounds less surprising than "write".

python/foxglove-sdk/src/lib.rs
gasmith: I use pyright with neovim, and it doesn't seem to interpret rst either. But this problem also exists for plain python libraries, where there's only one place to put a docstring, so... :shrug: 

I agree that we should take a more holistic look at cleaning up docs once things are a bit more stabilized.

eloff (approved):
Looks good!

PR #237, opened 2025-02-24
Rust ws server connection graph support

eloff:
Adds support for connection graphs to the websocket server. This follows the design in the C++ ws-protocol server very closely.

Clients can subscribe to updates, which sends the full state of the connection graph in response. [This is not documented](https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#subscribe-connection-graph-updates) , I think we should update the spec to document that, as it doesn't seem to me like it would work without it. The [C++ server behaves this way](https://github.com/foxglove/ws-protocol/blob/main/cpp/foxglove-websocket/include/foxglove/websocket/websocket_server.hpp#L1449).

Like with parameters, we trigger callbacks on the first subscriber and last unsubscriber to save the user having to do their own bookkeeping. We tell them when they need to start being concerned with the connection graph, and when they can stop.

It adds a method to the server, and the server handles for updating a connection graph. One passes a new, full connection graph in, and it will compute the difference with the last known state and send that difference to subscribers. Because it works with differences, that's why we need to send the full state when a subscriber first subscribes to the connection graph.

The idea is the client will probably maintain a ConnectionGraph object in the ServerListener that tracks the current graph, and after updating it will clone() it and pass it to publish_connection_graph. The server holds onto that copy so it can compute the diff the next time publish_connection_graph is called.


rust/foxglove/src/websocket.rs
```diff
@@ -780,14 +802,87 @@ impl ConnectedClient {
...
 
+    fn on_connection_graph_subscribe(&self, server: Arc<Server>) {
+        if !server.capabilities.contains(&Capability::ConnectionGraph) {
+            self.send_error("Server does not support connection graph capability".to_string());
+            return;
+        }
+
+        let is_subscribed = self.subscribed_to_connection_graph.load(Relaxed);
+        if is_subscribed {
+            tracing::info!(
```
bryfox, line 813: Maybe `debug`

rust/foxglove/src/websocket.rs
```diff
@@ -780,14 +802,87 @@ impl ConnectedClient {
...
+            if *subscriber_count == 0 {
+                if let Some(listener) = server.listener.as_ref() {
+                    listener.on_connection_graph_unsubscribe();
+                }
+            }
+        }
+
+        // Acquire the lock to sychronize with subscribe and with server.connection_graph_update.
+        let _guard = server.connection_graph.lock();
+        self.subscribed_to_connection_graph.store(false, Relaxed);
```
bryfox, line 873: Why are these accesses both Relaxed vs. Acquire/Release?

rust/foxglove/src/websocket.rs
```diff
@@ -780,14 +802,87 @@ impl ConnectedClient {
...
+    }
+
     /// Send an ad hoc error status message to the client, with the given message.
     fn send_error(&self, message: String) {
+        tracing::error!("{}", message);
```
bryfox, line 878: This seems surprising to me. The mechanism is to notify the client of an error, but I don't necessarily want to see errors in my server logs.

rust/foxglove/src/websocket.rs
```diff
@@ -780,14 +802,87 @@ impl ConnectedClient {
...
         self.send_status(Status::new(StatusLevel::Error, message));
     }
 
     /// Send an ad hoc warning status message to the client, with the given message.
     #[allow(dead_code)]
     fn send_warning(&self, message: String) {
+        tracing::warn!("{}", message);
```
bryfox, line 885: Same as above. If the idea is that the user can't see problems on the client, maybe these are better at debug level?

bryfox (commented):
Looking good. I left some comments but @gasmith will need to review as well.

Also tagging @achim-k in case he has any thoughts, since this follows the c++ implementation.

rust/foxglove/src/websocket/connection_graph.rs
```diff
@@ -0,0 +1,347 @@
+use std::collections::{HashMap, HashSet};
+
+use serde::Serialize;
+
+/// A HashMap where the keys are the topic or service name and the value is a set of string ids.
+pub type MapOfSets = HashMap<String, HashSet<String>>;
```
gasmith, line 6: It seems sort of weird to expose this type alias from the SDK, though I'm not strongly opposed to it. As an alternative, we might consider making the ConnectionGraph fields private, and providing public construction methods.

gasmith (commented):
I'm a little confused about when these updates are supposed to happen, and how they're supposed to be used. I was rather surprised to see that we're punting so much responsibility to the user. Don't we have enough state internally to generate and trigger these update messages?

rust/foxglove/src/websocket.rs
eloff: Maybe debug then. The C++ server does something similar.

eloff:
> I'm a little confused about when these updates are supposed to happen, and how they're supposed to be used. I was rather surprised to see that we're punting so much responsibility to the user. Don't we have enough state internally to generate and trigger these update messages?

I'm still very confused about how it's used, which is why my approach is just to re-implement what we have currently in c++ in the ws-protocol repo. I can't re-imagine or try to improve the API if I don't understand it. When we have actual users of the API and they come to us with feedback, we can revist then. This just gets us a a baseline, and I think mimicking current behavior is a good starting point.

Initially, like you, I thought we should be providing these updates based on the server state, connections, etc.

But this actually used in the rosbridge, and that info comes from ros, see:

https://github.com/foxglove/ros-foxglove-bridge/blob/main/ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp#L151
https://github.com/foxglove/ros-foxglove-bridge/blob/main/ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp#L381

I don't have a clue what a rosbridge is or how it's used, but it seems like it tracks the state and we just poll it and update the connection graph view from that.

rust/foxglove/src/websocket/connection_graph.rs
eloff: Yeah, that's probably better

rust/foxglove/src/websocket.rs
eloff: It believe it doesn't need anything stronger, it's only modified with the server.connection_graph mutex held, and doesn't gate access to other memory.

gasmith:
> > I'm a little confused about when these updates are supposed to happen, and how they're supposed to be used. I was rather surprised to see that we're punting so much responsibility to the user. Don't we have enough state internally to generate and trigger these update messages?
> 
> I'm still very confused about how it's used, which is why my approach is just to re-implement what we have currently in c++ in the ws-protocol repo. I can't re-imagine or try to improve the API if I don't understand it. When we have actual users of the API and they come to us with feedback, we can revist then. This just gets us a a baseline, and I think mimicking current behavior is a good starting point.

Fair point. I'm fine with mimicking the existing API rather than trying to design a replacement in a vacuum.

rust/foxglove/src/websocket.rs
```diff
@@ -797,14 +823,87 @@ impl ConnectedClient {
...
+        }
+
+        // Acquire the lock to sychronize with subscribe and with server.connection_graph_update.
+        let _guard = server.connection_graph.lock();
+        self.subscribed_to_connection_graph.store(false, Relaxed);
+    }
+
     /// Send an ad hoc error status message to the client, with the given message.
     fn send_error(&self, message: String) {
+        tracing::debug!("{}", message);
```
gasmith, line 899: Let's extend this message a bit. I'd rather see:

```
DEBUG [foxglove::weboscket] Sending error to client 127.0.0.1:48792: oh noes
```

Than:

```
DEBUG [foxglove::websocket] oh noes
```

Same goes for the warning below.

gasmith (approved):
LG

PR #239, opened 2025-02-24
python: Add support for websocket services

gasmith:
This adds support for ws services to the Python SDK.

See #230 for a more detailed explanation. The main difference between that PR and this one, is that this PR uses `tokio::task::spawn_blocking` to invoke service callbacks, and reworks the callback interface to more closely resemble the rust interface (a unified `Request` object instead of its individual components).


python/foxglove-sdk-examples/ws-services/main.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+    A handler for the service, adhering to the `ServiceHandler` type.
+
+    The handler should return a bytes object which will be sent back to the client.
+    """
+    log_request(client, request)
+    return b"{}"
+
+
+# A handler can also be defined as a lambda
```
bryfox, line 36: @eloff had mentioned in https://github.com/foxglove/foxglove-sdk/pull/230#discussion_r1966009573 that we could drop this example

python/foxglove-sdk/src/websocket_server.rs
```diff
@@ -291,10 +346,180 @@ impl From<PyCapability> for foxglove::websocket::Capability {
...
+    }
+
+    /// The request encoding.
+    #[getter]
+    fn encoding(&self) -> &str {
+        self.0.encoding()
+    }
+
+    /// The request payload.
+    #[getter]
```
bryfox, line 416: Nit: you could add `get_all` to the `pyclass()` to produce these getters

python/foxglove-sdk-examples/ws-services/main.py
```diff
@@ -0,0 +1,114 @@
...
+    Service,
+    ServiceHandler,
+    ServiceSchema,
+    start_server,
+)
+
+
+def logging_handler(
+    client: Client,
+    request: Request,
```
bryfox, line 25: I think I like this. You have to figure out what that object is, rather than having everything in the signature, but it feels a little more future-proof.

If we don't care to support lambdas (per another comment), then we could consider calling a function with kwargs, which seems a little more natural to me. I don't know that Callable can be annotated with that, but we could define a protocol like ServerListener. But that's still not as future-proof; I prefer what you did.

bryfox (approved):
Looks good to me. I tested as in #230.

I'd add a PR description for posterity (or link to the other PR).

python/foxglove-sdk-examples/ws-services/main.py
gasmith: Ah, right. Will do.

python/foxglove-sdk/src/websocket_server.rs
gasmith: I don't think that works in this case, because this is a tuple struct, and we're delegating to methods on the inner request.

python/foxglove-sdk-examples/ws-services/main.py
gasmith: Right, those were a lot of motivations for wrapping things up in a `Request` in rust too. Honestly, I was wondering the other day whether we should also fold `Client` into `Request` and make it simpler. We didn't do that in rust because `Client` has a lifetime attached, but perhaps we could switch to a `ClientView` and expose `Request::client_id()`.

For lambdas, I think as Dan pointed out, they're not particularly popular anymore. Regardless of that, I think its probably more conventional in python for callbacks to use positional arguments rather than kwargs, so that you don't force a particular arg naming.

bryfox: The other thing with `ClientView` was that we were leaving open the possibility of including information in addition to the `id`. A `request.client_id` seems perfectly sensible, but how real do you think the possibility is that we end up with other `client_*` properties?

gasmith: We might. I could imagine wanting to recover the client address, for example. But even if we do, no big deal. I think putting `client_*` methods on `Request` isn't substantially different from putting `*` methods on `Client`.

bryfox: 👍 If I saw 5-10 client-prefixed attrs, I'd think they should be nested in a client object, but I think that's unlikely. ID & addr seem reasonable. Getting rid of the `Client` type and only thinking about a `Request` arg seems like an improvement.

gasmith: Sounds good, I'll ticket that out separately, and we'll do the same on the rust side.

PR #240, opened 2025-02-25
python: implement additional ServerListener callbacks

bryfox:
This adds additional callbacks to python's ServerListener interface: `on_subscribe`, `on_unsubscribe`, `on_client_advertise`, and `on_client_unadvertise`.

I've updated the live_visualization example to monitor subscribers. This example should be moved to `foxglove-sdk-examples`; I can do so as part of this PR, but it's going to create noise, so I've left it alone for now.

I did not introduce a new "channel view" type for python; there, the ID types are identical. I don't see much value in a type that differs in name only; I think the context makes it clear.


bryfox:
I moved the live_visualization example as noted above, and tested running locally.

PR #242, opened 2025-02-25
rust: Don't expose websocket service IDs

gasmith:
Websocket service IDs are an implementation detail to accommodate `ws-protocol`. It's not something that our users need to be aware of. Service names need to be unique, and so that should be the handle by which a service is identified or removed. I realized this as I was trying to implement the python plumbing for `remove_services`.

Now we need two maps for indexing services: one by name, one by ID. They need to be synchronized, so I introduced a new `ServiceMap` struct.

### DIFF
--- a/rust/foxglove/src/websocket/service.rs
+++ b/rust/foxglove/src/websocket/service.rs
@@ -217,16 +217,6 @@ impl ServiceMap {
         assert!(prev.is_none());
     }
 
-    /// Returns true if the map contains a service with the provided name.
-    pub fn contains_name(&mut self, name: impl AsRef<str>) -> bool {
-        self.name.contains_key(name.as_ref())
-    }
-
-    /// Returns true if the map contains a service with the provided ID.
-    pub fn contains_id(&mut self, id: ServiceId) -> bool {
-        self.id.contains_key(&id)
-    }
-
     /// Removes a service by name.
     pub fn remove_by_name(&mut self, name: impl AsRef<str>) -> Option<Arc<Service>> {
         if let Some(id) = self.name.remove(name.as_ref()) {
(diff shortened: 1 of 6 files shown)


rust/foxglove/src/websocket/service.rs
```diff
@@ -182,3 +186,65 @@ impl Service {
...
+
+    /// Returns true if the map contains a service with the provided ID.
+    pub fn contains_id(&mut self, id: ServiceId) -> bool {
+        self.id.contains_key(&id)
+    }
+
+    /// Removes a service by name.
+    pub fn remove_by_name(&mut self, name: impl AsRef<str>) -> Option<Arc<Service>> {
+        if let Some(id) = self.name.remove(name.as_ref()) {
+            self.id.remove(&id)
```
bryfox, line 235: This don't need locking to coordinate with inserts because IDs are never re-used?

bryfox (approved):

rust/foxglove/src/websocket/service.rs
gasmith: Regarding the locking, this is rust's classic mutable-xor-shared. Because we hold a `&mut self`, we know there are no concurrent `&self`. For the websocket server, that guarantee is [provided by an `RwLock`](https://github.com/foxglove/foxglove-sdk/pull/242/files#diff-a7f22dab80c06b17a93e48fe59f89fad848400d6a7fd59c40a281af78504dac6R189).

To your other comment, it is true that IDs are not reused. But `ServiceId` is only 32-bit, so we _technically_ need to worry about overflow. Currently we don't handle that case, [we just panic](https://github.com/foxglove/foxglove-sdk/pull/242/files#diff-cb7c17349cc730d06a155dc240dc37bcaf9f05a3765dc26af25bce403af9a3aeR87).

bryfox: Ah, thanks for the explanation.

PR #243, opened 2025-02-25
Update python logging handling and level support

bryfox:
### Changelog

- Python: logging is no longer configured automatically on import, and defaults to `INFO`
- Python: `verbose_on` has been renamed to `set_log_level` and accepts the same argument values as `logging.setLevel()`

### Description

The primary change here is that `logging.basicConfig` is no longer called at the top-level of the foxglove module. This call is a no-op once logging is configured, and our library should not risk configuring logging unexpectedly. I considered going further and removing the call entirely, but I do think it's convenient for some use cases, and users are unlikely to be tripped up by configuration conflicts.

Other changes:

- `set_log_level` (renamed) accepts any of the standard `logging` level values, including `logging.INFO`, `"INFO"` and custom levels; it no longer accepts not our custom strings (including `"info"`). It no longer raises assertion errors, but `logging` will raise value errors given an unknown string.
- python logging defaults to "INFO", like rust

The examples that called `verbose_on` now do so with `DEBUG`.

Resolves FG-10629


python/foxglove-sdk/python/foxglove/__init__.py
```diff
@@ -167,36 +163,47 @@ def start_server(
...
 
-def _log_level_from_int(level: int) -> str:
-    log_levels = {10: "debug", 20: "info", 30: "warn", 40: "error"}
-    return log_levels.get(level, "unknown")
-
-
-def verbose_on(level: Union[int, str] = "debug") -> None:
+def verbose_on(level: Union[int, str] = "INFO") -> None:
     """
     Enable SDK logging.
```
gasmith, line 168: Could we call this `set_log_level`, perhaps? SDK logging is _always_ enabled, since we call `pyo3_log::init()` unconditionally during module init. We should document the fact that the default log level is `WARNING`. If callers want to adjust the log level, they can.

python/foxglove-sdk/python/foxglove/__init__.py
```diff
@@ -167,36 +163,47 @@ def start_server(
...
-            logging.INFO,
-            logging.WARN,
-            logging.ERROR,
-        ], ValueError("Invalid log level")
-        level = _log_level_from_int(level)
-    else:
-        assert level in ["debug", "info", "warn", "error"], ValueError(
-            "Invalid log level"
+    # This will raise a ValueError for invalid levels.
+    logging.basicConfig(level=level, format="%(asctime)s [%(levelname)s] %(message)s")
```
gasmith, line 174: It's a massive improvement to get this out of the module initialization path. I still feel a little uncomfortable calling it at all, because we're depending on the user to set up logging before they try to configure the foxglove log level.

python/foxglove-sdk/python/foxglove/__init__.py
bryfox: 👍 yes, I think that's an improvement.

python/foxglove-sdk/python/foxglove/__init__.py
bryfox: Same. It seems like a small but annoying footgun; on the other hand, one goal of the SDK is to make it very easy to get started. I'd be happy to take this out for now if you think that's the right tradeoff, and in either case gather feedback.

python/foxglove-sdk/python/foxglove/__init__.py
bryfox: What would you replace `verbose_off` with? I don't really want to introduce a special value here that means "off". We could consider "critical" to mean off, I suppose, rather than mapping it to the Error level in core. I think I'd prefer that to some other function like `disable_logging`. I think the "verbose" name here was chosen to avoid any confusion with message logging to sinks, and sticking with "log_level" hopefully avoids that confusion.

bryfox: I updated this as described above

python/foxglove-sdk/python/foxglove/__init__.py
```diff
@@ -167,37 +162,47 @@ def start_server(
...
-    return log_levels.get(level, "unknown")
-
-
-def verbose_on(level: Union[int, str] = "debug") -> None:
+def set_log_level(level: Union[int, str] = "INFO") -> None:
     """
     Enable SDK logging.
+
+    This function will call logging.basicConfig() for convenience in scripts, but in general you
+    should configure logging yourself: https://docs.python.org/3/library/logging.html
```
gasmith, line 170:
```suggestion
    should configure logging yourself before calling this function: https://docs.python.org/3/library/logging.html
```

python/foxglove-sdk/python/foxglove/__init__.py
```diff
@@ -214,6 +219,5 @@ def verbose_off() -> None:
     "log",
     "open_mcap",
     "start_server",
-    "verbose_off",
-    "verbose_on",
+    "set_log_level",
```
gasmith, line 222: Nit: Sort

gasmith (approved):

python/foxglove-sdk/python/foxglove/__init__.py
gasmith: Agreed. I suggested a small docstring clarification so that users understand the ordering requirement.

python/foxglove-sdk/python/foxglove/__init__.py
gasmith: I like the way this turned out.

PR #245, opened 2025-02-25
python: Add support for adding & removing services

gasmith:


python/foxglove-sdk/src/websocket_server.rs
```diff
@@ -393,6 +393,35 @@ impl PyWebSocketServer {
...
+    /// These services will be available for clients to use until they are removed with
+    /// :py:meth:`remove_services`.
+    ///
+    /// This method will fail if the server was not configured with :py:attr:`Capability.Services`.
+    ///
+    /// :param services: Services to add.
+    /// :type services: list[:py:class:`Service`]
+    pub fn add_services(&self, py: Python<'_>, services: Vec<PyService>) -> PyResult<()> {
+        if let Some(server) = &self.0 {
+            let services = services
```
eloff, line 408: Is it possible to easily move this into the allow threads closure? Not a big deal either way

python/foxglove-sdk/python/foxglove/tests/test_server.py
```diff
@@ -34,15 +34,34 @@ class DefaultServerListener(ServerListener):
...
     def test_services_interface(self) -> None:
+        test_svc = Service(
+            name="test",
+            schema=ServiceSchema(name="test-schema"),
+            handler=lambda *_: b"{}",
+        )
+        test2_svc = Service(
+            name="test2",
+            schema=ServiceSchema(name="test-schema"),
+            handler=lambda *_: b"{}",
```
eloff, line 45: Is this plumbed through to rust now? I probably missed when that happened

eloff (approved):
Looks good to me, thanks for the doc cleanups!

python/foxglove-sdk/src/websocket_server.rs
gasmith: Yep, we sure can.

python/foxglove-sdk/python/foxglove/tests/test_server.py
gasmith: Ah, sorry. I was probably moving too fast. #239 added support for services in python, #244 dropped the client argument from the callback.

PR #246, opened 2025-02-26
C++ setup

jtbandes:
### Changelog
None

### Docs

None

### Description

Some initial setup for C/C++ builds. Overall structure is:
- `c` folder with a Rust crate that builds a `staticlib`, `cdylib`, and uses [cbindgen](https://github.com/mozilla/cbindgen) to generate a C header
- `cpp` folder with a C++ wrapper around the C library, as well as tests and examples
  - Includes CMake scripts to build the tests and examples
  - Currently not planning to ship a prebuilt C++ library — users will have to link against the static/shared library and build the C++ sources in their own project.
- CI workflow that builds the static & shared libraries on several platforms and uploads them to action artifacts

The actual functionality is limited to constructing/starting/stopping a server. Will continue to add functionality in future PRs.


c/foxglove/src/lib.rs
```diff
@@ -0,0 +1,53 @@
+use std::ffi::{c_char, CStr};
+
+pub struct FoxgloveWebSocketServer {
+    server: Option<foxglove::WebSocketServer>,
+    handle: Option<foxglove::WebSocketServerBlockingHandle>,
+}
```
gasmith, line 6: An enum might be nice here:

```
pub enum FoxgloveWebSocketServer {
    Building(WebSocketServer),
    Running(WebSocketServerBlockingHandle),
}

File: c/foxglove/build.rs [hunk 2]
```diff
@@ -0,0 +1,11 @@
+use std::env;
+
+fn main() {
+    let crate_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
+
+    cbindgen::generate(crate_dir)
```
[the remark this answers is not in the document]
jtbandes: I'm not sure about this one, as it's kinda part of building this C crate? i.e. if I moved it to generate.ts then I would probably want to invoke that when building this package, which might seem backwards?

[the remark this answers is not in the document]
amacneil: true, I guess the main goal of `yarn generate` is that if you change the schemas you have one command to fix all the generated code. in this PR so far the generated code has no connection to foxglove schemas.

cpp/foxglove/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+target_include_directories(foxglove_c INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/../../c/foxglove/generated/include)
+
+file(GLOB_RECURSE foxglove_srcs CONFIGURE_DEPENDS
+    "include/*.hpp"
+    "src/*.hpp"
+    "src/*.cpp"
+)
+
+add_library(foxglove STATIC "${foxglove_srcs}")
+set_property(TARGET foxglove PROPERTY CXX_STANDARD 20)
```
indirectlylit, line 29: Are we using features which require C++20? I would have expected C++14 would be sufficient.

jtbandes: 2020 was already 5 years ago, do you think it's still too early to move forward with it?

indirectlylit: 🤷 my experience is c/c++ moves a lot slower. Would having this make it hard to incorporate an older library into a build, or is it possible to mix and match?

When I was building the C++ runtime of a project a couple years ago, I remember deciding on 14 because it included enough of the modern stuff I wanted, and there weren't major features after that I actually needed.

indirectlylit: (this is as opposed to cmake version, where I found a huge benefit to requiring newer cmake)

.github/workflows/c_cpp.yml
```diff
@@ -0,0 +1,74 @@
...
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - runner: ubuntu-24.04
+            target: aarch64-unknown-linux-gnu
+            lib_name: libfoxglove_c.a
```
amacneil, line 17: thoughts on calling it just `libfoxglove` instead of `libfoxglove_c`?

jtbandes: what should the C++ library be named then?

amacneil: Grok and chatgpt suggested calling the c library `libfoxglove` and the c++ library `libfoxglove++` or `libfoxglove-cpp`. They didn't think a suffix for the c variant was needed.

Not sure how I feel about special characters in the file name - it seems reasonably common (`libxml++`) but safer to avoid?

jtbandes: Could do `libfoxglove_c` and `libfoxglove_cpp` for consistency?

jtbandes: This may be somewhat moot as I'm currently not planning to ship prebuilt c++ libraries; people will just integrate it into their own build system. I stuck with the `foxglove_c` name, but currently the CI artifacts also include the `target` in the filename so they don't conflict. We can change this later if we clarify our distribution needs.

eloff:
> * Currently not planning to ship a prebuilt C++ library — users will have to link against the static/shared library and build the C++ sources in their own project.

So does this mean a header-only C++ library, that links to the C staticlib/dylib?

.github/workflows/c_cpp.yml
eloff: I like the idea of just libfoxglove. The fact that it's C is fairly obvious.

cpp/foxglove/CMakeLists.txt
eloff: I think 17 is a good compromise

c/include/foxglove-c/foxglove-c.h
```diff
@@ -0,0 +1,49 @@
...
+extern "C" {
+#endif // __cplusplus
+
+/**
+ * Create and start a server. The server must later be freed with `foxglove_server_destroy`.
+ *
+ * # Safety
+ * `name` and `host` must be null-terminated strings with valid UTF8.
+ */
+struct foxglove_websocket_server *foxglove_server_start(const char *name,
```
eloff, line 29: This will quickly get unwieldy, it would be better to pass in a C version of the ServerOptions struct. Don't have to address it now.

Maybe call this foxglove_start_server (verb first)

c/include/foxglove-c/foxglove-c.h
```diff
@@ -0,0 +1,49 @@
...
+ * Free a server created via `foxglove_server_start`.
+ *
+ * If the server has not already been stopped, it will be stopped automatically.
+ */
+void foxglove_server_destroy(struct foxglove_websocket_server *server);
+
+/**
+ * Stop and shut down a server.
+ */
+void foxglove_server_stop(struct foxglove_websocket_server *server);
```
eloff, line 43: maybe foxglove_stop_server (verb first?)

c/include/foxglove-c/foxglove-c.h
```diff
@@ -0,0 +1,49 @@
...
+struct foxglove_websocket_server *foxglove_server_start(const char *name,
+                                                        const char *host,
+                                                        uint16_t port);
+
+/**
+ * Free a server created via `foxglove_server_start`.
+ *
+ * If the server has not already been stopped, it will be stopped automatically.
+ */
+void foxglove_server_destroy(struct foxglove_websocket_server *server);
```
eloff, line 38: maybe foxglove_free_server, free is often used in C land, destroy sounds more like c++

eloff (approved):
LGTM, some minor bike-shedding type comments

c/include/foxglove-c/foxglove-c.h
jtbandes: Re verb first, my thinking was as we add more and more types, it’ll be nice to have related methods sorted near each other so that’s why I was thinking [type]_[method]

jtbandes:
> So does this mean a header-only C++ library, that links to the C staticlib/dylib?

Currently it includes cpp sources too, we could make it header-only but I’m not sure there’s much advantage to that?

.github/workflows/c_cpp.yml
jtbandes: If someone did want to build the c++ version into a shared library and install it, what would they call that?

eloff: libfoxglove_cpp or something like that, but optimize for the happy path

c/include/foxglove-c/foxglove-c.h
eloff: Ah, that's a good point. Carry on.

eloff:
> > So does this mean a header-only C++ library, that links to the C staticlib/dylib?
> 
> Currently it includes cpp sources too, we could make it header-only but I’m not sure there’s much advantage to that?

I seem to recall header-only libs being easier to setup in my build files, but that was a long time ago.

One less step, I'm not sure it matters that much.

eloff:
:ship: it

PR #247, opened 2025-02-26
rust: Wrap Duration and Timestamp schemas

gasmith:


rust/foxglove/src/lib.rs
```diff
@@ -155,6 +155,7 @@
 //! [tokio]: https://docs.rs/tokio/latest/tokio/
 
 #![warn(missing_docs)]
+#![cfg_attr(docsrs, feature(doc_auto_cfg))]
```
gasmith, line 158: This is so we get nice ``Available on crate feature `chrono` only`` annotations in our docs.

typescript/schemas/src/internal/generatePyclass.ts
```diff
@@ -492,73 +492,57 @@ export function generateTimeTypes(): string {
   return `
 /// A timestamp in seconds and nanoseconds
 ///
-/// :param seconds: The number of seconds since a user-defined epoch.
-/// :param nanos: The number of nanoseconds since the :py:attr:\`seconds\` value.
+/// :param sec: The number of seconds since a user-defined epoch.
```
gasmith, line 495: Just noticed that this says "user-defined"? That can't be right, can it?

rust/foxglove/src/schemas_wkt.rs
```diff
@@ -0,0 +1,238 @@
...
+    fn try_from(duration: std::time::Duration) -> Result<Self, Self::Error> {
+        let Ok(sec) = i32::try_from(duration.as_secs()) else {
+            return Err(FoxgloveError::DurationOutOfRange);
+        };
+        let nsec = duration.subsec_nanos();
+        Ok(Self { sec, nsec })
+    }
+}
+
+/// A timestamp, represented as an offset from the unix epoch.
```
bryfox, line 136: If you're not working with and converting from std/chrono types, I don't think we need to make this claim. MCAP is specifically agnostic about the choice of epoch.

But the limitations, comments, and fallible conversions make sense otherwise.

typescript/schemas/src/internal/generatePyclass.ts
bryfox: I just left a comment above, but this was intentional. The example epoch given in the mcap spec is robot boot time. My read of this is that it's perfectly fine to choose (and document) the UNIX epoch if users want to work with std & chrono, but that we don't otherwise need to make any assumptions about it.

rust/foxglove-proto-gen/src/lib.rs
```diff
@@ -199,6 +199,8 @@ pub fn generate_protos(proto_path: &Path, out_dir: &Path) -> anyhow::Result<()>
     }
 
     let mut config = prost_build::Config::new();
+    config.extern_path(".google.protobuf.Duration", "crate::schemas::Duration");
+    config.extern_path(".google.protobuf.Timestamp", "crate::schemas::Timestamp");
```
bryfox, line 203: I think I don't understand what this is doing. Aren't you also exporting these types from the schemas crate by publicly using `schemas_wkt` there?

bryfox (commented):
Nice.

I think a saturating, infallible conversion would be useful for a lot of users. But my vote would be to make it pretty explicitly opt-in via a custom trait. I'm not sure what that would mean for the python interface, but we can tackle that later.

typescript/schemas/src/internal/generatePyclass.ts
gasmith: Ah, I see. That makes sense. I'll update the documentation accordingly.

rust/foxglove-proto-gen/src/lib.rs
gasmith: Ah! By default, prost uses `::prost_types::Duration` and `::prost_types::Timestamp` as the types when it does codegen. This is just overriding that default to use our types instead.

The path is a bit funky. Publicly, we certainly want our wrappers to show up under `crate::schemas`. Internally, I had to put them under `crate::schemas_wkt` to cope with the fact that `src/schemas/` is exclusively for codegen, but that's an implementation detail I'd like to keep as private as possible, which is why I used the `crate::schemas::` path here. If you think it'd make things clearer, I could switch to `crate::schemas_wkt::`.

bryfox: Ah, I see. No, I think this is good — I completely misunderstood what `extern_path` was doing.

rust/foxglove/src/lib.rs
```diff
@@ -190,6 +190,7 @@ pub use mcap_writer::{McapWriter, McapWriterHandle};
 pub use metadata::{Metadata, PartialMetadata};
 pub(crate) use runtime::get_runtime_handle;
 pub use runtime::shutdown_runtime;
+pub use schemas_wkt::{RangeError, SaturatingFrom, SaturatingInto};
```
bryfox, line 193: Is this the only error we expose that isn't part of `FoxgloveError`? Should it be?

bryfox (approved):
Looks good to me. Matches the fields in our main documentation and is easier to work with.

One question about how we expose the error type for the conversion, but otherwise 👍 .

rust/foxglove/src/lib.rs
gasmith: It is the only one, so far. My justification for adding a new error type is that I wanted to build the `SaturatingFrom` trait implementations on top of it, and so I wanted a narrow error type to indicate which bound we've exceeded.

I don't love exposing the new error type from the SDK, but I don't see a way around it - it needs to be public. Maybe we could nest it under a module to avoid cluttering the top-level. If you like that idea, I'm open to suggestions for naming... `crate::convert::`?

bryfox: OK, makes sense. I'll defer to you and @eloff here. It looks like prost-types exposes similar errors at the top; on the other hand, `convert::RangeError` gives me some context about when/why I'd encounter that error, if I saw it in docs, etc.

eloff: I think I'd prefer if we make new FoxgloveError variant(s) instead of exporting a new error type.

Otherwise a caller could have a method returning FoxgloveError, but need to convert this error to be able to keep the same return type (or wrap both in some third error type).

gasmith: How about we add a `RangeError(#[from] RangeError)` variant on `FoxgloveError` to assist with that?

eloff: Maybe I'm overthinking this.

Probably the user would handle the error directly instead of passing it up the stack.

eloff (approved):

rust/foxglove/src/lib.rs
gasmith: Sounds good. I added the `From` implementation to convert `RangeError` to `FoxgloveError::Unspecified`, just for good measure.

PR #248, opened 2025-02-26
Enable rust-2024-compatibility lint rules

jtbandes:
### Changelog
None

### Docs

None

### Description

Enables lint rules to ease compatibility with the 2024 edition. A lot of these might require some manual inspection to decide what we want to do about them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,11 +1,11 @@
 [workspace]
 resolver = "2"
 members = [
-    "rust/foxglove",
-    "rust/foxglove-proto-gen",
-    "rust/examples/*",
-    "rust/examples-unstable/*",
-    "python/foxglove-sdk",
+  "rust/foxglove",
+  "rust/foxglove-proto-gen",
+  "rust/examples/*",
+  "rust/examples-unstable/*",
+  "python/foxglove-sdk",
 ]
 
 [workspace.dependencies]
@@ -20,3 +20,6 @@ tokio = { version = "1.40", features = ["macros", "rt-multi-thread", "signal", "
 tokio-tungstenite = "0.26"
 tokio-util = { version = "0.7", features = ["rt"] }
 tracing = { version = "0.1", features = ["log"] }
+
+[workspace.lints.rust]
+rust-2024-compatibility = "warn"
--- a/python/foxglove-sdk/Cargo.toml
+++ b/python/foxglove-sdk/Cargo.toml
@@ -8,6 +8,9 @@ publish = false
 name = "foxglove_py"
 crate-type = ["cdylib"]
 
+[lints]
+workspace = true
+
 [dependencies]
 bytes.workspace = true
 env_logger = "0.11.5"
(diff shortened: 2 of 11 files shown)


defunctzombie:
What's going on with this PR? Do we want this? Is this effort abandoned?

gasmith:
> What's going on with this PR? Do we want this? Is this effort abandoned?

It's worth doing. I'll take a look today.

c/src/parameter/tests.rs
```diff
@@ -82,7 +82,7 @@ fn test_bool() {
 }
 
 macro_rules! make_value {
-    ($ctor:ident, $value:expr) => {{
+    ($ctor:ident, $value:expr_2021) => {{
```
gasmith, line 85: We should just disable the `edition_2024_expr_fragment_specifier` lint. We don't currently have any macros that have special-case patterns for const expressions, and would generally prefer the 2024 semantics.

python/foxglove-sdk/src/lib.rs
```diff
@@ -229,10 +229,9 @@ fn open_mcap(
 
     let options = writer_options.map_or_else(McapWriteOptions::default, |opts| opts.into());
     let writer = BufWriter::new(file);
-    let handle = if let Some(context) = context {
-        McapWriter::with_options(options).context(&context.0)
-    } else {
-        McapWriter::with_options(options)
+    let handle = match context {
```
gasmith, line 232: We should disable the `if_let_rescope` lint as well. The rewrite is only desirable in the case where we're _intentionally_ abusing the broader scope of the `if-let` binding, which I don't think I've ever seen in the wild. I think it's very unlikely we'll run afoul of this convention in the future.

I've looked through all the rewrites here, and none of them actually need rewriting. And some of the rewrites really hurt readability (especially [this hunk](https://github.com/foxglove/foxglove-sdk/pull/248/files#diff-6442b9b09535fed1e455eccb7415a67ebbdf09296cfa660d37b08e0d35777abfR1008-R1044)).

rust/foxglove/src/log_macro.rs
```diff
@@ -12,11 +12,13 @@ impl ChannelPlaceholder {
...
-        // We can safely create a Channel from any Arc<RawChannel>
-        let channel = ManuallyDrop::new(Channel::<T>::from_raw_channel(channel_arc));
-        channel.log_with_meta(msg, metadata);
+        unsafe {
+            // Safety: we're restoring the Arc<RawChannel> we leaked into_raw in new()
+            let channel_arc = Arc::from_raw(channel_ptr as *mut RawChannel);
+            // We can safely create a Channel from any Arc<RawChannel>
+            let channel = ManuallyDrop::new(Channel::<T>::from_raw_channel(channel_arc));
+            channel.log_with_meta(msg, metadata);
+        }
```
gasmith, line 21:
```suggestion
        // Safety: we're restoring the Arc<RawChannel> we leaked into_raw in new()
        let channel_arc = unsafe { Arc::from_raw(channel_ptr as *mut RawChannel) };
        // We can safely create a Channel from any Arc<RawChannel>
        let channel = ManuallyDrop::new(Channel::<T>::from_raw_channel(channel_arc));
        channel.log_with_meta(msg, metadata);
```

jtbandes:
Sounds like @gasmith's proposed changes are significant enough that I don't think there's any value in keeping this PR open. Feel free to open a new one with the desired changes.

PR #249, opened 2025-02-26
rust: Improve service handler wrappers

gasmith:
Borrow some good ideas for wrapping callbacks from @eloff's asset handler work in #229.

In particular, I love the idea of handling the `tokio::spawn` and `tokio::task::spawn_blocking` on our side, because it handles the 99% case beautifully. Folks that want even finer control can always implement `Handler` themselves.

Changes here:
- Remove `handler_fn` (which took `(Request, Responder)` as arguments).
- Rename `sync_handler_fn` as `handler_fn`.
- Add `blocking_handler_fn` and `async_handler_fn` for blocking and async function handlers.
- Remove some useless `'static` bounds.

### DIFF
--- a/rust/examples/ws-services/src/server.rs
+++ b/rust/examples/ws-services/src/server.rs
@@ -25,25 +25,20 @@ pub async fn main(config: Config) -> Result<()> {
     // Simple services can be implemented with a closure.
     server
         .add_services([
-            Service::builder("/empty", empty_schema())
-                .sync_handler_fn(|_| anyhow::Ok(Bytes::new())),
+            Service::builder("/empty", empty_schema()).handler_fn(|_| anyhow::Ok(Bytes::new())),
             Service::builder("/echo", echo_schema())
-                .sync_handler_fn(|req| anyhow::Ok(req.into_payload())),
+                .handler_fn(|req| anyhow::Ok(req.into_payload())),
         ])
         .context("Failed to register services")?;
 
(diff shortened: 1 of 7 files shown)


rust/examples/ws-services/src/server.rs
```diff
@@ -99,6 +92,16 @@ fn set_bool_schema() -> ServiceSchema {
         .with_response("json", Schema::json_schema::<SetBoolResponse>())
 }
 
+async fn sleep_handler(_: Request) -> Result<Bytes, String> {
+    tokio::time::sleep(Duration::from_secs(1)).await;
+    Ok(Bytes::new())
+}
+
+fn blocking_handler(_: Request) -> Result<Bytes, String> {
+    std::thread::sleep(Duration::from_secs(1));
```
eloff, line 101: You could probably use shorter sleeps so the test runs faster

eloff (approved):
Nice! Now I feel like the time I spent on that API was worth it.

Looks like you also made some improvements by making the Arc an implementation detail and making the errors generic. I tried the latter and ran into problems, I want to give it another look tomorrow.

This looks great.

rust/examples/ws-services/src/server.rs
gasmith: Could, I suppose... but the client makes all of its `/sleep` calls in parallel, so it's only one second. Kind of nice to have that wait be user-visible, IMO.

PR #250, opened 2025-02-27
python: add close method to channels

bryfox:
Allow python implementations to explicitly close their channels, in order to stop advertisements to the app.

For both typed channels and BaseChannel, each python class stores the reference to the underlying channel as an option, and sets it to None when closed. Calling `log` after this results in debug logging, but does not raise an error.

I have not added this to any examples; I think this is a niche requirement which is not typically needed.


eloff (commented):
Looking good so far

gasmith (approved):
LG

PR #251, opened 2025-02-27
ros: Remove ImageMarkerArray

amacneil:
### Changelog
Drop support for `foxglove_msgs/ImageMarkerArray`

### Docs
None

### Description
`foxglove_msgs/ImageMarkerArray` was a hack that we used to support multiple image annotations on an image. It predates Foxglove schemas, and [foxglove_msgs/ImageAnnotations](https://docs.foxglove.dev/docs/visualization/message-schemas/image-annotations) has been our recommended way to display multiple image annotations for several years.


ros/foxglove_msgs/package.xml
```diff
@@ -19,7 +19,7 @@
   <build_depend condition="$ROS_VERSION == 2">rosidl_default_generators</build_depend>
   <build_depend>ros_environment</build_depend>
 
-  <depend>visualization_msgs</depend>
+  <depend>builtin_interfaces</depend>
```
jtbandes, line 22: Hm - this seems like it should be conditional on ROS 2. If we don't plan to ever publish for ROS 1 again, then we should remove all the conditional stuff.

amacneil: Good catch, updated.

I doubt we will publish any more ROS 1 updates but we can remove all that in another PR.

jtbandes (approved):
RIP 🪦

ros/foxglove_msgs/package.xml
defunctzombie: We should still be considerate of folks building ROS 1 things from source.

PR #253, opened 2025-02-27
readme: Add python and rust packages

amacneil:
- Update readme with with new python and rust SDKs (with disclaimer about beta status)
- Remove references to packages not contained in this repo

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,21 @@
 # Foxglove SDK
 
-This repo provides libraries and schemas to log and visualize multimodal data with [Foxglove](https://foxglove.dev).
+The Foxglove SDK allows you to log and visualize multimodal data with [Foxglove](https://foxglove.dev).
+
+> [!IMPORTANT]
+>
+> **This new Foxglove SDK is under active development.**
+>
+> It is ready for beta testing, but it is not yet fully documented and the API is subject to change.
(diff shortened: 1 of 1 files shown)


gasmith (approved):

README.md
```diff
@@ -11,56 +17,64 @@ Visit [Foxglove Docs](https://docs.foxglove.dev/) to get started.
...
+</tr>
+<tr>
 <td>
 
-Python implementation of the WebSocket protocol<br/>Repo: [foxglove/ws-protocol](https://github.com/foxglove/ws-protocol/tree/main/python)
+[foxglove_cpp](./cpp)
 
 </td>
+<td></td>
+<td>Foxglove SDK for C++</td>
```
jtbandes, line 53: a bit early to add these...but whatever floats your boat

jtbandes (approved):
I forget how we ended up with a HTML table here but wonder if it would be more manageable if using MD table syntax?

amacneil:
> I forget how we ended up with a HTML table here but wonder if it would be more manageable if using MD table syntax?

changed it a little while back because we were getting a huge markdown diff each time, due to prettier wanting to pad out each cell to the same width

README.md
amacneil: yeah I was going to wait to merge this, but I might just remove them and merge it sooner

amacneil: ok I removed c/c++ and added an additional disclaimer about beta testing

PR #254, opened 2025-02-28
python: improve documentation

bryfox:
This improves the python documentation & example; notably:

- Enum variants are documented (also in the stubs; IDEs will pick this up)
- ParameterValue nested classes are documented
- Added param types to docstrings
- All additional examples have been added to the docs
- The last remaining example was moved from sdk into sdk-examples

For the first two items, I resorted to manual documentation for the reasons noted in index.rst. This is unfortunate, and I think we're going to need to look into some sort of doc testing down the line, but it's better than missing documentation today.


python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -24,3 +65,44 @@ Channels
...
+
+      A decimal or integer value.
+
+   .. py:class:: Bool(value: bool)
+
+      A boolean value.
+
+   .. py:class:: Bytes(value: bytes)
+
+      A byte array
```
mguida22, line 100: Should this match the description In `_foglove_py/__init__.pyi`?

```suggestion
      A byte array, encoded as a base64-encoded string.
```

bryfox: Good catch. It's the other way around; the python implementation does the encoding for you. I've updated.

mguida22: Ah I see, in that case I think this is just missing a period.

mguida22 (approved):

python/foxglove-sdk/python/foxglove/__init__.py
```diff
@@ -155,12 +186,19 @@ def start_server(
     Start a websocket server for live visualization.
 
     :param name: The name of the server.
+    :type name: Optional[str]
     :param host: The host to bind to.
+    :type host: Optional[str]
     :param port: The port to bind to.
```
gasmith, line 192: We should add the default values here.

gasmith (approved):
LG

PR #255, opened 2025-02-28
Expose server port

jtbandes:
### Changelog
The server now exposes the port it's listening on, so you can pass a port of 0 and later determine the actual port used.

### Docs

None

### Description

Also tweaked a few memory orderings after discussing in slack – we can stick with the following defaults unless otherwise noted with a comment:
- load: Acquire
- store: Release
- fetch_add/compare_exchange etc.: AcqRel


jtbandes:
What's the magic incantation to update the generated type stubs? Also – do we need a lint step to enforce that they're up to date?

amacneil:
Pretty sure they are manually authored

bryfox:
> What's the magic incantation to update the generated type stubs? Also – do we need a lint step to enforce that they're up to date?

Yes, it's manual now, and relying on manual review (and unittest + mypy). I had a branch going to attempt to auto-generate them, but not there yet.

python/foxglove-sdk/CONTRIBUTING.md
```diff
@@ -6,12 +6,13 @@
...
 We use [Poetry](https://python-poetry.org/) to manage dependencies.
 
-Install Poetry:
+Install Poetry and Maturin:
 
 ```sh
 brew install pipx
 pipx ensurepath
 pipx install poetry
+pipx install maturin
```
bryfox, line 15: Was this required for you? maturin should be installed as part of `poetry install`, and then we run the commands with `poetry run maturin ...`

jtbandes: Might've done something wrong when I was getting things set up for the first time...I can remove it

rust/foxglove/src/websocket_server.rs
```diff
@@ -212,6 +214,13 @@ impl WebSocketServerHandle {
         self.0.runtime()
     }
 
+    /// Returns the local address that the server is listening on.
+    pub fn local_address(&self) -> Result<std::net::SocketAddr, FoxgloveError> {
```
gasmith, line 218: Is there a reason that rust exposes the full socketaddr, but other language bindings only expose the port number?

jtbandes: Hm, not exactly other than that I think I was scared to try and figure out how that type would be exposed to other languages.

Practically, do you know if the "IpAddr" part of the address ever changes from what was passed in?

How would you suggest to expose it? A string, or create some new wrapper type? Try to map to native networking APIs? (Not sure what that would look like on windows... I guess maybe windows also has sockaddr, but in a different header file? 😬 )

jtbandes: As an aside, when I worked on https://github.com/foxglove/foxglove-ios-bridge, my takeaway was that the server can be accessed at multiple IP addresses (on various network interfaces) but the port should always be the same.

gasmith: > Practically, do you know if the "IpAddr" part of the address ever changes from what was passed in?

I don't think I can answer authoritatively, but I would suspect that you could get a different address back in some situations on some platforms (e.g., with ipv6 and `::`).

I think I'd actually suggest the opposite: just expose the port from rust, unless someone actually has a use-case for recovering the rest of the sockaddr.

rust/foxglove/src/websocket.rs
```diff
@@ -1074,13 +1077,17 @@ impl Server {
...
+    }
+
     // Returns a handle to the async runtime that this server is using.
     pub fn runtime(&self) -> &Handle {
         &self.runtime
     }
 
-    // Spawn a task to accept all incoming connections and return
-    pub async fn start(&self, host: &str, port: u16) -> Result<String, FoxgloveError> {
+    // Spawn a task to accept all incoming connections and return the server's local port
```
gasmith, line 1089: nit: "... local addr"

rust/foxglove/src/websocket.rs
```diff
@@ -210,6 +210,8 @@ pub(crate) struct Server {
     /// It's analogous to the mixin shared_from_this in C++.
     weak_self: Weak<Self>,
     started: AtomicBool,
+    /// Local port the server is listening on, once it has been started
+    local_port: parking_lot::RwLock<Option<u16>>,
```
gasmith, line 214: Could make it an AtomicU16 now

rust/foxglove/src/websocket_server.rs
```diff
@@ -212,6 +214,11 @@ impl WebSocketServerHandle {
         self.0.runtime()
     }
 
+    /// Returns the local port that the server is listening on.
+    pub fn local_port(&self) -> Result<u16, FoxgloveError> {
```
gasmith, line 218: Just "port" to match the other language bindings?

rust/foxglove/src/websocket.rs
jtbandes: doh

gasmith (approved):
LG

PR #256, opened 2025-02-28
Enable a subset of clippy::pedantic lints

jtbandes:
### Changelog
None

### Docs

None

### Description

Enables some of the Pedantic lint group, and fixes the resulting errors.

I manually perused https://rust-lang.github.io/rust-clippy/rust-1.85.0/index.html?groups=pedantic and picked ones that seemed like they might be relatively obvious wins and not too controversial. Happy to adjust the list if these are not desired.


rust/examples/param-server/src/main.rs
```diff
@@ -76,16 +76,16 @@ impl ServerListener for ParamListener {
...
-                    param.value = old_param.value.clone();
-                    param.r#type = old_param.r#type.clone();
+                    param.value.clone_from(&old_param.value);
+                    param.r#type.clone_from(&old_param.r#type);
                 } else {
                     // Update the value
-                    old_param.value = param.value.clone();
-                    old_param.r#type = param.r#type.clone();
+                    old_param.value.clone_from(&param.value);
+                    old_param.r#type.clone_from(&param.r#type);
```
jtbandes, line 88: For some reason this lint only caught the instances in the `else` branch and not the `if` branch. I changed both because it seemed like an equivalent situation to me...

gasmith (approved):
Love it. Thanks for doing this!

PR #259, opened 2025-02-28
python: support asset serving

bryfox:
This adds support for an asset handler in the Python SDK. It follows the general approach of services: we expect a single function that returns the `bytes` of the asset, and call this function in a separate thread.

I also renamed "Request" -> "ServiceRequest" here. Even though the asset request is only a single URI, I think this clarifies what the request is for. It also localizes classes in the documentation.

I added a basic example which will serve "package://" assets from the local directory. We could include URDF descriptions in the example itself to make this easier to see out of the box; I haven't done so yet.


python/foxglove-sdk-examples/ws-asset-server/main.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def asset_handler(uri: str) -> bytes | None:
+    """
+    This will respond to "package://" asset requests from Foxglove by reading files from disk.
+    This example doesn't do any path validation or upward traversal prevention.
+    """
+    logging.debug(f"Asset request: {uri}")
+    if uri.startswith("package://"):
+        filepath = uri.replace("package://", "", 1)
```
eloff, line 14: package:///etc/passwd

I know it's an example, but maybe add a comment that this isn't secure for simplicity as an example

python/foxglove-sdk/src/websocket_server.rs
```diff
@@ -473,11 +483,42 @@ impl From<PyCapability> for foxglove::websocket::Capability {
     }
 }
 
+/// An asset server for live visualization.
+///
+/// The handler must be a callback function which takes the uri as its argument, and returns the
+/// asset `bytes`. If the handler returns `None`, a "not found" message will be sent to the client.
+/// If the handler raises an exception, the stringified exception message will be returned to the
+/// client as an error.
+pub struct AssetServer {
```
eloff, line 492: I think this could be simpler if you structured it as a closure passed as the blocking callback (if we're not using the builder you'd need to construct it yourself to set it in server options.)

At least it does the spawning and arc for you.

eloff (approved):
Looks good, maybe one possibility to simplify a little, but don't spend too long on it

python/foxglove-sdk-examples/ws-asset-server/main.py
bryfox: >     This example doesn't do any path validation or upward traversal prevention.

That's what I'd added; should I call it out more?

eloff: Nope, that will do the job

python/foxglove-sdk/src/websocket_server.rs
bryfox: I'm struggling to come up with the conversion for that. I'll merge this for now and take another look when I flesh out the URDF asset example.

python/foxglove-sdk/src/websocket_server.rs
```diff
@@ -356,6 +360,12 @@ pub fn start_server(
         server = server.services(services.into_iter().map(PyService::into));
     }
 
+    if let Some(asset_handler) = asset_handler {
+        server = server.fetch_asset_handler(Box::new(AssetServer {
```
jtbandes, line 364: Naming wise, the "AssetServer" name makes me think this is a standalone thing and not something I would necessarily expect to be passed in as an "asset handler"

could we call it something with AssetHandler in the name (I see AssetHandler is taken though) – PyAssetHandler? CallbackAssetHandler?

I guess using @eloff's suggestion of a closure would also resolve this concern because it would no longer need a name :)

python/foxglove-sdk-examples/ws-asset-server/main.py
```diff
@@ -0,0 +1,43 @@
...
+def main() -> None:
+    foxglove.set_log_level(logging.DEBUG)
+
+    server = foxglove.start_server(
+        asset_handler=asset_handler,
+    )
+
+    try:
+        while True:
+            # Send transforms for the model as needed, on a `FrameTransformsChannel`
```
jtbandes, line 35: stale comment?

python/foxglove-sdk-examples/ws-asset-server/main.py
```diff
@@ -0,0 +1,43 @@
...
+            time.sleep(1)
+
+    except KeyboardInterrupt:
+        server.stop()
```
jtbandes, line 39: feels like this could be nice to have as a helper method on the server, like `sleep_until_interrupted` or something like that?

gasmith: While it might be nice to have that sugar for these examples, I don't think it'd find much practical use in the real world. Most users of this SDK are going to start the server, _maybe_ register `server.stop()` as a pyexit hook if they're feeling fastidious, and then go about their _actual_ business.

python/foxglove-sdk/python/foxglove/__init__.py
```diff
@@ -19,9 +19,9 @@
     Parameter,
     ParameterType,
     ParameterValue,
-    Request,
     Schema,
     Service,
+    ServiceRequest,
```
gasmith, line 24: Good call on the rename.

python/foxglove-sdk/python/foxglove/__init__.py
```diff
@@ -183,7 +183,9 @@ def on_connection_graph_unsubscribe(self) -> None:
         return None
 
 
-ServiceHandler = Callable[["Request"], bytes]
+ServiceHandler = Callable[["ServiceRequest"], bytes]
+
+AssetHandler = Callable[[str], Optional[bytes]]
```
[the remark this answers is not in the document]
gasmith: In python, you can do whatever you want, which is why we spawn the callback on one of tokio's blocking threads. They can block as long as they want without causing any harm.

I don't have much experience with async python - and I get the impression it's not widely used - but if you really want to invoke an async function from the asset handler callback, I presume there's a way to get a handle to the asyncio eventloop and call `loop.run(my_async_fn)` to block and wait for it to complete.

python/foxglove-sdk/python/foxglove/_foxglove_py/__init__.pyi
```diff
@@ -180,7 +180,9 @@ AnyParameterValue = Union[
     ParameterValue.Dict,
 ]
 
-class Request:
+AssetHandler = Callable[[str], Optional[bytes]]
```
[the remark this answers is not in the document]
gasmith: I was wondering about that too. Feels like we should be able to do `from .. import AssetHandler, ServiceHandler` or so, to pull the type definition from the parent module.

python/foxglove-sdk/src/websocket_server.rs
```diff
@@ -538,11 +548,42 @@ impl From<PyCapability> for foxglove::websocket::Capability {
     }
 }
 
+/// An asset server for live visualization.
+///
+/// The handler must be a callback function which takes the uri as its argument, and returns the
+/// asset `bytes`. If the handler returns `None`, a "not found" message will be sent to the client.
+/// If the handler raises an exception, the stringified exception message will be returned to the
+/// client as an error.
+pub struct AssetServer {
```
gasmith, line 557: Doesn't need to be pub.

gasmith (approved):
LG

python/foxglove-sdk/python/foxglove/__init__.py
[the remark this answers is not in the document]
bryfox: I added docs for this. The callback is run in a separate thread. We decided against supporting async python interfaces in v1.

python/foxglove-sdk/python/foxglove/_foxglove_py/__init__.pyi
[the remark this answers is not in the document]
bryfox: I added a comment in the public module. Normally, that file imports everything from this stub, but we need to re-define the types for sphnix docs to work. I think this stub could probably import from the public file, but I don't think it should.

python/foxglove-sdk/src/websocket_server.rs
bryfox: Renamed to `CallbackAssetHandler` for now; I'll see if I can figure out conversions when I revisit the examples. This name isn't exposed to the python side, so I'd like to avoid the `Py` prefix.

python/foxglove-sdk-examples/ws-asset-server/main.py
bryfox: This was a placeholder instruction to the user; they'll need to publish the frame transforms.

I'm going to follow up with a more complete example, at which point I'll remove this.

python/foxglove-sdk/python/foxglove/__init__.py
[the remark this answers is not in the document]
jtbandes: > They can block as long as they want without causing any harm.

If they invoke something like a network call, will that result in the GIL being unlocked during the network call?

rust/foxglove/src/websocket.rs
```diff
@@ -876,6 +877,7 @@ impl ConnectedClient {
             let asset_responder = AssetResponder::new(Client::new(self), request_id, guard);
             handler.fetch(uri, asset_responder);
         } else {
+            tracing::error!("Server advertisd the Assets capability without providing a handler");
```
jtbandes, line 880:
```suggestion
            tracing::error!("Server advertised the Assets capability without providing a handler");
```

jtbandes (approved):

python/foxglove-sdk/python/foxglove/__init__.py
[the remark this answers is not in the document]
gasmith: Sure, just the same as if they had made the network call from any other context.

rust/foxglove/src/websocket.rs
bryfox: 😢 thanks

PR #260, opened 2025-02-28
python: Consolidate schema types

gasmith:
This change didn't go as far as I thought it would when I wrote the ticket.

It turns out that the (message_encoding + `Schema`) we use for channels is _really_ a (message_encoding + `Optional[Schema]`) because channels can be schema-less. The same is not true for service request/response schemas, which are mandatory.

Changes here:
- Dropped `SchemaDefinition`.
- The `Channel` constructor takes a message encoding and a `Schema`.
  - The message encoding is optional (_and ignored_) if the schema is a JSON schema.
  - The schema must be specified as a keyword argument, but may be `None` for schemaless.
- Plumbed `Schema` through the `BaseChannel` constructor.
- Removed the rejection of schemas with empty strings as encoding/data. If we really want to be that pedantic, we should do it in rust too.

I'm not thrilled with the result, but I thought I'd throw it up for review anyway. Feel free to nack.


python/foxglove-sdk-examples/live_visualization/main.py
```diff
@@ -111,11 +111,13 @@ def main() -> None:
...
-            schema_encoding="jsonschema",
+        schema=ChannelSchema(
             message_encoding="json",
-            schema_data=json.dumps(plot_schema).encode("utf-8"),
+            schema=Schema(
+                name="sine",
+                encoding="jsonschema",
+                data=json.dumps(plot_schema).encode("utf-8"),
+            ),
         ),
```
gasmith, line 121: An alternative to mull over:

```py
    sin_chan = foxglove.Channel(
        topic="/sine",
        message_encoding="json",
        schema=Schema(
            name="sine",
            encoding="jsonschema",
            data=json.dumps(plot_schema).encode("utf-8"),
        ),
    )
```

And for the case where you have a json schema as a plain object:

```py
    sin_chan = foxglove.Channel(
        topic="/sine",
        message_encoding="json", # optional?
        schema=json_schema,
    )
```

bryfox: I like this better. Solves the nested "schema" naming, and gets "message_encoding" out of the schema, which did seem a little strange.

> # optional?

I'd vote yes.

bryfox (commented):
I think this is an improvement. I started to write more feedback, but your added alternative seems to address this.

> Removed the rejection of schemas with empty strings as encoding/data.

👍 ...I think this was a holdover from an early python version; we should be consistent, and mcap allows empty schemas.

bryfox (approved):

PR #261, opened 2025-02-28
Disconnect slow client when control plane fills up

eloff:
There are two queues per connected client. One for data (log messages, sent with channel.log) with is lossy. When it's full, we drop the oldest queued message. But this doesn't work well for "control" messages where missing a message would be bad, like channel advertisements/unadvertisements, RPC responses, etc. So we have a second queue for these messages that isn't lossy.

But we still have to do *something* when the control plane queue fills up. If it's unbounded we consume resources until we crash the process. If it's bounded, but blocking, then slow clients can block other clients from making progress.

So we need to do something without blocking and without increasing the queue size. The only thing that really makes any sense is disconnect the slow client. It's obviously not a good experience for that client, but we have to protect the process and the other clients. So we make this a configurable message backlog size, and the if the user is getting disconnected, and they have the resources to burn, they can increase it. Conversely if the process is crashing with OOM they can decrease it.

To keep things simple there's just one configurable backlog size, and we use the same value for the data plane and control plane queues. They don't need to be the same size, but I'm not sure we need to expose two adjustable backlog sizes for the user. By default we use a backlog of 1024 messages.

When we disconnect the client, we tell them why with an error status message, and point them at the configurable backlog option.


rust/foxglove/src/websocket.rs
```diff
@@ -411,17 +412,24 @@ impl ConnectedClient {
...
         true
     }
 
-    fn on_disconnect(&self, server: &Arc<Server>) {
+    async fn on_disconnect(&self, server: &Arc<Server>) {
+        if self.cancellation_token.is_cancelled() {
+            let mut sender = self.sender.lock().await;
+            let status = Status::new(
+                StatusLevel::Error,
+                "disconnected because message backlog is full, consider increasing it".to_string(),
```
bryfox, line 426: "consider increasing it" sounds like it belongs more in the server log. I can't configure backlog size in the Foxglove app.

I'd also use sentence case here — "Disconnected".

bryfox (commented):
I can imagine test/sim use cases where there is only a single client, and where waiting is better than disconnecting entirely. So I'm not convinced this is always the right behavior.

However, I don't think anyone is very likely to run into the default limits with these 'control' messages, and we do provide a workaround. I'll let others weigh in and approve; I think this is a reasonable step for now.

rust/foxglove/src/websocket.rs
eloff: We can clarify that, but I really want to point the user at what knob they need to tune. They might have access to it (or if not they know who to talk to)

eloff:
> I can imagine test/sim use cases where there is only a single client, and where waiting is better than disconnecting entirely. So I'm not convinced this is always the right behavior.
> 
> However, I don't think anyone is very likely to run into the default limits with these 'control' messages, and we do provide a workaround. I'll let others weigh in and approve; I think this is a reasonable step for now.

I considered that, it is very tricky to implement though. I went down that mental rabbit hole for a bit and it looks hairy.

You have to be able to block everything that calls send_control_msg, which is a lot of functions, some called when processing messages from the client, and others invoked by sdk methods. Those may be in sync or async contexts. I think you'd need to provide both an async version of send_control_msg that blocks via await, and a sync version that blocks the thread, and then switch to the disconnect behavior when a second client connects (and maybe also switch back if the second last client disconnects.)

Not impossible, but maybe an opportunity for future improvement.

rust/foxglove/src/websocket.rs
```diff
@@ -322,6 +322,7 @@ pub(crate) struct ConnectedClient {
     /// Optional callback handler for a server implementation
     server_listener: Option<Arc<dyn ServerListener>>,
     server: Weak<Server>,
+    cancellation_token: CancellationToken,
```
gasmith, line 325: Add a description about how this is used. Maybe also give this a more self-descriptive name (`client_unresponsive`?).

gasmith (approved):
LG

PR #263, opened 2025-02-28
python: Add conversions for Duration and Timestamp

gasmith:
This change adds the following conversions into Duration and Timestamp for python:

- `Duration.from_secs(secs: float)`
- `Duration.from_timedelta(td: datetime.timedelta)`
- `Timestamp.from_epoch_secs(timestamp: float)`
- `Timestamp.from_datetime(dt: datetime.datetime)`

### DIFF
--- a/python/foxglove-sdk/python/foxglove/_foxglove_py/schemas_wkt.pyi
+++ b/python/foxglove-sdk/python/foxglove/_foxglove_py/schemas_wkt.pyi
@@ -48,7 +48,7 @@ class Timestamp:
         nsec: Optional[int] = None,
     ) -> "Timestamp": ...
     @staticmethod
-    def from_timestamp(timestamp: float) -> "Timestamp":
+    def from_epoch_secs(timestamp: float) -> "Timestamp":
         """
         Creates a :py:class:`Timestamp` from an epoch timestamp, such as is
         returned by :py:func:`time.time` or
--- a/python/foxglove-sdk/python/foxglove/_foxglove_py/schemas_wkt.pyi
+++ b/python/foxglove-sdk/python/foxglove/_foxglove_py/schemas_wkt.pyi
@@ -6,6 +6,9 @@ class Duration:
     A duration in seconds and nanoseconds
     """
 
+    sec: int
+    nsec: int
+
     def __new__(
         cls,
         sec: int,
@@ -41,6 +44,9 @@ class Timestamp:
     """
     A timestamp in seconds and nanoseconds
     """
+
+    sec: int
+    nsec: int
 
     def __new__(
         cls,
--- a/python/foxglove-sdk/python/foxglove/_foxglove_py/schemas_wkt.pyi
+++ b/python/foxglove-sdk/python/foxglove/_foxglove_py/schemas_wkt.pyi
@@ -1,3 +1,4 @@
+import datetime
 from typing import Optional
 
 class Duration:
@@ -10,6 +11,31 @@ class Duration:
         sec: int,
         nsec: Optional[int] = None,
     ) -> "Duration": ...
+    @staticmethod
+    def from_secs(secs: float) -> "Duration":
+        """
+        Creates a :py:class:`Duration` from seconds.
+
+        Raises `OverflowError` if the duration cannot be represented.
+
+        :param secs: Seconds
+        :type secs: float
+        :rtype: :py:class:`Duration`
+        """
+        ...
+
+    @staticmethod
+    def from_timedelta(td: datetime.timedelta) -> "Duration":
+        """
+        Creates a :py:class:`Duration` from a timedelta.
+
+        Raises `OverflowError` if the duration cannot be represented.
+
+        :param td: Timedelta
+        :type td: :py:class:`datetime.timedelta`
+        :rtype: :py:class:`Duration`
+        """
+        ...
 
 class Timestamp:
     """
@@ -21,3 +47,32 @@ class Timestamp:
         sec: int,
         nsec: Optional[int] = None,
     ) -> "Timestamp": ...
+    @staticmethod
+    def from_timestamp(timestamp: float) -> "Timestamp":
+        """
+        Creates a :py:class:`Timestamp` from an epoch timestamp, such as is
+        returned by :py:func:`time.time` or
+        :py:func:`datetime.datetime.timestamp`.
+
+        Raises `OverflowError` if the timestamp cannot be represented.
+
+        :param timestamp: Seconds since epoch
+        :type timestamp: float
+        :rtype: :py:class:`Timestamp`
+        """
+        ...
+
+    @staticmethod
+    def from_datetime(dt: datetime.datetime) -> "Timestamp":
+        """
+        Creates a UNIX epoch :py:class:`Timestamp` from a datetime object.
+
+        Naive datetime objects are presumed to be in the local timezone.
+
+        Raises `OverflowError` if the timestamp cannot be represented.
+
+        :param dt: Datetime
+        :type dt: :py:class:`datetime.datetime`
+        :rtype: :py:class:`Timestamp`
+        """
+        ...
--- a/python/foxglove-sdk/python/foxglove/_foxglove_py/schemas_wkt.pyi
+++ b/python/foxglove-sdk/python/foxglove/_foxglove_py/schemas_wkt.pyi
@@ -6,14 +6,15 @@ class Duration:
     A duration in seconds and nanoseconds
     """
 
-    sec: int
-    nsec: int
-
     def __new__(
         cls,
         sec: int,
         nsec: Optional[int] = None,
     ) -> "Duration": ...
+    @property
+    def sec(self) -> int: ...
+    @property
+    def nsec(self) -> int: ...
     @staticmethod
     def from_secs(secs: float) -> "Duration":
         """
@@ -45,14 +46,15 @@ class Timestamp:
     A timestamp in seconds and nanoseconds
     """
 
-    sec: int
-    nsec: int
-
     def __new__(
         cls,
         sec: int,
         nsec: Optional[int] = None,
     ) -> "Timestamp": ...
+    @property
+    def sec(self) -> int: ...
+    @property
+    def nsec(self) -> int: ...
     @staticmethod
     def from_epoch_secs(timestamp: float) -> "Timestamp":
         """
--- a/python/foxglove-sdk/python/foxglove/tests/test_time.py
+++ b/python/foxglove-sdk/python/foxglove/tests/test_time.py
@@ -5,7 +5,7 @@ from foxglove.schemas import Duration, Timestamp
 
 
 class TestTime(unittest.TestCase):
-    def test_duration_from_secs(self):
+    def test_duration_from_secs(self) -> None:
         self.assertEqual(Duration.from_secs(1.123), Duration(sec=1, nsec=123_000_000))
         self.assertEqual(Duration.from_secs(-0.123), Duration(sec=-1, nsec=877_000_000))
         self.assertEqual(Duration.from_secs(-1.123), Duration(sec=-2, nsec=877_000_000))
@@ -16,7 +16,7 @@ class TestTime(unittest.TestCase):
         with self.assertRaises(OverflowError):
             Duration.from_secs(1e42)
 
-    def test_duration_from_timedelta(self):
+    def test_duration_from_timedelta(self) -> None:
         td = datetime.timedelta(seconds=1, milliseconds=123)
         self.assertEqual(Duration.from_timedelta(td), Duration(sec=1, nsec=123_000_000))
 
@@ -38,7 +38,7 @@ class TestTime(unittest.TestCase):
         with self.assertRaises(OverflowError):
             Duration.from_timedelta(datetime.timedelta.max)
 
-    def test_timestamp_from_timestamp(self):
+    def test_timestamp_from_timestamp(self) -> None:
         self.assertEqual(
             Timestamp.from_timestamp(1.123), Timestamp(sec=1, nsec=123_000_000)
         )
@@ -49,7 +49,7 @@ class TestTime(unittest.TestCase):
         with self.assertRaises(OverflowError):
             Timestamp.from_timestamp(1e42)
 
-    def test_timestamp_from_datetime(self):
+    def test_timestamp_from_datetime(self) -> None:
         utc = datetime.timezone.utc
         dt = datetime.datetime(1970, 1, 1, tzinfo=utc)
         self.assertEqual(Timestamp.from_datetime(dt), Timestamp(sec=0))
--- a/python/foxglove-sdk/python/foxglove/tests/test_time.py
+++ b/python/foxglove-sdk/python/foxglove/tests/test_time.py
@@ -38,16 +38,16 @@ class TestTime(unittest.TestCase):
         with self.assertRaises(OverflowError):
             Duration.from_timedelta(datetime.timedelta.max)
 
-    def test_timestamp_from_timestamp(self) -> None:
+    def test_timestamp_from_epoch_secs(self) -> None:
         self.assertEqual(
-            Timestamp.from_timestamp(1.123), Timestamp(sec=1, nsec=123_000_000)
+            Timestamp.from_epoch_secs(1.123), Timestamp(sec=1, nsec=123_000_000)
         )
 
         with self.assertRaises(OverflowError):
-            Timestamp.from_timestamp(-1.0)
+            Timestamp.from_epoch_secs(-1.0)
 
         with self.assertRaises(OverflowError):
-            Timestamp.from_timestamp(1e42)
+            Timestamp.from_epoch_secs(1e42)
 
     def test_timestamp_from_datetime(self) -> None:
         utc = datetime.timezone.utc
--- a/python/foxglove-sdk/python/foxglove/tests/test_time.py
+++ b/python/foxglove-sdk/python/foxglove/tests/test_time.py
@@ -24,9 +24,18 @@ class TestTime(unittest.TestCase):
         )
 
         # argument conversions
-        Duration(sec=-(2**31))
-        Duration(sec=0, nsec=2**32 - 1)
-        Duration(sec=2**31 - 1, nsec=999_999_999)
+        d = Duration(sec=-(2**31))
+        self.assertEqual(d.sec, -(2**31))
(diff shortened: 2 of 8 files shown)


bryfox (approved):
Might be worth putting this in one of the examples (live-vizualization?) but looks good to me.

gasmith:
> Might be worth putting this in one of the examples (live-vizualization?) but looks good to me.

Oh yeah, I meant to do that, but forgot.

python/foxglove-sdk/python/foxglove/_foxglove_py/schemas_wkt.pyi
```diff
@@ -10,6 +11,31 @@ class Duration:
         sec: int,
         nsec: Optional[int] = None,
     ) -> "Duration": ...
+    @staticmethod
+    def from_secs(secs: float) -> "Duration":
```
jtbandes, line 15: I'm not a super heavy Python user but I wonder if it would be more pythonic to overload the constructor instead? e.g. `Duration(secs=...)` `Duration(timedelta=...)`?

rust/foxglove/src/schemas_wkt.rs
```diff
@@ -111,6 +111,45 @@ impl Duration {
     fn into_prost(self) -> prost_types::Duration {
         self.into()
     }
+
+    /// Creates a `Duration` from `f64` seconds, or fails if the value is unrepresentable.
+    pub fn try_from_secs_f64(secs: f64) -> Result<Self, RangeError> {
```
jtbandes, line 116: is it common to include the type name in the function name even if it's implied by the parameter type? (in other words, what about calling this `try_from_secs`?)

rust/foxglove/src/schemas_wkt.rs
```diff
@@ -111,6 +111,45 @@ impl Duration {
...
+                Err(RangeError::UpperBound)
+            }
+        } else {
+            let sec = secs as i32;
+            let nsec = ((secs - sec as f64) * 1e9) as i32;
+            if nsec < 0 {
+                assert!(sec <= 0);
+                Ok(Self {
+                    sec: sec - 1,
+                    nsec: u32::try_from(nsec + 1_000_000_000).expect("positive"),
```
jtbandes, line 134: nit: if this ever panics, we might want a more helpful error message than `"positive"`

On a related note, is there a nice rusty pattern for something like expect() but where you get to dynamically format a message (only when the try_from fails)? I've found it quite useful for our debugging flows to try and include as much info as possible in the error message, such as `"failed to convert 1234.567 to Duration"`

rust/foxglove/src/schemas_wkt.rs
```diff
@@ -242,6 +281,35 @@ impl Timestamp {
     fn into_prost(self) -> prost_types::Timestamp {
         self.into()
     }
+
+    /// Creates a `Timestamp` from `f64` timestamp seconds, or fails if the value is
+    /// unrepresentable.
+    pub fn try_from_timestamp_secs_f64(secs: f64) -> Result<Self, RangeError> {
```
jtbandes, line 287: why include `timestamp` in the name? Could this also be `try_from_secs`?

rust/foxglove/src/schemas_wkt.rs
```diff
@@ -242,6 +281,35 @@ impl Timestamp {
...
+                let sec = u32::MAX;
+                let nsec = ((secs - sec as f64) * 1e9) as u32;
+                Ok(Self { sec, nsec })
+            } else {
+                Err(RangeError::UpperBound)
+            }
+        } else {
+            let sec = secs as u32;
+            let nsec = ((secs - sec as f64) * 1e9) as u32;
+            Ok(Self { sec, nsec })
```
jtbandes, line 301: nit: since these two lines are identical to the other branch above, wonder if these could be combined if the logic is reordered

python/foxglove-sdk/python/foxglove/_foxglove_py/schemas_wkt.pyi
gasmith: Not really. Python doesn't do constructor overloading like C++ does. You only get one `__init__`, so if you want to overload, you end up with a ton of optional kwargs that you have to probe in order to guess which calling convention the caller was attempting.

rust/foxglove/src/schemas_wkt.rs
gasmith: It's not terribly common, but it does tend to happen for f32/f64. For example, `std::time::Duration` has [`try_from_secs_f32`](https://doc.rust-lang.org/std/time/struct.Duration.html#method.try_from_secs_f32) and [`try_from_secs_f64`](https://doc.rust-lang.org/std/time/struct.Duration.html#method.try_from_secs_f64). I presume the motivating factor is that you can't losslessly convert f32 to f64.

jtbandes: hmm, why did they decide to include these in the name if the language wouldn't let you pass the wrong type anyway?

rust/foxglove/src/schemas_wkt.rs
gasmith: Fair point.

I don't think there's a great rusty pattern that matches what you're looking for... maybe [`Result::inspect`](https://doc.rust-lang.org/std/result/enum.Result.html#method.inspect) and `panic!`:

```rust
u32::try_from(nsec + 1_000_000_000)
    .inspect(|e| panic!("expected {nsec} to be positive: {e}"))
    .unwrap()
```

python/foxglove-sdk/python/foxglove/_foxglove_py/schemas_wkt.pyi
jtbandes: I just learned about `@singledispatchmethod` 🤯   Let's never use that.

rust/foxglove/src/schemas_wkt.rs
gasmith: It's a public interface, so I wanted to leave open the possibility that there'd be a conversion from some other kind of seconds. Maybe that's going too far.

rust/foxglove/src/schemas_wkt.rs
gasmith: Yeah, can do.

rust/foxglove/src/schemas_wkt.rs
gasmith: Because sometimes you have an f32, and sometimes you have an f64. You want a conversion function for each of them, and those functions can't have the same name.

rust/foxglove/src/schemas_wkt.rs
bryfox: @bennetthardwick pointed me to this awhile ago, which he mentioned is suggested by clippy:

```rs
u32::try_from(nsec + 1_000_000_000)
    .unwrap_or_else(|e| panic!("{} {}", nsec, e))
```

gasmith: There it is!

rust/foxglove/src/schemas_wkt.rs
jtbandes: hm, interesting. I was mainly thrown off by this because it's already a member of the Timestamp type, so it felt redundant.

gasmith: maybe there's a better word for it... `try_from_epoch_f64`?

rust/foxglove/src/schemas_wkt.rs
jtbandes: ....gotcha, thanks for the ELI5. I was leaning on Swift/C++ experience and assuming Rust had function overloading 🤦

rust/foxglove/src/schemas_wkt.rs
gasmith: I think we should still call it "timestamp" in the python-facing interface, because that's consistent with the nomenclature from the `datetime` module.

gasmith: Done!

python/foxglove-sdk/src/schemas_wkt.rs
```diff
@@ -34,8 +95,8 @@ impl From<Timestamp> for foxglove::schemas::Timestamp {
 ///
 /// :param sec: The number of seconds in the duration.
 /// :param nsec: The number of nanoseconds in the positive direction.
-#[pyclass(module = "foxglove.schemas")]
-#[derive(Clone)]
+#[pyclass(module = "foxglove.schemas", eq)]
+#[derive(Clone, PartialEq)]
 pub struct Duration(pub(crate) foxglove::schemas::Duration);
```
jtbandes, line 100: Do the access specifiers here prevent the user from constructing a Timestamp/Duration out of raw values? If not, I wonder if we may need manual implementations of `==`, because the type system doesn't actually prevent `nsec` from being outside the [0, 1e9) range so there may be multiple representations of the same value. (Unless we are declaring those "invalid"?)

rust/foxglove/src/schemas_wkt.rs
```diff
@@ -242,6 +279,35 @@ impl Timestamp {
     fn into_prost(self) -> prost_types::Timestamp {
         self.into()
     }
+
+    /// Creates a `Timestamp` from seconds since epoch as an `f64`, or fails if the value is
+    /// unrepresentable.
+    pub fn try_from_epoch_secs_f64(secs: f64) -> Result<Self, RangeError> {
```
jtbandes, line 285: I still think there is some simplification to be had here, e.g. https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=07d918705f6b7ce9548066bb63a201a6

```rs
if secs < 0.0 {
    return Err(RangeError::LowerBound)
}
if secs >= u32::MAX as f64 + 1.0 {
    return Err(RangeError::UpperBound)
}
let sec = secs as u32;
let nsec = ((secs - sec as f64) * 1e9) as u32;
Ok(Self { sec, nsec })
```

jtbandes (approved):

rust/foxglove/src/schemas_wkt.rs
jtbandes: ...and maybe I've been staring at this too long, but why is `u32::MAX` the max integer value you're allowing? I guess it depends what "unrepresentable" means, but I don't think you'll lose precision on the `sec` part up until $2^{53}$ (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER)

python/foxglove-sdk/src/schemas_wkt.rs
gasmith: OK, now normalization is done eagerly when constructing `Duration` and `Timestamp`. Here's how things shook out:

- Rust:
  - Removed public visibility on `sec` and `nsec` fields.
  - Added getters for `sec` and `nsec` values.
  - Added `new` and `new_checked` constructors.
- Python:
  - Use `new_checked` constructor, and raise `OverflowError` when out of range.
  - Add getters for `sec` and `nsec` values.

python/foxglove-sdk/src/schemas_wkt.rs
```diff
@@ -1,13 +1,15 @@
...
+use pyo3::types::{timezone_utc, PyDateTime};
 
 /// A timestamp in seconds and nanoseconds
 ///
 /// :param sec: The number of seconds since a user-defined epoch.
 /// :param nsec: The number of nanoseconds since the :py:attr:\`sec\` value.
-#[pyclass(module = "foxglove.schemas")]
-#[derive(Clone)]
+#[pyclass(module = "foxglove.schemas", eq)]
+#[derive(Clone, PartialEq)]
```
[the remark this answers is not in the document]
gasmith: Added the derivation for `Eq` too. Can't hurt.

rust/foxglove/src/schemas_wkt.rs
gasmith: Nice. Simplified.

python/foxglove-sdk/python/foxglove/_foxglove_py/schemas_wkt.pyi
```diff
@@ -1,23 +1,84 @@
+import datetime
 from typing import Optional
 
 class Duration:
     """
     A duration in seconds and nanoseconds
     """
 
+    sec: int
+    nsec: int
```
jtbandes, line 10: If these are read-only (I think so because we use only `#[setter]`?) then it would be nice to expose them as such in the interface. I think the way you do that is just `@property def sec(self) -> int: ...`, I'm not sure if there's a more concise way

python/foxglove-sdk/python/foxglove/_foxglove_py/schemas_wkt.pyi
```diff
@@ -1,23 +1,84 @@
...
+
     def __new__(
```
jtbandes, line 12: just curious if there is a reason we defined these as `__new__` instead of `__init__`? I suspect since this is a .pyi only there's effectively no difference, other than I feel like `__new__` is more commonly customized in python

python/foxglove-sdk/python/foxglove/tests/test_time.py
```diff
@@ -0,0 +1,141 @@
...
+            Duration(sec=-2, nsec=1_000_000_001),
+            Duration(sec=-1, nsec=1),
+        )
+        self.assertEqual(
+            Duration(sec=-(2**31), nsec=1_000_000_001),
+            Duration(sec=-(2**31) + 1, nsec=1),
+        )
+
+        # argument conversions
+        Duration(sec=-(2**31))
```
jtbandes, line 27: should we test that all these roundtrip properly, i.e. that retrieving `.sec` actually gives `-(2**31)`?

python/foxglove-sdk/src/schemas_wkt.rs
```diff
@@ -1,26 +1,100 @@
...
+    #[pyo3(signature = (dt))]
+    fn from_datetime(py: Python, mut dt: Py<PyDateTime>) -> PyResult<Self> {
+        // If this is a naive datetime, presume the local timezone.
+        let tzinfo: Py<PyAny> = dt.getattr(py, "tzinfo")?;
+        if tzinfo.is_none(py) {
+            dt = dt.call_method0(py, "astimezone")?.extract(py)?;
+        }
+        let utc = timezone_utc(py);
+        let epoch = PyDateTime::new(py, 1970, 1, 1, 0, 0, 0, 0, Some(&utc)).unwrap();
+        let td = dt.call_method1(py, "__sub__", (epoch,))?;
```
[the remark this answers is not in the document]
jtbandes: I see it's also documented that "assumes there are exactly 3600*24 seconds in every day" 😅 
Might at least be worth some more comments on this stuff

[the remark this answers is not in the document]
jtbandes: It does look like perhaps `/` loses full precision while `//` is able to retain it:
```py
>>> (datetime.fromtimestamp(300*365*24*60*60,UTC)+timedelta(microseconds=1)-datetime.fromtimestamp(0,UTC))/timedelta(microseconds=1)
9460800000000000.0
>>> (datetime.fromtimestamp(300*365*24*60*60,UTC)+timedelta(microseconds=1)-datetime.fromtimestamp(0,UTC))//timedelta(microseconds=1)
9460800000000001
```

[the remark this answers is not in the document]
jtbandes: ooo and there's also divmod 😄  this might be my favorite so far since it gives us directly the result we want

```py
>>> divmod((datetime.fromtimestamp(300*365*24*60*60,UTC)+timedelta(microseconds=1)-datetime.fromtimestamp(0,UTC)), timedelta(seconds=1))
(9460800000, datetime.timedelta(microseconds=1))
```

python/foxglove-sdk/python/foxglove/_foxglove_py/schemas_wkt.pyi
gasmith: Right you are, nice catch.

python/foxglove-sdk/python/foxglove/_foxglove_py/schemas_wkt.pyi
gasmith: Dunno, I'm just following the convention from `schemas.pyi`, for better or worse.

In python, `__new__` is responsible for allocating an instance, and `__init__` is responsible for initializing ([docs](https://docs.python.org/3/reference/datamodel.html#classes)). When you're working in python, you typically implement `__init__`, except for unusual circumstances like metaclasses.

Since we're documenting the type interface, I'm not sure the distinction matters all that much.

ChatGPT says it's preferred to document the constructor as `__new__` for an extension class, but I'm not very convinced by its line of argument.

I think as long as IDEs and LSPs understand that this is the constructor interface, it probably doesn't matter too much. FWIW, `pyright` seems to understand it as such.

python/foxglove-sdk/python/foxglove/tests/test_time.py
gasmith: Yes, good idea.

python/foxglove-sdk/src/schemas_wkt.rs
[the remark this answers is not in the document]
gasmith: > Are we doing this stuff to avoid total_seconds intentionally?

Yes. If the caller is trying to convert a datetime, which is already integral, there's no point in bouncing that over to floating point and losing precision along the way.

> I see it's also documented that "assumes there are exactly 3600*24 seconds in every day"

Right. That's unix time for you. Every day is 86,400 seconds, [no exceptions](https://en.wikipedia.org/wiki/Unix_time#Leap_seconds).

> ooo and there's also divmod 😄 this might be my favorite so far since it gives us directly the result we want

It's possible, but it's a lot more fiddly work than what I currently have. We'd need to import `divmod` from `builtins`, construct the `timedelta(seconds=1)` divisor, call `divmod`, extract the `(u32, PyAny)` result, downcast the `PyAny` into a `PyDelta`, getattr microseconds and extract as u32, and finally multiply by 1000. Not sure I see the benefit.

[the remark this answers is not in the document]
gasmith: Maybe this is an argument for putting even more of the interface in pure python, rather than in rust. We could have a pure-python wrapper with various constructors around a more spartan rust class. Seems like a lot of work, though, to accommodate this one particular case.

PR #267, opened 2025-03-01
C/C++: add basic channel wrapper and log method

jtbandes:
### Changelog
C/C++ added Channel type.

### Docs

None

### Description
At least the basics of creating a schema+channel and logging messages seem to be working.

Some sadnesses/open areas to explore:
- Not sure if we want Schema to be its own ref-counted/unique_ptr'd entity, or just a plain ol' options struct. Right now I have a C++ options struct using C++ types (`std::string_view`) and a C options struct using C types (`const char*`). The conversion between these is super awkward. And the ownership/lifetime requirements of the string_views/pointers is not very clear (would need to just write it in comments).
- I want to mark some pointers as non-nullable on the C header side, but still check them for null in Rust to be safe. cbindgen will mark them as nonnull if they are `&T`, `&mut T`, or `NonNull<T>`. However, reading a bit into how the Rust side treats those types, I'm not sure that checking them for null is really possible (the compiler might be allowed to assume they are not?)  It seems like `Option<NonNull<T>>` might work, but that's kinda nuts when you think about it (could it be working by accident?)
- It feels silly and inefficient to wrap `Box` around a type that is already internally an `Arc`. Not sure there is any way around this though.
- I really would prefer `std::span` (C++20) over this ptr+length nonsense 😭 
- Need to tackle error handling across all these functions at some point – probably all will need to return some error by out-param? or return an error code?

### DIFF
--- a/c/src/lib.rs
+++ b/c/src/lib.rs
@@ -85,7 +85,9 @@ pub extern "C" fn foxglove_server_stop(server: Option<&mut FoxgloveWebSocketServ
 /// Create a new channel. The channel must later be freed with `foxglove_channel_free`.
 ///
 /// # Safety
-/// `topic` and `message_encoding` must be null-terminated strings with valid UTF8.
+/// `topic` and `message_encoding` must be null-terminated strings with valid UTF8. `schema` is an
+/// optional pointer to a schema. The schema and the data it points to need only remain alive for
+/// the duration of this function call (they will be copied).
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn foxglove_channel_create(
     topic: *const c_char,
--- a/c/src/lib.rs
+++ b/c/src/lib.rs
@@ -3,8 +3,18 @@
 #![warn(unsafe_attr_outside_unsafe)]
 
 use std::ffi::{c_char, CStr};
+use std::sync::Arc;
 
 pub struct FoxgloveWebSocketServer(Option<foxglove::WebSocketServerBlockingHandle>);
+pub struct FoxgloveChannel(Arc<foxglove::Channel>);
+
+#[repr(C)]
+pub struct FoxgloveSchema {
+    pub name: *const c_char,
+    pub encoding: *const c_char,
+    pub data: *const u8,
+    pub data_len: usize,
+}
 
 /// Create and start a server. The server must later be freed with `foxglove_server_free`.
 ///
@@ -19,25 +29,19 @@ pub unsafe extern "C" fn foxglove_server_start(
     host: *const c_char,
     port: u16,
 ) -> *mut FoxgloveWebSocketServer {
+    let name = unsafe { CStr::from_ptr(name) }
+        .to_str()
+        .expect("name is invalid");
+    let host = unsafe { CStr::from_ptr(host) }
+        .to_str()
+        .expect("host is invalid");
     Box::into_raw(Box::new(FoxgloveWebSocketServer(Some(
         foxglove::WebSocketServer::new()
-            .name(unsafe { CStr::from_ptr(name) }.to_str().unwrap())
-            .bind(unsafe { CStr::from_ptr(host) }.to_str().unwrap(), port)
+            .name(name)
+            .bind(host, port)
             .start_blocking()
             .expect("Server failed to start"),
     ))))
-}
-
-/// Get the port on which the server is listening.
-#[unsafe(no_mangle)]
-pub extern "C" fn foxglove_server_get_port(server: Option<&FoxgloveWebSocketServer>) -> u16 {
-    let Some(server) = server else {
-        panic!("Expected a non-null server");
-    };
-    let Some(ref handle) = server.0 else {
-        panic!("Server already stopped");
-    };
-    handle.port()
 }
 
 /// Free a server created via `foxglove_server_start`.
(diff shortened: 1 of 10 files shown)


cpp/foxglove/src/channel.cpp
```diff
@@ -0,0 +1,29 @@
...
+                     reinterpret_cast<const uint8_t*>(schema->data),
+                     schema->dataLen,
+                   })
+                 : nullptr
+        ),
+        foxglove_channel_free
+      ) {}
+
+void Channel::log(
+  const std::byte* data, size_t dataLen, uint64_t logTime, uint64_t publishTime, uint32_t sequence
```
eloff, line 22: Again you'll want some way to omit these time and sequence arguments. In C++ you could use function overloading maybe.

c/src/lib.rs
```diff
@@ -65,3 +81,79 @@ pub extern "C" fn foxglove_server_stop(server: Option<&mut FoxgloveWebSocketServ
...
+    let channel = channel.expect("channel is required");
+    if data.is_null() {
+        panic!("data is required");
+    }
+    channel.0.log_with_meta(
+        unsafe { std::slice::from_raw_parts(data, data_len) },
+        foxglove::PartialMetadata {
+            log_time: Some(log_time),
+            publish_time: Some(publish_time),
+            sequence: Some(sequence),
```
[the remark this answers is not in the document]
jtbandes: How important do we think it is to be able to set or not set each of these separately? Is it too much to ask people to just provide all 3? We could also provide two variants, one with metadata (all 3 fields) and one without any metadata. Otherwise, yeah we'll have to do the bool thing you described 😓

c/src/lib.rs
```diff
@@ -65,3 +81,79 @@ pub extern "C" fn foxglove_server_stop(server: Option<&mut FoxgloveWebSocketServ
...
+                .to_str()
+                .expect("schema name is invalid");
+            let encoding = CStr::from_ptr(schema.encoding)
+                .to_str()
+                .expect("schema encoding is invalid");
+            let data = std::slice::from_raw_parts(schema.data, schema.data_len);
+            foxglove::Schema::new(name, encoding, data)
+        })
+    };
+    Box::into_raw(Box::new(FoxgloveChannel(
```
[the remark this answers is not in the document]
jtbandes: Yes, agreed that sounds like it should work to remove this layer of indirection.

c/src/lib.rs
```diff
@@ -65,3 +81,79 @@ pub extern "C" fn foxglove_server_stop(server: Option<&mut FoxgloveWebSocketServ
...
+}
+
+/// Log a message on a channel.
+///
+/// # Safety
+/// `data` must be non-null, and the range `[data, data + data_len)` must contain initialized data
+/// contained within a single allocated object.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn foxglove_channel_log(
+    channel: Option<&mut FoxgloveChannel>,
```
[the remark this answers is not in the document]
jtbandes: I was hoping to end up with a nonnull annotation on the C header side as well, but not sure cbindgen will allow me to do that while retaining the Option on the rust side. I wonder how receptive they are to upstream PRs...

c/src/lib.rs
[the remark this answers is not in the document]
jtbandes: This is turning out trickier than I expect so I'll probably do it in a follow-up PR

cpp/foxglove/src/channel.cpp
```diff
@@ -0,0 +1,29 @@
+#include <foxglove-c/foxglove-c.h>
+#include <foxglove/channel.hpp>
+
+namespace foxglove {
+
+Channel::Channel(std::string topic, std::string messageEncoding, std::optional<Schema> schema)
```
achim-k, line 6: `const char *` for strings?

cpp/foxglove/include/foxglove/channel.hpp
```diff
@@ -0,0 +1,29 @@
...
+class Channel final {
+public:
+  Channel(std::string topic, std::string messageEncoding, std::optional<Schema> schema);
+
+  void log(
+    const std::byte* data, size_t dataLen, uint64_t logTime, uint64_t publishTime, uint32_t sequence
+  );
+
+private:
+  std::unique_ptr<foxglove_channel, void (*)(foxglove_channel*)> _impl;
```
achim-k, line 26: nit: could also use a raw pointer and free it in the destructor

cpp/foxglove/tests/test_server.cpp
```diff
@@ -11,3 +12,16 @@ TEST_CASE("Start and stop server") {
...
+TEST_CASE("Log a message") {
+  foxglove::WebSocketServerOptions options;
+  options.name = "unit-test";
+  options.host = "127.0.0.1";
+  options.port = 0;
+  foxglove::WebSocketServer server{options};
+  REQUIRE(server.port() != 0);
+
+  foxglove::Channel channel{"example", "json", std::nullopt};
+  const uint8_t data[] = {1, 2, 3};
```
achim-k, line 25: Does this not work?
```suggestion
  const std::byte data[] = {1, 2, 3};
```

jtbandes: nope, integer literals can't be assigned without a cast, so I could do `std::byte(1), ...`...

cpp/foxglove/src/channel.cpp
jtbandes: changed to string_view

cpp/foxglove/include/foxglove/channel.hpp
jtbandes: what's the advantage there?

PR #269, opened 2025-03-04
rust: Demote &Arc<Channel> to &Channel

gasmith:
I think we can relax `&Arc<Channel>` to just `&Channel` in `LogSink::log`.


bryfox (approved):

eloff (approved):
Thanks for doing this. The use of Arc<Channel> here used to be because we did an await inside a block_on for the websocket server. Now it's all sync because we use the channels.

PR #271, opened 2025-03-05
python: Don't override message_schema for json schemas

gasmith:
I watched @amacneil today as he did this:

```python
chan = Channel("topic", message_encoding="msgpack", schema={ ... })
```

That doesn't work as expected, because we automatically override the message_encoding to be json when you provide a schema that looks like a dict jsonschema. Which is a silly thing to do, in retrospect. The schema encoding does not imply a message encoding.

### DIFF
--- a/python/foxglove-sdk/python/foxglove/channel.py
+++ b/python/foxglove-sdk/python/foxglove/channel.py
@@ -118,12 +118,13 @@ def _normalize_schema(
     elif isinstance(schema, dict):
         if schema.get("type") != "object":
             raise ValueError("Only object schemas are supported")
-        message_encoding = message_encoding or "json"
-        schema = Schema(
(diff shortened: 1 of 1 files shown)


python/foxglove-sdk/python/foxglove/channel.py
```diff
@@ -118,11 +118,12 @@ def _normalize_schema(
     elif isinstance(schema, dict):
         if schema.get("type") != "object":
             raise ValueError("Only object schemas are supported")
-
-        return "json", Schema(
+        message_encoding = message_encoding or "json"
+        schema = Schema(
```
amacneil, line 122: overwriting `schema` here seems messy - better to give the var a different name

amacneil (approved):

amacneil (approved):

PR #272, opened 2025-03-05
rust: Use Bytes instead of Vec<u8> in protobufs

gasmith:
For bytes fields on protobufs, we should use `Bytes` instead of `Vec<u8>`. This enables zero-copy for buffers that are already `Bytes` of course, but also for types `T: AsRef<[u8]> + Send + 'static` (via [`Bytes::from_owner`](https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.from_owner)), such as you'd find for an `Arc<Vec<u8>>` or [Mmap](https://docs.rs/memmap/latest/memmap/struct.Mmap.html) or other various memory views.


bryfox (approved):
Should also end up being useful for #220.

PR #274, opened 2025-03-07
C/C++: add server callbacks for on_un/subscribe

jtbandes:
### Changelog
- C/C++: added server options with callback functions for subscribe and unsubscribe
- C/C++: added getter for channel id

### Docs

N/A

### Description

In addition to the features mentioned above:
- Updated C++ example server to log subscribe/unsubscribe events
- Added end-to-end C++ test using websocketpp asio client
- Changed some `string_view` to `string` because I learned that `string_view` is not necessarily null-terminated, but on the Rust side we assume it is


cpp/foxglove/src/server.cpp
```diff
@@ -1,11 +1,35 @@
...
+
+}  // anonymous namespace
+
 namespace foxglove {
 
 WebSocketServer::WebSocketServer(WebSocketServerOptions options)
-    : _impl(
-        foxglove_server_start(options.name.c_str(), options.host.c_str(), options.port),
+    : _callbacks(options.callbacks)
+    , _impl(
```
eloff, line 16: That's a brutal initializer, is there a more readable way to do this in C++? I mean maybe this is idiomatic, I try not to remember.

What if you make a method on WebSocketServerOptions that returns foxglove_server_options?

eloff (approved):
Some minor comments, otherwise looking great

cpp/foxglove/src/server.cpp
jtbandes: This got messy because I was trying to keep most of the initialization in the member-initializer-list, but I could just give up on that and put it in the constructor body...

cpp/foxglove/include/foxglove/channel.hpp
```diff
@@ -10,22 +10,26 @@ struct foxglove_channel;
...
+  std::string encoding;
+  const std::byte* data = nullptr;
+  size_t dataLen = 0;
 };
 
 class Channel final {
 public:
-  Channel(std::string_view topic, std::string_view messageEncoding, std::optional<Schema> schema);
+  Channel(
+    const std::string& topic, const std::string& messageEncoding, std::optional<Schema> schema
```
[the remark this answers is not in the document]
jtbandes: yeah - I guess I had two thoughts on that:
1. `const char*` (null terminated) seems pretty normal/expected for a C API
2. if we passed pointer+length, on the Rust side then we wouldn't be able to use `CStr::from_ptr`, which is not the end of the world, but on the other hand this is exactly what it's designed for.

PR #276, opened 2025-03-11
python: organize websocket code into submodule

bryfox:
### Changelog

- [python] Move websocket-related classes into a `websocket` module

### Docs
None
### Description

This re-organizes existing python code by introducing a `websocket` module. `start_server` is still in the top-level namespace, along with all other functions. I think we could go either way on this, but this sort of follows the rust organization, which has the server at the top level, with supporting classes in a websocket module.


eloff (approved):
Makes sense to me. I prefer having start server in the top level package like this, but I don't feel too strongly about it.

gasmith (approved):

PR #278, opened 2025-03-11
python: update & fix examples

bryfox:
### Changelog
None
### Docs
None
### Description

This updates and, in places, fixes the python examples.

- Remove lock files
- Add a CI job to run all examples
- Fix some issues uncovered by the above (typing in python 3.9; renamed class)

This also updates node in other CI jobs for consistency; v18 will be EOL soon.

Once the API has stabilized, I would like to reference published SDK versions with flexible constraints. This will make them easier to run, since the current setup requires a rust toolchain. Having a CI step in place will make sure examples continue to run correctly as we publish new versions.

The new CI job is intended to catch basic errors in the python examples. Because many of the examples run a live server indefinitely, these run for 5s and then exit. I put the script at the top-level since we have scripts there already, and because I don't want to clutter the examples directory with auxiliary stuff. These also run sequentially as noted, because I don't want to update examples to serve this CI step; I don't think it'll be the bottleneck yet. There's room for improvement down the line.


bryfox:
Failures here are because of the breaking module change from #276. We could release a 0.3.1 with that change, and update here; I don't foresee other breaking changes until we gather more feedback in the next couple of weeks.

Long term, I don't have a better plan for breaking changes other than constraining versions until releases are available. We could keep referencing the local SDK for now. The big downside in my mind is that people wanting to run these examples then need to also set up the rust toolchain. Open to thoughts here.

scripts/runPythonSdkExamples.ts
```diff
@@ -0,0 +1,91 @@
...
+      } else {
+        reject(new Error(`Failed to install dependencies for ${name}`));
+      }
+    });
+  });
+}
+
+async function newTempFile() {
+  const prefix = `${tmpdir()}${path.sep}`;
+  const dir = await mkdtemp(prefix);
```
gasmith, line 73: Should we clean this up on exit?

gasmith (approved):

bryfox:
I bumped the minor version of pyo3 because of [this build error](https://github.com/foxglove/foxglove-sdk/actions/runs/13835203473/job/38708441262?pr=278); will need to look into our versioning constraints at some point. Also bumped pyo3-log a minor version.

PR #279, opened 2025-03-11
rust: Rename LogContext->Context, LogSink->Sink

gasmith:
### Changelog
- `LogContext` renamed to `Context`
- `LogContext::global` renamed to `Context::get_default`.
- `Context` no longer implements `Default`, to avoid confusion.
- `LogSink` renamed to `Sink`


gasmith:
> Could you comment on the choice of `Namespace`?

After discussion, we've settled on keeping `Context` (but dropping the `Log` prefix).

PR #281, opened 2025-03-11
rust: Add subscription manager

gasmith:
### Changelog
None

### Related changes
- https://github.com/foxglove/foxglove-sdk/pull/292 
- https://github.com/foxglove/foxglove-sdk/pull/293

### Description
Updates:
- Removed synchronization, we'll do that in the `Context` with an `RwLock` around channels, sinks, and subs together.
- Removed the `ArcSwap` cache - pre-computing the SmallVec doesn't make much of a difference since it's small and we have to clone it anyway.
- Moved the module to `crate::context::subscriptions`.

This change adds a new type for tracking subscriptions of sinks to channels, but doesn't make use of it yet. That will come in a later PR.

We're planning to support two kinds of subscriptions: static subscriptions to all channels (e.g., for mcap), and dynamic subscriptions to particular channels (e.g., for websockets). We maintain these subscriptions as two separate sets. When subscriptions change, the `Context` will be responsible for querying the subscription manager in order to regenerate the `LogSinkSet` for affected channels.

Part of: FG-10743


rust/foxglove/src/subscription/maps.rs
```diff
@@ -0,0 +1,156 @@
...
+    }
+}
+
+/// A cached map from channel name to a set of interested subscribers.
+///
+/// This representation is different from the underlying system of record in [`SubscriptionMap`] in
+/// two significant ways:
+///
+///  - The `channel` map includes global subscribers.
+///  - Subscribers are stored as [`SubscriberSet`]s instead of hashmaps.
```
jtbandes, line 107: comment seems out of date:
- `channel name`
- `SubscriberSet`

rust/foxglove/src/subscription/maps.rs
```diff
@@ -0,0 +1,156 @@
...
+///
+pub(crate) struct SubscriberMap<T> {
+    /// The set of global subscribers.
+    global: SubscriberVec<T>,
+    /// A map from channel name to the set of subscribers interested in that channel, including global
+    /// subscribers.
+    channel: HashMap<ChannelId, SubscriberVec<T>>,
```
jtbandes, line 114: naming this `channel` is a bit confusing if this is not just channel-specific subscribers; I've become fond of the "V by K" naming scheme for maps, e.g. something like `all_by_channel_id`? (helps readability as `all_by_channel_id.get(id)` is clearer than `channel.get(id)`)

rust/foxglove/src/subscription/maps.rs
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /// Remove all global and per-channel subscriptions for a particular subscriber.
+    pub fn remove_subscriber<Q>(&mut self, key: &Q) -> bool
+    where
+        Q: ?Sized + Hash + Eq,
+        K: Borrow<Q>,
+    {
+        let mut removed = self.global.remove(key).is_some();
+        self.channel.retain(|_, subs| {
```
jtbandes, line 93: interesting - could unsubscribe_channels also be written using `retain`? might be slightly nicer for them to match

rust/foxglove/src/subscription/maps.rs
```diff
@@ -0,0 +1,156 @@
...
+}
+
+impl<K, V> Default for SubscriptionMap<K, V> {
+    fn default() -> Self {
+        Self {
+            global: HashMap::default(),
+            channel: HashMap::default(),
+        }
+    }
+}
```
jtbandes, line 26: could this be `#[derive(Default)]`?

rust/foxglove/src/subscription.rs
```diff
@@ -0,0 +1,111 @@
...
+mod tests;
+
+/// A set of distinct subscribers.
+///
+/// We use a [`SmallVec`] to avoid heap allocations for lookups, when the number of concurrent
+/// subscribers (i.e., sinks) is reasonably small.
+pub(crate) type SubscriberVec<T> = SmallVec<[T; 4]>;
+
+/// A structure for tracking and querying channel subscriptions.
+pub(crate) struct SubscriptionManager<K, V> {
```
jtbandes, line 27: I don't feel like I quite understand why this is generic over K,V -- if there's nothing "subscription"-related about the type requirements, why is it called a subscription manager? What would a `SubscriptionManager<u16, u8>` mean?

I wonder if it would make sense to only have one generic parameter and derive the keys from the values (i.e. the value type can implement some trait for getting its key), since it seems unlikely that we would want to use this with "out of line" keys that are unrelated to the values?

Another way of thinking about it: in a sense, I think this is more of a glorified "Set" of subscribers+subscriptions than it is a glorified "map"...

rust/foxglove/src/subscription.rs
```diff
@@ -0,0 +1,111 @@
...
+    /// Current subscriptions.
+    subscriptions: Mutex<SubscriptionMap<K, V>>,
+    /// Cached map from channel to interested subscribers.
+    subscribers: ArcSwap<SubscriberMap<V>>,
```
jtbandes, line 31: It seems like it would be incorrect/racy to ever perform a **store** on `subscribers` when not holding the subscriptions mutex. Is there a way to encode that better in the type system?

Relatedly, I am a bit skeptical of where we may end up using `has_subscribers()` and whether (because of the lack of locking) it will be easy to accidentally introduce races by assuming that a subscriber doesn't sneak in between calls to `has_subscribers()` and other functions...

rust/foxglove/src/subscription/maps.rs
```diff
@@ -0,0 +1,156 @@
...
+        key: K,
+        value: V,
+        channel_ids: impl IntoIterator<Item = ChannelId>,
+    ) -> bool {
+        let mut inserted = false;
+        for channel_id in channel_ids {
+            inserted |= self
+                .channel
+                .entry(channel_id)
+                .or_default()
```
jtbandes, line 55: thank goodness this is explicit, the implicit "or default" behavior in C++ is a huge footgun 😄

rust/foxglove/src/subscription/maps.rs
```diff
@@ -0,0 +1,156 @@
...
+
+impl<K, V> From<&SubscriptionMap<K, V>> for SubscriberMap<V>
+where
+    K: Hash + Eq + Clone,
+    V: Clone,
+{
+    fn from(value: &SubscriptionMap<K, V>) -> Self {
+        let global = value.global.values().cloned().collect();
+        let mut channel = HashMap::with_capacity(value.channel.len());
+        for (&channel_id, subs) in &value.channel {
```
jtbandes, line 134: would it make sense to write this as an iterator chain that ends up `collect()`ing into a HashMap, or would that not allow the language to reserve the capacity up front? (or did you choose not to do it that way for readability reasons?)

rust/foxglove/src/subscription/maps.rs
```diff
@@ -0,0 +1,156 @@
...
+    }
+}
+
+impl<K, V> From<&SubscriptionMap<K, V>> for SubscriberMap<V>
+where
+    K: Hash + Eq + Clone,
+    V: Clone,
+{
+    fn from(value: &SubscriptionMap<K, V>) -> Self {
+        let global = value.global.values().cloned().collect();
```
jtbandes, line 132: I find it at best a brain teaser, and at worst suspicious, that this conversion discards keys by using `values()`/`into_values()` – if we didn't need them in order to use the subscribers, what was the point of having them at all? (And especially if you consider that we let the user choose the key type – if they chose some custom interesting key type, would they be surprised that we end up ignoring their keys here?)

rust/foxglove/src/subscription.rs
jtbandes, line 1: not sure what naming conventions are generally like in rust, but I think it could be less confusing to keep the file names closer to the names of the types they implement, e.g. `subscription_manager.rs`

jtbandes (approved):
A little hard to fully grasp how this is going to play out as we use it in practice, but seems probably ok to move forward and evolve this over time. None of my comments are really blocking but I will be curious to read your rebuttals 😄  Overall I felt like the doc comments could use a bit more fleshing out to make it clearer what problems the classes are solving (e.g. "tracking and querying channel subscriptions" feels a bit light on context)

rust/foxglove/src/subscription/maps.rs
gasmith: It's possible, but it wouldn't be as efficient, since we'd have to iterate over all the keys in the map.

rust/foxglove/src/subscription/maps.rs
gasmith: Sadly no. I think in order for `#[derive(Default)]` to work, all generic type parameters also need to implement `Default`. I'm not sure why that's the case.

rust/foxglove/src/subscription.rs
gasmith: It doesn't really need to be generic. I put some of the rationale for why it ended up that way in the PR description. The main reason is that I didn't know what types I wanted when I started (and in fact `SinkId` is introduced in a PR stacked on top of this one), and it ended up being sort of useful for testing. I also thought that _maybe_ I'd be able to reuse it elsewhere, but that's looking less likely now. The generics don't add any overhead since rust is monomorphized, but I agree that it hurts readability.

rust/foxglove/src/subscription.rs
gasmith: > It seems like it would be incorrect/racy to ever perform a **store** on `subscribers` when not holding the subscriptions mutex. Is there a way to encode that better in the type system?

Nothing springs to mind.

> Relatedly, I am a bit skeptical of where we may end up using `has_subscribers()` and whether (because of the lack of locking) it will be easy to accidentally introduce races by assuming that a subscriber doesn't sneak in between calls to `has_subscribers()` and other functions...

I'll add some comments to describe the intended use for `has_subscribers()`. The basic idea is to use it as a guard around channel.log, so that you don't waste time serializing messages when nobody is listening. In that situation, if `has_subscribers()` returns false, and a subscriber sneaks in on another thread, it's not a big deal. Those things weren't synchronized to begin with.

There's perhaps a bit more to think about in the other direction. Consider [the proposed changes to `Channel::log`](https://github.com/foxglove/foxglove-sdk/pull/292/files#diff-46eacad04d875e34983c4dedaa20f6ef0b22b23500e1b1f63bccb8d9e4b0e4f4R141-R146), and suppose that a sink is unregistered between the call to `get_subscribers()` and `log_to_sinks()`. We'll deliver a message to a sink that isn't expecting one.

Maybe the answer to both of these things is to just store both the subscription and subscriber maps under an RwLock and call it a day.

rust/foxglove/src/subscription/maps.rs
gasmith: Yeah! We can definitely do it that way. I actually prefer the iterator style, just missed the opportunity to use it here.

The `.collect()` implementation usually reserves capacity on the basis of [`Iterator::size_hint`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint), which for a simple hashmap iterator ought to return the number of keys.

rust/foxglove/src/subscription/maps.rs
gasmith: We're using the key to ensure that we have sets of distinct subscribers, e.g. when we're adding a new subscription, or when we're merging the global and by-channel subscriptions to construct the `SubscriberMap`.

We're throwing the key away in the `SubscriberMap` because we don't need it any more. The subscriber map is immutable once constructed, and only serves as a cache for lookups. We want our lookup values to be `SmallVec` (aliased as `SubscriberVec` here): a small vector that usually avoids heap allocations. We could store `HashMap<K, V>` in the subscriber map, but then we need to do extra work to construct the return value:

```rust
impl SubscriptionManager<K, V> {
    ...
    pub fn get_subscribers(&self, channel_id: ChannelId) -> SubscriberVec<V> {
        self.subscribers
            .load()
            .get(channel_id)
            .values()
            .cloned()
            .collect()
    }
```

As opposed to what we have now:

```rust
    pub fn get_subscribers(&self, channel_id: ChannelId) -> SubscriberVec<V> {
        self.subscribers.load().get(channel_id).clone()
    }
```

rust/foxglove/src/subscription.rs
gasmith: I got rid of the generics.

gasmith:
I've simplified this change quite a bit. I realized that we probably want to do our synchronization elsewhere (like on `Context`), and that the `ArcSwap` cache isn't actually buying us any performance wins compared to an implementation that constructs the `SubscriberVec` dynamically. I also got rid of the generics, now that `SinkId` has merged.

jtbandes (approved):

rust/foxglove/src/context/subscriptions.rs
```diff
@@ -0,0 +1,127 @@
...
+        });
+        self.global.insert(sink_id, sink).is_none()
+    }
+
+    /// Adds subscriptions to the specified channels.
+    ///
+    /// Has no effect if the subscriber has a global subscription.
+    pub fn subscribe_channels(
+        &mut self,
+        sink: Arc<dyn Sink>,
```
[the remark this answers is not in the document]
gasmith: Right, that could totally be an `&Arc`. Let's do that.

> which raises the question of why Arc isn't Copy

My guess: Even though the clone is cheap, it's not a trivial copy, because it involves incrementing a refcount. And likewise, there's a non-trivial `Drop` to decrement the refcount.

rust/foxglove/src/context/subscriptions.rs
```diff
@@ -0,0 +1,127 @@
...
+                        self.by_channel.remove(&channel_id);
+                    }
+                }
+            }
+        }
+        removed
+    }
+
+    /// Remove all global and per-channel subscriptions for a particular subscriber.
+    pub fn remove_subscriber(&mut self, sink_id: SinkId) -> bool {
```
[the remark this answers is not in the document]
gasmith: I think I mildly prefer `remove_subscriber`; I feel like `unsubscribe_all` might be confused for "remove a global subscription".

rust/foxglove/src/context/subscriptions.rs
```diff
@@ -0,0 +1,123 @@
...
+                .insert(sink_id, sink.clone())
+                .is_none();
+        }
+        inserted
+    }
+
+    /// Removes subscriptions to the specified channels.
+    ///
+    /// Has no effect if the subscriber has a global subscription.
+    pub fn unsubscribe_channels(&mut self, sink_id: SinkId, channel_ids: &[ChannelId]) -> bool {
```
eloff, line 79: Same observation about return value

eloff (approved):
Minor nits only, this looks solid

PR #283, opened 2025-03-12
Support schemaless channels over websocket

bryfox:
This adds support for live visualization of JSON channels which don't define a schema. Currently, the SDK requires a schema to be defined for a websocket sink. A user can specify an empty schema, but that's a step we shouldn't require them to take.

The ws server will behave the same for schemaless channels using message encodings other than JSON, since Foxglove doesn't support this.


rust/foxglove/src/channel_builder.rs
```diff
@@ -38,9 +38,9 @@ impl<'a> ChannelBuilder<'a> {
     }
 
     /// Set the message encoding for the channel.
-    /// This is required for Channel, but not for [`TypedChannel`].
-    /// (it's provided by the [`Encode`] trait for [`TypedChannel`].)
-    /// The [well-known message encodings](https://mcap.dev/spec/registry#well-known-message-encodings) are preferred.
+    /// This is required for Channel, but not for [`TypedChannel`] (it's provided by the [`Encode`]
+    /// trait for [`TypedChannel`].) Foxglove supports several well-known message encodings:
+    /// https://docs.foxglove.dev/docs/visualization/message-schemas/introduction
```
gasmith, line 43: The URL should be wrapped in a `<>`.

gasmith (approved):

jtbandes (commented):
I am not totally sure about requiring `json` message encoding if no schema is given. If we end up adding support for other self-describing encodings like msgpack or cbor, we would probably want to allow those to be schemaless too. On the other hand, I can see how this error would currently serve to push users into the pit of success, so it might be defensible in that way...

cc @defunctzombie @achim-k for thoughts?

bryfox:
> If we end up adding support for other self-describing encodings like msgpack or cbor, we would probably want to allow those to be schemaless too.

Maybe the better approach is to check against the known encodings today, assuming the app will show a reasonable error if a new encoding is added which doesn't support empty schemas.

python/foxglove-sdk-examples/live-visualization/main.py
```diff
@@ -108,7 +108,7 @@ def main() -> None:
     tf_chan = FrameTransformsChannel("/tf")
     point_chan = PointCloudChannel("/pointcloud")
 
-    # Log dicts using JSON encoding
+    # Log dicts using JSON encoding. You can omit a schema if you specify message_encoding="json".
```
defunctzombie, line 111: @jtbandes ping'ed me on this so I'll share my thoughts:

I would remove this comment (about omitting a schema) from the example and encourage the practice of providing a schema. If you want to introduce a schema-less example then that would be a good place to go into more of the nuances of when you can do that. Keep this example focused on one thing.

As a non-python person I am not clear on the comment "Log dicts using JSON encoding" - this line does not mention json encoding anywhere. Is that the automatic choice when no message encoding is specified?

python/foxglove-sdk/python/foxglove/channel.py
```diff
@@ -32,7 +32,7 @@ def __init__(
             encoding is presumed to be "json".
         :param schema: A definition of your schema. Pass a :py:class:`Schema`
             for full control. If a dictionary is passed, it will be treated as a
-            JSON schema.
+            JSON schema. If not provided, the message_encoding must be "json".
```
defunctzombie, line 35: This is a bit backwards in my view - the message encoding (or more specifically the channel setup) is what decides whether you need a schema to deserialize not the other way around.

bryfox:
> I am not totally sure about requiring json message encoding if no schema is given. If we end up adding support for other self-describing encodings like msgpack or cbor, we would probably want to allow those to be schemaless too.

@jtbandes I updated this to instead check against an array of our encodings which are known to require schemas; does that resolve your concern?

rust/foxglove/src/websocket/protocol/server.rs
```diff
@@ -178,30 +181,46 @@ pub fn server_info(
...
-    };
+    if schema.is_none() && REQUIRING_SCHEMA.contains(&channel.message_encoding.as_str()) {
+        return Err(FoxgloveError::SchemaRequired);
+    }
 
-    Ok(json!({
-        "op": "advertise",
-        "channels": [Advertisement{
+    let advertisement = if let Some(schema) = schema {
+        let schema_data = if super::is_known_binary_schema_encoding(&schema.encoding) {
```
jtbandes, line 195: nit: it now seems odd that `is_known_binary_schema_encoding` is a function while `REQUIRING_SCHEMA` is a const array -- feel free to change mine if you prefer the array style (I don't think I have a strong preference)

rust/foxglove/src/websocket/protocol/server.rs
```diff
@@ -178,30 +181,46 @@ pub fn server_info(
...
+            BASE64_STANDARD.encode(&schema.data)
+        } else {
+            String::from_utf8(schema.data.to_vec())
+                .map_err(|e| FoxgloveError::Unspecified(e.into()))?
+        };
+
+        Ok::<Advertisement, FoxgloveError>(Advertisement {
```
jtbandes, line 202: curious if this was necessary for some reason?
```suggestion
        Ok(Advertisement {
```

jtbandes (approved):

rust/foxglove/src/websocket/protocol/server.rs
bryfox: Thanks, I didn't notice this after merging that change from main. I don't have a strong preference either; I'll rewrite mine with a function.

rust/foxglove/src/websocket/protocol/server.rs
bryfox: Yes, the compiler can't infer the error type; saying there are multiple From conversions. I poked at this a bit but didn't see a better way to resolve it — maybe @gasmith or @eloff have an idea.

PR #285, opened 2025-03-14
Change ClientChannel schema data to binary; unify base64 handling

jtbandes:
### Changelog
None

### Docs

None

### Description

The current websocket protocol uses JSON messages for advertisements (in both directions), which requires binary schemas to be base64-encoded. Currently the SDK handles base64-encoding when sending `protobuf` schemas to the client (although we missed also doing this for `flatbuffer` schemas), but ClientAdvertisements left the base64 handling up to the user (they would just receive a String which may or may not be base64-encoded). This PR adds base64 handling for the client advertisements as well so the SDK handles both directions and doesn't require the user to mess with base64.

- Changed `ClientChannel.schema` to `Option<Vec<u8>>`
- Added `JsonClientChannel` & `JsonClientAdvertise` types for deserialization, and `ClientChannel::try_from(JsonClientChannel)` to do the base64 decoding
- Treat `flatbuffer` as base64 as well as protobuf ([docs](https://docs.foxglove.dev/docs/connecting-to-data/frameworks/custom))


eloff (approved):
This looks good. It's nicer to take care of the base64 decoding here so the user doesn't have to do that.

rust/foxglove/src/websocket/protocol/client.rs
```diff
@@ -182,21 +188,80 @@ pub(crate) struct Unsubscribe {
...
+    type Error = ParseError;
+    fn try_from(channel: JsonClientChannel) -> Result<Self, ParseError> {
+        let schema = channel
+            .schema
+            .map(|schema| {
+                if channel
+                    .schema_encoding
+                    .as_ref()
+                    .map(|enc| is_known_binary_schema_encoding(enc.as_str()))
+                    .unwrap_or(false)
```
gasmith, line 246: Nit: Could do

```suggestion
                    .is_some_and(|enc| is_known_binary_schema_encoding(enc.as_str()))
```

You could even whittle that down to `.is_some_and(is_known_binary_schema_encoding)` if you make that function generic over `AsRef<str>`.

gasmith (approved):

PR #286, opened 2025-03-14
Remove ClientChannelView

jtbandes:
### Changelog
Rust: callbacks for on_client_advertise, on_client_unadvertise, and on_message_data now receive a `&ClientChannel` instead of a `ClientChannelView`.
Python: callback for on_client_advertise now receives a `ClientChannel` instead of a `ChannelView`; on_client_unadvertise and on_message_data now receive a `client_channel_id` instead of a `ChannelView`.

### Docs

None

### Description

Motivation: all the information about the client channel should be accessible to users in the callbacks.

Thoughts?


eloff (approved):
I like the change, it's how I originally wrote it and then changed my mind in the code review. I think the worry was exposing websocket protocol specific types in the callbacks, but those callbacks are very websocket and sometimes ros specific anyway.

For Python I see you opted not to expose the additional info other than the topic and id. We'll have to remember to come back and fix that. Unfortunately it all means allocating and copying in Python, there's no way around that - because Python is able to store a reference to ClientChannel in a way that escapes the callback. Rust cannot.

I think C and C++ have the same problems as Python, not sure how we deal with that there in performance may matter land. Maybe we give them a pointer or reference to an object with borrowed data (string_view or similar for c++?) and document that they have to copy it if they want to keep it after the callback returns.

jtbandes:
Updated to expose the info to Python as well.

python/foxglove-sdk/src/websocket.rs
```diff
@@ -24,6 +24,23 @@ pub struct PyChannelView {
     topic: Py<PyString>,
 }
 
+/// Information about a client channel.
+#[pyclass(name = "ClientChannel", module = "foxglove")]
+pub struct PyClientChannel {
+    #[pyo3(get)]
```
bryfox, line 30: You could add a [`get_all` parameter](https://pyo3.rs/v0.23.5/class.html?highlight=get_all#customizing-the-class) to the `#[pyclass]` attribute above if you want to avoid all of these individual getters. (We could do it elsewhere too; this just has a lot of them.)

python/foxglove-sdk/src/websocket.rs
```diff
@@ -77,22 +94,52 @@ impl ServerListener for PyServerListener {
...
-            channel.id.into(),
-            channel.topic.as_str(),
-        );
+        let client_info = PyClient {
+            id: client.id().into(),
+        };
+
+        let py_channel = PyClientChannel {
+            id: channel.id.into(),
+            topic: channel.topic.clone(),
```
bryfox, line 103: I think this does an extra copy before going to a python string, which you could avoid with the PyString type (see `PyChannelView`), but maybe this was intentional since it's infrequently called.

bryfox (approved):
I only took a quick look at the rust source since it'd been reviewed already, but this looks good to me.

python/foxglove-sdk/src/websocket.rs
jtbandes: Definitely was not intentional since I have no idea what I'm doing when it comes to pyo3. I guess this is enough motivation to change to PyString!

PR #287, opened 2025-03-14
C/C++: client publishing support

jtbandes:
### Changelog
C/C++: added callbacks for client advertise, publish, and unadvertise

### Docs

None

### Description

~Depends on #286~

Ran into some annoyances with exposing the capability flags with cbindgen...let me know if you have any better ideas!


c/include/foxglove-c/foxglove-c.h
```diff
@@ -16,33 +16,77 @@
...
+                          size_t payload_len,
+                          const void *context);
+  void (*on_client_unadvertise)(uint32_t client_id, uint32_t client_channel_id, const void *context);
 } foxglove_server_callbacks;
 
+typedef uint8_t foxglove_server_capability;
+/**
+ * Allow clients to advertise channels to send data messages to the server.
+ */
+#define foxglove_server_capability_CLIENT_PUBLISH (1 << 0)
```
gasmith, line 60: Is there a way to get a consistent screaming snake case here? I wasn't able to find a way to do it with the cbindgen config.

One option would be to declare the constant values outside of the `impl FoxgloveServerCapability` block in the rust lib, e.g.

```rust
/// Allow clients to advertise channels to send data messages to the server.
pub const FOXGLOVE_SERVER_CAPABILITY_CLIENT_PUBLISH: u8 = 1<< 0;
```

And note that you can save yourself a little typing by just calling it a u8, since that's all it is on the C side anyway. The main downside of doing it that way, is that it doesn't stay grouped with `typedef uint8_t foxglove_server_capability`.

gasmith (approved):
I'm probably not the most qualified person to review the C++, but this change looks good to me.

c/include/foxglove-c/foxglove-c.h
```diff
@@ -16,33 +16,82 @@
...
+
 typedef struct foxglove_websocket_server foxglove_websocket_server;
 
+typedef struct foxglove_client_channel {
+  uint32_t id;
+  const char *FOXGLOVE_NONNULL topic;
+  const char *FOXGLOVE_NONNULL encoding;
+  const char *FOXGLOVE_NONNULL schema_name;
+  const char *schema_encoding;
+  const void *schema;
```
eloff, line 63: Do people use void * in C instead of byte * ?

c/src/lib.rs
```diff
@@ -2,16 +2,90 @@
...
     pub callbacks: Option<&'a FoxgloveServerCallbacks>,
+    pub capabilities: FoxgloveServerCapability,
+    pub supported_encodings: *const *const c_char,
+    pub supported_encodings_count: usize,
+}
+
+#[repr(C)]
+pub struct FoxgloveClientChannel<'a> {
+    pub id: u32,
+    pub topic: &'a c_char,
```
eloff, line 83: So weird to see these as references instead of pointers

c/src/lib.rs
```diff
@@ -222,4 +332,71 @@ impl ServerListener for FoxgloveServerCallbacks {
...
+        let topic = CString::new(channel.topic.clone()).unwrap();
+        let encoding = CString::new(channel.encoding.clone()).unwrap();
+        let schema_name = CString::new(channel.schema_name.clone()).unwrap();
+        let schema_encoding = channel
+            .schema_encoding
+            .as_ref()
+            .map(|enc| CString::new(enc.clone()).unwrap());
+        let c_channel = FoxgloveClientChannel {
+            id: channel.id.into(),
+            topic: unsafe { &*topic.as_ptr() },
```
eloff, line 353: this would be simpler if they were pointers in the struct

eloff (approved):
Looks good, left some comments that don't need to be addressed

c/include/foxglove-c/foxglove-c.h
jtbandes: As far as I know C doesn't have a byte type (only C++ has std::byte) - so yes.  As one example, `memcpy` is 
```
void *memcpy(void *restrict dst, const void *restrict src, size_t n);
```

c/src/lib.rs
jtbandes: Yeah, I guess I was hoping that it would be helpful to have the nonnull annotation on the C header side. In practice I'm not sure it has really helped that much...can remove.

PR #289, opened 2025-03-14
python: migrate to pytest in sdk

bryfox:
### Changelog

None

### Description

This adds `pytest` to the SDK dev dependencies, and rewrites the tests from unittest.

Note that you no longer need to cd into the inner python directory to run tests.

Supports FG-10513.


gasmith (approved):
:pie:

PR #290, opened 2025-03-14
rust: Add SinkId for identifying sinks

gasmith:
### Changelog
- rust: Add `SinkId` and new mandatory `Sink::id()` trait method.

### Description
We've been using Arc pointers to compare log sinks, which is… fine. But it'd be a bit nicer if we just had a SinkId to use, like we have a ChannelId. That would give us a natural way of indexing sinks, which will be useful for both sink deregistration as well as subscription management.

### DIFF
--- a/rust/foxglove/src/sink.rs
+++ b/rust/foxglove/src/sink.rs
@@ -1,13 +1,29 @@
 use crate::channel::Channel;
 use crate::metadata::Metadata;
 use crate::FoxgloveError;
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
+
+/// Uniquely identifies a [`Sink`] in the context of this program.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct SinkId(u64);
+impl SinkId {
+    /// Allocates the next sink ID.
+    pub fn next() -> Self {
+        static NEXT_ID: AtomicU64 = AtomicU64::new(0);
+        let id = NEXT_ID.fetch_add(1, Ordering::Relaxed);
+        Self(id)
+    }
+}
 
 /// A [`Sink`] writes a message from a [`Channel`] to a destination.
 ///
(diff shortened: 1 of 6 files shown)


rust/foxglove/src/sink.rs
```diff
@@ -1,13 +1,29 @@
...
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 
+/// Uniquely identifies a [`Sink`] in the context of this program.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct SinkId(u64);
+impl SinkId {
+    /// Allocates the next sink ID.
+    pub fn next() -> Self {
+        static NEXT_ID: AtomicU64 = AtomicU64::new(0);
```
bryfox, line 13: Channel IDs start at 1 for good reason. It doesn't really matter, but maybe it makes sense to be consistent here.

jtbandes (approved):

bryfox (approved):
Makes sense to me

jtbandes:
Where is the actual comparison happening (and generally who is going to be using this)?

gasmith:
> Where is the actual comparison happening (and generally who is going to be using this)?

It's primarily for internal use. Currently, `Context` maintains a `LogSinkSet` on the basis of arc pointer equality ([ref](https://github.com/foxglove/foxglove-sdk/blob/b871118a131e9675e1b8381eda1684c37e0dc1ca/rust/foxglove/src/log_sink_set.rs#L30)). I'm working to replace that structure with a more sophisticated subscription manager (https://github.com/foxglove/foxglove-sdk/pull/281), and I'd much rather be able to index sinks by an integer identifier ([ref](https://github.com/foxglove/foxglove-sdk/pull/292/files#diff-f464984082a2f6bb88d548e8059b6e4dd78234875fc273a057f223f739d2784cR17-R18)).

eloff (approved):
This seems fine, it should make it easier to work with hashmaps

PR #292, opened 2025-03-14
rust: Use subscription manager

gasmith:
### Changelog
- Skip message serialization for a `TypedChannel` that has no subscribed sinks.
- rust: Add `Channel::has_sinks` and `TypedChannel::has_sinks`.
- rust: Hide `Sink` trait for now.

### Description
This change enhances the old channel->sinks linkage with the new subscription manager.

Structural changes:
- `Context` is now an `RwLock` wrapper around an inner struct.
  - Most updates impact multiple resources.
  - Difficult to reason about sequencing and raciness.
  - The lock should be mostly uncontended.
- `ContextInner`:
  - Holds a `Subscriptions` instance, which it uses to update channel sinks.
  - Indexes sinks by `SinkId`, instead of by `Arc` pointer.
  - Indexes channels by `ChannelId` in addition to topic, to support (un)subscribe-by-id.
- `Channel` no longer implements `Drop`; its lifecycle is managed by the `Context`.
- `LogSinkSet` is now an `ArcSwap` instead of `RwLock`, since we always replace the entire vec.
- `Sink::auto_subscribe()` method indicates whether a sink subscribes to all channels, default true.
- `Sink::add_channel()` now returns `bool`, for a dynamic subscriber to add a subscription synchronously.
- `TypedChannel` queries for subscribers before serialization.

For now, the `websocket::Server` sink still auto-subscribes to all channels. The next PR in this series will switch to using dynamic registration.

Part of: FG-10743

### DIFF
--- a/rust/foxglove/src/channel.rs
+++ b/rust/foxglove/src/channel.rs
@@ -128,6 +128,11 @@ impl Channel {
         self.sinks.store(sinks);
     }
 
+    /// Clears the set of subscribed sinks.
+    pub(crate) fn clear_sinks(&self) {
+        self.sinks.clear();
+    }
+
     /// Returns true if at least one sink is subscribed to this channel.
     pub fn has_sinks(&self) -> bool {
         !self.sinks.is_empty()
--- a/rust/foxglove/src/channel.rs
+++ b/rust/foxglove/src/channel.rs
@@ -1,11 +1,13 @@
-use crate::log_sink_set::LogSinkSet;
-use crate::{nanoseconds_since_epoch, Metadata, PartialMetadata, Sink};
-use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 use std::collections::BTreeMap;
 use std::sync::atomic::AtomicU32;
 use std::sync::atomic::Ordering::Relaxed;
-use std::sync::Arc;
+
+use serde::{Deserialize, Serialize};
+
+use crate::log_sink_set::LogSinkSet;
+use crate::sink::SmallSinkVec;
+use crate::{nanoseconds_since_epoch, Metadata, PartialMetadata};
 
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Debug, Deserialize, Serialize)]
 pub struct ChannelId(u64);
@@ -122,7 +124,7 @@ impl Channel {
     }
 
     /// Updates the set of sinks that are subscribed to this channel.
-    pub(crate) fn update_sinks(&self, sinks: Vec<Arc<dyn Sink>>) {
+    pub(crate) fn update_sinks(&self, sinks: SmallSinkVec) {
         self.sinks.store(sinks);
     }
 
--- a/rust/foxglove/src/channel.rs
+++ b/rust/foxglove/src/channel.rs
@@ -1,10 +1,11 @@
 use crate::log_sink_set::LogSinkSet;
-use crate::{nanoseconds_since_epoch, Metadata, PartialMetadata};
+use crate::{nanoseconds_since_epoch, Metadata, PartialMetadata, Sink};
 use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 use std::collections::BTreeMap;
 use std::sync::atomic::AtomicU32;
 use std::sync::atomic::Ordering::Relaxed;
+use std::sync::Arc;
 
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Debug, Deserialize, Serialize)]
 pub struct ChannelId(u64);
@@ -88,7 +89,6 @@ impl Schema {
 /// ```
 pub struct Channel {
     // TODO add public read-only accessors for these for the Rust API.
-    // TODO add a list of contexts here as well (or restrict to one context per channel?)
     pub(crate) sinks: LogSinkSet,
     /// id is a unique identifier for the channel inside the process,
     /// it's used to map a channel to a channel_id or subscription_id in log sinks.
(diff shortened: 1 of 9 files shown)


rust/foxglove/src/channel.rs
```diff
@@ -88,7 +89,6 @@ impl Schema {
 /// ```
 pub struct Channel {
     // TODO add public read-only accessors for these for the Rust API.
```
bryfox, line 91: Can we remove the above TODO as well? I'm not sure if it's stale, but if we still want it, let's file an issue instead.

rust/foxglove/src/context.rs
```diff
@@ -5,36 +5,159 @@ use std::sync::{Arc, LazyLock};
...
+        let topic = &channel.topic;
+        let Entry::Vacant(entry) = self.channels_by_topic.entry(topic.clone()) else {
+            return Err(FoxgloveError::DuplicateChannel(topic.clone()));
+        };
+        entry.insert(channel.clone());
+        self.channels.insert(channel.id(), channel.clone());
+
+        // Notify sinks of new channel.
+        for sink in self.sinks.values() {
+            sink.add_channel(&channel);
```
bryfox, line 39: Did the sink use to be responsible for more bookkeeping? "add_channel" now seems like it might be better named as "on_add_channel" or something like that?

rust/foxglove/src/sink.rs
```diff
@@ -43,4 +43,16 @@ pub trait Sink: Send + Sync {
...
+
+    /// Indicates whether this sink automatically subscribes to all channels.
+    ///
+    /// The default implementation returns true.
+    ///
+    /// A sink implementation may return false to indicate that it intends to manage its
+    /// subscriptions dynamically using
+    /// [`Context::subscribe_channels`][crate::Context::subscribe_channels] and
+    /// [`Context::unsubscribe_channels`][crate::Context::unsubscribe_channels].
+    fn auto_subscribe(&self) -> bool {
```
bryfox, line 55: I see this becomes useful in #293.

As a user of the SDK, under what circumstances might I want to use this? i.e., would it be useful to document this a bit more, or hide it from the public API if possible?

rust/foxglove/src/context.rs
```diff
@@ -5,36 +5,159 @@ use std::sync::{Arc, LazyLock};
...
+        }
+
+        true
+    }
+
+    /// Removes a sink from the context.
+    fn remove_sink(&mut self, sink_id: SinkId) -> bool {
+        // Remove sink's subscriptions. If this wasn't a no-op, update channel sinks.
+        if self.subs.remove_subscriber(sink_id) {
+            self.update_all_channel_sinks();
```
bryfox, line 95: I think I'm still wrapping my head around the full picture here. Why is this called unconditionally, but in `add_sink` the update is done only if `auto_subscribe` is true?

rust/foxglove/src/channel.rs
gasmith: Filed FG-10817

rust/foxglove/src/context.rs
```diff
@@ -5,36 +5,159 @@ use std::sync::{Arc, LazyLock};
...
+        if self.subs.unsubscribe_channels(sink_id, channel_ids) {
+            self.update_channel_sinks(channel_ids);
+        }
+    }
+
+    /// Updates the set of connected sinks on all channels.
+    fn update_all_channel_sinks(&self) {
+        for (id, channel) in &self.channels {
+            let sinks = self.subs.get_subscribers(*id);
+            channel.update_sinks(sinks.to_vec());
```
eloff, line 121: I thought get_subscribers returned a smallvec? Why do we need to_vec here?

rust/foxglove/src/context.rs
gasmith: I don't think so. This is just the trait method. I agree that `on_add_channel` sounds a little better. Filed FG-10818.

rust/foxglove/src/context.rs
gasmith: At the point we're removing a sink, we don't really care whether the sink auto-subscribes or not. We just want to remove _all_ of the sink's subscriptions. The `Subscriptions::remove_subscriber` method does that, by removing the sink from all of its internal maps.

Then, if at least one subscription was found, we need to update the corresponding channels. We could do fine-grained tracking to figure out which specific channels to update, but I'd prefer to postpone that optimization until we have good reason to make it.

eloff (approved):
Looks good and it makes sense to me.

We have (had?) a client ID and a sink id for connected clients. I think we could just collapse that into one id now if you haven't already done that.

rust/foxglove/src/sink.rs
gasmith: I don't think it deserves to be any more hidden than the rest of the `Sink` trait. If a user is implementing a `Sink`, they need to decide how they manage their subscriptions to channels: do they want everything, or do they want to be selective? I thought the docstring conveyed this pretty clearly, but I'm open to suggestions for improved wording.

That said, I don't think too many users will be implementing their own `Sink`s. That seems like a fairly niche case. We could hide the whole trait, along with `Context::add_sink` and `Context::remove_sink`, since users will attach our provided sink implementations to contexts through their respective builders.

rust/foxglove/src/context.rs
gasmith: Eh, because `LogSinkSet` still stores a `Vec` internally.

I added the smallvec back when I thought channels were going to be querying `Subscriptions` directly at `log` time to get a list of sinks. Now that we're using `LogSinkSet::for_each`, that optimization seemed less interesting because we're not returning a set of sinks on the hot path. I suppose it's still worthwhile for cache locality.

gasmith: Updated the `LogSinkSet` to store smallvecs.

rust/foxglove/src/sink.rs
```diff
@@ -32,19 +33,54 @@ pub trait Sink: Send + Sync {
...
+    /// time.
+    ///
+    /// For sinks that [auto-subscribe][Sink::auto_subscribe] to all channels, the return value of
+    /// this method is ignored.
+    ///
+    /// [ctx]: crate::Context
+    /// [sub]: crate::Context::subscribe_channels
+    fn add_channel(&self, _channel: &Arc<Channel>) -> bool {
+        false
+    }
```
gasmith, line 57: Because we're holding the `Context` lock during these callbacks, there's no way for a `Sink` implementation to synchronously subscribe to a channel during the `add_channel` callback. That feels like a major problem (though not for the websocket server, whose clients register subscriptions asynchronously).

I think having this callback return a boolean is a reasonable workaround. It's still a gnarly footgun, since attempting to call back into `Context` methods will simply _deadlock_, but on the other hand we're not really expecting too many folks to implement this interface themselves.

I was also thinking about how it would be nice to make these vectored APIs (e.g., it'd be much nicer to have a single batched channel advertisement when setting up a websocket client, rather than sending an advertisement per channel). I suppose if we go down that road, then this method would return `Option<Vec<ChannelId>>` to indicate the channels it wants to subscribe to?

bryfox: Above you mentioned hiding the Sink trait, which seems reasonable to me — keep the footgun internal, at least.

Maybe name this `should_add_channel` or something that implies side effects are unwanted?

> single batched channel advertisement

That's a good idea, and maybe another reason to hide the trait for now.

gasmith: OK, I'm convinced. I'm hiding it.

PR #293, opened 2025-03-14
rust: Manage websocket channel subscriptions dynamically

gasmith:
### Changelog
- Skip message serialization for a `TypedChannel` that has no subscribed websocket clients.

### Description
This change makes use of the new subscription manager, to teach the websocket to manage channel subscriptions dynamically based on client demand. The websocket no longer automatically subscribes to all channels. Instead, it only subscribes to channels that have one or more subscribed clients.

To make that happen, we move the implementation of `Sink` from `Server` to `ConnectedClient`. This is a bit simpler than the alternative, which would involve maintaining an aggregation layer for propagating client subscriptions back to the context. Anyway, it feels a bit more natural to treat the clients as sinks.

### DIFF
--- a/rust/foxglove/src/websocket.rs
+++ b/rust/foxglove/src/websocket.rs
@@ -1,18 +1,5 @@
 //! Websocket functionality
 
-use crate::channel::ChannelId;
-use crate::cow_vec::CowVec;
-pub use crate::websocket::protocol::client::{ClientChannel, ClientChannelId};
-pub(crate) use crate::websocket::protocol::client::{ClientMessage, Subscription, SubscriptionId};
-pub use crate::websocket::protocol::server::{
-    Parameter, ParameterType, ParameterValue, Status, StatusLevel,
-};
-use crate::{get_runtime_handle, Channel, FoxgloveError, Metadata, Sink, SinkId};
-use bimap::BiHashMap;
-use bytes::{BufMut, Bytes, BytesMut};
-use flume::TrySendError;
-use futures_util::{stream::SplitSink, SinkExt, StreamExt};
-use serde::Serialize;
 use std::collections::hash_map::Entry;
 use std::collections::HashSet;
 use std::sync::atomic::Ordering::{AcqRel, Acquire, Release};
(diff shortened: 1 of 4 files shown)


rust/foxglove/src/websocket.rs
```diff
@@ -972,6 +983,50 @@ impl ConnectedClient {
         let message = Message::binary(buf);
         self.send_control_msg(message);
     }
+
+    /// Advertises a channel to the client.
+    fn advertise_channel(&self, channel: &Arc<Channel>) {
```
bryfox, line 988: Do you think it's time for ConnectedClient to get its own module/file? This doesn't bother me with IDE support, but maybe would make the context more obvious when reading.

rust/foxglove/src/websocket.rs
```diff
@@ -972,6 +983,50 @@ impl ConnectedClient {
...
+
+        let message = match protocol::server::advertisement(channel) {
+            Ok(message) => message,
+            Err(err) => {
+                tracing::error!("Error creating advertise channel message to client: {err}");
+                return;
+            }
+        };
+
+        self.channels.write().insert(channel.id(), channel.clone());
```
bryfox, line 1005: This change (not storing until we have a successful advertisement) makes sense because the client is maintaining its own channels now. And that's OK because the context is also managing the channels at a higher level on behalf of the server, correct?

Note this code has changed a bit on main with the schemaless handling.

bryfox (approved):
Makes sense to me. I assume you've been running the examples to sanity check everything.

rust/foxglove/src/websocket.rs
eloff: +1 I've considering this for a while, the file is getting unwieldy for me at this length

rust/foxglove/src/websocket.rs
```diff
@@ -1354,59 +1364,31 @@ impl Server {
             }
         }
 
+        // Remove the client sink.
+        if let Some(context) = self.context.upgrade() {
```
eloff, line 1368: Maybe move this to on_disconnect with the other cleanup logic?

eloff (approved):
LGTM, only one minor suggestion

rust/foxglove/src/websocket.rs
gasmith: Right. I think it's correct to ignore the channel here if we're unable to construct an advertisement. Because if there's no way to advertise the channel, then from the client's point of view, the channel might as well not exist.

rust/foxglove/src/websocket.rs
gasmith: Mm, I guess I was trying to maintain parallelism with the call to `Context::add_sink` in `Server::register_client_and_advertise`. I'm inclined to leave it here for now, but I suspect that a bunch of this will get reorganized a bit when we move the client into its own module.

PR #294, opened 2025-03-14
Check for doc warnings

bryfox:
### Changelog
None
### Docs
None
### Description

This adds a CI step generate rust docs and fail if any warnings are encountered; also fixes a couple of warnings. The capability here was only a broken link, but this should guard against invalid identifiers as well.


jtbandes (approved):
🎉

PR #295, opened 2025-03-14
C/C++ Add mcap writer

eloff:
Implements basic mcap writing support, including plumbing the mcap library WriteOptions through to C and C++

This PR makes it possible to open, close, and free an mcap sink.


c/src/lib.rs
```diff
@@ -77,6 +80,119 @@ pub unsafe extern "C" fn foxglove_server_start(
...
+            .emit_attachment_indexes(self.emit_attachment_indexes)
+            .emit_metadata_indexes(self.emit_metadata_indexes)
+            .repeat_channels(self.repeat_channels)
+            .repeat_schemas(self.repeat_schemas)
+    }
+}
+
+pub struct FoxgloveMcapWriter(Option<foxglove::McapWriterHandle<BufWriter<File>>>);
+
+/// Create or open an MCAP file for writing. Must later be freed with `foxglove_mcap_close`.
```
jtbandes, line 137:
```suggestion
/// Create or open an MCAP file for writing. Must later be freed with `foxglove_mcap_free`.
```

cpp/foxglove/include/foxglove/mcap.hpp
```diff
@@ -0,0 +1,38 @@
...
+struct foxglove_mcap_writer;
+
+namespace foxglove {
+
+struct McapWriterOptions {
+  std::string_view path;
+  std::string_view profile;
+  std::string_view library;
+  uint64_t chunkSize;
+  bool useChunks;
```
jtbandes, line 15: I would recommend adding inline defaults for all the primitive types, iirc this helps avoid situations where we might accidentally read from uninitialized memory (plus it helps users to set sane default values if they don't care to change them)
```suggestion
  uint64_t chunkSize = 1024*768;
  bool useChunks = true;
```

cpp/foxglove/tests/test_mcap.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+  // Check if test.mcap file exists
+  REQUIRE(std::filesystem::exists("test.mcap"));
+
+  // Check that it contains the profile and library
+  std::ifstream file("test.mcap", std::ios::binary);
+  REQUIRE(file.is_open());
+  std::string content((std::istreambuf_iterator<char>(file)), std::istreambuf_iterator<char>());
+  REQUIRE(content.find("test_profile") != std::string::npos);
+  REQUIRE(content.find("test_library") != std::string::npos);
```
jtbandes, line 85: FYI, the catch2 library has some helpful matchers such as `ContainsSubstring`: https://github.com/catchorg/Catch2/blob/devel/docs/matchers.md#using-matchers

jtbandes (approved):

c/include/foxglove-c/foxglove-c.h
```diff
@@ -45,6 +47,28 @@ typedef struct foxglove_server_options {
   const struct foxglove_server_callbacks *callbacks;
 } foxglove_server_options;
 
+typedef struct foxglove_mcap_options {
+  const char *path;
+  size_t path_len;
```
eloff, line 52: @jtbandes What do you think about passing string data + size as two fields through the C API?

It's a little less natural if you have null-terminated strings (C) as you'd have to call strlen, but every other language using strings with a length. Even modern C usage might do that as well if they use a string library. This lets us use string_view for the C++ structs.

jtbandes: I don't think I have a strong preference since I rarely work with C APIs. I can't say I have seen it much before, but it does have a concrete benefit. I guess we should also then change the other ones I've added to be consistent with this (could be a separate PR)

eloff: Yeah, I think let's do that. You can create a ticket for it, and whoever gets to it first can do it

cpp/CMakeLists.txt
```diff
@@ -124,3 +124,10 @@ add_executable(example_server examples/ws-server/src/main.cpp)
 target_compile_options(example_server PUBLIC ${SANITIZER_COMPILE_OPTIONS})
 target_link_options(example_server PUBLIC ${SANITIZER_LINK_OPTIONS})
 target_link_libraries(example_server foxglove_cpp_static)
+
+add_executable(example_mcap examples/mcap/src/main.cpp)
+target_compile_options(example_mcap PUBLIC ${SANITIZER_COMPILE_OPTIONS})
+target_link_options(example_mcap PUBLIC ${SANITIZER_LINK_OPTIONS})
+set_property(TARGET example_mcap PROPERTY CXX_STANDARD 17)
+set_property(TARGET example_mcap PROPERTY CXX_STANDARD_REQUIRED True)
```
jtbandes, line 132: You can remove these two lines, I recently replaced them with a global cmake variable to fix some issues building the tests
```suggestion
```

cpp/examples/mcap/src/main.cpp
```diff
@@ -0,0 +1,31 @@
...
+  schema.data = reinterpret_cast<const std::byte*>(schemaData.data());
+  schema.dataLen = schemaData.size();
+  foxglove::Channel channel{"example", "json", std::move(schema)};
+
+  for (int i = 0; i < 100; ++i) {
+    std::string msg = "{\"val\": " + std::to_string(i) + "}";
+    channel.log(reinterpret_cast<const std::byte*>(msg.data()), msg.size());
+  }
+
+  // Optional, if you want to check for or handle errors
```
jtbandes, line 29: nit: this doesn't actually let you check for errors yet 😅   soon we will figure out error handling....

cpp/foxglove/include/foxglove/mcap.hpp
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+#include <memory>
+#include <string>
+
+#include "foxglove-c/foxglove-c.h"
```
jtbandes, line 6:
```suggestion
#include <foxglove-c/foxglove-c.h>
```

I'm curious for your opinion on this - I've been trying to avoid including this header from the C++ headers, because then it will pollute the global namespace with all of the C symbols and possibly lead to confusion if someone is trying to use the C++ library and not touch the C stuff. On the other hand, it is quite nice to be able to reuse things like enums. I guess we can always refine later, just something to think about...

yesterday I did see a post from someone suggesting that you could do something like `namespace detail { #include <foxglove-c/foxglove-c.h> }` to keep the C defs out of the global namespace ... but I'm not sure how kosher that is

cpp/foxglove/tests/test_mcap.cpp
```diff
@@ -0,0 +1,147 @@
...
+  FileCleanup cleanup("test.mcap");
+
+  foxglove::McapWriterOptions options = {};
+  options.path = "test.mcap";
+  options.create = true;
+  foxglove::McapWriter writer(options);
+  writer.close();
+
+  // Check if test.mcap file exists
+  REQUIRE(std::filesystem::exists("test.mcap"));
```
jtbandes, line 36: at some point we may want an e2e test that actually reads the file contents – for now, maybe it would be good to assert what the file size is? (just to prove that something was actually written)  But on the other hand that would lead to test failures if we ever change some details of the writer, like the default library string, so feel free to say no :)

cpp/foxglove/tests/test_mcap.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+  writer.close();
+
+  // Check if test.mcap file exists
+  REQUIRE(std::filesystem::exists("test.mcap"));
+
+  // Check that it contains the profile and library
+  std::ifstream file("test.mcap", std::ios::binary);
+  REQUIRE(file.is_open());
+  std::string content((std::istreambuf_iterator<char>(file)), std::istreambuf_iterator<char>());
```
jtbandes, line 85: nit: could make a helper function that reads the whole contents of a file path into a std::string, so we don't have to repeat this in each test

rust/examples-unstable/ws-stream-mcap/Cargo.toml
```diff
@@ -13,5 +13,5 @@ clap = { version = "4.5", features = ["derive"] }
 ctrlc = "3.4"
 env_logger = "0.11"
 foxglove = { path = "../../foxglove", features = ["unstable"] }
-mcap = "0.14"
+mcap = "0.14.1"
```
jtbandes, line 16: why is this one not 0.15 as well?

rust/foxglove/src/mcap_writer.rs
```diff
@@ -43,6 +43,7 @@ impl McapWriter {
     }
 
     /// Instantiates a new MCAP writer with the provided options.
+    /// The library option is ignored. chunk_size=0 is treated
```
jtbandes, line 46: incomplete comment

Also, is ignoring the library option a bug? Should we not expose an option if it is ignored?

jtbandes (approved):
🎉

jtbandes:
PR description could use an update as well

cpp/examples/mcap/src/main.cpp
eloff: Yeah, we'll have to think how we want to do that

cpp/foxglove/include/foxglove/mcap.hpp
eloff: The namespace trick seems like it works, we can go with that for now

cpp/foxglove/tests/test_mcap.cpp
eloff: maybe that the file size is bigger than X would be less fragile

I tried adding our mcap c++ library as a test dependency, but that road quickly became painful. At some point we should probably do it though. Cargo is so, so nice.

cpp/foxglove/tests/test_mcap.cpp
eloff: It's easier to hit tab. Cursor is ruining me.

rust/foxglove/src/mcap_writer.rs
eloff: It's not our type, it's from the mcap library. That's an argument maybe for duplicating the type and providing a conversion instead.

cpp/foxglove/include/foxglove/mcap.hpp
```diff
@@ -0,0 +1,41 @@
...
+struct foxglove_mcap_writer;
+
+namespace foxglove {
+
+struct McapWriterOptions {
+  std::string_view path;
+  std::string_view profile;
+  uint64_t chunkSize = 0;
+  FoxgloveMcapCompression compression = FoxgloveMcapCompression_None;
+  bool useChunks = false;
```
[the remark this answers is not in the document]
eloff: Yeah, you're right. I copied the defaults from our rust mcap library instead

c/src/lib.rs
```diff
@@ -187,6 +190,132 @@ pub unsafe extern "C" fn foxglove_server_start(
...
+        let profile = std::str::from_utf8(unsafe {
+            std::slice::from_raw_parts(self.profile as *const u8, self.profile_len)
+        })
+        .expect("profile is invalid");
+
+        let compression = match self.compression {
+            FoxgloveMcapCompression::None => None,
+            FoxgloveMcapCompression::Zstd => Some(Compression::Zstd),
+            FoxgloveMcapCompression::Lz4 => Some(Compression::Lz4),
+        };
```
jtbandes, line 235: I wonder what happens if someone on the C side puts an incorrect/unsupported value in the enum? Seems like rust isn't forcing us to have an "other" case ...

jtbandes (approved):

c/src/lib.rs
eloff: Bad things probably, I should add a default case

PR #298, opened 2025-03-19
rust: Remove arbitrary sleeps from tests

gasmith:
### Description
Now that we have websocket clients represented as individual sinks, we can remove a couple arbitrary 100ms sleeps from our tests, and instead periodically poll the number of subscribed sinks on a channel.

While we're at it, we might as well get rid of some other arbitrary sleeps as well, by polling on `RecordingServerListener` state, or simply waiting to receive messages on the websocket client.

I ran tests in a loop on my machine for ten minutes and didn't see any failures.

### DIFF
--- a/rust/foxglove/src/testutil.rs
+++ b/rust/foxglove/src/testutil.rs
@@ -240,8 +240,5 @@ pub async fn assert_eventually(cond: impl Fn() -> bool) {
         }
     })
     .await;
-    assert!(
-        result.is_ok(),
-        "condition not met within the specified timeout"
-    );
+    assert!(result.is_ok(), "condition not met within {timeout:?}");
 }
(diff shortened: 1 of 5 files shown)


rust/foxglove/src/testutil.rs
```diff
@@ -199,3 +222,26 @@ impl ServerListener for RecordingServerListener {
...
+
+/// Asserts that `cond` returns true within 50ms, polling every 1ms.
+///
+/// When using this, you probably want to wrap the things you're testing in `dbg!()` macros so that
+/// you get helpful debug logs when a test fails:
+///
+/// ```
+/// assert_eventually(|| dbg!(x.len()) == 2);
+/// ```
+pub async fn assert_eventually(cond: impl Fn() -> bool) {
```
bryfox, line 234: 👍

bryfox (approved):

PR #302, opened 2025-03-20
C/C++: address review comments from client publishing

jtbandes:
### Changelog
None

### Docs

None

### Description

I meant to include this in #287, but forgot to actually push the last commit before merging...


eloff (approved):
Looks like an improvement to me

PR #303, opened 2025-03-20
Python: improve high-level log API 

eloff:
Improves the Python `log(topic, message)` function:

Add more restrictive typings for the message, add a common base class for all foxglove schema types (to better support the typings), add support for dict, list, bytes, and str messages.

`dict | list` are serialized to json in a schemaless channel with message_encoding='json' 

`bytes | str` are serialized as-is, with str encoded to utf8 first

### DIFF
--- a/python/foxglove-sdk-examples/live-visualization/main.py
+++ b/python/foxglove-sdk-examples/live-visualization/main.py
@@ -9,11 +9,6 @@ from math import cos, sin
 import foxglove
 import numpy as np
 from foxglove import Channel, Schema
-from foxglove.channels import (
-    FrameTransformsChannel,
-    PointCloudChannel,
-    SceneUpdateChannel,
-)
 from foxglove.schemas import (
     Color,
     CubePrimitive,
@@ -134,11 +129,6 @@ def main() -> None:
         supported_encodings=["json"],
     )
 
-    # Log messages having well-known Foxglove schemas using the appropriate channel type.
-    box_chan = SceneUpdateChannel("/boxes")
-    tf_chan = FrameTransformsChannel("/tf")
-    point_chan = PointCloudChannel("/pointcloud")
-
     # Log messages with a custom schema and any encoding.
     sin_chan = Channel(
         topic="/sine",
@@ -171,7 +161,8 @@ def main() -> None:
 
             json_chan.log(json_msg)
 
-            tf_chan.log(
+            foxglove.log(
+                "/tf",
                 FrameTransforms(
                     transforms=[
                         FrameTransform(
@@ -190,7 +181,8 @@ def main() -> None:
                 )
             )
 
-            box_chan.log(
+            foxglove.log(
+                "/boxes",
                 SceneUpdate(
                     entities=[
                         SceneEntity(
@@ -215,7 +207,10 @@ def main() -> None:
                 )
             )
 
-            point_chan.log(make_point_cloud())
+            foxglove.log(
+                "/pointcloud",
+                make_point_cloud(),
+            )
 
             # Or use high-level log API without needing to manage explicit Channels.
             foxglove.log(
--- a/python/foxglove-sdk/python/foxglove/channel.py
+++ b/python/foxglove-sdk/python/foxglove/channel.py
@@ -113,14 +113,15 @@ def log(topic: str, message: JsonMessage | list[Any] | bytes | str | schemas.Fox
         elif isinstance(message, (dict, list)):
             channel = Channel(topic, message_encoding="json")
         else:
+            assert isinstance(message, schemas.FoxgloveSchema), "unsupported message type"
             channel_name = f"{schema_name}Channel"
             channel_cls = getattr(channels, channel_name, None)
             if channel_cls is not None:
                 channel = channel_cls(topic)
-            if channel is None:
-                raise ValueError(
-                    f"No Foxglove schema channel found for message type {type(message).__name__}"
-                )
+        if channel is None:
+            raise ValueError(
+                f"No Foxglove schema channel found for message type {schema_name}"
+            )
         _channels_by_topic[topic] = channel
 
     channel.log(message)
--- a/python/foxglove-sdk/python/foxglove/channel.py
+++ b/python/foxglove-sdk/python/foxglove/channel.py
@@ -1,7 +1,7 @@
 import json
 from typing import Any, Dict, Optional, Union
 
-from ._foxglove_py import BaseChannel, Schema, channels
+from ._foxglove_py import BaseChannel, Schema, channels, schemas
 
 JsonSchema = Dict[str, Any]
 JsonMessage = Dict[str, Any]
@@ -53,7 +53,7 @@ class Channel:
 
     def log(
         self,
-        msg: Union[JsonMessage, bytes],
+        msg: Union[JsonMessage, list[Any], bytes, str],
         log_time: Optional[int] = None,
         publish_time: Optional[int] = None,
         sequence: Optional[int] = None,
@@ -62,17 +62,19 @@ class Channel:
         Log a message on the channel.
 
         :param msg: the message to log. If the channel uses JSON encoding, you may pass a
-            dictionary. Otherwise, you are responsible for serializing the message.
+            dictionary or list. Otherwise, you are responsible for serializing the message.
         """
-        if isinstance(msg, bytes):
-            return self.base.log(msg, log_time, publish_time, sequence)
-
-        if self.message_encoding == "json":
+        if self.message_encoding == "json" and isinstance(msg, (dict, list)):
             return self.base.log(
                 json.dumps(msg).encode("utf-8"), log_time, publish_time, sequence
             )
 
-        raise ValueError(f"Unsupported message type: {type(msg)}")
+        if isinstance(msg, str):
+            msg = msg.encode("utf-8")
+        if isinstance(msg, bytes):
+            return self.base.log(msg, log_time, publish_time, sequence)
+
+        raise TypeError(f"Unsupported message type: {type(msg)}")
 
     def close(self) -> None:
         """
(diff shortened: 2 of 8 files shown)


typescript/schemas/src/internal/generatePyclass.test.ts
```diff
@@ -63,7 +66,7 @@ describe("generatePyclass", () => {
         /// :param field_nested_array: A nested array field
         ///
         /// See https://docs.foxglove.dev/docs/visualization/message-schemas/example-message
-        #[pyclass(module = "foxglove.schemas")]
+        #[pyclass(module = "foxglove.schemas", extends = FoxgloveSchema)]
```
bryfox: pyo3 doesn't (yet) support extending a custom type

rust/foxglove-proto-gen/src/lib.rs
```diff
@@ -205,7 +205,14 @@ pub fn generate_protos(proto_path: &Path, out_dir: &Path) -> anyhow::Result<()>
     config.bytes(["."]);
 
     let mut fds = config
-        .load_fds(&proto_files, &[proto_path])
+        .load_fds(
+            &proto_files,
+            &[
+                proto_path,
+                PathBuf::from("/usr/include"), // Standard location on Linux
+                PathBuf::from("/usr/local/include"), // Standard location on macOS and some Linux distros)
```
bryfox, line 213: What other protos are we needing to include?

bryfox (commented):
I like the more flexible logging for dicts, etc.

I'd like to see the typed channel live on in another example.

Not sure what to do about the base class — maybe a union for now?

rust/foxglove-proto-gen/src/lib.rs
eloff: It failed on the google Timestamp proto, which is in /usr/include for me. Adding this fixed the issue.

I still have issues with flatbuffer.

It's pretty painful that we need protoc and flatc installed to run make in the python dir.

typescript/schemas/src/internal/generatePyclass.test.ts
bryfox: ah, thanks — I must have been thinking of extending classes defined in python, which I ran into with ServerListener

rust/foxglove-proto-gen/src/lib.rs
bryfox: 🤔 I thought that we configured Prost to use our custom types when it encounters those definitions:

https://github.com/foxglove/foxglove-sdk/blob/8ae24c4f32ab01fb23d710e3eca6e11ec1ba0a24/rust/foxglove-proto-gen/src/lib.rs#L202-L203

> I still have issues with flatbuffer.
> It's pretty painful that we need protoc and flatc installed to run make in the python dir

but that's not for SDK; I guess I'm not clear how this relates to foxglove-proto-gen

eloff: Do you get this error when you run `yarn generate`?

Adding the include paths to foxglove-proto-gen fixed it for me.

```
thread 'main' panicked at rust/foxglove-proto-gen/src/bin.rs:12:65:
Failed to generate protos: Failed to load protos

Caused by:
    protoc failed: google/protobuf/timestamp.proto: File not found.
    foxglove/CircleAnnotation.proto:7:1: Import "google/protobuf/timestamp.proto" was not found or had errors.
    foxglove/CircleAnnotation.proto:14:3: "google.protobuf.Timestamp" is not defined.
    ```

File: typescript/schemas/src/internal/generatePyclass.test.ts (hunk 1 above)
Comment (eloff) [reply typescript/schemas/src/internal/generatePyclass.test.ts]: Yeah that doesn't work sadly

File: python/foxglove-sdk/python/foxglove/channel.py [hunk 3]
```diff
@@ -87,24 +90,45 @@ def close(self) -> None:
...
+    Creates a new channel the first time called for a given topic.
+    For foxglove types in the schemas module, this creates a typed channel.
+    For bytes and str, this creates a simple schemaless channel and logs the bytes as-is.
+    For dict and list, this creates a schemaless json channel.
+
+    The type of the message must be kept consistent for each topic or an error will be raised.
+    This can be avoided by creating and using the channels directly instead of using this function.
+
+    Note: currently this always creates a new channel for the given topic on the first call,
+    even if a channel already exists, which will raise an error.
```
gasmith, line 107: I didn't quite understand this the first time I read it. Maybe "even if a channel with the same topic was created by some other means".

It's sort of a bummer that we don't support creating a custom channel (e.g., a json or binary channel _with_ a schema attached), and then using the high-level `foxglove.log` to log to it later. Perhaps we could put every channel we create into `_channels_by_topic` - regardless of how the channel is created.

rust/foxglove-proto-gen/src/lib.rs
gasmith: I don't have this problem either. How did you install protoc & flatc?

python/foxglove-sdk/python/foxglove/channel.py
```diff
@@ -87,24 +90,45 @@ def close(self) -> None:
 _channels_by_topic: Dict[str, Channel] = {}
 
 
-def log(topic: str, message: Any) -> None:
-    channel: Optional[Channel] = _channels_by_topic.get(topic, None)
+def log(
+    topic: str,
+    message: Union[JsonMessage, list[Any], bytes, str, schemas.FoxgloveSchema],
```
gasmith, line 95: Probably an area for future improvement, but it might be nice if we offered a base class for messages, sort of like `Encode` in rust, so that folks could create their own.

gasmith (commented):
Looking good!

python/Makefile
```diff
@@ -37,6 +37,8 @@ generate-protobuf:
 		--python_out=foxglove_schemas_protobuf \
 		--mypy_out=foxglove_schemas_protobuf \
 		--proto_path ../../schemas/proto \
+		--proto_path /usr/include \
+		--proto_path /usr/local/include \
```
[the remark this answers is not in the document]
eloff: I feel like it used to work without this on my machine too. I don't know why it's an issue all of a sudden

python/foxglove-sdk/python/foxglove/channel.py
eloff: Yeah something like that seems like it would make sense. It's kind of a secondary shadow global context, but I don't really see how to avoid it for Python.

python/foxglove-sdk/python/foxglove/channel.py
eloff: I think being able to create a Channel instance with an encoder object that can serialize messages would resolve it. Having a kind of a base class is more invasive for the user's types without offering a commensurate benefit.

gasmith: Good point, that's probably a bit cleaner. And it could be composed with the high-level API, if you take my suggestion of allowing a user to declare a channel explicitly, and then reference it by topic-name in the high-level `log` API.

python/foxglove-sdk/python/foxglove/channel.py
```diff
@@ -87,24 +90,45 @@ def close(self) -> None:
 _channels_by_topic: Dict[str, Channel] = {}
 
 
-def log(topic: str, message: Any) -> None:
-    channel: Optional[Channel] = _channels_by_topic.get(topic, None)
+def log(
```
gasmith, line 93: Should we also plumb through the other keyword arguments from `Channel.log`?

python/foxglove-sdk-examples/live-visualization/main.py
```diff
@@ -134,11 +129,6 @@ def main() -> None:
         supported_encodings=["json"],
     )
 
-    # Log messages having well-known Foxglove schemas using the appropriate channel type.
-    box_chan = SceneUpdateChannel("/boxes")
-    tf_chan = FrameTransformsChannel("/tf")
-    point_chan = PointCloudChannel("/pointcloud")
```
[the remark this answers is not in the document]
eloff: I added some typed channel examples to live-viz and mcap, I think it will get more eyeballs that way (it's also less code overall)

python/foxglove-sdk/python/foxglove/channel.py
eloff: Yeah good point. Another reason to implement that in the future.

python/foxglove-sdk/python/foxglove/channel.py
eloff: Good idea

python/foxglove-sdk/python/foxglove/channel.py
```diff
@@ -87,24 +90,49 @@ def close(self) -> None:
...
+    topic: str,
+    message: Union[JsonMessage, list[Any], bytes, str, schemas.FoxgloveSchema],
+    log_time: Optional[int] = None,
```
bryfox, line 96: I think we should require kwargs for everything after messages; these ints run together. i.e., `(topic, message, *, log_time...`)

eloff: We don't currently do that for Channel.log, and there is a performance cost to that in Python. If we do decide to go that route, we should do that across the board.

gasmith: I think we _should_ do it across the board. Using explicit keywords for optional arguments is pretty standard hygiene in python. Is the overhead significant?

eloff: It allocates, populates, and frees a hashtable (dict) for each log call AFAIK

I know Python is slow anyway, but I'm in favor of leaving it as-is and letting the user decide if they want to call it with positional or keyword arguments.

bryfox: I see we also have a bug in the stub interface for typed channels — those do require kwargs. I'll open a separate bugfix PR for that.

bryfox: Bug fix: #309

Follow-up for base channel: FG-10841

rust/foxglove-proto-gen/src/lib.rs
gasmith: I disagree, mildly. If there's some other way by which protobuf is discovering these files, hardcoding these paths might disrupt a build environment that deliberately installs the protobuf toolchain elsewhere (e.g., to support side-by-side versioning).

python/foxglove-sdk/python/foxglove/channel.py
```diff
@@ -87,24 +91,55 @@ def close(self) -> None:
...
+    message: Union[JsonMessage, list[Any], bytes, str, schemas.FoxgloveSchema],
+    *,
+    log_time: Optional[int] = None,
+    publish_time: Optional[int] = None,
+    sequence: Optional[int] = None,
+) -> None:
+    """Log a message on a topic.
+
+    Creates a new channel the first time called for a given topic.
+    For foxglove types in the schemas module, this creates a typed channel.
```
bryfox, line 105: capitalize "Foxglove"

bryfox: could link to the typed channels module to give "typed channel" some context. something like:


```
(see :py:mod:`foxglove.channels`)
```

bryfox (commented):
Couple of nits on documentation, but this looks good to me.

Any progress on the proto path issue? Can we back those changes out for now?

rust/foxglove-proto-gen/src/lib.rs
eloff: That's a good point. We can make a ticket to revisit though, I don't have time to get to the root of it today, and I want to get this in today for the hack week.

gasmith: Since the problem seems to be isolated to your local environment, I'm inclined to suggest that we remove the path overrides before merging this change.

eloff:
> Couple of nits on documentation, but this looks good to me.
> 
> Any progress on the proto path issue? Can we back those changes out for now?

Backed out the proto path changes, fixed docs, should be good to go

python/foxglove-sdk/python/foxglove/channel.py
```diff
@@ -1,8 +1,13 @@
 import json
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Optional, Union, cast
 
+<<<<<<< HEAD
+from . import schemas
+from ._foxglove_py import BaseChannel, Schema, channels
```
bryfox, line 6: Sorry, this is from #310 — hopefully you can delete this line; otherwise, I changed the references to `_foxglove.BaseChannel` etc.

bryfox (approved):
Looks good to me once that merge conflict noise is cleaned up.

PR #304, opened 2025-03-20
rust: Fix websocket service schema encoding

gasmith:
### Description
When I implemented websocket services, I didn't know about the Service Call panel in the UI, and so I did all my testing with a custom client in rust. Turns out I botched the request/response schema encoding.

This change fixes the schema encoding, using the same logic that we use for encoding channel schemas. The ws-services example is now compatible with the UI.

### DIFF
--- a/rust/foxglove/src/websocket/protocol/server.rs
+++ b/rust/foxglove/src/websocket/protocol/server.rs
@@ -314,16 +314,15 @@ impl<'a> TryFrom<&'a service::MessageSchema> for AdvertiseServiceMessageSchema<'
 pub(crate) fn advertise_services<'a>(services: impl IntoIterator<Item = &'a Service>) -> String {
     let services: Vec<_> = services
         .into_iter()
-        .filter_map(|s| {
-            AdvertiseService::try_from(s)
-                .inspect_err(|e| {
-                    error!(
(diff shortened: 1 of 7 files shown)


bryfox (commented):
Looks good, minor comments on the example

bryfox (approved):

rust/foxglove/src/websocket/protocol/server.rs
```diff
@@ -185,6 +187,23 @@ fn is_schema_required(message_encoding: &str) -> bool {
         || message_encoding == "cdr"
 }
 
+/// Encodes schema data, based on message encoding.
+///
+/// For binary encodings, the schema data is base64-encoded. For other encodings, the schema must
+/// be valid UTF-8, or this function will return an error.
+fn encode_schema_data<'a>(
+    message_encoding: &str,
```
jtbandes, line 195: nit: this is actually a schema encoding, not a message encoding (there's a small voice inside me that wants to encourage a newtype for both of these so we don't confuse them, but I'm going to silence it 😅)

gasmith: :facepalm: whoops, nice catch.

PR #305, opened 2025-03-20
rust: Cope with buggy websocket service/asset handlers

gasmith:
### Changelog
- rust: Better handling for buggy websocket service/asset handlers

### Description
Improve the websocket server (and fetch asset) behavior, by ensuring that an error response is sent to the client, even if the handler fails to send one. This might happen because the handler panicked, or simply due to a flaw in implementation.


rust/foxglove/src/websocket.rs
```diff
@@ -819,7 +819,9 @@ impl ConnectedClient {
             return;
         };
 
-        // Prepare the responder and the request.
+        // Prepare the responder and the request. No failures past this point. If the responder is
+        // dropped without sending a response, it will send a generc "internal server error" back
+        // ot the client.
```
jtbandes, line 824:
```suggestion
        // to the client.
```

rust/foxglove/src/websocket/service/response.rs
```diff
@@ -45,11 +39,40 @@ impl Responder {
...
+    }
+}
+
+impl Drop for Responder {
+    fn drop(&mut self) {
+        if let Some(inner) = self.0.take() {
+            // The service call handler has dropped its responder without responding. This could be
+            // due to a panic or some other flaw in implementation. Reply with a generic error
+            // message.
+            inner.respond(Err("Internal server error".into()))
```
jtbandes, line 61: If it could also be a logic error, I would recommend that the error message should have a little more detail, such as `Internal server error: service was dropped without sending a response`

rust/foxglove/src/websocket/service/response.rs
```diff
@@ -45,11 +39,40 @@ impl Responder {
...
+        }
+    }
+}
+
+struct Inner {
+    client: Arc<ConnectedClient>,
+    service_id: ServiceId,
+    call_id: CallId,
+    encoding: String,
+    _guard: SemaphoreGuard,
```
jtbandes, line 71: curious what the `_` is meant to indicate, when the access level is not different from the other members?

jtbandes (approved):

rust/foxglove/src/websocket/service/response.rs
gasmith: It indicates that the field is intentionally never read. Often used for synchronization guards like this, or [phantom data](https://doc.rust-lang.org/std/marker/struct.PhantomData.html).

bryfox (approved):
I reviewed the asset responder update in the context of the rest of the PR; LGTM.

jtbandes (approved):

rust/foxglove/src/websocket/service/response.rs
eloff: Such a C++ question :)

rust/foxglove/src/websocket/tests.rs
```diff
@@ -1381,6 +1426,8 @@ async fn test_fetch_asset() {
                 |_client, uri: String| {
                     if uri.ends_with("error") {
                         Err("test error".to_string())
+                    } else if uri.ends_with("panic") {
+                        panic!("oh no")
```
eloff, line 1430: I checked the code twice, I don't understand how this works.

Doesn't the panic kill the process? I don't see where we recover it.

eloff (approved):
Looks great, ship it

rust/foxglove/src/websocket/tests.rs
gasmith: Answered in slack too, but for posterity: This panic happens inside a tokio task, so only the task exits. The rest of the program continues running.

PR #306, opened 2025-03-20
rust: Miscellaneous tidying

gasmith:
### Description
Odds and ends included here:
- For ws examples, use the crate name as the ws server name.
- Add `Display` impl for newtyped IDs.
- Hide `new()` constructors for IDs that users shouldn't create.
- Adjust `Debug` impls for `Schema` and `Request` to elide payloads.
- Derive `Clone` and `Eq` for a few public structures where it makes sense.
- Fix client-publish example by advertising support for JSON encoding

### DIFF
--- a/rust/examples/client-publish/src/main.rs
+++ b/rust/examples/client-publish/src/main.rs
@@ -23,16 +23,26 @@ struct Cli {
     port: u16,
     #[arg(long, default_value = "127.0.0.1")]
     host: String,
-    #[arg(long, default_value = "example-json-server")]
-    server_name: String,
 }
 
 struct ExampleCallbackHandler;
 impl ServerListener for ExampleCallbackHandler {
-    fn on_message_data(&self, _client: Client, channel: &ClientChannel, message: &[u8]) {
+    fn on_message_data(&self, client: Client, channel: &ClientChannel, message: &[u8]) {
         let json: serde_json::Value =
             serde_json::from_slice(message).expect("Failed to parse message");
-        println!("Received message on channel {}: {json}", channel.id);
+        println!(
+            "Client {} published to channel {}: {json}",
+            client.id(),
+            channel.id
+        );
+    }
+
+    fn on_client_advertise(&self, client: Client, channel: &ClientChannel) {
+        println!(
+            "Client {} advertised channel: {}",
+            client.id(),
+            channel.topic
+        );
     }
 }
 
@@ -44,10 +54,11 @@ async fn main() {
     let args = Cli::parse();
 
     let server = WebSocketServer::new()
-        .name("client-publish")
+        .name(env!("CARGO_PKG_NAME"))
         .bind(args.host, args.port)
         .capabilities([Capability::ClientPublish])
         .listener(Arc::new(ExampleCallbackHandler))
+        .supported_encodings(["json"])
         .start()
         .await
         .expect("Failed to start server");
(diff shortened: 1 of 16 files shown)


rust/foxglove/src/channel_builder.rs
```diff
@@ -6,6 +6,7 @@ use crate::{Channel, Context, Encode, FoxgloveError, Schema};
 
 /// ChannelBuilder is a builder for creating a new [`Channel`] or [`TypedChannel`].
 #[must_use]
+#[derive(Debug, Clone)]
```
jtbandes, line 9: why would one want to clone a channel builder?

rust/foxglove/src/websocket/protocol/client.rs
```diff
@@ -227,7 +227,7 @@ pub(crate) struct JsonClientChannel {
 }
 
 /// Information about a channel advertised by the client
-#[derive(Debug, PartialEq)]
+#[derive(Debug, Clone, PartialEq)]
```
jtbandes, line 230: maybe Eq as well?

On the other hand, both PartialEq and Eq kind of would indicate that it's acceptable to compare ClientChannels that came from different clients – which is it not.

rust/foxglove/src/websocket/service.rs
```diff
@@ -31,6 +31,14 @@ impl ServiceId {
     pub fn new(id: u32) -> Self {
         Self(id)
     }
+
+    /// Allocates the next service ID.
+    pub fn next() -> Self {
+        static NEXT_ID: AtomicU32 = AtomicU32::new(1);
+        let id = NEXT_ID.fetch_add(1, Ordering::Relaxed);
+        assert_ne!(id, 0);
```
jtbandes, line 39: maybe add a bit more detail:
```suggestion
        assert_ne!(id, 0, "ServiceId overflowed");
```

rust/foxglove/src/websocket/service/request.rs
```diff
@@ -9,7 +9,7 @@ use crate::websocket::ClientId;
 use super::{CallId, Service};
 
 /// A service call request.
-#[derive(Debug, Clone)]
+#[derive(Clone)]
```
jtbandes, line 12: TIL this feature is coming 🔜: https://github.com/rust-lang/rust/issues/121050

rust/examples/client-publish/src/main.rs
```diff
@@ -23,16 +23,26 @@ struct Cli {
...
             serde_json::from_slice(message).expect("Failed to parse message");
-        println!("Received message on channel {}: {json}", channel.id);
+        println!(
+            "Client {} published to channel {}: {json}",
+            client.id(),
+            channel.id
+        );
+    }
+
+    fn on_client_advertise(&self, client: Client, channel: &ClientChannel) {
```
jtbandes, line 40: maybe also include unadvertise?

jtbandes (approved):

rust/foxglove/src/websocket/protocol/client.rs
gasmith: Yeah, probably deserves to be `Eq`.

We currently use the `PartialEq` for tests, so we _could_ hide it under a `#[cfg_attr(test, ...)]`, but on the other hand I feel like that's maybe getting _too_ pedantic.

rust/foxglove/src/websocket/service/request.rs
gasmith: Thank god hurry up already

PR #309, opened 2025-03-21
python: fix stub definitions for typed channel logging

bryfox:
### Changelog
[python] Fix: typed channel stubs correctly indicate kwargs

### Description

The `log()` function for typed channels use kwargs for optional arguments after the initial message. Example:

https://github.com/foxglove/foxglove-sdk/blob/4360f833c87c29aa175e9cc53199920708abcded/python/foxglove-sdk/src/generated/channels.rs#L86

This fixes the generated stub definitions to match.

This also adds a "generated" header to the channels source, which was missing.


gasmith (approved):
:ship:

PR #311, opened 2025-03-21
Add rust high level log! macro - type-safe logging without channels

eloff:
I didn't think it was possible, but it looks like I have found a way to get type-safe logging without channels in rust without a runtime cost.

The magic is in create_channel, which is called at most once for any given log site. It creates the Channel or fetches an existing one from the Context for the topic. It compares the schema and message_encoding and panics if they differ, making it type-safe.

One catch is the Channel, once fetched or created, will live until the end of the process. It can be added or removed from sinks or contexts, but it can't ever be freed. It would be possible to provide a method to free it, but I don't think that makes any sense.

The type of the log $msg for a given log site can't change in Rust AFAIK. It's always the same type (even if that type were dynamic, e.g. 'Box<dyn Encode>' it's always a 'Box<dyn Encode>').

I added optional "named arguments" like keyword argument syntax in Python for log_time, publish_time, sequence.

I didn't add a context argument, one because it would double the number of cases in the macro, which are already many. But also because I think it's not that useful. If you want to specify the context, it's better to explicitly create the channel.


rust/foxglove/src/log_macro.rs
```diff
@@ -0,0 +1,350 @@
...
+
+impl ChannelPlaceholder {
+    pub fn new(channel: Arc<RawChannel>) -> *mut Self {
+        Arc::into_raw(channel) as *mut Self
+    }
+
+    pub unsafe fn log<T: Encode>(channel_ptr: *mut Self, msg: &T, metadata: PartialMetadata) {
+        // Safety: we're restoring the Arc<Channel> we leaked into_raw in new()
+        let channel_arc = Arc::from_raw(channel_ptr as *mut RawChannel);
+        // We can safely create a TypedChannel from any Arc<Channel>
```
gasmith, line 17:
```suggestion
        // Safety: we're restoring the Arc<RawChannel> we leaked into_raw in new()
        let channel_arc = Arc::from_raw(channel_ptr as *mut RawChannel);
        // We can safely create a Channel from any Arc<RawChannel>
```

gasmith (approved):
Looks awesome. Should we add this to the library rustdoc, or do you want to do that in a separate change?

rust/foxglove/src/context.rs
```diff
@@ -239,11 +239,16 @@ impl Context {
     pub fn unsubscribe_channels(&self, sink_id: SinkId, channel_ids: &[ChannelId]) {
         self.0.write().unsubscribe_channels(sink_id, channel_ids);
     }
+
+    /// Removes all channels and sinks from the context.
+    pub(crate) fn clear(&self) {
+        self.0.write().clear();
+    }
```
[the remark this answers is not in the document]
eloff: It's pub(crate) so users can't use it and we don't need to document it externally

rust/foxglove/src/lib.rs
```diff
@@ -85,6 +83,31 @@
...
+//! let channel = Channel::new("/log")?;
+//! channel.log(&Log{
+//!     message: "Hello, Foxglove!".to_string(),
+//!     ..Default::default()
+//! });
+//!
+//! // Flush and close the MCAP file.
+//! mcap.close()?;
+//! # Ok(()) }
+//!
```
gasmith, line 107:
```suggestion
//! ```
//!
```

rust/foxglove/src/lib.rs
```diff
@@ -85,6 +83,31 @@
 //!
 //! [MCAP channel]: https://mcap.dev/guides/concepts#channel
 //!
+//! In the example above, log! creates a Channel("/log") behind the scenes on the first call.
```
gasmith, line 86:
```suggestion
//! In the example above, `log!` creates a channel behind the scenes on the first call.
```

rust/foxglove/src/lib.rs
```diff
@@ -9,21 +9,19 @@
...
-//! To record messages, you need at least one sink, and at least one channel. In this example, we
-//! create an MCAP file sink, and a channel for [`Log`](`crate::schemas::Log`) messages on a topic
-//! called `/log`. Then we write one log message and close the file.
+//! To record messages, you need at least one sink. In this example, we
+//! create an MCAP file sink, and log a [`Log`](`crate::schemas::Log`) message
+//! on a topic called `/log`. We write one log message and close the file.
 //!
 //! ```no_run
-//! use foxglove::{Channel, McapWriter};
+//! use foxglove::{McapWriter, Channel, log};
```
gasmith, line 17:
```suggestion
//! use foxglove::{McapWriter, log};
```

gasmith (approved):
Yep, I like it.

PR #312, opened 2025-03-21
python: fix up the live viz example

bryfox:
### Changelog
None

### Description

This cleans up the live viz example and fixes a bug with the example subscriber handling. Subscription callbacks are called for each topic, so keep track of what topics clients are subscribed to.

### DIFF
--- a/python/foxglove-sdk-examples/live-visualization/main.py
+++ b/python/foxglove-sdk-examples/live-visualization/main.py
@@ -1,7 +1,6 @@
 import datetime
 import json
 import logging
-import math
 import struct
 import time
 from math import cos, sin
@@ -51,7 +50,8 @@ plot_schema = {
 
 class ExampleListener(ServerListener):
     def __init__(self) -> None:
-        self.subscribers: set[int] = set()
+        # Map client id -> set of subscribed topics
+        self.subscribers: dict[int, set[str]] = {}
 
     def has_subscribers(self) -> bool:
         return len(self.subscribers) > 0
@@ -66,7 +66,9 @@ class ExampleListener(ServerListener):
         We'll use this and on_unsubscribe to simply track if we have any subscribers at all.
         """
         logging.info(f"Client {client} subscribed to channel {channel.topic}")
-        self.subscribers.add(client.id)
+        if client.id not in self.subscribers:
+            self.subscribers[client.id] = set()
+        self.subscribers[client.id].add(channel.topic)
 
     def on_unsubscribe(
         self,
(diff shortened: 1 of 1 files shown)


python/foxglove-sdk-examples/live-visualization/main.py
```diff
@@ -228,6 +229,10 @@ def main() -> None:
 
             time.sleep(0.05)
 
+            while not listener.has_subscribers():
+                time.sleep(1)
+                continue
```
gasmith, line 234: Hm, I wonder if we should expose `Channel::has_sinks` in python? I guess a new `Context::has_sinks` would be more ideal, but that would also mean exposing `Context`.

python/foxglove-sdk-examples/live-visualization/main.py
```diff
@@ -51,7 +50,8 @@
 
 class ExampleListener(ServerListener):
     def __init__(self) -> None:
-        self.subscribers: set[int] = set()
+        # Map client id -> set of subscribed topics
+        self.subscribers: dict[int, set[str]] = {}
```
eloff, line 54:
```suggestion
        self.subscribers: dict[int, set[str]] = defaultdict(set)
```

Add `from collections import defaultdict`

And then you can just do `self.subscribers[client.id].add(topic)`, the set will be created if one doesn't exist yet.

eloff (approved):
One optional change you can make to improve code readability a little

python/foxglove-sdk-examples/live-visualization/main.py
```diff
@@ -66,7 +66,9 @@ def on_subscribe(
         We'll use this and on_unsubscribe to simply track if we have any subscribers at all.
         """
         logging.info(f"Client {client} subscribed to channel {channel.topic}")
-        self.subscribers.add(client.id)
+        if client.id not in self.subscribers:
+            self.subscribers[client.id] = set()
+        self.subscribers[client.id].add(channel.topic)
```
gasmith, line 71:
```suggestion
        self.subscribers.setdefault(client.id, set()).add(channel.topic)
```

gasmith (approved):

python/foxglove-sdk-examples/live-visualization/main.py
bryfox: Maybe. The SDK already avoids most of the work if you simply call log() and don't worry about this state, so this example is somewhat artificial. I actually considered removing it entirely. Should we wait to see what pain points users have?

gasmith: Yeah, let's wait.

python/foxglove-sdk-examples/live-visualization/main.py
bryfox: 👍 but I went with the `setdefault` suggestion below

PR #313, opened 2025-03-21
C/C++: typed channels for builtin schemas

jtbandes:
### Description

Current proposed structure:
- Add a `foxglove_typed_channel` type, with an enum to specify which builtin type is used
- On the Rust side, store an enum with the rust TypedChannel as a payload. The `log` function takes a `void*` for the message struct, and switches on the rust enum to know which type of message to interpret it as.


eloff (commented):
This is looking really good, I like the approach!

eloff:
Closing in favor of https://github.com/foxglove/foxglove-sdk/pull/413

PR #314, opened 2025-03-21
python: use published SDK for examples

bryfox:
### Changelog

python: update examples to use the published package

### Description

We no longer require local build environments to run these python examples.

### DIFF
--- a/python/foxglove-sdk-examples/live-visualization/pyproject.toml
+++ b/python/foxglove-sdk-examples/live-visualization/pyproject.toml
@@ -5,6 +5,7 @@ description = "Example of live visualization in Foxglove"
 authors = [{ name = "Foxglove", email = "support@foxglove.dev" }]
 readme = "README.md"
 requires-python = ">=3.9"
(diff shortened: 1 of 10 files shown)


python/foxglove-sdk-examples/live-visualization/pyproject.toml
```diff
@@ -5,6 +5,7 @@ description = "Example of live visualization in Foxglove"
 authors = [{ name = "Foxglove", email = "support@foxglove.dev" }]
 readme = "README.md"
 requires-python = ">=3.9"
+dependencies = ["foxglove-sdk>=0.4.0", "numpy>=2.0"]
```
amacneil, line 8: Is there a way we could override this in CI so that the examples are tested (or at least type checked) against the main library?

bryfox: I don't think this is well-supported in poetry, so I updated the script to install the local version when prompted. 

https://github.com/foxglove/foxglove-sdk/actions/runs/14111136584/job/39530020221#step:13:39

gasmith (approved):

PR #316, opened 2025-03-24
Fix memory ordering and remove unnecessary Arc

Muon:
### Changelog
Fix memory ordering bug in Rust quickstart example.

### Docs

Related: FG-10853

### Description

Using `Ordering::Relaxed` can, on platforms with weak memory ordering, lead to the Ctrl-C handler never causing the program to exit, if it is handled on a separate thread to the main thread, since a store with `Ordering::Relaxed` does not ensure that another thread reading that value will ever actually read what got stored, just that it will either read the old value or the new value.

### DIFF
--- a/rust/examples/quickstart/src/main.rs
+++ b/rust/examples/quickstart/src/main.rs
@@ -1,9 +1,6 @@
 use std::{
     ops::Add,
-    sync::{
-        atomic::{AtomicBool, Ordering},
-        Arc,
-    },
+    sync::atomic::{AtomicBool, Ordering},
 };
 
 use foxglove::McapWriter;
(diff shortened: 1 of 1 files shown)


bennetthardwick:
If the store and load are reordered such that the loop still reads false when the signal comes through it will read true on the next iteration of the loop. This logic isn't relying on strong memory ordering so I don't think this is necessary.

Muon:
It may not. `Relaxed` does not guarantee that the other thread will ever read the updated value.

Muon:
Hmm, actually I'm wrong, though it's kinda worse than I expected: https://stackoverflow.com/a/70593598

`Relaxed` *is* okay... because none of the ops actually guarantee visibility happening ever.

PR #317, opened 2025-03-24
Preserve schema name in encodings

bryfox:
### Changelog
Fix: the schema name for GeoJSON channels capitalized correctly

### Description

foxglove-proto-gen renames the GeoJSON schema to GeoJson to match structure naming conventions, but incorrectly writes that as the schema name. This preserves the renaming behavior, but fixes the schema name for compatibility with the app.


eloff (approved):
I knew there was a reason I changed the name, but this is a better fix.

LGTM

PR #318, opened 2025-03-25
Add custom-protobuf example

defunctzombie:
### Changelog
TBD

### Docs
TBD

### Description

</td><td>

</td></tr></table>


eloff:
I think this can be closed now, as foxglove derive implements the spirit of this PR, feel free to re-open if that's not true

PR #319, opened 2025-03-25
rust: Expose ChannelId struct

achim-k:
### Changelog
rust: Expose `ChannelId` struct

### Docs
TBD

### Description
In my custom Sink, I'd like to use a `HashMap<ChannelId, T>`. This PR exposes the `ChannelId` struct publicly to support that use case.


gasmith (approved):
:ship:

eloff (approved):
Yeah, this should be public

PR #320, opened 2025-03-25
C++: fix schema lifetime issue in Channel constructor

wimagguc:
### Changelog

None

### Docs

None

### Description

Updates the C++ wrapper's `Channel` implementation to deep copy the schema data into an internal `std::vector<uint8_t>` — this ensures that the memory remains valid for the lifetime of the channel, so the copied data can be passed to the C API at the user's convenience.

The `foxglove::Channel` constructor previously stored a raw pointer to schema data passed via the Schema struct. This was a problem because if the original memory was invalidated after construction, it could lead to dangling pointers, which lead to a corrupted schema, which lead to this error:

### DIFF
--- a/cpp/foxglove/include/foxglove/channel.hpp
+++ b/cpp/foxglove/include/foxglove/channel.hpp
@@ -4,6 +4,7 @@
 #include <memory>
 #include <optional>
 #include <string>
+#include <vector>
 
 struct foxglove_channel;
 
@@ -32,6 +33,7 @@ public:
 
 private:
   std::unique_ptr<foxglove_channel, void (*)(foxglove_channel*)> _impl;
+  std::vector<uint8_t> _storedSchema;
 };
 
 }  // namespace foxglove
--- a/cpp/foxglove/src/channel.cpp
+++ b/cpp/foxglove/src/channel.cpp
@@ -8,14 +8,25 @@ Channel::Channel(
 )
     : _impl(nullptr, foxglove_channel_free) {
   foxglove_schema cSchema = {};
+  std::optional<foxglove_schema*> schemaPtr = std::nullopt;
+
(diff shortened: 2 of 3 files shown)


jtbandes (commented):
The schema should never be accessed again after the `foxglove_channel_create` call. Internally the rust core should take a copy of it. So I'm not sure why it would be needed for the C++ Channel to store a copy that it never uses again?

eloff (changes requested):
The rust side should be copying this data. If it's not, that's a bug in `foxglove_channel_create`. The C++ side shouldn't be keeping a copy of the schema.

The code is:

```
let schema = unsafe {
        schema.as_ref().map(|schema| {
            let name = CStr::from_ptr(schema.name)
                .to_str()
                .expect("schema name is invalid");
            let encoding = CStr::from_ptr(schema.encoding)
                .to_str()
                .expect("schema encoding is invalid");
            let data = std::slice::from_raw_parts(schema.data, schema.data_len);
            foxglove::Schema::new(name, encoding, data)
        })
    };
```
    
The problem is Schema::new converts data to a Cow using into(). Which strangely in this case is borrowing the data instead of copying it (treating it like a static byte array). We need to construct the Cow outside Schema::new as an owned Cow (copying) and pass that instead.

Confirmed with this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=f857dfaf1f02061bd48dd4cca3201986

wimagguc:
> The rust side should be copying this data. If it's not, that's a bug in `foxglove_channel_create`. The C++ side shouldn't be keeping a copy of the schema.

Yup, indeed, this is a bug in `foxglove_channel_create` — closing this PR in favor of fixing the bug at that level. Thanks both!

PR #321, opened 2025-03-25
Fix lifetime issues

jtbandes:
### Changelog
C++: fixed a potential use-after-free by adding a missing copy.

### Docs

None

### Description

Reported in #320 

It was intended that the Schema data should never be used again after the Channel constructor, but because it was held by a Cow, it was not actually being copied. Introduce an explicit `to_owned()` to fix the issue.


eloff (approved):
LGTM

PR #322, opened 2025-03-25
Add prebuilt C library for iOS & simulator

jtbandes:
### Changelog
C/C++: added prebuilt `.a`/`.dylib`s for iOS & iOS Simulator

### Docs

None

### Description

I tried adding the other Apple targets too (watchOS, tvOS, visionOS) but they weren't supported with this current CI workflow setup, probably because they are Tier 3 targets: https://rust-lang.github.io/rustup-components-history/


eloff (approved):
LGTM

wimagguc (approved):
Oh this hasn't merged, I just grabbed the artifacts from the CI? Thanks for these!

PR #324, opened 2025-03-26
Add `foxglove::Encode` derive macro

defunctzombie:
### Changelog
Add a `foxglove::Encode` derive macro to implement the `Encode` trait for logging custom structs using a binary serialization format.

### Docs
TBD

### Description

```rust
#[derive(Serialize, JsonSchema)]
enum MessageLevel {
    Debug,
    Info,
}

#[derive(Serialize, JsonSchema)]
struct Message {
    level: MessageLevel,
    msg: String,
    count: u32,
}
```

</td><td>

```rust
#[derive(foxglove::Encode)]
enum MessageLevel {
    Debug,
    Info,
}

#[derive(foxglove::Encode)]
struct Message {
    level: MessageLevel,
    msg: String,
    count: u32,
}
```

</td></tr></table>

#### Background
When using the Rust SDK to log data, one might want to define app-specific structs for logging. The current recommended approach for this directs the user to derive the `Serialize` and `JsonSchema` traits for their struct. The foxglove SDK then provides a specialization of the `Encode` trait for structs which have implemented these two other traits.

```rust
#[derive(serde::Serialize, schemars::JsonSchema)]
struct Message {
    msg: String,
    count: u64,
}
```

This has some downsides:
- Field names are repeated in every message (leading to message size bloat for smaller messages and high-rate data)
- "byte" fields are serialized as base64 strings
- No support for u64 (larger integers are not properly deserialized in the app)

I believe there is a better way forward; to instead guide the user down a path that serializes their structs using one of the binary formats we support in the app.

#### Proposal

This proposal introduces a new `foxglove::Encode` derive macro. The purpose of this macro is to implement the `Encode` trait for a struct and thereby making it possible to log the struct on a channel.

> ~Note: Looking at prior art like serde and builtins, the macro should be called `Encode` since it "derives" the Encode trait when used. There is a small awkwardness with this and using it for enums but I think that can be overlooked. I would not go with `Loggable` but did not know how to properly export it from the foxglove crate and not collide with the existing `Encode` trait.~

The proposed macro generates an implementation that serializes the struct using the protobuf serialization format.

> Note: I consider the choice of protobuf an "implementation detail" and the user does not need to "worry" about. However, it can also be seen as an important call-out that the users recorded data serializes to a well-known open-source format so it can also be easily consumed in scripts or processing pipelines.

### Implementation

Rust requires that macros be in their own crate. For this I've introduced the `foxglove-derive` create (following the pattern I saw here: https://github.com/tokio-rs/prost and here: https://github.com/serde-rs/serde/tree/master/serde_derive)

This crate will be a dependency of the `foxglove` crate which means if we want to publish them together we will need to publish it first and ensure the versions stay in-sync.

### TODO

- [x] Settle on the macro name
- [x] Move the `ProtobufField` trait and impls to a subdirectory/namespace (I think this will aid organization since these are not really meant to be used by outside folks as part of the "public" interface).
- [x] Implement `ProtobufField` for the remaining rust built-in types we want to support (add test coverage accordingly)
- [x] Implement an encoded_len interface for writers
- [x] Update the publish steps to keep crates in sync (derive must publish first)


rust/foxglove-derive/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+    }
+}
+
+fn derive_enum_impl(input: DeriveInput) -> TokenStream {
+    let name = &input.ident;
+
+    // Extract variants from the enum
+    let variants = match &input.data {
+        Data::Enum(data) => &data.variants,
+        _ => unreachable!(),
```
gasmith, line 25: Could avoid the `unreachable!` by passing the enum we unpacked on L13.

rust/foxglove-derive/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+    let name = &input.ident;
+    let name_str = name.to_string();
+    let package_name = name_str.to_lowercase();
+    let full_name = format!("{}.{}", package_name, name_str);
+
+    // Extract fields from the struct
+    let fields = match &input.data {
+        Data::Struct(data) => match &data.fields {
+            Fields::Named(fields) => &fields.named,
+            _ => panic!("Only named fields are supported"),
```
gasmith, line 88: We should probably use [`compile_error!`](https://doc.rust-lang.org/std/macro.compile_error.html) here.

rust/foxglove-derive/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+                #(#field_encoders)*
+
+                // Write the length as a varint
+                let len = buf.len();
+                let mut len_value = len as u64;
+                while len_value >= 0x80 {
+                    out.put_u8((len_value as u8) | 0x80);
+                    len_value >>= 7;
+                }
+                out.put_u8(len_value as u8);
```
gasmith, line 167: Could do something like this:

```suggestion
                let mut len_value = len as u64;
                ::foxglove::ProtobufField::write(&len_value, buf);
```

That's essentially a hack to get at the (inlined) `prost::encoding::encode_varint` that we want. Maybe we should re-export `foxglove::protobuf::encode_varint` under a `#[doc(hidden)]` and then we can just call that directly. Ditto for the enum variant encoding above.

rust/foxglove-derive/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+                0 // Varint, same as integers
+            }
+
+            fn write(&self, buf: &mut impl foxglove::bytes::BufMut) {
+                let mut value = *self as u64;
+                while value >= 0x80 {
+                    buf.put_u8((value as u8) | 0x80);
+                    value >>= 7;
+                }
+                buf.put_u8(value as u8);
```
gasmith, line 57: This should be i32 to match the descriptor type. Maybe also reuse the existing implementation:

```suggestion
                let value = *self as i32;
                ::foxglove::ProtobufField::write(&value, buf);
```

rust/foxglove-derive/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+            let mut value = foxglove::prost_types::EnumValueDescriptorProto::default();
+            value.name = Some(String::from(#variant_name));
+            value.number = Some(#variant_value);
+            enum_desc.value.push(value);
+        }
+    });
+
+    // Generate implementation
+    let expanded = quote! {
+        impl foxglove::ProtobufField for #name {
```
gasmith, line 42: Throughout the proc macro, we should always use the fully-qualified `::foxglove` to refer to the foxglove crate, because there might be a local `mod foxglove` that takes precedence. We could add a `mod foxglove {}` to our tests to cover this.

rust/foxglove-derive/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+
+            fn type_name() -> Option<String> {
+                Some(stringify!(#name).to_string())
+            }
+        }
+
+        impl foxglove::Encode for #name {
+            type Error = std::io::Error;
+
+            fn get_schema() -> Option<foxglove::Schema> {
```
gasmith, line 199: Feels like another good target for memoization.

rust/foxglove/README.md
```diff
@@ -9,26 +9,9 @@ Foxglove app.
 [Foxglove]: https://docs.foxglove.dev/
 [MCAP]: https://mcap.dev/
 
-# Overview
```
gasmith, line 12: :+1:

rust/foxglove/src/protobuf.rs
```diff
@@ -0,0 +1,277 @@
...
+/// type definitions in Foxglove.
+///
+/// # Arguments
+///
+/// * `file_descriptor_set` - A reference to the Protocol Buffers FileDescriptorSet to serialize
+///
+/// # Returns
+///
+/// A `Vec<u8>` containing the binary protobuf encoding of the FileDescriptorSet
+pub fn prost_file_descriptor_set_to_vec(
```
gasmith, line 16: Maybe `#[doc(hidden)]`? Doesn't feel like our core line of business.

rust/foxglove/src/protobuf.rs
```diff
@@ -0,0 +1,277 @@
...
+}
+
+// Implement ProtobufField for u64 that serializes the value as a varint in protobuf serialization
+impl ProtobufField for u64 {
+    fn field_type() -> ProstFieldType {
+        ProstFieldType::Uint64
+    }
+
+    fn wire_type() -> u32 {
+        0
```
gasmith, line 94: If there isn't an existing enumeration we can use in `prost_types`, can we define one of our own?

rust/foxglove-derive/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+    let variant_descriptors = variants.iter().enumerate().map(|(i, v)| {
+        let variant_name = v.ident.to_string();
+        let variant_value = i as i32;
+        quote! {
+            let mut value = foxglove::prost_types::EnumValueDescriptorProto::default();
+            value.name = Some(String::from(#variant_name));
+            value.number = Some(#variant_value);
+            enum_desc.value.push(value);
+        }
+    });
```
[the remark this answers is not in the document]
gasmith: Oh also we should update our enum test to include enums with (non-zero, negative) discriminants.

rust/foxglove-derive/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+                    #(#enum_defs)*
+                }
+
+                {
+                    let mut nested_type = &mut message.nested_type;
+                    #(#message_defs)*
+                }
+
+                Some(message)
+            }
```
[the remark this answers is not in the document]
defunctzombie: In theory you only construct this when constructing a new channel right? So typically once at program startup?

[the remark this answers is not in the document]
gasmith: Yes, but we also have lazily-constructed channels, and the high-level log macro that constructs the schema [once per callsite](https://github.com/foxglove/foxglove-sdk/blob/26028f93a18f67743a53f0f6d42baf793b97158c/rust/foxglove/src/log_macro.rs#L101).

There's also the recursive bit. For multiple messages that nest the same inner types, we end up deriving the inner type's schema repeatedly.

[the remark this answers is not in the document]
defunctzombie: I would say that the construction of "once per callsite" is wild and we should fix that.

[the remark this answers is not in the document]
gasmith: That's just for the `log!` macro, which is mostly there as a convenience; you can construct channels in other ways. For `log!`, you have to construct the schema to determine whether you have an existing channel with a matching schema. I don't think there's any way around that.

rust/foxglove-derive/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "foxglove-derive"
+version = "0.0.1"
```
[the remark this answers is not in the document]
bryfox: TBD. I think that would be the nicest way to do it, but it means we need to wait on the crates.io release as part of our release process (— correct me if I'm wrong). I know we had an issue once related to docs.rs queue times, but maybe it's OK to wait on crates.io? Obviously this needs to be dealt one way or another with before we merge.

rust/foxglove-derive/src/lib.rs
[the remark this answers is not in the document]
bryfox: Done in #396

rust/foxglove-derive/src/lib.rs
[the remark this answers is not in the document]
bryfox: gasmith fixed the `log!` bug in #400.

I memoized the descriptor (and schema, per below comment) using a `OnceLock`; let me know if you had something else in mind.

rust/foxglove-derive/src/lib.rs
bryfox: Done in #396

rust/foxglove-derive/src/lib.rs
bryfox: Done in #396; I re-exported `encode_varint` at the top level (as with `prost_file_descriptor_set_to_vec`).

rust/foxglove-derive/src/lib.rs
bryfox: I don't understand the i32 comment; this should be writing a varint, right? In #396 I re-used the varint fn as suggested above.

rust/foxglove-derive/src/lib.rs
bryfox: Done in #396

rust/foxglove/src/protobuf.rs
bryfox: There is one; updated in #396

rust/foxglove-derive/src/lib.rs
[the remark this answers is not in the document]
gasmith: Oh right, I had forgotten about protobuf's requirements about default enum variants. I suppose this would mostly be a concern for deserialization. Anyone constructing the enum in rust is going to supply a value (which may be the `#[default]`-tagged variant), although I suppose an unset `Option<T>` enum will serialize as 0.

I think at a minimum, we should include user-facing documentation to indicate that we're using protobuf as the underlying serialization format. We can use that opportunity to call out gotchas like this one.

As I suggested above, I think we'll probably want to augment this proc macro to catch various invalid enum definitions that we can't handle properly, and display a nice error message. We can piggy-back on that work to detect the case where there is on zero-valued enum variant. Future work, though; not blocking this PR.

bryfox:
I believe this is ready to go but requires publishing a 0.7.1 version of the foxglove-derive crate; our publishing token in CI is not authorized for publishing a new crate. There's an 'empty' crate in #426 we can publish, which should allow CI checks here to start passing.

PR #329, opened 2025-03-31
python: Channel.close() should remove channel from context

gasmith:
### Changelog
- python: Fixed a bug with Channel.close(), which wasn't unregistering the channel

### Description
A straightforward change to make `Channel.close()` behave as advertised by unregistering the channel from the context.


eloff (approved):
LGTM

PR #330, opened 2025-03-31
Optimize buffering in log_to_sinks

eloff:
### Changelog

Improve logging performance by ~20%

### Description

This actually simplifies the code. I had to bring in the SmallBytes crate for a BufMut wrapper around SmallVec.

### DIFF
--- a/rust/foxglove/Cargo.toml
+++ b/rust/foxglove/Cargo.toml
@@ -32,6 +32,7 @@ serde_repr = "0.1.19"
 serde_with = { version = "3.12.0", features = ["macros", "base64"] }
 serde.workspace = true
 smallvec = "1.14.0"
+smallbytes = "0.1.0"
 strum = { version = "0.26", features = ["derive"] }
 thiserror.workspace = true
 tokio-tungstenite.workspace = true
--- a/rust/foxglove/src/channel.rs
+++ b/rust/foxglove/src/channel.rs
@@ -112,6 +112,9 @@ impl<T: Encode> Channel<T> {
     fn log_to_sinks(&self, msg: &T, metadata: PartialMetadata) {
         // Try to avoid heap allocation by using a stack buffer.
         let mut buf: SmallBytes<STACK_BUFFER_SIZE> = SmallBytes::new();
+        if let Some(estimated_size) = msg.encoded_len() {
+            buf.reserve(estimated_size);
+        }
 
         msg.encode(&mut buf).unwrap();
         self.inner.log_to_sinks(&buf, metadata);
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -691,6 +691,7 @@ dependencies = [
  "serde_json",
  "serde_repr",
  "serde_with",
+ "smallbytes",
  "smallvec",
  "strum 0.26.3",
  "tempfile",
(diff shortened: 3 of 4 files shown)


rust/foxglove/Cargo.toml
```diff
@@ -32,6 +32,7 @@ serde_repr = "0.1.19"
 serde_with = { version = "3.12.0", features = ["macros", "base64"] }
 serde.workspace = true
 smallvec = "1.14.0"
+smallbytes = "0.1.0"
```
gasmith, line 35: Hm, this crate appears to be [fresh abandonware](https://users.rust-lang.org/t/new-crate-smallbytes-smallvec-impl-bufmut/112889)...

... but it looks like the stars have aligned, and `impl BufMut for SmallVec` was [recently added](https://github.com/servo/rust-smallvec/pull/371) to `smallvec`, [to be released with v2.0.0](https://github.com/servo/rust-smallvec/releases/tag/v2.0.0-alpha.11)!

rust/foxglove/src/channel.rs
```diff
@@ -110,29 +111,10 @@ impl<T: Encode> Channel<T> {
...
-
-        match msg.encode(&mut cursor) {
-            Ok(()) => {
-                // Compute the written amount of bytes
-                let written = cursor.as_ptr() as usize - stack_buf.as_ptr() as usize;
-                self.inner.log_to_sinks(&stack_buf[..written], metadata);
-            }
-            Err(_) => {
-                // Likely the stack buffer was too small, so fall back to a heap buffer.
-                let mut size = msg.encoded_len().unwrap_or(STACK_BUFFER_SIZE * 2);
```
gasmith, line 124: I'm a little surprised we don't benefit more from the `encoded_len()` hint. Which message sizes did you benchmark against?

gasmith (commented):
Feels like this might be a nice change to bring in `divan`, so we can see some before & after numbers.

rust/foxglove/src/channel.rs
eloff: I was testing with 440 kg message sizes.

In practice what happens is everything goes into the stack buffer, then the data field is serialized, it either fits or it's too big and it grows the buffer to fit. Remaining fields won't be enough to cause it to grow again.

I think we could still use encoded_len with reserve or similar, but it doesn't make a difference.

rust/foxglove/Cargo.toml
eloff: Nice. I think let's start with this and then we can remove the smallbytes dependency later.

gasmith: Sure. Let's file a linear ticket so we don't forget.

rust/foxglove/src/channel.rs
gasmith: Well, what if you had _several_ big fields? Couldn't each one theoretically trigger a new allocation as the serializer attempts to write them, one after another?

eloff: I added a call to reserve(). It might be a little slower in some cases as encoded_len() is recursive and not that cheap for prost. Should be minor compared to the memory copy cost.

rust/foxglove/Cargo.toml
eloff: Added FG-10965

PR #331, opened 2025-04-01
rust: Replace static_channel! with LazyChannel

gasmith:
### Changelog
- rust: Replace `static_channel!` with `LazyChannel`

### Description
This change attempts to address some of the feedback we've received recently:
- `static_channel!` is weird, and merely obscures typing.
- Lazily-initialized channels are a footgun.

This change aims to address both concerns by introducing a new `LazyChannel` struct. This removes the macro weirdness, and updates the naming to make the laziness more overt.

The result is a little more verbose, but I think the clarity justifies the keystrokes:

```rust
// Before
static_channel!(pub(crate), TF, FrameTransform);

// After
pub(crate) static TF: LazyChannel<FrameTransform> = LazyChannel::new("/tf");
```

Fixes: FG-10886
Fixes: FG-10913

### DIFF
--- a/rust/examples/mcap/src/main.rs
+++ b/rust/examples/mcap/src/main.rs
@@ -3,7 +3,7 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
 use clap::{Parser, ValueEnum};
-use foxglove::McapWriter;
+use foxglove::{LazyChannel, McapWriter};
 use mcap::{Compression, WriteOptions};
 use std::time::Duration;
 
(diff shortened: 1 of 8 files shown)


rust/examples/mcap/src/main.rs
```diff
@@ -48,7 +48,7 @@ struct Message {
     count: u32,
 }
 
-foxglove::static_channel!(pub MSG_CHANNEL, "/msg", Message);
+static MSG_CHANNEL: LazyChannel<Message> = LazyChannel::new("/msg");
```
eloff, line 51: Nice, yes, this is better

eloff (approved):
This is great. Could we also do something similar for a raw channel?

gasmith:
> This is great. Could we also do something similar for a raw channel?

I suppose so... something like this?

```rust
static CH: LazyRawChannel = LazyRawChannel::new("/raw", "message_encoding")
    .schema("name", "schema_encoding", b"data");
````

Muon (approved):
I like it!

eloff:
> > This is great. Could we also do something similar for a raw channel?
> 
> I suppose so... something like this?
> 
> ```rust
> static CH: LazyRawChannel = LazyRawChannel::new("/raw", "message_encoding")
>     .schema("name", "schema_encoding", b"data");
> ```

Yeah, something like that

PR #332, opened 2025-04-01
Fix build errors with cmake

eloff:
I think this might do it. Leaning heavily on Claude for this fix. I barely understand it, but it works on my machine (tm)


eloff:
Now the Python build is failing, working on it...

gasmith (approved):

PR #333, opened 2025-04-01
Add Timestamp::now() convenience method

eloff:
I really wanted a Timestamp::now() method during the hack week. The try_from SystemTime works, but it's verbose.

### DIFF
--- a/rust/foxglove/src/schemas_wkt.rs
+++ b/rust/foxglove/src/schemas_wkt.rs
@@ -339,6 +339,15 @@ impl Timestamp {
         Self::new_checked(sec, nsec).unwrap()
     }
 
+    /// Returns the current timestamp using SystemTime
+    pub fn now() -> Self {
+        let now = std::time::SystemTime::now();
+        let duration = now.duration_since(std::time::UNIX_EPOCH).unwrap();
+        let sec = duration.as_secs() as u32;
+        let nsec = duration.subsec_nanos();
+        Self::new(sec, nsec)
+    }
+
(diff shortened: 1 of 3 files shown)


rust/foxglove/src/schemas_wkt/tests.rs
```diff
@@ -281,3 +281,14 @@ fn test_timestamp_from_system_time() {
...
 }
+
+#[test]
+fn test_timestamp_now() {
+    let now = std::time::SystemTime::now();
+    let before = Timestamp::try_from(now).unwrap();
+
+    let now = Timestamp::now();
+
+    assert!(now.sec() > before.sec() || (now.sec() == before.sec() && now.nsec() >= before.nsec()));
```
gasmith, line 292: Hm, maybe we should implement `PartialOrd`...

rust/foxglove/src/schemas_wkt.rs
```diff
@@ -339,6 +339,12 @@ impl Timestamp {
         Self::new_checked(sec, nsec).unwrap()
     }
 
+    /// Returns the current timestamp using SystemTime
```
gasmith, line 342:
```suggestion
    /// Returns the current timestamp using [`SystemTime::now`][std::time::SystemTime::now].
```

rust/foxglove/src/schemas_wkt/tests.rs
eloff: Yeah, seems like a good idea

gasmith (approved):

PR #334, opened 2025-04-01
rust: Add Channel::close

gasmith:
### Changelog
- rust: Add `Channel::close`

### Description
As part of making `Context` public, I realized that we probably don't want to expose `Context::remove_channel_for_topic()`. We should really be removing channels by ID, to avoid TOCTOU errors. We should expose this functionality publicly as `Channel::close()`, to mirror the python `Channel.close`.

While I was doing that, I realized that [my fix for python's `Channel.close`](https://github.com/foxglove/foxglove-sdk/pull/329) should've also removed a lie in the docstrings ("Destroying all references to the channel will also close it") and completely missed updating the generated typed channels. So this change fakes care of those things too.

The python SDK issues a (debug-level) warning when attempting to log on a channel that was explicitly closed. In this change, we move that warning down to rust, so that it's common. With a few changes:
 - The message is now issued at _warning_ level, because we don't want users to miss it.
 - Add a per-channel warning throttler with a 5s interval.
 - Handle the case where a channel is implicitly closed because the context is dropped.
 - Python no longer has to wrap channels in Option<>.

Fixes: FG-11010
Fixes: FG-10877

### DIFF
--- a/rust/foxglove/src/channel/raw_channel.rs
+++ b/rust/foxglove/src/channel/raw_channel.rs
@@ -1,14 +1,22 @@
 //! A raw channel.
 
 use std::collections::BTreeMap;
-use std::sync::atomic::AtomicU32;
-use std::sync::atomic::Ordering::Relaxed;
+use std::sync::atomic::Ordering::{Acquire, Relaxed, Release};
+use std::sync::atomic::{AtomicBool, AtomicU32};
 use std::sync::{Arc, Weak};
+use std::time::Duration;
+
+use parking_lot::Mutex;
+use tracing::warn;
 
 use super::ChannelId;
 use crate::log_sink_set::LogSinkSet;
 use crate::sink::SmallSinkVec;
+use crate::throttler::Throttler;
 use crate::{nanoseconds_since_epoch, Context, Metadata, PartialMetadata, Schema};
+
+/// Interval for throttled warnings.
+static WARN_THROTTLER_INTERVAL: Duration = Duration::from_secs(10);
 
 /// A log channel that can be used to log binary messages.
 ///
@@ -32,6 +40,8 @@ pub struct RawChannel {
     metadata: BTreeMap<String, String>,
     message_sequence: AtomicU32,
     sinks: LogSinkSet,
+    closed: AtomicBool,
+    warn_throttler: Mutex<Throttler>,
 }
 
 impl RawChannel {
@@ -51,6 +61,8 @@ impl RawChannel {
             metadata,
             message_sequence: AtomicU32::new(1),
             sinks: LogSinkSet::new(),
+            closed: AtomicBool::new(false),
+            warn_throttler: Mutex::new(Throttler::new(WARN_THROTTLER_INTERVAL)),
         })
     }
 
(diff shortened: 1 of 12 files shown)


rust/foxglove/src/channel/raw_channel.rs
```diff
@@ -81,16 +96,49 @@ impl RawChannel {
...
-    pub(crate) fn update_sinks(&self, sinks: SmallSinkVec) {
-        self.sinks.store(sinks);
+    /// Closes the channel, removing it from the context.
+    ///
+    /// You can use this to explicitly unadvertise the channel to sinks that subscribe to channels
+    /// dynamically, such as the [`WebSocketServer`][crate::WebSocketServer].
+    ///
+    /// Attempts to log on a closed channel will elicit a throttled warning message.
+    pub fn close(&self) {
+        if !self.is_closed() {
```
gasmith, line 106: We could make this a `compare_exchange`, but `Context::remove_channel` has its own synchronization, and I figured it was better to flip the `closed` boolean in just one place, rather than two.

rust/foxglove/src/context.rs
```diff
@@ -180,13 +188,13 @@ impl Context {
...
         self.0.write().add_channel(channel)
     }
 
-    /// Removes the channel for the specified topic.
-    pub fn remove_channel_for_topic(&self, topic: &str) -> bool {
-        self.0.write().remove_channel_for_topic(topic)
+    /// Removes a channel from the context.
+    pub(crate) fn remove_channel(&self, channel_id: ChannelId) -> bool {
+        self.0.write().remove_channel(channel_id)
     }
```
gasmith, line 198: I chose to restrict these to `pub(crate)` to ensure that the Channel's `context` linkage is always correct. Maybe it's worth trying to express that more concretely with types, but I didn't want to get too carried away.

For now, the only way to add a channel to a context is by constructing it (via the `ChannelBuilder`). You can remove a channel either by calling `Channel::close`, or by dropping the context entirely.

rust/foxglove/src/context.rs
```diff
@@ -197,25 +205,29 @@ impl Context {
     /// present and future channels on the context. Otherwise, the sink is expected to manage its
     /// subscriptions dynamically with [`Context::subscribe_channels`] and
     /// [`Context::unsubscribe_channels`].
+    #[doc(hidden)] // Hidden until Sink is public
```
gasmith, line 208: Whoops, these bled over from my work on making Context public.

rust/foxglove/src/throttler.rs
```diff
@@ -0,0 +1,46 @@
...
+    /// Create a new throttler with the specified duration.
+    pub const fn new(interval: Duration) -> Self {
+        Self {
+            interval,
+            next_at: None,
+        }
+    }
+
+    /// Returns true if the action is allowed now, and updates the internal timestamp.
+    pub fn try_acquire(&mut self) -> bool {
```
eloff, line 19: This is nice

rust/foxglove/src/context.rs
eloff: This should be a comment right here in the code

rust/foxglove/src/channel/raw_channel.rs
eloff: That seems like the best choice

eloff (approved):
This looks great

PR #335, opened 2025-04-02
rust: Make Context public, add LazyContext

gasmith:
### Changelog
- rust: Add Context and LazyContext

### Description
This change unhides `Context` from the public API. The bulk of the change is:
- Updating documentation to describe contexts.
- Add shortcut methods for building channels & sinks from a context.
- Add a `LazyContext` for use with `LazyChannel`.

Fixes: FG-10885
Fixes: FG-10854

### DIFF
--- a/rust/foxglove/src/context/lazy_context.rs
+++ b/rust/foxglove/src/context/lazy_context.rs
@@ -1,0 +1,66 @@
+use std::ops::Deref;
+use std::sync::{Arc, LazyLock};
+
+use crate::{Encode, LazyChannel, LazyRawChannel};
+
+use super::Context;
+
+static DEFAULT_CONTEXT: LazyContext = LazyContext::new();
+
+/// A context that is initialized lazily upon first use.
+///
+/// This is intended to be used with [`LazyChannel`][crate::LazyChannel] to create static channels
+/// attached to static contexts.
+///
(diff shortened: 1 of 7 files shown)


rust/foxglove/src/context/lazy_context.rs
```diff
@@ -0,0 +1,66 @@
...
+    pub const fn channel<T: Encode>(&'static self, topic: &'static str) -> LazyChannel<T> {
+        LazyChannel::new(topic).context(self)
+    }
+
+    /// Creates a new lazily-initialized raw channel in this context.
+    pub const fn raw_channel(
+        &'static self,
+        topic: &'static str,
+        message_encoding: &'static str,
+    ) -> LazyRawChannel {
```
eloff, line 55: Should this return a builder? There's no way to set a schema via this interface.

I guess one could use the builder directly if one needs that.

gasmith: The `LazyRawChannel` is actually a builder, because it's lazy. You can do:

```rust
static CTX: LazyContext = LazyContext::new();
static CH: LazyRawChannel =
    CTX.raw_channel("/topic", "json")
        .schema("topic-schema", "jsonschema", b"{...}");
```

PR #336, opened 2025-04-03
Enable sccache again now that issue has been fixed upstream

eloff:
Now that sccache 0.10.0 is published to pip, we can enable the cache again


eloff:
Honestly, the build time seems slightly better without the cache than with it. I think we just leave it off.

PR #338, opened 2025-04-03
Call `on_unsubscribe` when client disconnects

bryfox:
### Changelog

When a WS client disconnects, a server listener will receive an `on_unsubscribe` event for all subscribed channels.

### Description

A Foxglove (WS) client may disconnect without explicitly unsubscribing; a server listener should be made aware of this via the `on_unsubscribe` callback.

### DIFF
--- a/rust/foxglove/src/websocket/tests.rs
+++ b/rust/foxglove/src/websocket/tests.rs
@@ -504,6 +504,62 @@ async fn test_log_only_to_subscribers() {
 
 #[traced_test]
 #[tokio::test]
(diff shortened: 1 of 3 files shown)


rust/foxglove/src/websocket/tests.rs
```diff
@@ -502,6 +502,62 @@ async fn test_log_only_to_subscribers() {
     server.stop().await;
 }
 
+#[traced_test]
```
gasmith, line 505: Probably doesn't need to be a traced test

gasmith (approved):

PR #342, opened 2025-04-04
Move websocket support into a feature (enabled by default)

bryfox:
### Changelog

Rust: move websocket functionality and Tokio dependencies behind a "live_visualization" feature, enabled by default

### Description

This allows rust SDK users to remove the dependency on tokio if live viz support is not needed.

The `mcap` example has been updated to omit this feature, and CI now has a `build --no-default-features` step.


rust/foxglove/src/lib.rs
```diff
@@ -244,9 +246,9 @@
 //!
 //! # Requirements
 //!
-//! The Foxglove SDK depends on [tokio] as its async runtime with the `rt-multi-thread`
-//! feature enabled. Refer to the tokio documentation for more information about how to configure
-//! your application to use tokio.
+//! With the `live_visualization` feature (enabled by default), the Foxglove SDK depends on [tokio]
+//! as its async runtime with the `rt-multi-thread` feature enabled. Refer to the tokio
+//! documentation for more information about how to configure your application to use tokio.
```
gasmith, line 251: We should add a separate top-level section called `Feature flags`, with a  list of features that we exposes in bullet-point form. For example, [here's how tokio does it](https://docs.rs/tokio/latest/tokio/#feature-flags).

Incidentally, I think we can relax the wording around tokio. It's totally fine to bring your own tokio runtime with whatever config you like. What we should say is that _if you choose to use the "blocking" version of the websocket server_, we'll spawn our own multi-threaded runtime. This documentation probably belongs best with the blocking websocket server itself, rather than here.

gasmith (approved):
Love it!

gasmith (approved):

PR #343, opened 2025-04-04
rust: Add foxglove-ws-protocol

gasmith:


rust/foxglove-ws-protocol/src/schema.rs
```diff
@@ -0,0 +1,117 @@
...
+    #[error("Schema data is not valid base64")]
+    NotBase64,
+    /// Missing schema or schema_encoding.
+    #[error("Missing schema or schema_encoding")]
+    MissingSchema,
+}
+
+/// Returns true if a schema is required for this message encoding.
+pub(crate) fn is_schema_required(message_encoding: &str) -> bool {
+    message_encoding == "flatbuffer"
```
bryfox, line 31: This is fairly specific to the Foxglove app; we decided to 'default to true' for hypothetical future encodings. I don't think there's anything to change, but something to consider if we're publishing a crate.

.github/workflows/ci.yml
```diff
@@ -33,10 +33,16 @@ jobs:
       # Validate that we can build against the MSRV (minimum specified rust version).
       - run: cargo +1.83.0 build -p foxglove --verbose
       - run: cargo clippy --no-deps --all-targets --tests -- -D warnings
+      - run: cargo +nightly rustdoc -p foxglove-ws-protocol --all-features -- -D warnings --cfg docsrs
       - run: cargo +nightly rustdoc -p foxglove --all-features -- -D warnings --cfg docsrs
       - run: cargo test --features unstable --verbose
         timeout-minutes: 10
+      - run: cargo publish --package foxglove-ws-protocol --dry-run
       - run: cargo publish --package foxglove --dry-run
+      - run: cargo publish --package foxglove-ws-protocol
```
bryfox, line 42: Assuming SDK will want to depend on the same version of ws-protocol, we'll need to figure out some step here to wait for this to be published? We can punt on that for now, but something to consider.

bryfox (commented):
Ongoing discussion elsewhere about whether to publish a crate for this, but the changes look great.

Should we consider putting the client implementation behind a feature, so that it doesn't need to be included? (e.g., SDK does not need any of the client functionality today outside of tests.)

rust/foxglove-ws-protocol/src/client.rs
```diff
@@ -0,0 +1,154 @@
...
+    /// Parses a client message from JSON.
+    pub fn parse_json(json: &'a str) -> Result<Self, ParseError> {
+        let msg = serde_json::from_str::<JsonMessage>(json)?;
+        Ok(msg.into())
+    }
+
+    /// Parses a client message from a binary buffer.
+    pub fn parse_binary(mut data: &'a [u8]) -> Result<Self, ParseError> {
+        if data.is_empty() {
+            Err(ParseError::EmptyBinaryMessage)
```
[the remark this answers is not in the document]
gasmith: Right, we'll update the callsite, so that the behavior doesn't change.

rust/foxglove-ws-protocol/src/client/advertise.rs
```diff
@@ -0,0 +1,199 @@
...
+        ChannelBuilder {
+            id,
+            topic: topic.into(),
+            encoding: encoding.into(),
+            schema: None,
+        }
+    }
+
+    /// Returns the decoded schema data.
+    pub fn schema(&self) -> Result<Vec<u8>, crate::schema::DecodeError> {
```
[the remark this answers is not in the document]
gasmith: Good idea; let's call it `decode_schema()`. Applied the same change to server::advertise::Channel.

rust/foxglove-ws-protocol/src/schema.rs
gasmith: Yeah, I was wondering about that too. Less of an issue now that we're keeping it internal.

gasmith (commented):
> Ongoing discussion elsewhere about whether to publish a crate for this, but the changes look great.

Sounds like the consensus is to move it inside, so I've done that.

> Should we consider putting the client implementation behind a feature, so that it doesn't need to be included? (e.g., SDK does not need any of the client functionality today outside of tests.)

It's a bit hard to separate that out, with the way that it's written, but I don't think it would be worth it.

We _do_ depend on client message types for parsing, we just don't need the constructors or serialization support. And likewise, we don't need deserialization support for server messages. We could wall that off with `cfg` and `cfg_attr` attributes, but I think it'd be quite a mess.

All that dead code will be omitted from the final binary anyway. Feature flags are better for controlling transitive dependencies, which in this case are the same for both client and server usage patterns.

rust/foxglove/Cargo.toml
```diff
@@ -49,6 +56,7 @@ clap = { version = "4.5", features = ["derive"] }
 env_logger = "0.11.5"
 futures-util = "0.3.31"
 insta = { version = "1.42.2", features = ["json"] }
+maplit = "1.0.2"
```
bryfox, line 59: note we have a `collection!` macro which might take care of these use cases

rust/foxglove/src/websocket.rs
```diff
@@ -43,6 +43,8 @@ use service::{CallId, Service, ServiceId, ServiceMap};
 mod tests;
 #[cfg(all(test, feature = "unstable"))]
 mod unstable_tests;
+#[allow(dead_code)]
```
bryfox, line 46: Is this temporary, or is there a reasonable way to narrow this down to just the parts we don't need?

bryfox (approved):
👍

rust/foxglove/src/websocket.rs
gasmith: This is just temporary until the next PR where we switch over.

rust/foxglove/Cargo.toml
gasmith: Ah, yeah, will use that.

gasmith: I guess I'm inclined to leave it as maplit, so that we can easily separate this out into a separate crate later without untangling dependencies on the rest of the foxglove crate.

Does `collection!` provide any additional benefits over maplit? Maybe we switch everything to maplit?

bryfox: >  I'm inclined to leave it as maplit, so that we can easily separate this out into a separate crate later

Makes sense.

> Does collection! provide any additional benefits over maplit? Maybe we switch everything to maplit?

I don't think so, assuming we like the explicit macro names. I'm in favor of using the same thing everywhere, rather than someday trying to remember why we're inconsistent if we don't end up separating that crate.

gasmith: OK, I'll switch the remaining collection! users to to maplit in a separate PR.

PR #347, opened 2025-04-07
Fail WS handshake if client doesn't request any subprotocol

bryfox:
### Changelog
None

### Description

This updates/fixes the WS server handling of handshakes in the failure cases. If a client doesn't request the supported subprotocol (or doesn't include the protocol header at all), the upgrade fails and the connection is closed with a 400.

Previously, we relied on well-behaved clients to close the connection when their requested protocol wasn't supported, and we left the connection open if the client didn't request any protocol.


gasmith (approved):
LG

PR #348, opened 2025-04-07
rust: Refactor websocket connection graph management

gasmith:
### Description
This is another change towards separating the websocket server from the connected client. Just as with parameters, we don't really need to maintain state in two places, and the server is the natural location because (a) the connection graph state is shared, and (b) the server needs to track how many clients are subscribed in order to deliver unsubscribe callbacks.

This change also fixes a bug: we weren't removing a client's graph subscription on disconnect. I added unit test coverage for this case (c2 disconnects without sending unsubscribe).

It was convenient to move the list of subscribers into the `ConnectionGraph` itself, so that it'd be under the server's mutex without having to introduce a new struct. A little wasteful to have an extra hashset in there, but it's not like this weird corner of the SDK is going to get a lot of use anyway.

I also removed [the "take-and-update" approach](https://github.com/foxglove/foxglove-sdk/pull/348/files#diff-a7f22dab80c06b17a93e48fe59f89fad848400d6a7fd59c40a281af78504dac6L783-L787) to building the initial `ConnectionGraphUpdate`, in favor of adding a `ConnectionGraph::diff<'a>(&self, other: &'a ConnectionGraph) -> ConnectionGraphDiff<'a>` that we can use more generally. It's less performant, but I think the readability is worth it.

### DIFF
--- a/rust/foxglove/src/testutil.rs
+++ b/rust/foxglove/src/testutil.rs
@@ -1,5 +1,6 @@
 //! Test utilities.
 
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::time::Duration;
 
 use parking_lot::Mutex;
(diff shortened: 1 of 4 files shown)


eloff (approved):
LGTM

Sacrifices performance for readability in some places, but that seems like a reasonable tradeoff.

PR #349, opened 2025-04-07
python: allow logging on schemaless channels

bryfox:
### Changelog
Python: 'schema' and 'message_encoding' are now both optional when defining a channel.

### Description

One stated use case for the SDK is that a user should be able to log messages for custom schemas without needing to care about the underlying encoding or serialization. This enables faster development for exploratory use cases, and makes it easier to test out ideas in the app.

In the Rust SDK, we support this use case by providing the `Encode` trait for custom structs that implement [JsonSchema](https://docs.rs/schemars/latest/schemars/trait.JsonSchema.html), which can be used with typed channels. We've discussed doing something similar in the Python SDK; perhaps annotations that allow us to dynamically produce schemas with minimal effort from the user. However, this feels less pythonic to me than allowing users to log dicts and taking care of the encoding for them. We took a similar stance with the "high-level" API — a user logs their data to [untyped] channels. The SDK is already partway there; we support this behavior on channels if you explicitly choose "json" encoding, or provide an "any object" schema. And the app supports generally schemaless logging using JSON.

If we move forward with some Rust solution that allows logging of custom schemas with binary protocols (#318), then we can consider something like that for other languages as well.

### DIFF
--- a/python/foxglove-sdk-examples/live-visualization/main.py
+++ b/python/foxglove-sdk-examples/live-visualization/main.py
@@ -144,6 +144,7 @@ def main() -> None:
     )
 
     # If you want to use JSON encoding, you can also specify the schema and log messages as dicts.
+    # Dicts can also be logged without specifying a schema.
     json_chan = Channel(topic="/json", schema=plot_schema)
 
     img_chan = RawImageChannel(topic="/image")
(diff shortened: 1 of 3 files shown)


eloff (approved):
LGTM

We probably need to support something similar in rust

gasmith (approved):

PR #350, opened 2025-04-07
C/C++ error handling

eloff:
Adds error returns and error details to the C APIs which can fail.

This is a rework of the error handling based on feedback from James. Instead of returning error strings to C, we can return an error type enum (which we just map directly from the enum members in FoxgloveError in the crate root.)

The reason is we don't really need the full string errors. An error code is enough, plus we log the full string message. The user can debug based on the error code, and if that's not enough, enable the logs and check them. It's not quite as convenient for debugging, but it's a lot more convenient for library usage and we don't have to allocate and free errors on the heap.

Additionally the C++ API was changed to modern C++ practice of using Result-like return values instead of exceptions. This also means constructors had to be replaced with static create methods instead, so that it can return the result. For this I used tl::expected which is an implementation of std::expected in C++ 23. https://en.cppreference.com/w/cpp/utility/expected

One annoying thing is our error enum is duplicated as a C enum and then duplicated again as a c++ enum. A bit of a pain to update correctly, so I opened a ticket to generate these instead so they don't get out of sync.

I added some tests to exercise the error handling in the C++ API.

One change I made is to remove the foxglove_server_free and foxglove_mcap_free functions, preferring to use just the stop/close methods. There's nothing one can do with a closed/stopped server or mcap writer, so there's no point separating stop/close from free.


c/src/lib.rs
```diff
@@ -251,53 +283,107 @@ impl FoxgloveMcapOptions {
...
         std::slice::from_raw_parts(options.path as *const u8, options.path_len)
     })
-    .expect("path is invalid");
+    .map_err(|e| foxglove::FoxgloveError::ValueError(format!("path is invalid: {}", e)))?;
 
     // Safety: this is safe if the options struct contains valid strings
-    let mcap_options = unsafe { options.to_write_options() };
+    let mcap_options = unsafe { options.to_write_options() }?;
+
+    println!("create, truncate: {}, {}", options.create, options.truncate);
```
bryfox, line 328: 🔪

c/src/lib.rs
```diff
@@ -251,53 +283,107 @@ impl FoxgloveMcapOptions {
...
+        if options.truncate {
+            file_options.create(true).truncate(true);
+        } else {
+            file_options.create_new(true);
+        }
+    } else if options.truncate {
+        file_options.truncate(true);
+    } else {
+        // Append doesn't make sense with mcap
+        return Err(foxglove::FoxgloveError::IoError(std::io::Error::new(
```
bryfox, line 342: How much benefit do you see to supporting create=false & truncate=true (requiring that a file already exist)? The downside is that these options open the possibility for this unsupported case.

If it's a net benefit, should python have a similar interface? Right now, it only provides the equivalent of a truncate option.

bryfox (commented):
Seems good to me overall; I think the error pointer approach is straightforward.

c/src/lib.rs
eloff: That's a good point, I don't think that combination makes sense.

I think just a truncate option is enough, which just describes whether we error or truncate an existing file.

Create can be assumed to always be true.

cpp/examples/mcap/src/main.cpp
```diff
@@ -4,7 +4,6 @@
 int main(int argc, const char* argv[]) {
   foxglove::McapWriterOptions options = {};
   options.path = "test.mcap";
-  options.create = true;
```
jtbandes, line 7: is this just for testing?

c/src/lib.rs
```diff
@@ -533,3 +664,97 @@ impl foxglove::websocket::ServerListener for FoxgloveServerCallbacks {
...
+    magic: u32,
+    pub message: *const c_char,
+}
+
+impl From<foxglove::FoxgloveError> for FoxgloveError {
+    fn from(error: foxglove::FoxgloveError) -> Self {
+        let message = CString::new(error.to_string()).unwrap();
+        Self {
+            // This gives a little protection against people passing an unintialized
+            // FoxgloveError into foxglove_error_free, which is an easy mistake to make in C.
```
jtbandes, line 731: what makes this mistake easy - is it that we never assign the error if no error was encountered? What if we just assigned a null message (i.e. initialize it) every time there was no error?

c/src/lib.rs
```diff
@@ -533,3 +664,97 @@ impl foxglove::websocket::ServerListener for FoxgloveServerCallbacks {
...
+            magic: FOXGLOVE_ERROR_MAGIC,
+            kind: FoxgloveErrorKind::from(&error),
+            message: message.into_raw(),
+        }
+    }
+}
+
+/// Free an error returned from a call to a foxglove.
```
jtbandes, line 739: typo: "to a foxglove"

cpp/foxglove/include/foxglove/error.hpp
```diff
@@ -0,0 +1,38 @@
+
```
jtbandes, line 1:
```suggestion
#pragma once
```

cpp/foxglove/include/foxglove/error.hpp
```diff
@@ -0,0 +1,38 @@
...
+  DuplicateChannel,
+  DuplicateService,
+  MissingRequestEncoding,
+  ServicesNotSupported,
+  ConnectionGraphNotSupported,
+  IoError,
+  McapError,
+};
+
+class FoxgloveError : public std::exception {
```
jtbandes, line 26: I don't think this really matters much, just a habit I've ended up with to avoid surprises
```suggestion
class FoxgloveError : public std::exception final {
```

cpp/foxglove/src/error.cpp
```diff
@@ -0,0 +1,51 @@
+#include <foxglove-c/foxglove-c.h>
+#include <foxglove/error.hpp>
+
+namespace foxglove {
+
+FoxgloveError::FoxgloveError(const foxglove_error&& error)
+    : std::exception()
+    , _impl(new foxglove_error(std::move(error))) {}
```
jtbandes, line 8: nit: prefer `make_unique`

cpp/foxglove/src/channel.cpp
```diff
@@ -14,9 +15,13 @@ Channel::Channel(
...
     cSchema.data_len = schema->dataLen;
   }
-  _impl.reset(
-    foxglove_channel_create(topic.data(), messageEncoding.data(), schema ? &cSchema : nullptr)
-  );
+  foxglove_error cError = {};
+  _impl.reset(foxglove_channel_create(
+    topic.data(), messageEncoding.data(), schema ? &cSchema : nullptr, &cError
+  ));
+  if (_impl.get() == nullptr) {
```
jtbandes, line 22: nit: [can be shortened](https://en.cppreference.com/w/cpp/memory/unique_ptr/operator_bool) (applies to a few other places too)
```suggestion
  if (!_impl) {
```

cpp/foxglove/src/error.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+FoxgloveError::~FoxgloveError() {
+  foxglove_error_free(_impl.get());
```
jtbandes, line 11: this threw me off at first - but I now understand that the free function is freeing the message and has nothing to do with deleting the `new`'d memory from the constructor (the unique_ptr takes care of that). Could be worth some comments

cpp/foxglove/include/foxglove/error.hpp
```diff
@@ -0,0 +1,38 @@
...
+public:
+  FoxgloveError(const foxglove_error&& error);
```
jtbandes, line 28: also not a big deal but another surprise-avoiding habit (only matters for constructors with 1 argument iirc)
```suggestion
  explicit FoxgloveError(const foxglove_error&& error);
```

jtbandes: Separately, this line caught my eye as `const foo&&` is not commonly used.....but reading more about it, I think you could argue that it's correct/defensible to use here https://www.sandordargo.com/blog/2021/08/18/const-rvalue-references

jtbandes (approved):
Seems like a good approach; I'll stamp so as to unblock merging whenever you feel it's ready.

I don't necessarily have opinions on this, but wanted to point out that C++ does have `error_code`/`error_condition` types that might allow you to keep the C++ API more aligned with how the C API works. But exceptions have their benefits too (not cluttering up the function signatures). If some day someone asks us for `-fno-exceptions` support, then...well I don't want to think about that 😄 

I also think it's worth asking how important it really is to expose the error kind – or if just a message would suffice for users?

c/cbindgen.toml
```diff
@@ -21,24 +21,24 @@ after_includes = """
...
-
-typedef struct foxglove_channel foxglove_channel;
 """
 
 [ptr]
 non_null_attribute = "FOXGLOVE_NONNULL"
 
 [enum]
-prefix_with_name = true
+rename_variants = "QUALIFIED_SCREAMING_SNAKE_CASE"
```
eloff, line 30: FOXGLOVE_ERROR_VALUE_ERROR vs foxglove_error_Value_Error, seems much better to me and also fits the google style guide

c/cbindgen.toml
```diff
@@ -21,24 +21,24 @@ after_includes = """
...
 
+prefix_with_name = false
 [export.rename]
 FoxgloveWebSocketServer = "foxglove_websocket_server"
 FoxgloveChannel = "foxglove_channel"
 FoxgloveSchema = "foxglove_schema"
 FoxgloveServerCallbacks = "foxglove_server_callbacks"
 FoxgloveServerOptions = "foxglove_server_options"
 FoxgloveMcapOptions = "foxglove_mcap_options"
+FoxgloveMcapCompression = "foxglove_mcap_compression"
```
eloff, line 40: I guess I left this out in my earlier PR

c/include/foxglove-c/foxglove-c.h
```diff
@@ -19,24 +18,20 @@
...
 
-
-typedef struct foxglove_channel foxglove_channel;
-
-
 /**
  * Allow clients to advertise channels to send data messages to the server.
  */
-#define FOXGLOVE_SERVER_CAPABILITY_CLIENT_PUBLISH (1 << 0)
+#define FOXGLOVE_SERVER_CAPABILITY_CLIENT_PUBLISH 1
```
eloff, line 24: I was getting some linter errors about bitwise ops on signed integers, I didn't figure out a fix other than just hard-coding them

cpp/examples/ws-server/src/main.cpp
```diff
@@ -57,7 +57,12 @@ int main(int argc, const char* argv[]) {
   options.callbacks.onClientUnadvertise = [](uint32_t clientId, uint32_t clientChannelId) {
     std::cerr << "Client " << clientId << " unadvertised channel " << clientChannelId << '\n';
   };
-  foxglove::WebSocketServer server{options};
+  auto serverResult = foxglove::WebSocketServer::create(std::move(options));
```
eloff, line 60: This seems to be a canonical way to create class instances if the constructor can fail and return an expected result. But it's new to me, so I can't be certain.

c/include/foxglove-c/foxglove-c.h
bryfox: It looks like the tidy config rule that you commented out is formatted incorrectly — I get an error about the config itself.

This seems to work for me (and allows the bitshift when I run `clang-tidy`):

```
CheckOptions:
  hicpp-signed-bitwise.IgnorePositiveIntegerLiterals: true
```

cpp/CMakeLists.txt
```diff
@@ -57,6 +57,13 @@ FetchContent_Declare(asio
 )
 FetchContent_MakeAvailable(asio)
 
+FetchContent_Declare(expected
+  GIT_REPOSITORY https://github.com/TartanLlama/expected.git
```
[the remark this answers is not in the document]
eloff: Yes, I'll add a ticket for that

cpp/foxglove/include/foxglove/mcap.hpp
```diff
@@ -3,11 +3,13 @@
...
 #include <string>
 
+enum foxglove_error : uint8_t;
+enum class FoxgloveError : uint8_t;
 struct foxglove_mcap_writer;
 
 namespace foxglove {
 
-enum class McapCompression {
+enum McapCompression : uint8_t {
```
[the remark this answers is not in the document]
eloff: Ah, now I remember why. cbindgen doesn't output enums with the class keyword, and this comes from C, so this forward declaration has to match exactly.

cpp/CMakeLists.txt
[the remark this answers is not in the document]
eloff: FG-11244

cpp/.clang-tidy
```diff
@@ -19,8 +19,7 @@ Checks:
...
   - -readability-magic-numbers
-
-CheckOptions:
-  hicpp-signed-bitwise:
-    # See discussion at https://stackoverflow.com/a/58845898/23649 & https://stackoverflow.com/a/78343833/23649
-    IgnorePositiveIntegerLiterals: true
+# CheckOptions:
+#   hicpp-signed-bitwise:
+#     # See discussion at https://stackoverflow.com/a/58845898/23649 & https://stackoverflow.com/a/78343833/23649
+#     IgnorePositiveIntegerLiterals: true
```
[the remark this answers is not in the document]
bryfox: Yeah, hard-coded seems fine for now. FWIW, I this worked in llvm version 18.1.3 (ubuntu) and 20.1.3 (mac).

cpp/foxglove/include/foxglove/mcap.hpp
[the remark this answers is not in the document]
bryfox: Looks like there's an option for it?

https://github.com/mozilla/cbindgen/blob/b105f550ed0d62e9207cb7da694ca8b988f32f18/docs.md?plain=1#L909-L911

Not a big deal, but if we can't use that, I wonder if we should rename the fields since they get hoisted up to the foxglove namespace (I think).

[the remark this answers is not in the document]
eloff: Oh nice! I'll try it

c/src/lib.rs
```diff
@@ -251,111 +325,94 @@ impl FoxgloveMcapOptions {
...
-#[unsafe(no_mangle)]
-pub extern "C" fn foxglove_server_get_port(server: Option<&FoxgloveWebSocketServer>) -> u16 {
-    let Some(server) = server else {
-        panic!("Expected a non-null server");
+        tracing::error!("foxglove_mcap_close called with null writer");
+        return FoxgloveError::ValueError;
     };
-    let Some(ref handle) = server.0 else {
-        panic!("Server already stopped");
+    let Some(writer) = writer.take() else {
```
gasmith, line 401:
```suggestion
    // Safety: undo the Box::into_raw from foxglove_mcap_open
    let writer = unsafe { Box::from_raw(writer) };
    let Some(writer) = writer.take() else {
```

c/src/lib.rs
```diff
@@ -251,111 +325,94 @@ impl FoxgloveMcapOptions {
...
-        .expect("Failed to open file");
-    let writer = foxglove::McapWriter::with_options(mcap_options)
-        .create(BufWriter::new(file))
-        .expect("Failed to create writer");
-    Box::into_raw(Box::new(FoxgloveMcapWriter(Some(writer))))
+        .map_err(foxglove::FoxgloveError::IoError)?;
+
+    let writer = foxglove::McapWriter::with_options(mcap_options).create(BufWriter::new(file))?;
+    // We can avoid this double indirection if we refactor McapWriterHandle to move the context into the Arc
+    // and then add into_raw and from_raw methods like with the WebSocketServerHandle
```
gasmith, line 383: Perhaps you meant `RawChannel` instead of `WebSocketServerHandle`?

c/src/lib.rs
```diff
@@ -529,3 +629,101 @@ impl foxglove::websocket::ServerListener for FoxgloveServerCallbacks {
...
+            foxglove::FoxgloveError::MissingRequestEncoding(_) => {
+                FoxgloveError::MissingRequestEncoding
+            }
+            foxglove::FoxgloveError::ServicesNotSupported => FoxgloveError::ServicesNotSupported,
+            foxglove::FoxgloveError::ConnectionGraphNotSupported => {
+                FoxgloveError::ConnectionGraphNotSupported
+            }
+            foxglove::FoxgloveError::IoError(_) => FoxgloveError::IoError,
+            foxglove::FoxgloveError::McapError(_) => FoxgloveError::McapError,
+            _ => FoxgloveError::Unspecified,
```
gasmith, line 677: Hm, maybe we should turn off the `#[non_exhaustive]` on `foxglove::FoxgloveError` for this build (with a `cfg_attr`?), to ensure that this enum stays in sync.

cpp/foxglove/include/foxglove/server.hpp
```diff
@@ -69,16 +72,20 @@ struct WebSocketServerOptions {
...
 
 private:
-  WebSocketServerCallbacks _callbacks;
-  std::unique_ptr<foxglove_websocket_server, void (*)(foxglove_websocket_server*)> _impl;
+  WebSocketServer(
+    foxglove_websocket_server* server, std::unique_ptr<WebSocketServerCallbacks> callbacks
+  );
+
+  std::unique_ptr<WebSocketServerCallbacks> _callbacks;
+  std::unique_ptr<foxglove_websocket_server, foxglove_error (*)(foxglove_websocket_server*)> _impl;
```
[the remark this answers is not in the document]
gasmith: Seems like a legit suggestion?

cpp/foxglove/include/foxglove/mcap.hpp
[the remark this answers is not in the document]
eloff: I already tried that one, for whatever reason it doesn't work the way we think it should.

It'd be nice to fix that ourselves after the header is generated, because otherwise you do get global namespace pollution from the enum values, among other things. If we did that we wouldn't need  to duplicate the enum in the C++ header files to avoid that.

I added a ticket for us to fix this: FG-11215

gasmith (approved):

c/src/lib.rs
eloff: Good catch

c/src/lib.rs
eloff: The truth is stranger, I added those methods to the WebSocketServerHandle and eliminated the Boxed Arc entirely, just returning the Arc pointer instead. But it's kind of ugly since those have to be pub doc(hidden) methods, and the double alloc and double indirection doesn't matter at all for performance, so I backed it out and didn't update this comment.

c/src/lib.rs
eloff: I don't know how to do that, would you add a feature that's off by default but turned on when including foxglove as a dependency in the C crate?

cpp/foxglove/include/foxglove/server.hpp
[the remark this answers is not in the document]
eloff: Either way the return value is ignored here, so I don't think it matters

c/src/lib.rs
gasmith: Yeah, something like that. It'd be pretty ugly. I guess it's a moot point, though, if you're planning to codegen these enums anyway. :)

c/include/foxglove-c/foxglove-c.h
eloff: works for me too, thanks!

cpp/foxglove/include/foxglove/error.hpp
eloff: It was muscle memory for me, maybe incorrectly. I'll have to read the blog post and think it over

PR #353, opened 2025-04-08
rust: Minor doc cleanup

gasmith:
I spotted a few minor improvements we could make while looking at the published 0.5.0 docs.


eloff (approved):
LGTM

PR #354, opened 2025-04-08
rust: Fix build for --no-default-features

gasmith:
### Description
CI was building the whole workspace with `--no-default-features`, which means that the build was picking up all the feature flags from the examples when building the foxglove crate. This change fixes CI to build the foxglove crate alone with the `--no-default-features` flag.


bryfox (approved):
Good catch, thanks

PR #355, opened 2025-04-08
rust: Use new ws protocol implementation

gasmith:
### Description
In this change, we switch over to using the new ws protocol implementation. The change is fairly mechanical, and shouldn't introduce any functional changes.

I used the opportunity to split a few things from `websocket::protocol` into `websocket::X`. These are basically just newtypes and conversions.

The tests shrank in size, perform stricter validation, and I find them a little more readable. Not quite the eye-watering result I was hoping for, but I think it's still a reasonable improvement.


rust/foxglove/src/websocket.rs
```diff
@@ -466,12 +451,22 @@ impl ConnectedClient {
...
             self.send_error("Server does not support clientPublish capability".to_string());
             return;
         }
 
+        let channels: Vec<_> = message
+            .channels
+            .into_iter()
+            .filter_map(|c| {
+                ClientChannel::try_from(c)
+                    .inspect_err(|e| warn!("Failed to parse advertised channel: {e:?}"))
```
bryfox, line 465: nit: we have `tracing::` qualifiers pretty much everywhere else in this file; maybe we can standardize now. (Is there a clippy rule to help here?)

rust/foxglove/src/websocket/channel.rs
```diff
@@ -0,0 +1,86 @@
+use super::ws_protocol::{client, schema, server};
```
bryfox, line 1: I'm a little lost in this file, but maybe it's just me. There's a lot of code for ClientChannel, which is distinct from server advertisements, raw channels, etc. And also conversions between things that live elsewhere (`server::advertise::Channel` / `RawChannel`; schemas).

bryfox (approved):
Minor comments, but 👍

rust/foxglove/src/websocket.rs
gasmith: No, no clippy lint for that, AFAIK.

rust/foxglove/src/websocket/channel.rs
gasmith: Fair point. I can split this up a bit more. The bulk of this file is related to `ClientChannel` and `ClientChannelId`, and conversions from client advertisements. So maybe this just becomes `client_channel.rs`. The last two conversions are for server advertisements, which belong elsewhere.

PR #356, opened 2025-04-08
rust: Replace collection! with maplit

gasmith:


bryfox (approved):
Do you prefer the explicit `maplit::` because it's a macro, or some other reason?

gasmith:
> Do you prefer the explicit `maplit::` because it's a macro, or some other reason?

I don't have a strong preference. I think this was right after you pointed out that we use the fully-qualified form for tracing macros, and so that's probably how I ended up here.

PR #358, opened 2025-04-08
Remove publish_time and sequence from metadata options

bryfox:
This proposes to remove `publish_time` and `sequence` from `Metadata`, which users can currently provide to `log()`.

#### Publish Time

`publish_time` is a message field in MCAP but does not have universal support in the Foxglove app. Some panels do support this (e.g. plot), but also support user-defined timestamp fields, where users are free to define their semantics. Data imported to a Foxglove primary site is ordered by log_time. The SDK should not include this only to support writing to an MCAP field; the MCAP spec dictates that log_time should be used if publish_time is unavailable.

#### Sequence

`sequence` is a message field in MCAP. It's useful for detecting missing, out-of-order, or discontinuous messages. The existing implementation is problematic because users are free to set sequence numbers in some log calls, but not others, which would result in an out-of-sequence stream with possible duplications. In ROS 2, `std_msgs/Header` [no longer includes a sequence](https://design.ros2.org/articles/changes.html) because it was inconsistent in ROS 1. Data streamed from a Foxglove primary site produces its own sequence numbers.

### Notes

For the Rust SDK, where users work with metadata structs directly, I've left `log_time` inside those structs. Arguably, it's less useful now that there's only a single field — there's no longer a reason for `::default`, and a future addition will likely mean manual changes for those users. Still, I think the migration would be easier, and the struct better communicates the intent.


rust/foxglove/src/channel/raw_channel.rs
```diff
@@ -172,15 +172,9 @@ impl RawChannel {
 
     /// Logs a message with additional metadata.
     pub(crate) fn log_to_sinks(&self, msg: &[u8], opts: PartialMetadata) {
-        let mut metadata = Metadata {
-            sequence: opts.sequence.unwrap_or_else(|| self.next_sequence()),
```
gasmith, line 176: We can get rid of the `next_sequence` method too.

gasmith (commented):
Mm, that's some lovely red.

Small comment about sequence numbering for mcap, otherwise LGTM.

rust/foxglove/src/mcap_writer/mcap_sink.rs
```diff
@@ -59,9 +59,10 @@ impl<W: Write + Seek> WriterState<W> {
             .write_to_known_channel(
                 &mcap::records::MessageHeader {
                     channel_id: mcap_channel_id,
-                    sequence: metadata.sequence,
+                    sequence: channel.next_sequence(),
```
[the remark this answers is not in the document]
bryfox: 👍 I think each mcap channel also needs its own sequence, so I added a map<ChannelId, AtomicU32>.

rust/foxglove/src/mcap_writer/mcap_sink.rs
```diff
@@ -6,22 +6,34 @@ use std::collections::hash_map::Entry;
...
             channel_map: HashMap::new(),
+            channel_sequence: HashMap::new(),
         }
     }
 
+    fn next_sequence(&mut self, channel_id: ChannelId) -> u32 {
+        self.channel_sequence
+            .entry(channel_id)
+            .or_insert_with(|| AtomicU32::new(1))
+            .fetch_add(1, Relaxed)
```
gasmith, line 34: Since we've got a `&mut self`, we don't need the atomic:

```rust
        *self.channel_sequence
            .entry(channel_id)
            .and_modify(|v| *v += 1)
            .or_insert(1)
```

gasmith (approved):
Minor nit. LG!

PR #360, opened 2025-04-09
rust: Add websocket connection poller

gasmith:
This should be the last step in separating the client from the server, so they can move into different modules.

The idea is to encapsulate the poll loop, so that the server doesn't need to care about its inner workings. It just calls an `async fn Poller::run(self)` which consumes self and returns when the client connection is closed.

While we're at it, we might as well move the whole `WebSocketStream` into the poller, too. That way, we can easily ensure that the connection is shutdown in the appropriate sequence - i.e., there are no races between the poller's send loop and [some external attempt to send a close message](https://github.com/foxglove/foxglove-sdk/pull/360/files#diff-a7f22dab80c06b17a93e48fe59f89fad848400d6a7fd59c40a281af78504dac6L896-L897). But to accomplish that, we need better signalling to indicate _why_ the connection is being closed. For that, we introduce a `ShutdownReason` enum and use a oneshot channel to deliver it to the poll loop.

A surprising upshot of this rearrangement is that `Server::stop` is no longer async, which has two big implications:
- We don't need `WebSocketBlockingHandle` anymore, because `stop()` was the only async method.
- We can implement `Drop` for `WebSocketServerHandle`, if we want to.

### DIFF
--- a/rust/foxglove/src/websocket.rs
+++ b/rust/foxglove/src/websocket.rs
@@ -141,6 +141,7 @@ pub(crate) struct ConnectedClient {
     weak_self: Weak<Self>,
     sink_id: SinkId,
     context: Weak<Context>,
+    poller: parking_lot::Mutex<Option<Poller>>,
     /// A cache of channels for `on_subscribe` and `on_unsubscribe` callbacks.
     channels: parking_lot::RwLock<HashMap<ChannelId, Arc<RawChannel>>>,
     data_plane_tx: flume::Sender<Message>,
@@ -243,7 +244,7 @@ impl ConnectedClient {
         websocket: WebSocketStream<TcpStream>,
         addr: SocketAddr,
         message_backlog_size: usize,
-    ) -> (Arc<Self>, Poller) {
+    ) -> Arc<Self> {
         let (data_plane_tx, data_plane_rx) = flume::bounded(message_backlog_size);
         let (control_plane_tx, control_plane_rx) = flume::bounded(message_backlog_size);
         let (shutdown_tx, shutdown_rx) = oneshot::channel();
@@ -253,6 +254,7 @@ impl ConnectedClient {
             weak_self: weak_self.clone(),
             sink_id: SinkId::next(),
             context: context.clone(),
+            poller: parking_lot::Mutex::default(),
             channels: parking_lot::RwLock::default(),
             data_plane_tx,
             data_plane_rx: data_plane_rx.clone(),
@@ -264,14 +266,14 @@ impl ConnectedClient {
             server: server.clone(),
             shutdown_tx: parking_lot::Mutex::new(Some(shutdown_tx)),
         });
-        let poller = Poller {
+        *client.poller.lock() = Some(Poller {
             client: client.clone(),
             websocket,
             data_plane_rx,
             control_plane_rx,
             shutdown_rx,
-        };
-        (client, poller)
+        });
+        client
     }
 
     fn id(&self) -> ClientId {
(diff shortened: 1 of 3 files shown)


rust/foxglove/src/websocket.rs
```diff
@@ -178,6 +292,17 @@ impl ConnectedClient {
         &self.weak_self
     }
 
+    fn addr(&self) -> SocketAddr {
+        self.addr
+    }
+
+    /// Shuts down the connection by signalling the [`Poller`] to exit.
+    fn shutdown(&self, reason: ShutdownReason) {
+        if let Some(shutdown_tx) = self.shutdown_tx.lock().take() {
```
bryfox, line 301: I assume the lock/take is to prevent this from being called twice. Does tokio::oneshot not guarantee that for us?

rust/foxglove/src/websocket.rs
```diff
@@ -881,7 +993,7 @@ impl Server {
         Ok(local_addr)
     }
 
-    pub async fn stop(&self) {
+    pub fn stop(&self) {
```
bryfox, line 996: Do you think we should release a patch version with some of the fixes & doc updates prior to the next breaking change?

bryfox (approved):
Nice!

rust/foxglove/src/websocket.rs
bryfox: NM, this is not breaking. I see you also have #362 for breaking changes.

rust/foxglove/src/websocket.rs
gasmith: Sort of, but it does so in a sync-free way [by consuming self](https://docs.rs/tokio/latest/tokio/sync/oneshot/struct.Sender.html#method.send). Since the `ConnectedClient` has shared references floating around everywhere, we can't have shutdown consume self, or even take `&mut self`, so we're forced to do a `Mutex<Option<Sender>>` or similar.

rust/foxglove/src/websocket.rs
```diff
@@ -145,29 +143,145 @@ pub(crate) struct ConnectedClient {
...
+            shutdown_tx: parking_lot::Mutex::new(Some(shutdown_tx)),
+        });
+        let poller = Poller {
+            client: client.clone(),
+            websocket,
+            data_plane_rx,
+            control_plane_rx,
+            shutdown_rx,
+        };
+        (client, poller)
```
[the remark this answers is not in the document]
gasmith: Done!

rust/foxglove/src/websocket/tests.rs
```diff
@@ -138,7 +138,7 @@ async fn test_client_connect() {
         ServerInfo::new("mock_server").with_session_id("mock_sess_id")
     );
 
-    server.stop().await;
+    server.stop();
```
eloff, line 141: It will be really nice when we can just drop the server handle to stop the server and clean it up, we can remove all these stop calls at the end of the tests

rust/foxglove/src/websocket.rs
```diff
@@ -143,31 +141,145 @@ pub(crate) struct ConnectedClient {
...
+/// - Receiving messages from the websocket and invoking [`ConnectedClient::handle_message`].
+/// - Waiting for a shutdown signal, and closing the websocket.
+struct Poller {
+    websocket: WebSocketStream<TcpStream>,
+    data_plane_rx: flume::Receiver<Message>,
+    control_plane_rx: flume::Receiver<Message>,
+    shutdown_rx: oneshot::Receiver<ShutdownReason>,
+}
+
+enum ShutdownReason {
```
eloff, line 173: This is a nice way of handling shutdown for different reasons, it solves a lot of problems very neatly!

rust/foxglove/src/websocket.rs
```diff
@@ -143,31 +141,145 @@ pub(crate) struct ConnectedClient {
...
+        let ws_tx_loop = async {
+            while let Ok(msg) = tokio::select! {
+                msg = self.control_plane_rx.recv_async() => msg,
+                msg = self.data_plane_rx.recv_async() => msg,
+            } {
+                if let Err(err) = ws_tx.send(msg).await {
+                    tracing::error!("Error sending message to client {addr}: {err}");
+                }
+            }
+            unreachable!("ConnectedClient holds queues");
```
eloff, line 211: Checks out

eloff (approved):
This looks great, I don't see any issues with it

PR #362, opened 2025-04-09
rust: Remove WebSocketServerBlockingHandle

gasmith:
Follow-up from #360. We can now remove the `WebSocketServerBlockingHandle` entirely, since there are no more async methods on `WebSocketServerHandle`.

We could also implement a `Drop` handler for `WebSocketServerHandle`, to stop the server when the handle is dropped. This would be better-aligned with the `McapWriter` behavior. I think the main reason we didn't do that in the past, is because of the fact that `stop()` was async. If we decide to implement Drop, I'll do it in a separate PR, but I'm curious to hear what others think.


bryfox (approved):
Do we want to publish a patch release with fixes and doc updates before merging this one?

> If we decide to implement Drop, I'll do it in a separate PR, but I'm curious to hear what others think.

I think consistency with the mcap sink is nice, but:

- it's less important for server to close cleanly; today, you can't restart the server, so it's assumed as long as you need it
- it creates some awkwardness, especially in other languages; you have to retain a server handle but there's nothing you need to do with it. Same thing with mcap, I guess, but I might expect to be able to say "start_server" and have that live as long as the process

I'm not yet convinced of either direction.

eloff (approved):
LGTM

It's so nice to combine the two handle types, they were so close to the same thing, just that pesky stop method.

I think we should also implement drop, I can't think why we'd want the server to outlive the handle ever.

gasmith:
OK, filed https://linear.app/foxglove/issue/FG-11152. Feel free to pile on there with more pros/cons.

PR #363, opened 2025-04-10
rust: Finish splitting up websocket.rs

gasmith:
This is the capstone change to break up websocket.rs into more manageable pieces. This change, despite its size, is purely code motion and a sprinkling of `pub`s.


bryfox (approved):
I didn't do a line-by-line of the moved content, but otherwise LG — please let me know if there's any place that needs specific review.

gasmith:
> I didn't do a line-by-line of the moved content, but otherwise LG — please let me know if there's any place that needs specific review.

Nope, it's a pretty boring change.

Sort of disappointing that github doesn't recognize moved hunks and render them as such.

PR #364, opened 2025-04-10
Add context to C and C++

eloff:
This adds a new C function to create a context and free a context.

It adds a C++ wrapper around the context, and provides a way to pass it when creating channels, mcap writers, and websocket servers. It now uses shared_ptr under the hood, so it can be trivially copied (like the Arc<Context> in Rust).

The mcap and websocketserver use options structs in their creation, so I added a Context to each options struct as a way to set the context. And set that to a default, null context, which gets replaced by the global context on the Rust side. Channels don't use an options struct, so I added an overloaded constructor with 4 params where the final one is the context, again defaulting to a null Context.

Then I updated all the tests to use a context scoped to the test so they don't pollute the global context anymore (yay!)


cpp/foxglove/include/foxglove/context.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+public:
+  Context();
+
+  static Context get_default();
+
+private:
+  explicit Context(const foxglove_context* context);
+
+  inline const ContextInner* get_inner() const {
```
jtbandes, line 27: nit: inline is implied here (2nd sentence: https://en.cppreference.com/w/cpp/language/inline)
```suggestion
  const ContextInner* get_inner() const {
```

cpp/foxglove/include/foxglove/context.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+class Context final {
+  friend class McapWriterOptions;
+  friend class Channel;
+  friend class WebSocketServerOptions;
+
+public:
+  Context();
+
+  static Context get_default();
```
jtbandes, line 22: naming: I think we were starting to use camelCase for members (at least data members) in some other places, so might want to double-check for consistency (I don't really have a strong opinion of which way is right for C++, although I find mixing camel for types and snake for methods to be a little weird)

cpp/foxglove/include/foxglove/server.hpp
```diff
@@ -59,17 +62,25 @@ struct WebSocketServerCallbacks {
...
+  uint16_t port = 0;
   WebSocketServerCallbacks callbacks;
   WebSocketServerCapabilities capabilities = WebSocketServerCapabilities(0);
   std::vector<std::string> supportedEncodings;
+
+  WebSocketServerOptions() = default;
+  explicit WebSocketServerOptions(const Context& context);
+
+private:
+  const ContextInner* context = nullptr;
```
jtbandes, line 78: the direct storage of the raw pointer, as well as the fact that we how have to define our own constructor, makes me a little uneasy...  Could we make `Context` copyable (or use shared_ptr) and store it directly instead? I know, a shared_ptr-to-Arc is a bit crazy, but it seems like it might improve the C++ ergonomics quite a bit

cpp/foxglove/include/foxglove/context.hpp
```diff
@@ -0,0 +1,34 @@
...
+struct foxglove_context;
+
+namespace foxglove {
+
+typedef foxglove_context ContextInner;
+
+class Context final {
+  friend class McapWriterOptions;
+  friend class Channel;
+  friend class WebSocketServerOptions;
```
jtbandes, line 17: I was going to say having to use `friend` seems like something to generally avoid, but I see the google c++ style guide allows it "within reason" https://google.github.io/styleguide/cppguide.html#Friends

A real 🤯 -- "If they’re used properly, they _enhance_ encapsulation" https://isocpp.org/wiki/faq/friends#friends-and-encap

jtbandes (commented):
Lots of bikeshedding could be done here, but there's something to be said for an implementation that works 👌

cpp/foxglove/include/foxglove/context.hpp
eloff: That's a good question, I'd rather we stick to rust naming TitleCase for types and snake_case for methods and members. But that's just my preference. I'd like to see what the others say.

cpp/foxglove/include/foxglove/server.hpp
gasmith: I'm inclined to agree with that. We're effectively making copies of the pointer with get_inner(), so using unique_ptr seems a little weird.

cpp/foxglove/include/foxglove/context.hpp
eloff: I was using it as an alternative to pub(crate) which we don't have in C++

cpp/foxglove/tests/test_mcap.cpp
```diff
@@ -124,10 +125,40 @@ std::string readFile(const std::string& path) {
...
+  FileCleanup cleanup("test.mcap");
+  auto context1 = foxglove::Context::create();
+  auto context2 = foxglove::Context::create();
+
+  foxglove::McapWriterOptions options{context1};
+  options.path = "test.mcap";
+  auto writer = foxglove::McapWriter::create(options);
+  REQUIRE(writer.has_value());
+
+  // Write message, but on default global context
```
bryfox, line 138: This comment looks stale — both contexts are explicit

cpp/foxglove/include/foxglove/server.hpp
```diff
@@ -62,9 +65,12 @@ struct WebSocketServerCallbacks {
...
 
 struct WebSocketServerOptions {
+  friend class WebSocketServer;
+
+  Context context;
   std::string name;
-  std::string host;
-  uint16_t port;
+  std::string host = "127.0.0.1";
+  uint16_t port = 0;
```
bryfox, line 73: Python and Rust default to port 8765, which matches the default connection setting in the app; can we do the same here?

c/src/lib.rs
```diff
@@ -396,7 +400,15 @@ unsafe fn do_foxglove_mcap_open(
         .open(path)
         .map_err(foxglove::FoxgloveError::IoError)?;
 
-    let writer = foxglove::McapWriter::with_options(mcap_options).create(BufWriter::new(file))?;
+    let mut builder = foxglove::McapWriter::with_options(mcap_options);
+    if !context.is_null() {
+        let context =
+            ManuallyDrop::new(unsafe { Arc::from_raw(context as *const foxglove::Context) });
```
bryfox, line 406: Do the Safety docs need updating on the pub fn?

bryfox (commented):
This looks good to me, but should get other eyes

c/src/lib.rs
```diff
@@ -550,6 +573,28 @@ pub unsafe extern "C" fn foxglove_channel_log(
     FoxgloveError::Ok
 }
 
+pub struct FoxgloveContext(#[allow(unused)] foxglove::Context);
```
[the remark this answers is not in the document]
eloff: Woah

gasmith (approved):

PR #365, opened 2025-04-11
Batch websocket channel advertisements

gasmith:
When a websocket is established, we can advertise channels in batches, rather than one at a time.

### DIFF
--- a/rust/foxglove/src/sink.rs
+++ b/rust/foxglove/src/sink.rs
@@ -4,7 +4,7 @@ use std::sync::Arc;
 use smallvec::SmallVec;
 
 use crate::metadata::Metadata;
-use crate::{FoxgloveError, RawChannel};
+use crate::{ChannelId, FoxgloveError, RawChannel};
 
 /// Uniquely identifies a [`Sink`] in the context of this program.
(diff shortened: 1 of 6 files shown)


rust/foxglove/src/sink.rs
```diff
@@ -52,18 +52,26 @@ pub trait Sink: Send + Sync {
...
     /// [sub]: crate::Context::subscribe_channels
-    fn add_channel(&self, _channel: &Arc<RawChannel>) -> bool {
-        false
+    fn add_channels(&self, _channel: &[&Arc<RawChannel>]) -> Option<Vec<ChannelId>> {
+        None
+    }
+
+    /// Called when a new channel is made available within the [`Context`][crate::Context].
+    ///
+    /// See [`Sink::add_channels`] for more details.
```
bryfox, line 70: Maybe worth keeping the note about the boolean return value from the old documentation

bryfox (approved):
LGTM. I wonder if this is worth a benchmark to cover use cases with hundreds/thousands of channels, but not needed in this PR in either case.

rust/foxglove/src/sink.rs
gasmith: Ah, good point.

PR #367, opened 2025-04-14
Doc updates

bryfox:
This improves generated documentation and code comments in a few places:

- Remove the explicit "unstable" note from python docs
- resolves FG-10910: Add links to schema documentation to explain their use in visualization
- resolves FG-10867: clarify the reason for renaming of GeoJSON


eloff (approved):
LGTM

PR #368, opened 2025-04-14
python: fix default schema names; provide topic and schema getters

bryfox:
This improves the handling of schema definitions which aren't given a name. Previously, we defaulted to a hard-coded string. This produces warnings in the app if you create multiple channels with unnamed schemas. Now, we'll generate based on a hash of the schema contents so that differing schemas will have different names, with reasonable probability. I used base64 encoding and prefixed with "schema-" so that labeling within the app is clear to users.

This also exposes `topic` and `schema_name` on the Channel, and improves the serialized representation of the channel class.

### DIFF
--- a/python/foxglove-sdk/python/foxglove/channel.py
+++ b/python/foxglove-sdk/python/foxglove/channel.py
@@ -1,4 +1,6 @@
+import hashlib
 import json
+from base64 import b64encode
 from typing import Any, Dict, Optional, Union, cast
 
 from . import _foxglove_py as _foxglove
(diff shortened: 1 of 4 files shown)


python/foxglove-sdk/python/foxglove/channel.py
```diff
@@ -164,13 +179,24 @@ def _normalize_schema(
...
             ),
         )
 
     raise TypeError(f"Invalid schema type: {type(schema)}")
+
+
+def _default_schema_name(data: bytes) -> str:
+    # Provide a consistent, readable, and reasonably unique name for a given schema so the app can
+    # identify it to the user.
+    hash = hashlib.shake_128(data).digest(6)
```
eloff, line 201: About 1 in a billion chance of collision for 1000 schemas. Should be enough.

python/foxglove-sdk/src/lib.rs
```diff
@@ -135,6 +135,14 @@ impl BaseChannel {
         Ok(())
     }
 
+    fn topic(&self) -> &str {
+        self.0.topic()
+    }
+
+    fn schema_name(&self) -> Option<&str> {
+        self.0.schema().map(|s| s.name.as_str())
```
eloff, line 143: Btw you can also use the ? Operator on options if the function returns an option.

No need to change this, just pointing it out in case you didn't know.

eloff (approved):
LGTM

PR #370, opened 2025-04-15
python: fix channel interface to provide ID

bryfox:
This adds `id()` to the Python channel interface. We publish subscription notifications on the server callback interface with these IDs, but they were not meaningful until now.

This also updates the typed channel interface with public getters for schema and topic name to match the raw channel getters added in #368.

### DIFF
--- a/python/foxglove-sdk/python/foxglove/_foxglove_py/__init__.pyi
+++ b/python/foxglove-sdk/python/foxglove/_foxglove_py/__init__.pyi
@@ -42,7 +42,7 @@ class BaseChannel:
         metadata: Optional[List[Tuple[str, str]]] = None,
     ) -> "BaseChannel": ...
     def id(self) -> int:
-        """The id of the channel"""
+        """The unique ID of the channel"""
         ...
 
     def topic(self) -> str:
--- a/python/foxglove-sdk/python/foxglove/_foxglove_py/__init__.pyi
+++ b/python/foxglove-sdk/python/foxglove/_foxglove_py/__init__.pyi
@@ -41,6 +41,10 @@ class BaseChannel:
         schema: Optional["Schema"] = None,
         metadata: Optional[List[Tuple[str, str]]] = None,
     ) -> "BaseChannel": ...
+    def id(self) -> int:
+        """The id of the channel"""
+        ...
+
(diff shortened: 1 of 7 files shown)


python/foxglove-sdk/python/foxglove/channel.py
```diff
@@ -59,9 +59,11 @@ def __init__(
         _channels_by_topic[topic] = self
 
     def __repr__(self) -> str:
-        return (
-            f"Channel(topic='{self.base.topic()}', schema='{self.base.schema_name()}')"
-        )
+        return f"Channel(id='{self.id()}', topic='{self.topic()}', schema='{self.schema_name()}')"
```
gasmith, line 62: Maybe we don't need the quotes around the ID?

```suggestion
        return f"Channel(id={self.id()}, topic='{self.topic()}', schema='{self.schema_name()}')"
```

typescript/schemas/src/internal/generatePyclass.ts
```diff
@@ -594,7 +609,7 @@ impl ${channelClass} {
     }
 
     fn __repr__(&self) -> String {
-        format!("${channelClass}(topic='{}')", self.0.topic()).to_string()
+        format!("${channelClass}(id='{}',topic='{}')", self.id(), self.topic()).to_string()
```
gasmith, line 612: The `, ` separator is preferred over `,`; and I think we can drop the quotes around ID (as above).

```suggestion
        format!("${channelClass}(id={}, topic='{}')", self.id(), self.topic()).to_string()
```

python/foxglove-sdk/python/foxglove/_foxglove_py/__init__.pyi
```diff
@@ -41,6 +41,10 @@ class BaseChannel:
         schema: Optional["Schema"] = None,
         metadata: Optional[List[Tuple[str, str]]] = None,
     ) -> "BaseChannel": ...
+    def id(self) -> int:
+        """The id of the channel"""
```
gasmith, line 45: To be consistent with the typed channel docstrings:

```suggestion
        """The unique ID of the channel."""
```

gasmith (approved):
LG, just nits.

PR #371, opened 2025-04-15
c++: Remove usage of CStr

bryfox:
### Changelog
None

### Description

This replaces usage of `CStr` with a `FoxgloveString` type having an associated length, per https://github.com/foxglove/foxglove-sdk/pull/295#discussion_r2002095320. I also updated the MCAP options which used separate length parameters.

### DIFF
--- a/c/build.rs
+++ b/c/build.rs
@@ -8,5 +8,5 @@ fn main() {
         .write_to_file("include/foxglove-c/foxglove-c.h");
 
     println!("cargo:rerun-if-changed=cbindgen.toml");
-    println!("cargo:rerun-if-changed=src/lib.rs");
+    println!("cargo:rerun-if-changed=src/");
 }
--- a/c/build.rs
+++ b/c/build.rs
@@ -8,4 +8,5 @@ fn main() {
         .write_to_file("include/foxglove-c/foxglove-c.h");
 
     println!("cargo:rerun-if-changed=cbindgen.toml");
+    println!("cargo:rerun-if-changed=src/lib.rs");
 }
--- a/c/include/foxglove-c/foxglove-c.h
+++ b/c/include/foxglove-c/foxglove-c.h
@@ -103,7 +103,7 @@ typedef struct foxglove_string {
    */
   const char *data;
   /**
-   * Number of characters in the string.
+   * Number of bytes in the string
    */
   size_t len;
 } foxglove_string;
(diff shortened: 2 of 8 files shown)


eloff (dismissed):
This looks good, but we need some bike shedding on separate arrays of lengths vs arrays of structures. After standup maybe

c/include/foxglove-c/foxglove-c.h
```diff
@@ -101,11 +101,14 @@ typedef uint8_t foxglove_server_capability;
...
 typedef struct foxglove_server_options {
   const char *name;
+  size_t name_len;
   const char *host;
+  size_t host_len;
   uint16_t port;
   const struct foxglove_server_callbacks *callbacks;
   foxglove_server_capability capabilities;
   const char *const *supported_encodings;
+  const size_t *supported_encoding_lengths;
```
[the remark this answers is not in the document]
bryfox: Per discussion: we'll go with the struct approach; I'll update all usage in this PR.

c/build.rs
```diff
@@ -8,4 +8,5 @@ fn main() {
         .write_to_file("include/foxglove-c/foxglove-c.h");
 
     println!("cargo:rerun-if-changed=cbindgen.toml");
+    println!("cargo:rerun-if-changed=src/");
```
gasmith, line 11: Does this do anything useful? I'd expect cargo to run after changing stuff under src/ by default.

c/include/foxglove-c/foxglove-c.h
```diff
@@ -94,6 +94,20 @@ typedef struct foxglove_mcap_writer foxglove_mcap_writer;
...
+/**
+ * A string with associated length.
+ */
+typedef struct foxglove_string {
+  /**
+   * Pointer to valid UTF-8 data
+   */
+  const char *data;
+  /**
+   * Number of characters in the string.
```
gasmith, line 106: Nit: s/characters/bytes/, perhaps?

gasmith (approved):
LG

c/include/foxglove-c/foxglove-c.h
bryfox: 👍 This is correct; characters is not. I added a test.

c/build.rs
bryfox: Yeah, before this change, when I ran `make` from the C++ directory, the C header would not update as expected.

Based on https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed, I think by adding the line above, we may have disabled the default behavior, requiring this? If it seems off, though, I'll dig further.

eloff: I ran into trouble with this as well, I was running touch on the toml file and manually regenerating it by running cargo check.

gasmith: Ah, got it. Sounds good.

PR #372, opened 2025-04-16
Add C++ src to assets

bryfox:
This adds C++ source files to the SDK release assets.  In #369 we decided to include headers as well as the libraries, and I should have included C++ source as well.

To test, I have created a draft release with the additional assets.

### DIFF
--- a/.github/workflows/c_cpp.yml
+++ b/.github/workflows/c_cpp.yml
@@ -140,10 +140,12 @@ jobs:
         run: |
           mkdir -p artifacts/foxglove/lib
           mkdir -p artifacts/foxglove/include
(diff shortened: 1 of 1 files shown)


gasmith (commented):
Is it necessary/useful to have the cpp sources for building a C++ application against the SDK? I would've expected that the headers & libraries were sufficient.

bryfox:
> Is it necessary/useful to have the cpp sources for building a C++ application against the SDK? I would've expected that the headers & libraries were sufficient.

That's what I'd been thinking, but we only build C libraries. Users are expected to include the cpp sources into their own builds (per initial docs, CMake comments, and #246).

gasmith (approved):

PR #373, opened 2025-04-17
Add RawAudio schema.

nidanin:
### Changelog
Add a RawAudio schema type.

### Docs
Docs needed but forthcoming.

### Description
Add a new `RawAudio` schema message type.

We'll start with just a raw PCM audio message type because it's easier to encode and play back. Once we've worked out the rest of the audio processing playback pipeline we can apply what we learn there to the inevitably more complex `CompressedAudio` message type we'll also need.


python/foxglove-sdk/src/generated/schemas.rs
```diff
@@ -73,6 +82,63 @@ pub(crate) enum LocationFixPositionCovarianceType {
...
+        sample_rate: u32,
+        number_of_channels: u32,
+    ) -> Self {
+        Self(foxglove::schemas::Audio {
+            timestamp: timestamp.map(Into::into),
+            data: data
+                .map(|x| Bytes::copy_from_slice(x.as_bytes()))
+                .unwrap_or_default(),
+            format: format as i32,
+            description: data
```
nidanin, line 116: Looks like there's a bug in the code to write these out. It's reusing `data` instead of `description`. I can try to fix this but maybe somebody more familiar with the code knows what's going wrong here?

nidanin: @bryfox Looks like you have some experience here. Any ideas?

nidanin: Actually I think this is the culprit:
https://github.com/foxglove/foxglove-sdk/blob/e9617cccbdf09f313b9912356a30a4da9294ec01/typescript/schemas/src/internal/generatePyclass.ts#L176

defunctzombie:
> 1. Should the format be an enum or just a string with restricted values?

Prior art here is a string with restricted values. We've found this most flexible to the various other "serialization" formats we tend to support with our schemas.

> 2. `sample_rate` and `num_channels` seem to only be required for PCM formats.

What does "required" mean in this context?

Do you think PCM formats would be a thing someone wants to use? I ask this not knowing anything about what "PCM formats" means.

> 3. `description` seems to only be required for some formats.

Same question about what "required" means.

> 6. All the w3codec PCM formats are interleaved. Seems like we should just follow this for simplicity

For the layperson what does that mean with respect to how the data is organized? And would it be interleaved from a typical c++ app/encoder where someone might be recording this data?

defunctzombie:
For image and video data we have `CompressedImage`, `CompressedVideo`, and `RawImage`. Is there a similar analog for Audio? Is there a typical convention for naming? i.e. `CompressedImage` is a term of art similar to `RawImage`.

typescript/schemas/src/internal/schemas.ts
```diff
@@ -1,5 +1,57 @@
...
+  fields: [
+    {
+      name: "timestamp",
+      type: { type: "primitive", name: "time" },
+      description: "Timestamp of audio frame",
+    },
+    {
+      name: "data",
+      type: { type: "primitive", name: "bytes" },
+      description: `Audio frame data`,
```
defunctzombie, line 30: Is there more we need to say about the data here? For context look at the details we provide on `CompressedVideo` formats and the things we expect. Is that true here as well? Anything about the amount or lack thereof that should or should not be present here? Anything detailed per-format that we should indicate?

nidanin:
> > 2. `sample_rate` and `num_channels` seem to only be required for PCM formats.
> 
> What does "required" mean in this context?

They're needed to interpret the data correctly because the data itself is just a sequence of sample values so we need some other indicator of how many channels and what sample rate it represents.

> Do you think PCM formats would be a thing someone wants to use? I ask this not knowing anything about what "PCM formats" means.

They're the easiest to capture and encode because no extra codecs are required but they're the least space-efficient. I suspect at least some customers will want this.

> > 3. `description` seems to only be required for some formats.
> 
> Same question about what "required" means.

For example, FLAC requires it:
https://www.w3.org/TR/webcodecs-flac-codec-registration/#audiodecoderconfig-description

And it's meaningful for AAC:
https://www.w3.org/TR/webcodecs-aac-codec-registration/#audiodecoderconfig-description

> > 6. All the w3codec PCM formats are interleaved. Seems like we should just follow this for simplicity
> 
> For the layperson what does that mean with respect to how the data is organized? And would it be interleaved from a typical c++ app/encoder where someone might be recording this data?

I'm not sure which is more common in the C++/robotics world. I'd need to some some reason. Converting between interleaved & non-interleaved is straightforward though.

nidanin:
ROS message type quite basic:

https://github.com/ros-drivers/audio_common/blob/master/audio_common_msgs/msg/AudioInfo.msg

nidanin:
Linear requests so far seem to include PCM, mp3 and ogg:
https://linear.app/foxglove/project/audio-support-53746050eca5/customer-requests

defunctzombie:
I don't see what new headache it would cause that anyone who starts out with raw images and then switches to jpeg or maybe video would not encounter. Either way the consumer of the data is going to need to understand it. I don't feel strongly about whether it is one or two schemas here only that I offer it as an example where we separated by schema because the field needs were different.

nidanin:
> I don't see what new headache it would cause that anyone who starts out with raw images and then switches to jpeg or maybe video would not encounter.

If I switch at some point and then want to do data platform stream of data that changed formats can I do it? Or would I need a new topic name and schema?

defunctzombie:
Today we can't do it because we don't change schema on a topic on a data source once it is set (last I recall) but its not an absolute limitation of the app (some plumbing would need to change). This would also be true if you made a breaking change to a schema that does not take kindly to that - like a ROS schema by introducing a field. In your example I would say it might be pretty uncommon for a schema change not to also imply a software update and likely a completely different "run" of a robot.

nidanin:
I think you have more background and context on this than I do. We could certainly do separate `RawAudio` and `CompresedAudio` schemas. That would have the advantage of having fewer fields that are only relevant for some formats.

nidanin:
Would that also cause problems for layouts if I start out using PCM and then switch to compressed and then have to change topics?

defunctzombie:
"problems" is a complex statement. Like if you switch topic names then yea maybe layouts have to be adjusted depending on how the panel handles selecting the audio topic. The general reality is that if you change your topics that tends to mean you have to go update your layouts.

nidanin:
> "problems" is a complex statement. Like if you switch topic names then yea maybe layouts have to be adjusted depending on how the panel handles selecting the audio topic. The general reality is that if you change your topics that tends to mean you have to go update your layouts.

Ok. Well I don't really have a strong opinion here. Maybe somebody with more background the schemas and their use should make the call on just `Audio` vs both `RawAudio` & `CompressedAudio`.

defunctzombie:
I am gonna say you need to drive the discussion here. What might help is a writeup/table describing what fields the various formats we need/want to support *will* require a schema to have and some thoughts on how someone would pack their info into the schema. I view the way we need to drive this forward is to present the universe of things we definitely know today and what the options are. Ideally you pair this with a recommendation and the tradeoffs that went into making the decision.

nidanin:
@defunctzombie To make sure I understand the considerations here, the advantage of having distinct `RawAudio` vs `CompressedAudio` types is to omit the unneeded fields for dev ergonomics, not space right? The extra space taken up per-message by an unused `sample_rate`, `num_channels`, and empty `description` field isn't really consequential but asking users to fill them out even for formats where they're not needed is undesirable?

defunctzombie:
The space taken would depend on the serialization format - ROS1 (Possibly ROS2) formats for example don't have optional fields and would always have some value for these fields that take up space. So I think space can be a consideration; figuring out how much of a consideration it is though is part of the research here. Whether it is consequential I think we could back up with numbers like: "A PCM audio message at 30hz would have 33ms of audio data per message which would be ## bytes of data and ## bytes of metadata and ## bytes of unused data". The same statement can be made for the other kinds of audio formats. With this you can develop a real number on what % overhead unused fields might or might not have. Same is true if you extend this to 100hz data or slow it down to 1hz data.

nidanin:
I'm not sure if I should keep appending comments here or create and link to a separate proposal doc for discussion but here are some rough figures:

A typical low latency DAW configuration will use a buffer size of ~512 samples, or around 11ms of latency at 44.1khz. This seems like a reasonable lower-bound for the latency we should try to achieve in the app. So assuming a message size of 512 mono 16 bit samples, that's a block size of 512 * 16 = 8192 bytes. If the `sample_rate` and `num_channel` fields are uint32 that adds up to 8 bytes so the overhead will be only ~ 0.1%. I'm not sure how much space an empty `description` would take up though.

If we also consider low-bitrate lossy formats where the compression could be around 10x then that overhead would rise to about 1% of the total message size.

I suspect for most robotics applications much higher latencies than this would be acceptable so the overhead would be less in that case.

nidanin:
Also looking more closely at the w3codecs spec, there are some not-very well documented details around specific construction of the `description` field and container formats for opus & aac.

I think prototyping this stuff in a browser risks building in too many web assumptions. I suggest the way forward here is to build some encoders in C++ or Python vs a proposed schema and see what it takes to encode audio that we can actually play back in a test audio panel.

nidanin:
Also, if we want to ship something as quickly as possible, we could start with just the `RawAudio` message type, where the `sample_rate` and `num_channels` are always required and the `description` is not needed and ship that and then proceed to the `CompressedAudio` message type which has more complex concerns around message formatting and `description`.

defunctzombie:
I like that idea - how can we pressure test that sample_rate and num_channels are required? And that a user writing that data to the message would have it? With `RawAudio` it would also be good to understand the data to metadata ratio so we can understand the metadata "overhead" of including it in every message. iirc that was one design decision that the ROS audio common package did differently (separate data and metadata) - I lean towards including - but for example  the CameraCalibration and CompressedImage data is separate topics (though one can very correctly point out that CameraCalibration is separate from decoding the image data itself and that all that metadata is present in the image messages). Same idea could apply here.

nidanin:
`sample_rate` and `num_channels` are definitely required for raw PCM audio. I suppose there's some way we could embed them in channel metadata instead of in every block? I think the overhead in the case of PCM audio would be so small it's not worth doing anything heroic to reduce it.

defunctzombie:
@foxymiles is this ready for review or still draft?

nidanin:
@defunctzombie Still draft. I could use a little help using this branch in app & mcap repos to make sure it's correct.

defunctzombie:
You'll want to update the `@foxglove/schemas` dependency in the app and mcap repos to reference github:

```

"@foxglove/schemas": "github:foxglove/schemas#workspace=@foxglove/schemas&commit=a806c88
```

[https://yarnpkg.com/protocol/git#workspaces-support](https://yarnpkg.com/protocol/git#workspaces-support)

schemas/jsonschema/index.ts
```diff
@@ -4590,6 +4590,50 @@ export const Quaternion = {
   }
 };
 
+export const RawAudio = {
+  "title": "foxglove.RawAudio",
+  "description": "A single frame of an audio bit stream",
```
bryfox, line 4595: nit: is "bitstream" (one word) more common?

bryfox (commented):
I read the comment discussion about the schema design and don't have anything to offer. The generated schemas & channels look as expected to me. I can help test if needed, but this looks good to me. I'll defer to @defunctzombie for approval.

schemas/jsonschema/index.ts
nidanin: I actually started out with `bitstream` but the spell checker didn't like it so I changed it. I agree that I've seen `bitstream` used more often though.

typescript/schemas/src/internal/schemas.ts
```diff
@@ -1,5 +1,38 @@
...
+const RawAudio: FoxgloveMessageSchema = {
+  type: "message",
+  name: "RawAudio",
+  description: "A single frame of an audio bit stream",
+  fields: [
+    {
+      name: "timestamp",
+      type: { type: "primitive", name: "time" },
+      description: "Timestamp of the audio frame",
+    },
```
snosenzo, line 12: Should it have a `frame_id`?

I was thinking it could be used for something like delivering spatial audio to teleoperators depending on a perspective in a scene and also used to show visualizations of sound being detected by devices in the 3D scene.

However this is certainly not a main use case and might cause some confusion. @defunctzombie has also pointed out that we don't have a way to support optional fields, so this would likely have to be filled with an empty string in most cases.

ros/foxglove_msgs/ros1/RawAudio.msg
```diff
@@ -0,0 +1,19 @@
+# foxglove_msgs/RawAudio
+# A single frame of an audio bit stream
+
+# Generated by https://github.com/foxglove/foxglove-sdk
+
+# Timestamp of the audio frame
```
defunctzombie, line 6: Do we need to say it is the start of the audio frame? Since an audio frame does last for some duration?

defunctzombie (approved):

typescript/schemas/src/internal/schemas.ts
nidanin: We do support more conventional multi-channel now by setting `number_of_channels` > 2. I guess this would allow some use cases that that alone doesn't?

I suppose we could also add a new `SpatialAudio` message type or something like that.

ros/foxglove_msgs/ros1/RawAudio.msg
nidanin: Yes it makes sense to clarify that. I'll update.

nidanin:
I'm also updating the language to refer to an audio `block` instead of an audio `frame` because `frame` has a common meaning in audio APIs that means one sample * number of channels.

PR #374, opened 2025-04-17
python: Add MCAP write options

bryfox:
### Changelog

python: Add MCAP write options
python: [deprecation] `MCAPWriter` moved to the `foxglove.mcap` module

### Description

This exposes MCAP writer options in the python SDK. I omitted some options which are not useful today: file metadata and attachments are not used; `disable_seek` is only useful if you control the rust writer; `library` is used by the SDK. I think we should consider removing these from C++ and Rust as appropriate.

The options here otherwise match the rust writer options; note that this differs from the python mcap library.

This also introduces a `foxglove.mcap` submodule. MCAPWriter is re-exported from `foxglove` to avoid a breaking change, but is documented as deprecated.

### DIFF
--- a/python/foxglove-sdk/python/docs/api/index.rst
+++ b/python/foxglove-sdk/python/docs/api/index.rst
@@ -8,6 +8,7 @@ foxglove
 
 .. automodule:: foxglove
    :members:
+   :exclude-members: MCAPWriter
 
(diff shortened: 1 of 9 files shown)


gasmith (approved):
:ship:

PR #375, opened 2025-04-18
Fix generation of data fields in python class generation.

nidanin:
### Changelog
Fix an issue with naming of fields in generation of python classes.

### Docs
None

### Description
The code to generate python classes was generating fields with the hardcoded name of `data` for the field instead of the field name. This didn't get flagged as an error because we happen to use the name `data` for this kind of field in our existing schema.


bryfox (approved):
👍

gasmith (approved):

PR #378, opened 2025-04-19
Python protobuf example

amacneil:
Add python example including:

- Foxglove schemas
- JSON with jsonschema
- Schemaless JSON
- Custom protobuf schema


eloff (approved):
LGTM

PR #379, opened 2025-04-21
c++: Add connection graph

bryfox:
### Changelog

C++: Add connection graph support to the WS server

### Description

This adds connection graph support for C++ and an `example_connection_graph` target. The graph can be rendered in the Topic Graph panel of Foxglove.


cpp/foxglove/include/foxglove/server.hpp
```diff
@@ -22,6 +23,22 @@ struct ClientChannel {
   size_t schemaLen;
 };
 
+class ConnectionGraph final {
```
eloff, line 26: Should go in it's own .hpp/.cpp files

c/include/foxglove-c/foxglove-c.h
```diff
@@ -287,6 +294,63 @@ void foxglove_internal_register_cpp_wrapper(void);
...
+ * Set a subscribed topic and its associated subscriber ids.
+ * Overwrites any existing topic with the same name.
+ *
+ * # Safety
+ * `topic`, and each ID in `subscriber_ids` must adhere to the safety rules of `FoxgloveString`.
+ * `subscriber_ids_count` must be the number of elements in the `subscriber_ids` array.
+ */
+foxglove_error foxglove_connection_graph_set_subscribed_topic(struct foxglove_connection_graph *FOXGLOVE_NONNULL graph,
+                                                              struct foxglove_string topic,
+                                                              const struct foxglove_string *subscriber_ids,
```
eloff, line 337: Should we do this, or should we create a FoxgloveStringArray type?

eloff (commented):
This is looking good so far

c/include/foxglove-c/foxglove-c.h
bryfox: I can go either way on this.

There's no practical safety difference, right? I assume a caller would create them in exactly the same way, so it's a matter of preference for the interface? If we had functions that accepted multiple arrays of strings, I could see a concrete benefit, but I don't see anything like that coming up.

I imagine passing `size_t` args next to arrays is reasonably common — moreso than with the sized string interfaces which led to foxglove_string. So this doesn't seem too bad to me; and we won't end up with different array types for each underlying type.

I guess the other consideration is what we want the callback interface to look like; e.g., how you receive a list of parameter names.

Do you have a preference?

c/src/connection_graph.rs
```diff
@@ -0,0 +1,271 @@
...
+
+pub struct FoxgloveConnectionGraph(pub(crate) foxglove::websocket::ConnectionGraph);
+
+impl Default for FoxgloveConnectionGraph {
+    fn default() -> Self {
+        Self(foxglove::websocket::ConnectionGraph::new())
+    }
+}
+
+/// Create a new `FoxgloveConnectionGraph`.
```
gasmith, line 11: Ultra-pedantry: For docstrings on public C API functions, if we want to refer to type names directly, we should use the mangled snake_case form. In this particular case, I think I'd prefer to just say "Creates a new connection graph" and let the type signature speak for itself.

c/src/connection_graph.rs
```diff
@@ -0,0 +1,271 @@
...
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn foxglove_connection_graph_create(
+    graph: *mut *mut FoxgloveConnectionGraph,
+) -> FoxgloveError {
+    let graph_box = Box::new(FoxgloveConnectionGraph::default());
+    unsafe { *graph = Box::into_raw(graph_box) };
+
+    if graph.is_null() {
+        return FoxgloveError::Unspecified;
+    }
```
gasmith, line 24: Is this actually a concern? Rust just panics if it fails to allocate. Maybe in the future we'll have [`Box::try_new`](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.try_new), but I think for now it makes sense to just unconditionally return `FoxgloveError::Ok`.

c/src/connection_graph.rs
```diff
@@ -0,0 +1,271 @@
...
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn foxglove_connection_graph_free(graph: *mut FoxgloveConnectionGraph) {
+    drop(unsafe { Box::from_raw(graph) });
+}
+
+/// Set a published topic and its associated publisher ids.
+/// Overwrites any existing topic with the same name.
+///
+/// # Safety
+/// `topic`, and each ID in `publisher_ids` must adhere to the safety rules of `FoxgloveString`.
```
gasmith, line 42: We should mention that we copy the strings out of these pointers, and so the pointers only need to be valid for the duration of this function call.

c/src/lib.rs
```diff
@@ -287,6 +300,29 @@ pub extern "C" fn foxglove_server_stop(
...
+    graph: Option<&mut FoxgloveConnectionGraph>,
+) -> FoxgloveError {
+    let Some(server) = server else {
+        tracing::error!("foxglove_server_publish_connection_graph called with null server");
+        return FoxgloveError::ValueError;
+    };
+    let Some(graph) = graph else {
+        tracing::error!("foxglove_server_publish_connection_graph called with null graph");
+        return FoxgloveError::ValueError;
+    };
```
gasmith, line 315: What's the idea behind accepting an `Option<&mut T>` instead of just `&mut T`?

c/src/lib.rs
```diff
@@ -287,6 +300,29 @@ pub extern "C" fn foxglove_server_stop(
...
+    let Some(server) = server.as_ref() else {
+        tracing::error!("foxglove_server_publish_connection_graph called with closed server");
+        return FoxgloveError::SinkClosed;
+    };
+    match server.publish_connection_graph(graph.0.clone()) {
```
gasmith, line 320: Could we consume the graph instead of cloning it, so that the caller doesn't need to call `foxglove_connection_graph_free`?

cpp/foxglove/include/foxglove/server/connection_graph.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+struct foxglove_connection_graph;
+
+namespace foxglove {
+
+class ConnectionGraph final {
+public:
+  ConnectionGraph();
+
+  foxglove_connection_graph& impl();
```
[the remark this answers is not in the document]
eloff: Yeah, C++ doesn't have Rust's pub(crate) which is so useful. I vote we allow friend for that purpose.

I never subscribed to the puritan style of OOO in C++ and Java where each class is a strict boundary layer.

c/src/connection_graph.rs
gasmith: Ach, misread. In that case, it needs to happen before we dereference. ;)

c/src/lib.rs
eloff: The pattern is used all over in the C API, it converts a null pointer to None so you can check for that.

c/src/connection_graph.rs
bryfox: Ah, thanks; I misunderstood the earlier feedback. I fixed the ordering and kept the error return value since that's the pattern we've adopted elsewhere.

c/src/lib.rs
bryfox: Maybe? Right now (in C++) the unique_ptr deals with the freeing so you don't need to think about it. But you also don't need to allocate a new connection graph if you want to re-publish it periodically (see example).

c/src/connection_graph.rs
eloff: Btw, ValueError is what we used elsewhere, indicating an issue with a parameter

bryfox: 👍 makes sense

c/src/connection_graph.rs
```diff
@@ -19,11 +19,12 @@ pub unsafe extern "C" fn foxglove_connection_graph_create(
     graph: *mut *mut FoxgloveConnectionGraph,
 ) -> FoxgloveError {
     let graph_box = Box::new(FoxgloveConnectionGraph::default());
-    unsafe { *graph = Box::into_raw(graph_box) };
-
     if graph.is_null() {
         return FoxgloveError::Unspecified;
```
eloff: Use ValueError like the other cases, it indicates an issue with the params

eloff (approved):
This looks good to me, I just left a comment about fixing the error code for the null check

cpp/foxglove/include/foxglove/server/connection_graph.hpp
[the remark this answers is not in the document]
bryfox: OK, I've declared WebSocketServer a friend of ConnectionGraph. I think that makes sense here; the latter is really only meaningful in the context of the server.

bryfox:
I think I've addressed all feedback now, and per https://github.com/foxglove/foxglove-sdk/pull/379#discussion_r2054665633 replaced the impl getter with a friend declaration.

PR #381, opened 2025-04-23
rust: Parameter::string() should not produce a byte array

gasmith:
### Changelog
- rust: Fixed Parameter::string() constructor, which was producing a byte array

### Description
The `Parameter::string()` constructor was mistakenly returning a byte array, when it should've been returning a string.

This change also updates `Parameter::float64_array()` to accept an `impl IntoIterator<Item = f64>` and updates the rust `param-server` example.

### DIFF
--- a/rust/foxglove/src/websocket/ws_protocol/parameter.rs
+++ b/rust/foxglove/src/websocket/ws_protocol/parameter.rs
@@ -68,7 +68,7 @@ impl Parameter {
     }
 
     /// Creates a new parameter with a float64 array value.
-    pub fn float64_array(name: impl Into<String>, values: Vec<f64>) -> Self {
+    pub fn float64_array(name: impl Into<String>, values: impl IntoIterator<Item = f64>) -> Self {
         Self {
             name: name.into(),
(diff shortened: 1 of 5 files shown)


rust/foxglove/src/websocket/ws_protocol/parameter.rs
```diff
@@ -80,7 +80,11 @@ impl Parameter {
 
     /// Creates a new parameter with a string value.
     pub fn string(name: impl Into<String>, value: impl Into<String>) -> Self {
-        Self::byte_array(name, value.into().into_bytes())
+        Self {
+            name: name.into(),
+            r#type: None,
+            value: Some(ParameterValue::String(value.into().into_bytes())),
```
bryfox, line 86: Is it correct that we're converting to bytes and then base64-encoding? Why would I use a string parameter and not a byte array in that case? Maybe this should be using the utf8 string directly?

bryfox: oh, I see #382 now

bryfox (approved):

PR #382, opened 2025-04-23
String parameters must not be base64-encoded

gasmith:
### Changelog
Breaking changes:
- String websocket parameters are not base64-encoded

### Description
I realized while plumbing C++ parameters that we're not representing string parameter values properly. The [example](https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#example-5) from the ws-protocol spec clearly shows that strings values are only supposed to be encoded when the parameter type is specified as `byte_array`.

This change:
- Replaces `ParameterValue::String(Vec<u8>)` with `ParameterValue::String(String)`.
- Adds `Parameter::decode_byte_array()` helper.
  - I didn't extend this to python yet, because we don't have the encode side either. (cf. FG-11299)
- Replaces python `ParameterValue.Bytes(value: bytes)` with `ParameterValue.String(value: str)`.

### DIFF
--- a/python/foxglove-sdk/python/foxglove/_foxglove_py/websocket.pyi
+++ b/python/foxglove-sdk/python/foxglove/_foxglove_py/websocket.pyi
@@ -148,10 +148,15 @@ class ParameterValue:
 
         def __new__(cls, value: float) -> "ParameterValue.Number": ...
 
-    class Bytes:
-        """A byte array."""
+    class String:
+        """
+        A string value.
 
-        def __new__(cls, value: bytes) -> "ParameterValue.Bytes": ...
+        For parameters of type :py:attr:ParameterType.ByteArray, this is a
+        base64 encoding of the byte array.
+        """
+
+        def __new__(cls, value: str) -> "ParameterValue.String": ...
 
     class Array:
         """An array of parameter values."""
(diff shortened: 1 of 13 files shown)


python/foxglove-sdk/python/foxglove/_foxglove_py/websocket.pyi
```diff
@@ -148,10 +148,15 @@ class ParameterValue:
...
+    class String:
+        """
+        A string value.
+
+        For parameters of type :py:attr:ParameterType.ByteArray, this is a
+        base64 encoding of the byte array.
+        """
 
-        def __new__(cls, value: bytes) -> "ParameterValue.Bytes": ...
+        def __new__(cls, value: str) -> "ParameterValue.String": ...
```
bryfox, line 159: We should probably provide both Bytes and String in python, right? I guess I added this interface originally, and thought I was simplifying since strings (then) appeared to be base-64 encoded bytes anyway.

bryfox (approved):
One question, but LG

python/foxglove-sdk/python/foxglove/_foxglove_py/websocket.pyi
gasmith: Yeah, that's FG-11299. What we want is a `Parameter.bytes()` constructor, like we have in rust with [`Parameter::byte_array`](https://github.com/foxglove/foxglove-sdk/pull/381/files#diff-04b8e72b3f6524c3a38dbd312982a4a43e9229e73a1f07544dd81dc218930f5fL86-L93). The `ParameterValue.String` variant can only be a string, but it's the outer `Parameter.type` that determines how it is interpreted.

PR #385, opened 2025-04-24
Support channels with a shared topic

bryfox:
### Changelog

Do not error or panic if topic names are re-used between incompatible channels.

### Description

This makes a few adjustments to our handling of re-used topic names within a context's channels. 

At a high level, there's a bit of a mismatch between the handling of channels and topics in Foxglove and MCAP. MCAP doesn't dictate anything here. The app doesn't have an explicit concept of a Channel; it instead relies on topic names. There's a slight complicating factor in that the Rust MCAP library has a notion of a "canonical channel"; if you create a second channel with identical topic, schema, etc., then you'll get back a reference to the canonical channel. It appears to be the only MCAP library to behave this way.

The motivation here (and initial bug report) is to ease playback of MCAP data which was created from other sources, which may contain multiple channels with the same topic name. I tried to take a pragmatic approach here that may err on the side of being too flexible. Instead of returning an error (or panicking), SDK will log a warning that a duplicate topic name was encountered, and I updated documentation to clarify that topic names *should* be unique for compatibility. 

With duplicate topics in play, this brings live viz from the SDK more in line with MCAP playback (in terms of app behavior).

#### Considerations / future enhancements

We could introduce something like a "strict mode" (opt-in) which does panic when encountering state that won't be fully compatible with the app. That could arguably make initial development or testing easier when app compatibility is the goal.

We could consider adopting the "canonical channel" behavior for the WS Server as well, in order to coalesce channels that share a schema. Currently, the app will only subscribe to one channel associated with a topic. We could instead investigate enhancements in the app for subscriptions to shared topic names (and the equivalent for file playback).

I did consider keeping the behavior stricter for now. (I don't think panicking should be the default behavior in these cases, but I'm less convinced that these shouldn't be errors.) But I think the approach here is more practical in the near term.


rust/foxglove/src/channel/lazy_channel.rs
```diff
@@ -17,6 +17,8 @@ use super::{Channel, RawChannel};
 /// initialized, which is guaranteed to happen when the channel is first used. If you need to
 /// ensure the channel is initialized _before_ using it, you can use [`LazyChannel::init`].
 ///
+/// You should choose a unique topic name per channel for compatibility with the Foxglove app.
```
eloff, line 20: Per channel with a different schema

rust/foxglove/src/channel_builder.rs
```diff
@@ -83,9 +84,6 @@ impl ChannelBuilder {
     /// Builds a [`Channel`].
     ///
     /// `T` must implement [`Encode`].
-    ///
-    /// Returns [`FoxgloveError::DuplicateChannel`] if a channel with the same topic already
-    /// exists.
     pub fn build<T: Encode>(mut self) -> Result<Channel<T>, FoxgloveError> {
```
eloff, line 87: This can't fail now, should we not return a result type here?

rust/foxglove/src/context.rs
```diff
@@ -22,20 +23,30 @@ struct ContextInner {
 }
 impl ContextInner {
     /// Returns the channel for the specified topic, if there is one.
+    ///
+    /// If multiple channels use the same topic name, this will return the first channel that was
+    /// added to this context.
```
eloff, line 28: I was wondering how you'd handle this, seems the best option.

rust/foxglove/src/context.rs
```diff
@@ -22,20 +23,30 @@ struct ContextInner {
...
         let topic = channel.topic();
-        let Entry::Vacant(entry) = self.channels_by_topic.entry(topic.to_string()) else {
-            return Err(FoxgloveError::DuplicateChannel(topic.to_string()));
-        };
-        entry.insert(channel.clone());
+
         self.channels.insert(channel.id(), channel.clone());
 
+        // Index the channel by topic name if we haven't seen it before
+        if !self.channels_by_topic.contains_key(topic) {
```
eloff, line 40: Use entry like the original code to avoid doing the lookup twice

rust/foxglove/src/context.rs
```diff
@@ -218,9 +229,17 @@ impl Context {
         Arc::clone(LazyContext::get_default())
     }
 
-    /// Returns a channel builder for a channel in this context.
-    pub fn channel_builder(self: &Arc<Self>, topic: impl Into<String>) -> ChannelBuilder {
-        ChannelBuilder::new(topic).context(self)
+    /// Creates a builder for a channel in this context. Results in a
+    /// [`FoxgloveError::DuplicateChannel`] if a channel with the same topic already exists.
```
eloff, line 233: This should be allowed now? Seems inconsistent.

Maybe it's to support log!, but maybe we could think of a better API for that anyway.

rust/foxglove/src/mcap_writer/mcap_sink.rs
```diff
@@ -65,7 +70,8 @@ impl<W: Write + Seek> WriterState<W> {
                 mcap_channel_id
             }
         };
-        let sequence = self.next_sequence(channel_id);
+
+        let sequence = self.next_sequence(mcap_channel_id);
```
bryfox, line 74: This is a bugfix which accounts for the "canonical channel" handling of the rust mcap writer.

rust/foxglove/src/context.rs
bryfox: I kept it because it seemed like it was steering people in the right direction, but I agree with you; I think I should update this to be consistent and rely on warnings and documentation.

rust/foxglove/src/context.rs
```diff
@@ -22,20 +23,30 @@ struct ContextInner {
...
         self.channels.insert(channel.id(), channel.clone());
 
+        // Index the channel by topic name if we haven't seen it before
+        if !self.channels_by_topic.contains_key(topic) {
+            self.channels_by_topic
+                .insert(topic.to_string(), channel.clone());
+        } else {
+            warn!(
+                "Channel with topic {} already exists in this context; use a unique topic for each channel",
+                topic
```
gasmith, line 46: Nit: Could do `".. {topic} .."`

rust/foxglove/src/channel/lazy_channel.rs
gasmith: I think I prefer the original wording. Even if the app technically supports having multiple channels with the same schema on the same topic, it doesn't support it _well_, because it doesn't give you any way to determine which channel a particular message came from.

gasmith (commented):
I think this looks good, and I think it strikes the right balance of flexibility and user-guidance. I agree with most of Dan's suggestions.

bryfox:
OK, thanks — I think I've addressed the feedback above. Perhaps still an open question about the wording [in this thread](https://github.com/foxglove/foxglove-sdk/pull/385#discussion_r2058757765).

rust/foxglove/src/channel_builder.rs
bryfox: Done. Note this resulted in changes to the python classes too (which is good; our examples weren't catching anyway). I added a unit test that should catch the expectation there.

gasmith (approved):
:ship:

PR #386, opened 2025-04-24
C/C++: Add support for websocket parameter subscriptions

gasmith:
### Changelog
- C/C++: Add support for websocket parameters

### Description
This change is a follow-up on #383, to finish adding support for websocket parameter subscriptions and publishing.

I tried to remove a little boilerplate for the server setup in our tests. While I was doing that, I noticed that the C++ WebSocketServer is RAII - it shuts down the server when it is dropped. So I got rid of some explicit `server.stop()` calls too.

### DIFF
--- a/c/src/lib.rs
+++ b/c/src/lib.rs
@@ -280,10 +280,22 @@ pub struct FoxgloveServerCallbacks {
             params: *const FoxgloveParameterArray,
         ) -> *mut FoxgloveParameterArray,
     >,
+    pub on_parameters_subscribe: Option<
+        unsafe extern "C" fn(
+            context: *const c_void,
+            param_names: *const *const c_char,
+            param_names_len: usize,
+        ),
+    >,
+    pub on_parameters_unsubscribe: Option<
+        unsafe extern "C" fn(
+            context: *const c_void,
+            param_names: *const *const c_char,
+            param_names_len: usize,
+        ),
+    >,
     pub on_connection_graph_subscribe: Option<unsafe extern "C" fn(context: *const c_void)>,
     pub on_connection_graph_unsubscribe: Option<unsafe extern "C" fn(context: *const c_void)>,
-    // pub on_parameters_subscribe: Option<unsafe extern "C" fn()>
-    // pub on_parameters_unsubscribe: Option<unsafe extern "C" fn()>
 }
 unsafe impl Send for FoxgloveServerCallbacks {}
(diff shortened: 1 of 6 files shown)


c/src/lib.rs
```diff
@@ -195,10 +195,22 @@ pub struct FoxgloveServerCallbacks {
             params: *const FoxgloveParameterArray,
         ) -> *mut FoxgloveParameterArray,
     >,
+    pub on_parameters_subscribe: Option<
+        unsafe extern "C" fn(
+            context: *const c_void,
+            param_names: *const *const c_char,
```
eloff, line 201: What about using foxglove string here? It seems inconsistent to use c strings in some places and strings with length elsewhere.

c/src/lib.rs
```diff
@@ -195,10 +195,22 @@ pub struct FoxgloveServerCallbacks {
             params: *const FoxgloveParameterArray,
         ) -> *mut FoxgloveParameterArray,
     >,
+    pub on_parameters_subscribe: Option<
```
eloff, line 198: Maybe add a note that the strings are only valid until the callback returns.

cpp/foxglove/src/server.cpp
```diff
@@ -110,6 +111,32 @@ FoxgloveResult<WebSocketServer> WebSocketServer::create(
         return array.release();
       };
     }
+    if (callbacks->onParametersSubscribe) {
+      cCallbacks.on_parameters_subscribe = [](
+                                             const void* context,
+                                             const char* const* _param_names,
+                                             size_t param_names_len
+                                           ) {
+        std::vector<std::string> param_names;
```
eloff, line 120: Why not string_view here? User can copy them to a string if they need to.

cpp/foxglove/tests/test_server.cpp
```diff
@@ -101,17 +101,33 @@ class WebSocketClient {
...
 
+foxglove::WebSocketServer start_server(foxglove::WebSocketServerOptions&& options) {
+  auto result = foxglove::WebSocketServer::create(std::move(options));
+  REQUIRE(result.has_value());
+  auto server = std::move(result.value());
+  REQUIRE(server.port() != 0);
+  return server;
+}
+
+foxglove::WebSocketServer start_server(
```
eloff, line 112: This is nice

bryfox (approved):
👍 . Looks like existing feedback has been addressed.

PR #389, opened 2025-04-25
python: Improved websocket parameters

gasmith:
### Changelog
- python: Improved websocket parameters

### Description
The purpose of this change is to make websocket parameters more ergonomic in python, by implementing conversions to and from native python types.

The basic idea is that we can infer the (type, value) from the type information on a python object, so that the caller doesn't need to specify it explicitly. Likewise, we can convert (type, value) back to a single native python object using a `get_value()` method.

The implementation gets just a tiny bit hairy, because we need to introduce some shim types to accomplish the conversions, but generally, pyo3's trait system works in our favor.

### DIFF
--- a/python/foxglove-sdk-examples/ws-param-server/main.py
+++ b/python/foxglove-sdk-examples/ws-param-server/main.py
@@ -10,13 +10,7 @@ import time
 from typing import List, Optional
 
 import foxglove
-from foxglove.websocket import (
-    Capability,
-    Client,
-    Parameter,
-    ParameterType,
-    ParameterValue,
-)
+from foxglove.websocket import Capability, Client, Parameter
 
 
 class ParameterStore(foxglove.websocket.ServerListener):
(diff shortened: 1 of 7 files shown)


python/foxglove-sdk-examples/ws-param-server/main.py
```diff
@@ -72,21 +66,21 @@ def main() -> None:
     foxglove.set_log_level(logging.DEBUG)
 
     initial_values: list[Parameter] = [
+        Parameter("p0"),
```
bryfox, line 69: Since there are breaking changes here, I think we'll want to bump the minimum version requirement in the example's pyproject.toml. I guess that'll be 0.7.0, but obviously that won't resolve until the next release. CI currently runs examples using the latest version (referenced by path), so maybe this will pass CI and we can go ahead with that?

Or we can stick to a backwards-compatible version (which looks like `raw()`) for now, and update examples in another release?

python/foxglove-sdk/python/foxglove/_foxglove_py/websocket.pyi
```diff
@@ -115,9 +120,30 @@ class Parameter:
...
         *,
-        type: Optional["ParameterType"] = None,
-        value: Optional["AnyParameterValue"] = None,
+        value: Optional["AnyNativeParameterValue"] = None,
     ) -> None: ...
+    @staticmethod
+    def raw(
+        name: str,
+        *,
+        type: Optional["ParameterType"],
```
bryfox, line 129: Seems like this should have a default value of `None`. Probably `value` should too, since it's optional, though I'm not sure what use cases this serves; maybe forcing `value=None` is appropriate.

python/foxglove-sdk/python/foxglove/_foxglove_py/websocket.pyi
```diff
@@ -115,9 +120,30 @@ class Parameter:
         self,
         name: str,
         *,
-        type: Optional["ParameterType"] = None,
-        value: Optional["AnyParameterValue"] = None,
+        value: Optional["AnyNativeParameterValue"] = None,
     ) -> None: ...
+    @staticmethod
+    def raw(
```
bryfox, line 126: I'm not sure about the name "raw"; a user doesn't need to be aware of the underlying implementation, so from that point of view, what's "raw" about these explicit types?

Could we make `new()` accept a union including these raw types as well, or is that problematic because of the explicit "type" argument?

bryfox (commented):
Much nicer, thanks — I guess I called this out in #231 but neglected to file an issue.

I left some questions about the example and interface, which I think boil down to whether we can have a backward-compatible change for now.

python/foxglove-sdk-examples/ws-param-server/main.py
gasmith: I reverted the compatibility breakage, so that we don't need to worry about this for now.

python/foxglove-sdk/python/foxglove/_foxglove_py/websocket.pyi
gasmith: Agreed. I've removed the "raw" nomenclature.

bryfox (approved):
👍

PR #390, opened 2025-04-26
Document AVIF image support

amacneil:
### Changelog
Document support for AVIF image format.


defunctzombie (approved):
📿 there aren't edge cases we aren't aware of with supporting this format

PR #391, opened 2025-04-28
Generate Timestamp & Duration schemas

bryfox:
### Changelog
None

### Docs
Docs PR forthcoming

### Description

This adds `Duration` and `Timestamp` to `FoxgloveMessageSchema` definitions (rather than treating them as primitives), so that we can begin documenting them as such in FG-11348.

Message schemas are now documented as part of the SDK (https://docs.foxglove.dev/docs/sdk/schemas), and this matches the "Timestamp" terminology used there.

In some cases, our types use the name "Time" rather than "Timestamp". That hasn't changed here; documentation will link to equivalent reference implementations.

For the most part, the generated schemas are unchanged. However:

- JSONSchemas now include Timestamp and Duration
- for FlatBuffers, I removed the documented `sec` range which was outside the range of int32.
- Some definitions now inherit the schema & field descriptions for the time-related types


python/foxglove-sdk/python/foxglove/schemas/__init__.py
```diff
@@ -100,8 +100,11 @@
     Vector3,
 ]
 
+
+Time = Timestamp
```
gasmith, line 104: I wonder whether we should do this the other way around. The `Timestamp` name feels like a protobuf-specific quirk, based on our decision to use that underlying well-known type for encoding.

gasmith (approved):

python/foxglove-sdk/python/foxglove/schemas/__init__.py
amacneil: I actually think `Timestamp` is probably a better name for it (worth doing an internal poll on it). E.g. would be nice to standardize on Timestamp everywhere in the schemas + SDK + app and not have an alias like this.

I'm not sure how practical it would be to update our flatbuffer and IDL schemas to use `Timestamp` instead of `Time` - the app could support both but it might cause some disruption.

typescript/schemas/src/internal/types.ts
```diff
@@ -1,6 +1,7 @@
 export type FoxglovePrimitive =
   | "string"
   | "float64"
+  | "int32"
   | "uint32"
   | "boolean"
   | "bytes"
```
amacneil, line 7: should we remove `time` and `duration` from this enum if they are now real schemas?

typescript/schemas/src/internal/types.ts
```diff
@@ -38,6 +39,7 @@ export type FoxgloveMessageSchema = {
   name: string;
   description: string;
   rosEquivalent?: keyof typeof import("@foxglove/rosmsg-msgs-common").ros1;
+  protoEquivalent?: "Timestamp" | "Duration";
```
amacneil, line 42: would be better for the docs if we referred to them by the fully qualified name like `google.protobuf.Timestamp`

typescript/schemas/src/internal/schemas.ts
```diff
@@ -1,5 +1,45 @@
...
+      description: "The number of nanoseconds in the positive direction",
+    },
+  ],
+};
+
+const Time: FoxgloveMessageSchema = {
+  type: "message",
+  name: "Time",
+  description: "A timestamp composed of seconds and nanoseconds",
+  rosEquivalent: "std_msgs/Time",
```
defunctzombie, line 27: How do we support ROS2 equivalent? which is `builtin_interfaces/Time` ?

defunctzombie (approved):
👍 one open question on how we handled the `ros2 equivalent` behavior

python/foxglove-sdk/python/foxglove/schemas/__init__.py
amacneil: @defunctzombie would it be possible to standardize our name here on `Timestamp` vs `Time`?

I guess in the app its currenty normalized to `Time`? Not sure what that affects tho.

defunctzombie: Standardize in what sense? I don't think we lean too heavily on the string "Time" vs "Timestamp" in the code that handles message parsing and schemas - it is more prevalent in documentation and various type names.

Here's a few places where we reference "time" types in the docs:

- https://docs.foxglove.dev/docs/visualization/message-schemas/built-in-types#time
- https://docs.foxglove.dev/extension-api/type-aliases/Time
- https://docs.foxglove.dev/extension-api/type-aliases/MessageEvent#receivetime

It would be quite a lift to change the `Time` type to `Timestamp` in the various apis - likely we would alias and look at moving over piece by piece so `Time` would never fully go away but we could use `Timestamp` as the type in typescript.

typescript/schemas/src/internal/types.ts
bryfox: It's not clear to me what "primitive" means since it looks like pretty much everything here represents time as a sec/nsec pair; maybe it's more of an indication to use "built-in" types?

We could remove it here and adjust all the code to generate identical schemas, but since this is a public export, I'm inclined to leave it alone for now.

typescript/schemas/src/internal/schemas.ts
bryfox: Good question.

https://github.com/foxglove/foxglove-sdk/blob/aa7ab0727e054900642ffe2cc218bf684f3a0a53/typescript/schemas/src/internal/generateRos.ts#L32

Rather than duplicating those hardcoded values in our documentation too, I think I'll add a `ros2Equivalent`.

python/foxglove-sdk/python/foxglove/schemas/__init__.py
gasmith: > Some inconsistency is unavoidable since we want to use native types where possible (e.g., ROS; protobuf).

What's the value of using the native protobuf Timestamp and Duration types, which have different field names, different component types, and different ranges of representable values? It just seems like a source of confusion.

gasmith: I'd like to challenge that a little.

If I'm writing with the foxglove SDK, and reading with some third-party tooling, I'd be confused by the fact that I logged my time in one format (the foxglove schema), and I'm reading it in a different format (the protobuf well-known types).

If I'm writing with third-party tooling, and reading with foxglove app, presumably the app already knows how to interpret either kind of timestamp. If not, it should. If we suppose reading MCAPs with the SDK as has been suggested, then I think it's the same argument as the paragraph above: I'd rather read exactly what I wrote, not a translation of it. Especially considering that foxglove timestamp/duration types are less expressive than the protobuf counterparts.

bryfox: I've captured the above points (and out-of-band discussion) in FG-11450. This is specifically related to serialization considerations in the SDK, and need not block documentation of Foxglove schemas. I think the naming question (Timestamp vs Time) remains unanswered, but I'll start a new thread for that.

bryfox:
We have one remaining question about naming — do we want to document the schema as `Time` or `Timestamp`? @amacneil [stated a preference for Timestamp](https://github.com/foxglove/foxglove-sdk/pull/391#discussion_r2071982760); an internal poll yielded no other votes. We use "Time" for documenting the extension API, and in user scripts. We use "Timestamp" in the SDK schema.

My reasoning for "Time" is in the description. I'll augment the documentation to describe some of the nuance here — naming, support for well-known types, differences with flatbuffers. I don't think an alias in the SDK (for discoverability) is a bad idea, but can remove that.

If we want to instead use "Timestamp", I'll close this PR and file an issue to first deal with that naming in the app where needed.

amacneil:
Copying my comment from slack

I think its worth renaming the schema to Timestamp and either updating the app or aliasing Time to Timestamp in the app.
this aligns better with industry terminology (timestamp == datetime, time == time sans date)

- [Protobuf Timestamp](https://protobuf.dev/reference/php/api-docs/Google/Protobuf/Timestamp.html)
- [Apache Arrow Timestamp](https://arrow.apache.org/docs/python/timestamps.html)
- [Postgres Timestamp](https://www.postgresql.org/docs/current/datatype-datetime.html)
- ros 2 calls the [schema Time and the field name ](https://docs.ros2.org/foxy/api/std_msgs/msg/Header.html)stamp - ros doing its own thing as always

bryfox:
I've updated the schema names to Timestamp. I kept the "Time" name in a couple of definitions to avoid a breaking change right now (see PR description), and to unblock the documentation work. I'll file follow-up issues for this.

python/foxglove-sdk/python/foxglove/schemas/__init__.py
```diff
@@ -111,6 +113,7 @@
     "CubePrimitive",
     "CylinderPrimitive",
     "Duration",
+    "Duration",
```
gasmith, line 116: Duplicate

gasmith (dismissed):

jtbandes:
What's the latest on this PR?

bryfox:
> What's the latest on this PR?

Internal discussions failed to drive consensus on either `Time` or `Timestamp` as the name. It's still on me to move this forward.

amacneil:
> Internal discussions failed to drive consensus on either Time or Timestamp as the name. It's still on me to move this forward.

For the sake of moving forward then lets go with `Timestamp`. App is still gonna need to accept both for backwards compatibility.

bryfox:
> For the sake of moving forward then lets go with Timestamp. App is still gonna need to accept both for backwards compatibility.

👍 I think this makes even more sense now given the new docs organization. I've brought this up-to-date with the main branch, and updated the description here. I have also updated a documentation PR to render previews for this proposed change.

typescript/schemas/src/internal/exportTypeScriptSchemas.ts
```diff
@@ -24,16 +25,18 @@ export function exportTypeScriptSchemas(
...
-  schemas.set("Duration", DURATION_TS);
   schemas.set("Time", TIME_TS);
 
   const allSchemaNames = [
     ...Object.values(foxgloveMessageSchemas),
     ...Object.values(foxgloveEnumSchemas),
-  ].sort((a, b) => a.name.localeCompare(b.name));
+  ]
+    .map((schema) => schema.name)
+    .concat(["Time"])
```
bryfox, line 35: Previously, neither Time nor Duration were exported from the index; if that is intentional, I can omit them.

python/foxglove-sdk/python/foxglove/schemas/__init__.py
jtbandes, line 1: it'd be nice to mark this and other generated files as linguist-generated with .gitattributes so they don't default to visible in diffs

typescript/schemas/src/internal/types.ts
```diff
@@ -33,11 +26,16 @@ export type FoxgloveMessageField = {
...
+// Flatbuffers and OMG IDL use "Time" instead of "Timestamp" for backwards compatibility.
 export type FoxgloveMessageSchema = {
   type: "message";
   name: string;
   description: string;
   rosEquivalent?: keyof typeof import("@foxglove/rosmsg-msgs-common").ros1;
+  ros2Equivalent?: keyof typeof import("@foxglove/rosmsg-msgs-common").ros2jazzy;
+  protoEquivalent?: "google.protobuf.Timestamp" | "google.protobuf.Duration";
+  flatbuffersEquivalent?: "Time";
+  omgIdlEquivalent?: "Time";
```
jtbandes: imo the "fooEquivalent" naming was intended to mean that the schema should be replaced with a well-known schema from that framework. For flatbuffers and omgidl, I don't think that really applies since we aren't using an existing upstream type — i.e. I wouldn't expect these equivalents to be something we are actually vending a schema for.

I think you could either call these e.g. `flatbuffersName/omgidlName` (since the name is what's changing) or you could hard-code that logic in generate.ts since it seems we already have some hard-coded logic related to Time & Duration in there

scripts/generate.ts
```diff
@@ -162,16 +166,25 @@ async function main({ clean }: { clean: boolean }) {
...
     await fs.mkdir(path.join(outDir, "flatbuffer"), { recursive: true });
     await fs.writeFile(path.join(outDir, "flatbuffer", "ByteVector.fbs"), BYTE_VECTOR_FB);
+    // The name "Time" is kept for backwards compatibility.
     await fs.writeFile(path.join(outDir, "flatbuffer", "Time.fbs"), TIME_FB);
     await fs.writeFile(path.join(outDir, "flatbuffer", "Duration.fbs"), DURATION_FB);
 
-    for (const schema of Object.values(foxgloveMessageSchemas)) {
+    // We use slightly different definitions for Time and Duration than the standard schema
+    // - Duration uses int32 for nsec instead of uint32
+    // - Time documents specific ranges
```
jtbandes: nit: I don't think this documentation is a valid reason to deviate in how they are generated...The comments were [copied from protobuf anyway](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/duration.proto#L102-L115). 😄

typescript/schemas/src/internal/generateOmgIdl.ts
```diff
@@ -10,32 +10,19 @@ function primitiveToIdl(type: Exclude<FoxglovePrimitive, "time" | "duration">) {
...
-};
-
-};
-`;
+export function omgIdlSchemaName(schema: FoxgloveSchema): string {
+  if (schema.type === "message" && schema.omgIdlEquivalent) {
+    return schema.omgIdlEquivalent;
+  }
+  return schema.name;
+}
```
jtbandes: why is this one done differently from flatbuffers?

scripts/generate.ts
```diff
@@ -162,16 +166,25 @@ async function main({ clean }: { clean: boolean }) {
   await logProgress("Generating FlatBuffer definitions", async () => {
     await fs.mkdir(path.join(outDir, "flatbuffer"), { recursive: true });
     await fs.writeFile(path.join(outDir, "flatbuffer", "ByteVector.fbs"), BYTE_VECTOR_FB);
+    // The name "Time" is kept for backwards compatibility.
     await fs.writeFile(path.join(outDir, "flatbuffer", "Time.fbs"), TIME_FB);
```
jtbandes: What type of backwards compatibility are we interested in? Should we offer both old & new names like you are doing for jsonschemas? If we use Timestamp then you would also be able to delete the custom logic around generating the import statements.

typescript/schemas/src/internal/generateFlatbufferSchema.ts
```diff
@@ -99,17 +100,19 @@ export function generateFlatbuffers(
...
-            break;
-          case "primitive":
-            if (field.type.name === "time") {
+            if (field.type.schema.name === "Timestamp") {
               type = "Time";
               imports.add(`Time`);
-            } else if (field.type.name === "duration") {
+            } else if (field.type.schema.name === "Duration") {
               type = "Duration";
               imports.add(`Duration`);
```
jtbandes: hm, any reason this needs to stay as Duration instead of foxglove.Duration?  If not, then you can delete this branch and the `else` branch will do the right thing

typescript/schemas/src/internal/generateJsonSchema.ts
```diff
@@ -1,50 +1,63 @@
...
+      },
+    };
+  }
+  if (schema.name === "Duration") {
+    return {
+      type: "object",
+      title: "duration",
+      properties: {
+        sec: { type: "integer" },
+        nsec: { type: "integer", minimum: 0, maximum: 999_999_999 },
```
jtbandes, line 40: I'm now wondering why we didn't allow these to be negative...maybe because of ROS?

It's unfortunate how the different frameworks are inconsistent :(
Hopefully at the end of the day the values are treated the same no matter where they came from?

typescript/schemas/src/internal/generateProto.ts
```diff
@@ -80,20 +82,20 @@ export function generateProto(
...
-        break;
-      case "primitive":
-        if (field.type.name === "time") {
+        if (field.type.schema.name === "Timestamp") {
           qualifiers.push("google.protobuf.Timestamp");
           imports.add(`google/protobuf/timestamp`);
-        } else if (field.type.name === "duration") {
+        } else if (field.type.schema.name === "Duration") {
           qualifiers.push("google.protobuf.Duration");
           imports.add(`google/protobuf/duration`);
```
jtbandes: would it make sense to reference protoEquivalent here instead?

typescript/schemas/src/internal/generateRos.ts
```diff
@@ -163,7 +174,11 @@ export function generateRosMsgDefinition(
       }
 
       case "nested":
-        if (field.type.schema.rosEquivalent != undefined) {
+        if (field.type.schema.name === "Timestamp") {
+          fieldType = "time";
+        } else if (field.type.schema.name === "Duration") {
+          fieldType = "duration";
+        } else if (field.type.schema.rosEquivalent != undefined) {
           fieldType = field.type.schema.rosEquivalent;
```
jtbandes, line 182: seems weird to me that we aren't using rosEquivalent for this

typescript/schemas/src/internal/generateSdkRustCTypes.ts
```diff
@@ -43,24 +41,37 @@ function toTitleCase(name: string) {
   return name.toLowerCase().replace(/(?:^|_)([a-z])/g, (_, letter: string) => letter.toUpperCase());
 }
 
+// We use special FoxgloveTimestamp and FoxgloveDuration types for the time and duration fields.
```
jtbandes, line 44: why not generate these too?

typescript/schemas/src/internal/generateTypeScript.ts
```diff
@@ -9,38 +9,36 @@ function primitiveToTypeScript(type: Exclude<FoxglovePrimitive, "time" | "durati
...
       return ["Float32Array", "Float64Array"];
+    case "int32":
+      return ["Int32Array"];
     case "uint32":
       return ["Uint32Array"];
   }
 }
 
+/**
+ * An alias to `Timestamp` for backwards compatibility.
```
jtbandes: not exactly an alias right now 😄   should it be?

typescript/schemas/src/internal/types.ts
```diff
@@ -1,11 +1,4 @@
-export type FoxglovePrimitive =
-  | "string"
-  | "float64"
-  | "uint32"
-  | "boolean"
-  | "bytes"
-  | "time"
-  | "duration";
```
jtbandes, line 8: after seeing the impact of this elsewhere, it feels like we went from having special-case logic to deal with the fact that time/duration were almost-but-not-quite-primitives, to now having special-case logic to deal with the fact that they are almost-but-not-quite-schemas. I think my takeaway would be that having them be a separate `{ type: "time" }` (i.e. not lumping them in with all other `nested` types) might have resulted in less special casing. But none of this really matters too much as long as the right results are generated, because this api is all internal anyway. (hence how we never have cleaned up the `type: { type: ...` wart

yarn.lock
```diff
@@ -764,6 +764,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@foxglove/message-definition@npm:^0.3.1":
```
jtbandes, line 477: let's try to avoid having 2 versions of this; a `yarn dedupe` might be enough but sometimes it takes manually hacking the yarn.lock file

yarn.lock
```diff
@@ -795,11 +802,21 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@foxglove/rosmsg@npm:^5.0.4":
```
jtbandes: also have 2 versions of this now

scripts/generate.ts
bryfox: Aha, that explains why the one comment was wrong. I guess it doesn't make sense to document different semantics here anyway, unless the app is doing something different. I'll switch over to a generated Time type but keep the Duration override for now because of the int type.

bryfox: Oh, actually, we use structs instead of tables for these. We could special-case the handling of the int type, and the table/struct generation if you think that's better. For now I'm going to leave things as they are (using the manually-defined types) and update the comment here.

scripts/generate.ts
bryfox: I'm assuming that flatbuffer codegen must produce different output if we change the names; they're effectively a new type that would break people's integrations.

I think that might be a step we want to take, but I would also like to avoid doing so in this PR.

The JSON schemas don't offer the old "Time" name. I added both Timestamp and Duration here, but we didn't output anything for those before.

typescript/schemas/src/internal/generateJsonSchema.ts
bryfox: It is unfortunate. I don't know why we used unsigned, but ROS is as good a guess as I could have.

> Hopefully at the end of the day the values are treated the same no matter where they came from?

I presume that's true for the app. I don't know how people are using these JSON schemas in particular, but restricting the range seems like a reasonable starting point, at least.

typescript/schemas/src/internal/generateRos.ts
bryfox: I was a little confused by this, and why we output "time" for ROS1 ([example](https://github.com/foxglove/foxglove-sdk/blob/9ccc0bb074288757df14c2c28f2b24a5f23d63b9/ros/src/foxglove_msgs/ros1/CameraCalibration.msg#L7)) instead of the rosEquivalent; I guess that's some internal ROS1 thing?

Now that we also have ros2Equivalent, you're right — using it makes more sense. I'll re-work it.

bryfox: It looks like we explicitly chose not to do this in https://github.com/foxglove/foxglove-sdk/pull/115, and we have tests asserting that the outputs here are specifically "time" and "duration" rather than the equivalent types.

I'm going to leave this alone unless you or @achim-k think there's no longer a reason for this.

typescript/schemas/src/internal/generateSdkRustCTypes.ts
bryfox: It'll require some more special casing, but should be doable. I opened FG-12637 as a follow-up.

typescript/schemas/src/internal/generateTypeScript.ts
bryfox: Oops, thanks — an earlier version of this did alias, but I backed that out to avoid changes to TS for now.

typescript/schemas/src/internal/types.ts
bryfox: Yeah, you're right. I think it's possible that we'll be able to remove some or most of this special casing down the line with a breaking release. At that point, I hope keeping in "nested" will feel reasonable.

yarn.lock
bryfox: Oh, thanks. This cleaned up a lot — should I add a `yarn dedupe --check` to CI?

typescript/schemas/src/internal/generateOmgIdl.ts
bryfox: No reason; I updated flatbuffers to work a little more like this

yarn.lock
jtbandes: Yeah maybe -- I think the challenge is that it makes dependabot PRs sometimes not merge cleanly, since dependabot sadly [doesn't run dedupe](https://github.com/dependabot/dependabot-core/issues/5830). I believe in some repos (maybe only internal repos?) we have a CI job that will run yarn dedupe on dependabot PRs.

typescript/schemas/src/internal/exportTypeScriptSchemas.ts
jtbandes: seems fine / probably desirable to have them exported

typescript/schemas/src/internal/generatePyclass.ts
```diff
@@ -39,6 +39,9 @@ export function generateSchemaPrelude(): string {
  * Generate a `pyclass`-annotated struct or enum definition for the given schema.
  */
 export function generatePyclass(schema: FoxgloveSchema): string {
+  if (!shouldGenerateSchemaClass(schema)) {
+    return "";
+  }
```
jtbandes, line 44: nit: is this needed if `filter` is being used elsewhere?

typescript/schemas/src/internal/schemas.ts
```diff
@@ -1,13 +1,55 @@
...
+  fields: [
+    {
+      name: "sec",
+      type: { type: "primitive", name: "int32" },
+      description: "The number of seconds in the duration",
+    },
+    {
+      name: "nsec",
+      type: { type: "primitive", name: "uint32" },
+      description: "The number of nanoseconds in the positive direction",
```
jtbandes, line 19: Doesn't have to happen in this PR, but I think we could use more detail on what it means by "in the positive direction", maybe some examples

jtbandes (approved):
Looks good!

It's probably not super critical, but the deduped yarn.lock still shows multiple versions of `@foxglove/rosmsg` and `@foxglove/message-definition` (i.e. new versions being added without the old ones being removed)

typescript/schemas/src/internal/generatePyclass.ts
bryfox: As-is, yeah; this is called in each of the three exported functions. We're generating a python class (.py), a stub (.pyi), and a module file (__init__.py). I could move this up to `generate`, but it seems like knowledge of that should live here.

typescript/schemas/src/internal/schemas.ts
bryfox: I copied this from our existing docs. It's a little weird since we're not consistent about signed/unsigned; I'll file a follow-up issue.

bryfox:
> the deduped yarn.lock still shows multiple versions of @foxglove/rosmsg and @foxglove/message-definition

I see; this is because I bumped `@foxglove/rosmsg-msgs-common` to get updated ros definitions. 070dfae bumps `@foxglove/rosmsg` (dev dependency) as well to bring these in line.

PR #392, opened 2025-04-29
Add readability-identifier-naming rules

bryfox:
### Changelog
c++: Breaking: parameters such as mcap writer options are now snake_case

### Description

This fixes clang-tidy settings to include our C++ header files as well, and fixes some of the resulting warnings.

This also adds `readability-identifier-naming` casing options to clang-tidy, which accounts for most of the changes here.

The settings generally Google's style guide but use `camelBack` for methods rather than initial caps. That's what we've favored so far; the main change here, then, is to use snake_case for variables, etc. The other change I added was to trailing underscore for private members; since there are some specific restrictions related to leading underscores, it seemed easier to enforce the latter.

The casing changes here do involve some breaking changes, and are up for debate.

I omitted global options, since I don't see a reason for us to include any.

### DIFF
--- a/cpp/.clang-tidy
+++ b/cpp/.clang-tidy
@@ -27,11 +27,11 @@ CheckOptions:
   readability-identifier-naming.AbstractClassCase: CamelCase
   readability-identifier-naming.ClassCase: CamelCase
   readability-identifier-naming.ClassConstantCase: CamelCase
-  readability-identifier-naming.ClassMemberCase: CamelCase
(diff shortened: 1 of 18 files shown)


gasmith (approved):
:ship:

bryfox:
Based on discussions elsewhere, I think this is the style we want to go with, but @eloff or @jtbandes if you disagree with anything public-facing here, I can revisit prior to our next release.

PR #394, opened 2025-04-30
Add c types for foxglove schemas and conversion from c types to native types

eloff:
This is part of adding support for the foxglove schemas for C++

This adds the generated C structs and types for the foxglove schemas. It also adds generated logging and conversion methods. They work by converting the C struct to a temporary Rust struct, wrapped in ManuallyDrop and using unsafe to borrow the C types as much as possible. This is not possible for arrays of nested schema structs, which are fairly common, because the C schema struct and Rust schema struct don't have the same layout, so we have to allocate a Vector and convert them.

Because some parts are allocated (owned) and some parts are not, we need a custom free function in some cases.

This can be improved on in the future: see FG-11381 for details.

There are no tests for this at current, there's a ticket to track that here: FG-11383

The C++ side of this is coming in the near future.


typescript/schemas/src/internal/generateRust.ts
```diff
@@ -0,0 +1,236 @@
...
+        }
+      }
+      return lines.join("\n");
+    })
+    .join("\n\n")}
+}
+
+impl ${name} {
+${name.endsWith("Primitive") ? "" : `
+  unsafe fn borrow_option_to_native(msg: Option<&Self>) -> Result<ManuallyDrop<foxglove::schemas::${name}>, foxglove::FoxgloveError> {
```
bryfox, line 113: Is it practical/useful to add Safety docs to these generated functions?

typescript/schemas/src/internal/generateRust.ts
```diff
@@ -0,0 +1,236 @@
...
+              return `${dstName}: unsafe { self.${srcName}.as_ref() }.map(|&m| m.into())`;
+            }
+            return `${dstName}: self.${srcName}`;
+          case "enum":
+            return `${dstName}: self.${srcName} as i32`;
+          case "nested":
+            return `${dstName}: unsafe { self.${srcName}.as_ref().map(|m| m.borrow_to_native()) }.transpose()?.map(ManuallyDrop::into_inner)`;
+        }
+      })
+      .join(",      \n")}
```
bryfox, line 159: nit: trailing spaces?

typescript/schemas/src/internal/generateRust.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  unsafe fn borrow_to_native(&self) -> Result<ManuallyDrop<foxglove::schemas::${name}>, foxglove::FoxgloveError> {
+    Ok(ManuallyDrop::new(foxglove::schemas::${name} {
+    ${fields
+      .map((field) => {
+        const srcName = escapeId(toSnakeCase(field.name));
+        const dstName = escapeId(toSnakeCase(field.name));
+        if (field.array != undefined) {
+          if (typeof field.array === "number") {
+            return `${dstName}: unsafe { Vec::from_raw_parts(self.${srcName}.as_ptr() as *mut ${(field.type.type === "primitive") ? primitiveToRust(field.type.name) : "todo!()"}, self.${srcName}.len(), self.${srcName}.len()) }`;
```
bryfox, line 129: can we replace the todo! with a TS assertion/error? Looks like we're not generating this (yet?).

bryfox (commented):
Coming together... left some minor comments.

c/src/generated_types.rs
```diff
@@ -0,0 +1,2927 @@
...
+
+#[unsafe(no_mangle)]
+pub extern "C" fn foxglove_channel_log_vector3(
+    channel: Option<&FoxgloveChannel>,
+    msg: Option<&Vector3>,
+    log_time: Option<&u64>,
+) -> FoxgloveError {
+    // Safety: we're borrowing from the msg, but discard the borrowed message before returning
+    match unsafe { Vector3::borrow_option_to_native(msg) } {
+        Ok(msg) => log_msg_to_channel(channel, &*msg, log_time),
```
gasmith, line 2921: Just to make sure I understand: This is safe because the msg is guaranteed to be entirely on the stack, correct?

c/src/generated_types.rs
```diff
@@ -0,0 +1,2927 @@
...
+                .filter_map(|m| unsafe {
+                    m.as_ref()
+                        .map(|m| m.borrow_to_native().map(ManuallyDrop::into_inner))
+                })
+                .collect::<Result<Vec<_>, _>>()?,
+            points: unsafe { std::slice::from_raw_parts(self.points, self.points_count) }
+                .iter()
+                .filter_map(|m| unsafe {
+                    m.as_ref()
+                        .map(|m| m.borrow_to_native().map(ManuallyDrop::into_inner))
```
gasmith, line 973: I think we may have a problem here. Suppose one of these `borrow_to_native()` calls fails, and it's not the first one in the array. The elements that we've successfully converted are no longer protected by `ManuallyDrop`, which means that they'll get dropped, and we'll double-free the nested structs.

It looks like we need to keep the `ManuallyDrop` until we've successfully converted all the elements. If we encounter an error, we need to rollback the successfully-converted elements with `free_points_annotation()`.

c/src/generated_types.rs
```diff
@@ -0,0 +1,2927 @@
...
+                })
+                .collect::<Result<Vec<_>, _>>()?,
+            texts: unsafe { std::slice::from_raw_parts(self.texts, self.texts_count) }
+                .iter()
+                .filter_map(|m| unsafe {
+                    m.as_ref()
+                        .map(|m| m.borrow_to_native().map(ManuallyDrop::into_inner))
+                })
+                .collect::<Result<Vec<_>, _>>()?,
```
gasmith, line 982: Another flavor of the same problem above. If we fail to convert a `TextAnnotation`, we have a `Vec<PointsAnnotation>` that we need to manually free.

typescript/schemas/src/internal/generateRust.ts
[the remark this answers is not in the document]
eloff: I guess this is technically targeting C. There's a generateCpp coming in the next PR.

typescript/schemas/src/internal/generateRust.ts
eloff: Yeah, I swapped it out for throwing an error, haven;t updated this branch yet

c/src/generated_types.rs
```diff
@@ -0,0 +1,2927 @@
...
+    }
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn foxglove_channel_log_camera_calibration(
+    channel: Option<&FoxgloveChannel>,
+    msg: Option<&CameraCalibration>,
+    log_time: Option<&u64>,
+) -> FoxgloveError {
+    // Safety: we're borrowing from the msg, but discard the borrowed message before returning
```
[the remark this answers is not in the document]
eloff: I've been doing it wrong for years apparently :laughing: 

I'll add safety docs to the methods

c/src/generated_types.rs
eloff: More specifically it's safe because it doesn't escape the function call, and everything it borrows from outlives the function call.

If the protobuf types had borrowed versions with a lifetime, we could express all of this in a way it could verify. This is really working around limitations of the prost crate (which somehow I find myself doing every single time I use the crate...)

c/src/generated_types.rs
eloff: Ughh, you're right

PR #395, opened 2025-04-30
rust: Rework websocket responder payload types

gasmith:
### Changelog
- rust: Breaking: Changed websocket FetchAsset and Service handler types

### Description
I originally proposed we use `bytes::Bytes` for websocket responders, which works fine for callback wrappers that need to return an owned/static value, but it's too restrictive for the underlying responder methods, which would be perfectly happy with any `T: AsRef<[u8]>`.

And so, the core of this change is updating [`service::Responder::respond`](https://github.com/foxglove/foxglove-sdk/pull/395/files#diff-0dc13b18578dee88e4a12ad6685a031c1b807dbe8ced9cb5151cc6253e93c9d4R52-R55) and [`AssetResponder::respond`](https://github.com/foxglove/foxglove-sdk/pull/395/files#diff-48005a706a82df3099ed9cd1dd2586acb02efc3d24d32abfd702dabc4173528dR80-R84) to accept a `T: AsRef<[u8]>`.

Since that change breaks compatibility (albeit in a mild way), we might as well go a bit further:
- If we accept an `E: Display`, we can simplify several callers.
- Callback wrappers can return a `T: AsRef<[u8]>`.

### DIFF
--- a/rust/foxglove/src/websocket/service/handler.rs
+++ b/rust/foxglove/src/websocket/service/handler.rs
@@ -1,6 +1,6 @@
-use std::{fmt::Display, future::Future, sync::Arc};
-
-use bytes::Bytes;
+use std::fmt::Display;
+use std::future::Future;
+use std::sync::Arc;
 
 use crate::websocket::service::{Request, Responder};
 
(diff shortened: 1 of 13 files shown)


rust/foxglove/src/websocket/service/handler.rs
```diff
@@ -28,76 +28,86 @@ pub trait SyncHandler: Send + Sync {
...
     fn call(&self, request: Request, responder: Responder) {
         let result = SyncHandler::call(self, request);
-        responder.respond(result.map_err(|e| e.to_string()));
+        responder.respond(result);
     }
 }
 
 /// A wrapper around a function that serves as a service call handler.
-pub(crate) struct HandlerFn<F, E>(pub F)
+pub(crate) struct HandlerFn<F, T, E>(pub F)
```
bryfox, line 49: nit: "T" was unclear to me without context below. This is the response type, so maybe "R" if we want to stick with single letters?

bryfox (approved):
Looks good to me

rust/foxglove/src/websocket/service/handler.rs
gasmith: I suppose we could; I guess I was thinking `T` as in `Result<T, E>` as it's schematized in the docs.

PR #396, opened 2025-04-30
 Flesh out derive macro for Encode

bryfox:
This picks up where #324 left off, and is based on that branch. I've opened a separate PR to aid review assuming a reader is already familiar with that one.

At a high level, this makes the following changes.

0. Rename Loggable→Encode
1. Support for additional protobuf field types (ints, bytes, etc); see protobuf.rs
2. Fix support for &str with lifetimes
3. Support generic structs
4. Do not panic if writing a large value (string, bytes) would overflow the buffer

I also re-added documentation (and an example) for deriving JSON schemas. I think this is still valuable for some use cases; for example, the `mcap` CLI supports `--json` formatting on messages.

Not addressed here, but could consider in a future PR:

- Putting this behind a "derive" feature
- Some mechanism to override the protobuf package or schema name (perhaps an attr macro?)

Note we will need to come up with a viable release plan prior to merging this feature. I think the derive crate will need to be published first, and awaited on, before we continue to publish the primary crate.


defunctzombie:
> * Putting this behind a "derive" feature

This would allow us to omit the "protobuf" dependency during install?

> I also re-added documentation (and an example) for deriving JSON schemas.

Does this also mean we depend on serde? Is the only reason we do that for json stuff?

Since `serde` is a popular package I think its viable to use it and tell people to derive it - but the requirement to then go learn _yet another_ package (schemars) for json schema stuff seems meh to me but maybe this package is also considered "common". I'd imagine we handle it in a manner similar to this Encode stuff were we maybe expose an alias and take more control over how users expose this data?

Overall I won't fight this JSON stuff too much but I think it is full of more pitfalls than avoiding it.

rust/examples/mcap/src/main.rs
```diff
@@ -41,13 +41,20 @@ impl From<CompressionArg> for Option<McapCompression> {
...
 }
 
+#[derive(Debug, serde::Serialize, schemars::JsonSchema)]
+struct JsonMessage {
+    msg: String,
+    count: u32,
+}
+
 static MSG_CHANNEL: LazyChannel<Message> = LazyChannel::new("/msg");
+static JSON_CHANNEL: LazyChannel<JsonMessage> = LazyChannel::new("/json");
```
defunctzombie, line 57: I am so/so on this. On one hand it is cool to see you can write binary and JSON to a file - on the other it makes the example more "noisy"

rust/foxglove/src/lib.rs
```diff
@@ -117,14 +117,15 @@
...
+//!
+//! The easiest way to do this is to derive the [`Encode`] trait, which will generate a schema
+//! and allow you to log your struct to a channel. This currently uses protobuf encoding.
 //!
 //! ```no_run
-//! #[derive(foxglove::Loggable)]
-//! struct Custom {
-//!     msg: String,
+//! #[derive(foxglove::Encode)]
+//! struct Custom<'a> {
```
defunctzombie, line 127: Why would someone do this instead of using the `String` type like before?

defunctzombie (commented):
:clap:

bryfox:
> > * Putting this behind a "derive" feature
> 
> This would allow us to omit the "protobuf" dependency during install?

It'd remove the dependency on the derive crate, but I think we depend on prost anyway. (I would probably also include this in the default features.)

> > I also re-added documentation (and an example) for deriving JSON schemas.
> 
> Does this also mean we depend on serde? Is the only reason we do that for json stuff?

We depend on serde-json for WS as well. It's possible that if you omitted the live-visualization feature, and didn't want JSON serialization for MCAP, we could make this optional.

> Since `serde` is a popular package I think its viable to use it and tell people to derive it - but the requirement to then go learn _yet another_ package (schemars) for json schema stuff seems meh to me but maybe this package is also considered "common". I'd imagine we handle it in a manner similar to this Encode stuff were we maybe expose an alias and take more control over how users expose this data?

In the common case, you don't really need to learn `schemars`; you derive JsonSchema and get our implementation. We could also consider putting this behind a feature, I suppose; I don't think that dependency is otherwise needed.

I agree with your feedback about adding JSON to the example; we should keep that simple. But I do think having JSON as an option is useful for tooling, and presumably broader type support (at least right now).

defunctzombie (approved):

rust/foxglove-derive/Cargo.toml
```diff
@@ -11,9 +11,11 @@ proc-macro2 = "1.0"
 quote = "1.0"
 syn = { version = "2.0", features = ["full", "extra-traits"] }
 prost-types.workspace = true
+tracing.workspace = true
```
gasmith, line 14: I don't think this belongs here, unless we're somehow using tracing at macro derivation time (which we shouldn't do).

rust/foxglove-derive/tests/scalar_fields_test.rs
```diff
@@ -1,60 +1,168 @@
...
+#[test]
+fn test_insufficient_bytes_buffer_warns() {
+    let test_struct = TestMessageBytes {
+        bytes: bytes::Bytes::from_static(&[1, 2, 3, 4]),
+    };
+    let mut buf = FixedSizeBuffer::with_capacity(3);
+    test_struct.encode(&mut buf).expect("Failed to encode");
+
+    assert!(logs_contain("Failed to write bytes"));
+    assert!(logs_contain("insufficient buffer capacity"));
```
gasmith, line 165: Maybe make a note about where these logs come from, since they don't originate from this crate.

rust/foxglove/src/protobuf.rs
```diff
@@ -143,20 +265,16 @@ impl ProtobufField for String {
...
-        // Write the length as a varint
-        let len = bytes.len();
-        let mut len_value = len as u64;
-        while len_value >= 0x80 {
-            buf.put_u8((len_value as u8) | 0x80);
-            len_value >>= 7;
+        // Write the length as a varint, followed by the data
+        if let Err(e) = prost::encoding::encode_length_delimiter(self.len(), buf) {
+            warn!("Failed to write string: {}", e);
+            return;
```
gasmith, line 271: It's inconsistent that we fail "gracefully" here, but panic elsewhere (e.g., the `bool` implementation that panics on OOM for `buf.put_u8(*self as u8)`).

At a higher level, I don't think that a warning is sufficient. If we fail to serialize the complete message, the buffer content is corrupt and should be discarded. Seems like if we want to handle this case, we need to plumb errors properly.

rust/foxglove-derive/src/lib.rs
```diff
@@ -136,7 +151,8 @@ fn derive_struct_impl(input: DeriveInput) -> TokenStream {
 
     // Generate the output tokens
     let expanded = quote! {
-        impl foxglove::ProtobufField for #name {
+        #[automatically_derived]
+        impl #impl_generics foxglove::ProtobufField for #name #ty_generics #where_clause {
```
gasmith, line 155: The fact that this works with generic types is a bit of a mind-blower, but I suppose it makes sense upon reflection (hah). Definitely precludes our ability to generate descriptors at compile time, but maybe that was too much of a pie-in-the-sky dream anyway.

rust/foxglove/src/protobuf.rs
bryfox: Propagating an error rather than warning makes sense.

You're also right about the inconsistency, but I'm a little torn on the right direction here. Dealing with the case of a potentially large buffer seemed useful; is it worth doing this check for every byte? I see our typed Channel implementation unwraps when encoding. Are there practical cases where a buffer is full but the logging system can otherwise make progress?

rust/foxglove-derive/src/lib.rs
```diff
@@ -52,12 +52,7 @@ fn derive_enum_impl(input: &DeriveInput, data: &DataEnum) -> TokenStream {
             }
 
             fn write(&self, buf: &mut impl ::foxglove::bytes::BufMut) {
-                let mut value = *self as u64;
-                while value >= 0x80 {
-                    buf.put_u8((value as u8) | 0x80);
-                    value >>= 7;
-                }
-                buf.put_u8(value as u8);
+                ::foxglove::encode_varint(*self as u64, buf);
```
defunctzombie, line 55: maybe worth namespacing all the custom protobuf encoding stuff into a `protobuf` namespace?

Or another trick I've seen recently - the `subtle` namespace:

https://v8docs.nodesource.com/node-20.3/d1/d40/namespacecppgc_1_1subtle.html

https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle

rust/foxglove/src/protobuf.rs
bryfox: I think we should make a call one way or the other, but I want to cap the scope of this stacked PR. I filed FG-11427 to track this.

rust/foxglove-derive/src/lib.rs
bryfox: Done (as `protobuf`)

bryfox:
I think I've addressed the feedback above, except for FG-11427 noted above.

I also fixed an issue with message and enum definitions being duplicated in .proto descriptors: https://github.com/foxglove/foxglove-sdk/pull/396/commits/fcc97da52365195af22c1f441678ea8e584d2584

I think this is ready for review & merge into the base. The last issue that _needs_ to be solved is a deployment strategy/update that makes the derive crate available.

rust/foxglove/src/protobuf.rs
gasmith: I think we need to make a choice before we release, because as written, this implementation leads to corruption, where we produce partial or completely invalid serializations. Note that the latter is possible because we may fail to extend a buffer for serializing a large object, `warn!` about it, but then subsequently succeed in extending the buffer for a smaller subsequent object in the same message.

bryfox: Done; I left the additional checks and tests in place (modified to panic) since we'll revisit this.

defunctzombie:
> The last issue that _needs_ to be solved is a deployment strategy/update that makes the derive crate available.

I looked at serde and prost and they both version their derive crates along with their other crates. Maybe look at some other notable frameworks that have been around a while and see if a pattern emerges. Or go with what serde and prost do - easy to stop doing it later.

rust/foxglove-derive/src/lib.rs
```diff
@@ -2,63 +2,61 @@ extern crate proc_macro;
...
 /// Derive macro for enums and structs allowing them to be logged to a Foxglove channel.
-#[proc_macro_derive(Loggable)]
+#[proc_macro_derive(Encode)]
 pub fn derive_loggable(input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as DeriveInput);
 
     match &input.data {
-        Data::Enum(_) => derive_enum_impl(input),
+        Data::Enum(data) => derive_enum_impl(&input, data),
         Data::Struct(_) => derive_struct_impl(input),
```
gasmith, line 17: Nit: We'd benefit from the same thing here, by capturing `Data::Struct(data)` and passing it to `derive_struct_impl()`.

rust/foxglove-derive/src/lib.rs
```diff
@@ -2,63 +2,61 @@ extern crate proc_macro;
...
 
-            fn write(&self, buf: &mut impl foxglove::bytes::BufMut) {
-                let mut value = *self as u64;
-                while value >= 0x80 {
-                    buf.put_u8((value as u8) | 0x80);
-                    value >>= 7;
-                }
-                buf.put_u8(value as u8);
+            fn write(&self, buf: &mut impl ::foxglove::bytes::BufMut) {
+                ::foxglove::protobuf::encode_varint(*self as u64, buf);
```
gasmith, line 55: I'm still astonished that this works properly for `TestEnum::ValueNeg`.

rust/foxglove/src/protobuf.rs
```diff
@@ -139,24 +270,16 @@ impl ProtobufField for String {
...
-        let len = bytes.len();
-        let mut len_value = len as u64;
-        while len_value >= 0x80 {
-            buf.put_u8((len_value as u8) | 0x80);
-            len_value >>= 7;
+        // Write the length as a varint, followed by the data
+        prost::encoding::encode_length_delimiter(self.len(), buf).expect("Failed to write string");
+        if buf.remaining_mut() < self.len() {
+            panic!("Failed to write string; insufficient buffer capacity");
         }
```
gasmith, line 281: I don't think you need to panic explicitly. `BufMut::put_slice()` will panic if you try to write beyond the end of the buffer.

The more I think about this, the more I feel that having an `encoded_len()` method to get the minimum encoded length is a hard requirement, because otherwise the caller doesn't know how to size the buffer. See prost's [`Message::encode_to_vec` implementation](https://docs.rs/prost/latest/src/prost/message.rs.html#63-71).

I know this PR is already quite large, so feel free to defer, but we should handle this before we release.

bryfox: I left the panics in place for the traced tests (though I guess only the bytes variant exercises that path). These will go away prior to merging main; I'm going to open PRs to address the above and also the crate publishing.

PR #397, opened 2025-04-30
C/C++: Add support for services

gasmith:
# Based on PR #386

## Chain of upstream PRs as of 2025-04-30

* PR #383:
  `main` ← `gasmith/fg-10680-cpp-params`

  * PR #386:
    `gasmith/fg-10680-cpp-params` ← `gasmith/fg-10680-cpp-params-sub`

    * **PR #397 (THIS ONE)**:
      `gasmith/fg-10680-cpp-params-sub` ← `gasmith/fg-10679-cpp-services`


cpp/foxglove/src/server/service.cpp
```diff
@@ -0,0 +1,122 @@
...
+  c->schema.data = reinterpret_cast<const uint8_t*>(this->schema.data);
+  c->schema.data_len = this->schema.data_len;
+}
+
+/**
+ * ServiceSchema implementation.
+ */
+void ServiceSchema::writeTo(
+  foxglove_service_schema* c,
+  // NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
```
bryfox, line 26: This seems like a reasonable lint; are you ignoring because (req, res) is common?

bryfox (commented):
Looking great!

I'm unable to build the services example, because of errors which appear to be described by https://github.com/nlohmann/json/issues/4490 — there's no longer a base template for std::char_traits, and it fails on std::byte. It looks like we're using the latest release of that lib, though, which contains that fix?

cpp/foxglove/src/server/service.cpp
gasmith: This lint was tripping before, but doesn't seem to be tripping anymore... so, hooray? ¯\_(ツ)_/¯

The lint is supposed to be complaining about the fact that there are two adjacent arguments of the same type (or convertible to the same type), and so it's easy to swap the ordering. Nothing to do with argument naming. The lint _should_ be tripping on this line.

In actual fact, I don't really care about the ordering. I just need buffer space - preferably on the stack. Maybe I can have the caller provide an array of two schemas.

gasmith:
> I'm unable to build the services example, because of errors which appear to be described by [nlohmann/json#4490](https://github.com/nlohmann/json/issues/4490) — there's no longer a base template for std::char_traits, and it fails on std::byte. It looks like we're using the latest release of that lib, though, which contains that fix?

Weird. Maybe try a make clean, if you haven't already? CI seems to be able to build with clang 18.1.3, and obviously I can too with clang 19.1.7.

eloff (approved):
LGTM, minor comments only

c/src/service.rs
```diff
@@ -0,0 +1,286 @@
...
+        drop(unsafe { FoxgloveService::from_raw(service) });
+    }
+}
+
+/// Overrides the default response encoding.
+///
+/// # Safety
+/// - `responder` must be a pointer to a `foxglove_service_responder` obtained via the
+///   `foxglove_service.handler` callback. This value is moved into this function, and must not
+///   accessed afterwards.
```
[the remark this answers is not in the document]
gasmith: Whoops, copy/paste error actually. Nice catch.

PR #403, opened 2025-05-05
C/C++: Add support for fetch asset handler

gasmith:
### Changelog
- C/C++: Add support for websocket fetch asset handler

### Description
This change adds support for websocket fetch asset callbacks in C & C++, with unit tests.


bryfox (approved):
👍 Nice, this looks good to me. I haven't run anything manually, but will comment once I figure out what's going on with the base branch for me.

c/src/fetch_asset.rs
```diff
@@ -0,0 +1,96 @@
...
+    /// # Safety
+    /// - The raw pointer must have been obtained from [`Self::into_raw`]
+    unsafe fn from_raw(ptr: *mut Self) -> Box<Self> {
+        unsafe { Box::from_raw(ptr) }
+    }
+}
+
+#[derive(Clone)]
+pub(crate) struct FetchAssetHandler {
+    context: *const c_void,
```
eloff: Same comment about the name, if you rename it in the previous PR don't forget this one as well

cpp/foxglove/include/foxglove/server/fetch_asset.hpp
```diff
@@ -0,0 +1,68 @@
...
+///
+/// This is the means by which a fetch asset implementation responds to a
+/// request from a client. Each request is paired with a unique responder
+/// instance, and must be used exactly once.
+class FetchAssetResponder final {
+public:
+  /// @brief Sends asset data to the client.
+  ///
+  /// @param data Asset data.
+  void respondOk(const std::vector<std::byte>& data) && noexcept;
```
eloff, line 22: A data pointer and length would be more flexible here, rather than requiring a vector

eloff (approved):
LGTM

Some minor comments only

cpp/foxglove/include/foxglove/server/fetch_asset.hpp
gasmith: Good point. Done!

PR #404, opened 2025-05-05
C++: Catch callback exceptions

gasmith:
If a user-provided callback in C++ raises an exception, that currently causes an inscrutable panic in rust followed by a SIGABRT. Instead, we should catch exceptions, log a message to stderr, and move on.

### DIFF
--- a/cpp/foxglove/src/server.cpp
+++ b/cpp/foxglove/src/server.cpp
@@ -4,7 +4,6 @@
 #include <foxglove/server.hpp>
 #include <foxglove/server/connection_graph.hpp>
 
-#include <iostream>
 #include <type_traits>
 
 namespace foxglove {
@@ -32,7 +31,7 @@ FoxgloveResult<WebSocketServer> WebSocketServer::create(
         try {
           (static_cast<const WebSocketServerCallbacks*>(context))->onSubscribe(channel_id);
         } catch (const std::exception& exc) {
-          std::cerr << "onSubscribe callback failed: " << exc.what() << "\n";
+          warn() << "onSubscribe callback failed: " << exc.what();
         }
       };
     }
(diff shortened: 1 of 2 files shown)


cpp/foxglove/src/server.cpp
```diff
@@ -62,28 +74,44 @@ FoxgloveResult<WebSocketServer> WebSocketServer::create(
...
+        }
       };
     }
     if (callbacks->onConnectionGraphUnsubscribe) {
       c_callbacks.on_connection_graph_unsubscribe = [](const void* context) {
-        (static_cast<const WebSocketServerCallbacks*>(context))->onConnectionGraphUnsubscribe();
+        try {
+          (static_cast<const WebSocketServerCallbacks*>(context))->onConnectionGraphUnsubscribe();
+        } catch (const std::exception& exc) {
+          std::cerr << "onConnectionGraphUnsubscribe callback failed: " << exc.what() << "\n";
```
bryfox, line 113: Would it be useful to prefix all of these with "Foxglove server" or similar?

bryfox (approved):

cpp/foxglove/src/server.cpp
gasmith: Good idea. Moreover, we should probably be consistent about how we issue warnings from the C++ shim, because these are not the only ones.

I took a stab at adding a `WarnStream`, which is basically a wrapper around an ostringstream, whose destructor shoves a prefixed & newlined message into `std::cerr`. If that looks good to you, we can apply it elsewhere too.

bryfox: 👍 that's an improvement. I'm now wondering if we should consider plumbing these through the C library (pending #405) so that we end up with the same formatting & filtering as anything from that library. But maybe that's more overhead than it's worth.

gasmith: Yeah! That idea occurred to me too. It'd be straightforward enough to replace that `std::cerr << ...` with a `foxglove_log_warning(...)` or `foxglove_log(FOXGLOVE_LOG_LEVEL_WARNING, ...)`. In fact, the more I think of it, the more I like it.

gasmith: We'll save that for another day though. https://linear.app/foxglove/issue/FG-11441

PR #405, opened 2025-05-06
C++: Provide an interface to set the level of SDK stderr logging

bryfox:
### Changelog
C++: Provide an interface to set the level of SDK stderr logging

### Description

This adds a C++ function to set the filter level for internal logging from the SDK.

Naming: corresponds to python's `set_log_level`. I originally considered including "severity" in the name to distinguish this from logging messages onto channels. The term "severity" is used in Boost's 'trivial' logging. Other alternatives considered: many permutations involving "stderr", "filter", "min", "severity", and "level".

I added these to foxglove.hpp/cpp. I think it's reasonable to have share some cross-cutting functionality in a single file, and I think we could add "error" here. (I also considered putting this in errors.hpp since they relate to stderr logging.) "config.hpp" is another option.

I also updated examples to make use of these as appropriate.


c/src/logging.rs
```diff
@@ -0,0 +1,36 @@
...
+pub enum FoxgloveLogSeverityLevel {
+    Off = 0,
+    Debug = 1,
+    Info = 2,
+    Warn = 3,
+    Error = 4,
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn foxglove_set_log_severity_level(level: FoxgloveLogSeverityLevel) {
```
gasmith, line 15: Let's add a docstring. We should indicate that this should be called at least once, early, during program initialization. We should also document the environment variables that the logger uses.

Maybe we should have a separate call for initializing logging? I guess it's not strictly necessary, as long as this one is well documented.

gasmith, line 15: I'll cast a vote for `foxglove_set_log_level`. I know we're trying to distinguish from logging on channels, but I don't think "severity" helps. If we ever introduce channel-log-levels, those could just as easily be interpreted as "severity". Maybe "internal", "console", or "debug" would be more effective. But I still think a well-written docstring would be sufficient.

gasmith, line 15: Since there's no canonical or de-facto standard C++ logging facility, I'm guessing we're going to need to expose more knobs in the future. A few off the top of my head:

- Log to stdout instead of stderr
- Disable colors
- Set format string

With `env_logger`, we get some of these controls (colors, at least) as environment variables that override the hardcoded configuration, and so maybe we don't need to plumb those through for now.

c/src/logging.rs
```diff
@@ -0,0 +1,36 @@
...
+    static INIT: Once = Once::new();
+    INIT.call_once(|| {
+        let initial_level = match &level {
+            FoxgloveLogSeverityLevel::Off => "off",
+            FoxgloveLogSeverityLevel::Debug => "debug",
+            FoxgloveLogSeverityLevel::Info => "info",
+            FoxgloveLogSeverityLevel::Warn => "warn",
+            FoxgloveLogSeverityLevel::Error => "error",
+        };
+        let env = env_logger::Env::default().default_filter_or(initial_level);
```
gasmith, line 25: Instead of the `default_filter_or` (which would use the `RUST_LOG` environment variable), let's do 

```rust
.filter_or("FOXGLOVE_LOG", initial_level)
.write_style_or("FOXGLOVE_LOG_STYLE", "auto")
```

c/src/logging.rs
```diff
@@ -0,0 +1,36 @@
...
+        env_logger::init_from_env(env);
```
gasmith, line 26: For controlling stdout/stderr, we can use [`Builder::target`](https://docs.rs/env_logger/latest/env_logger/struct.Builder.html#method.target):

```rust
Builder::from_env(env).target(...).init()
```

Eventually that's where we'd plug in the formatting overrides too, but I think we can save that for another day.

c/src/logging.rs
```diff
@@ -0,0 +1,36 @@
...
+    });
+
+    log::set_max_level(match level {
+        FoxgloveLogSeverityLevel::Off => LevelFilter::Off,
+        FoxgloveLogSeverityLevel::Debug => LevelFilter::Debug,
+        FoxgloveLogSeverityLevel::Info => LevelFilter::Info,
+        FoxgloveLogSeverityLevel::Warn => LevelFilter::Warn,
+        FoxgloveLogSeverityLevel::Error => LevelFilter::Error,
+    });
```
gasmith, line 35: Does this actually work? I'm pretty sure that the following usage will only result in error-level logging:

```c
foxglove_set_log_level(FOXGLOVE_LOG_LEVEL_ERROR);
foxglove_set_log_level(FOXGLOVE_LOG_LEVEL_DEBUG);
foxglove_server_start(&options);
```

On the flip side, I'm not sure how important it is to control log level dynamically at runtime. It's probably fine to just offer a function that initializes the logging facility with a static log level.

c/src/logging.rs
bryfox: Great idea; maybe "FOXGLOVE_LOG_LEVEL" for the former name?

c/src/logging.rs
bryfox: Thanks; it does not. For now, I updated the docs to indicate subsequent calls have no effect.

bryfox:
@gasmith thanks — I think I've addressed the feedback above.

To allow configuration via environment variables, I hooked into the existing "register" function which is called when setting up a sink. I'm not sure if that's exactly what you had in mind, and I'm happy to revisit.

c/src/logging.rs
```diff
@@ -0,0 +1,57 @@
...
+/// Initialize SDK logging with the given severity level.
+///
+/// The SDK logs informational messages to stderr. Any messages below the given level are not
+/// logged.
+///
+/// This function should be called before other Foxglove initialization to capture output from all
+/// components. Subsequent calls will have no effect.
+///
+/// Log level may be overridden with the FOXGLOVE_LOG_LEVEL environment variable: "debug", "info",
+/// "warn", or "error".
```
gasmith, line 21:
```suggestion
/// Log level may be overridden with the FOXGLOVE_LOG_LEVEL environment variable: "debug", "info",
/// "warn", "error", or "off". The default level is "info".
```

cpp/foxglove/include/foxglove/foxglove.hpp
```diff
@@ -0,0 +1,40 @@
...
+///
+/// The SDK logs informational messages to stderr. Any messages below the given level are not
+/// logged. Note that this does not affect logging of messages to MCAP or Foxglove.
+///
+/// This function should be called before other Foxglove initialization to capture output from all
+/// components. Subsequent calls will have no effect.
+///
+/// As long as you initialize one logging sink (WebSocket server or MCAP), log level may instead be
+/// configured via a `FOXGLOVE_LOG_LEVEL` environment variable, with one of the values "debug",
+/// "info", "warn", or "error". Default is "info".
```
gasmith, line 28:
```suggestion
/// "info", "warn", "error", or "off". Default is "info".
```

c/src/lib.rs
```diff
@@ -643,6 +645,13 @@ pub unsafe extern "C" fn foxglove_context_free(context: *const FoxgloveContext)
 #[unsafe(no_mangle)]
 pub extern "C" fn foxglove_internal_register_cpp_wrapper() {
     foxglove::library_version::set_sdk_language("cpp");
+
+    let log_config = std::env::var("FOXGLOVE_LOG_LEVEL")
+        .or_else(|_| std::env::var("FOXGLOVE_LOG_STYLE"))
+        .ok();
+    if log_config.is_some() {
+        foxglove_set_log_level(logging::FoxgloveLogLevel::Debug);
```
gasmith, line 653: I think we want info by default.

```suggestion
        foxglove_set_log_level(logging::FoxgloveLogLevel::Info);
```

gasmith (approved):
LG!

PR #406, opened 2025-05-06
C/C++: Fix deleted operator warnings for Channel

gasmith:


bryfox:
OK, I see this with clang-18.

gasmith:
I'm running clang 19.1.7.

PR #407, opened 2025-05-06
Document C++ interface

bryfox:
### Changelog
C++: Minor breaking: remove `testId` from Channel.

### Description

This adds documentation for the public C++ interface, and a doxygen config file to allow generation of basic documentation. 2baef9 shows the actual config file changes from the default generated file.

We're not locked in to using Doxygen; other tools support at least subset of the syntax. I opted for "///"-style comments which is familiar, and slightly easier to copy from Rust. Note one advantage of using javadoc ("/**") or qt-style ("/*!") blocks is that Doxygen has an "autobrief" config that lets us omit the `@brief` command.

I also removed `testId()` from channel, which was never called, and was identical to `id()` regardless.

Future PRs will augment documentation and add generation & deployment steps to CI.

This supports FG-11290.


cpp/foxglove/include/foxglove/channel.hpp
```diff
@@ -13,34 +13,67 @@ struct foxglove_context;
...
 
   Channel(const Channel&) = delete;
   Channel& operator=(const Channel&) = delete;
 
+  /// @brief Default move constructor.
   Channel(Channel&& other) noexcept = default;
+  /// @brief Default move assignment.
   Channel& operator=(Channel&& other) noexcept = default;
-
+  /// @brief Default estructor.
```
gasmith, line 76:
```suggestion
  /// @brief Default destructor.
```

cpp/foxglove/include/foxglove/channel.hpp
```diff
@@ -13,34 +13,67 @@ struct foxglove_context;
...
 
+  /// @brief Uniquely identifies a channel in the context of this program.
+  ///
+  /// @return The ID of the channel.
   [[nodiscard]] uint64_t id() const;
 
   Channel(const Channel&) = delete;
   Channel& operator=(const Channel&) = delete;
 
+  /// @brief Default move constructor.
```
gasmith, line 72: Do we need to document default implementations explicitly?

cpp/foxglove/include/foxglove/mcap.hpp
```diff
@@ -12,45 +12,74 @@ struct foxglove_context;
...
 
+/// @brief An MCAP writer, used to log messages to an MCAP file.
 class McapWriter final {
 public:
+  /// @brief Create a new MCAP writer.
+  /// @param options The options for the MCAP writer.
+  /// @return A new MCAP writer.
   static FoxgloveResult<McapWriter> create(const McapWriterOptions& options);
 
+  /// @brief Stops logging events, flushes buffered data, and returns the writer.
```
gasmith, line 73: Doesn't return the writer.

cpp/foxglove/include/foxglove/server.hpp
```diff
@@ -15,16 +15,28 @@ struct foxglove_connection_graph;
...
+  /// @brief The schema of the channel.
   const std::byte* schema;
+  /// @brief The length of the schema of the channel.
   size_t schema_len;
 };
 
+/// @brief The capabilities of a WebSocket server.
+///
+/// A server may advertise certain capabilities to clients and provide related functionality
+/// in WebSocketServer::callbacks.
```
gasmith, line 39:
```suggestion
/// in WebSocketServerCallbacks.
```

cpp/foxglove/include/foxglove/server.hpp
```diff
@@ -54,39 +66,85 @@ inline WebSocketServerCapabilities operator&(
   return WebSocketServerCapabilities(uint8_t(a) & uint8_t(b));
 }
 
+/// @brief The callback interface for a WebSocket server.
```
gasmith, line 69: We should mention that these are invoked by the websocket client's poll loop, and that the implementations should be as low-latency as possible. Whatever wording we used in rust should apply here, minus references to tokio::spawn_blocking etc.

gasmith (approved):

cpp/foxglove/include/foxglove/channel.hpp
bryfox: Sort of; I'd like to leave "warn as error" enabled, and I wasn't find able to find a way to ignore these. I think we could write symbols into the Doxyfile, but that seems cumbersome.

It turns out the default destructors don't warn on missing docs, so I removed those.

PR #408, opened 2025-05-07
Add encoding_len to ProtobufField

bryfox:
This builds on #324, and is a follow-up to #396.

The primary change is to add a required `encoded_len` method to the ProtobufField trait. This provides the length of encoded data to `Encode::encode`, which will now return an error if the remaining buffer is too short to hold the written data.

This also adds a compiler check for unit-only enums; any variant with fields will now provide a nicer error message.

This also updates the documentation, and notes the protobuf limitation around enum values.

This also fixes the encoding of tags, which were previously written as a single byte. I capped the derived field numbers to 2047, which makes it a little easier to think about encoded_len and avoids thinking about reserved ranges or other limits.

### DIFF
--- a/rust/foxglove-derive/src/lib.rs
+++ b/rust/foxglove-derive/src/lib.rs
@@ -291,8 +291,7 @@ fn derive_struct_impl(input: &DeriveInput, data: &DataStruct) -> TokenStream {
             }
 
             fn encode(&self, buf: &mut impl ::foxglove::bytes::BufMut) -> Result<(), Self::Error> {
-                let total_len = 0 #(+ #field_message_lengths)*;
-                if total_len > buf.remaining_mut() {
+                if self.encoded_len().is_some_and(|len| len > buf.remaining_mut()) {
                     return Err(::foxglove::FoxgloveError::EncodeError(
                         "insufficient buffer".to_string(),
                     ));
@@ -301,6 +300,10 @@ fn derive_struct_impl(input: &DeriveInput, data: &DataStruct) -> TokenStream {
                 // The top level message is encoded without a length prefix
                 #(#field_encoders)*
                 Ok(())
+            }
+
+            fn encoded_len(&self) -> Option<usize> {
+                Some(#tags_encoded_len #(+ #field_message_lengths)*)
             }
         }
     };
--- a/rust/foxglove-derive/src/lib.rs
+++ b/rust/foxglove-derive/src/lib.rs
@@ -121,6 +121,15 @@ fn derive_struct_impl(input: &DeriveInput, data: &DataStruct) -> TokenStream {
     let mut field_encoders = Vec::new();
     let mut field_message_lengths = Vec::new();
 
+    // Field number + wire type must fit into a u32, but there is also a much lower reserved
+    // range starting at 19,000. We should need to encode a much smaller space in practice; if
+    // we limit to 2047, then each encoded tag will take at most two bytes.
+    // Fields 1-15 use a single byte for the tag; fields 16-2047 use two bytes.
+    // https://protobuf.dev/programming-guides/proto3/#assigning
+    let max_one_byte_field_number = 15;
+    let max_field_number = 2047;
+    let mut tags_encoded_len: usize = 0;
+
     // If a struct nests multiple values of the same enum or message type, we
     // only define them once, based on name.
     let mut enum_defs: HashMap<&syn::Type, proc_macro2::TokenStream> = HashMap::new();
@@ -131,9 +140,21 @@ fn derive_struct_impl(input: &DeriveInput, data: &DataStruct) -> TokenStream {
         let field_type = &field.ty;
         let field_number = i as u32 + 1;
 
+        if field_number > max_field_number {
+            return TokenStream::from(quote! {
+                compile_error!("Too many fields to encode");
+            });
+        }
+
         field_message_lengths.push(quote! {
             ::foxglove::protobuf::ProtobufField::encoded_len(&self.#field_name)
         });
+
+        tags_encoded_len += if field_number <= max_one_byte_field_number {
+            1
+        } else {
+            2
+        };
 
         enum_defs.entry(field_type).or_insert_with(|| quote! {
             if let Some(enum_desc) = <#field_type as ::foxglove::protobuf::ProtobufField>::enum_descriptor() {
@@ -230,7 +251,7 @@ fn derive_struct_impl(input: &DeriveInput, data: &DataStruct) -> TokenStream {
             }
 
             fn encoded_len(&self) -> usize {
-                0 #(+ #field_message_lengths)*
+                #tags_encoded_len #(+ #field_message_lengths)*
             }
         }
 
--- a/rust/foxglove-derive/src/lib.rs
+++ b/rust/foxglove-derive/src/lib.rs
@@ -78,7 +78,7 @@ fn derive_enum_impl(input: &DeriveInput, data: &DataEnum) -> TokenStream {
             }
 
             fn encoded_len(&self) -> usize {
-                prost::encoding::encoded_len_varint(*self as u64)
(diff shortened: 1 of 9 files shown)


rust/foxglove/src/protobuf.rs
```diff
@@ -151,6 +172,10 @@ impl ProtobufField for u8 {
     fn write(&self, buf: &mut impl bytes::BufMut) {
         encode_varint((*self).into(), buf);
     }
+
+    fn encoded_len(&self) -> usize {
+        prost::encoding::encoded_len_varint(*self as u64)
```
defunctzombie, line 177: Is a u8 ever written as a varint?

rust/foxglove/src/protobuf.rs
```diff
@@ -233,6 +282,10 @@ impl ProtobufField for bool {
     fn write(&self, buf: &mut impl bytes::BufMut) {
         buf.put_u8(*self as u8);
     }
+
+    fn encoded_len(&self) -> usize {
+        prost::encoding::encoded_len_varint(*self as u64)
```
defunctzombie, line 287: bool would always be 1 byte?

rust/foxglove-derive/src/lib.rs
```diff
@@ -68,6 +76,10 @@ fn derive_enum_impl(input: &DeriveInput, data: &DataEnum) -> TokenStream {
             fn type_name() -> Option<String> {
                 Some(stringify!(#name).to_string())
             }
+
+            fn encoded_len(&self) -> usize {
+                ::foxglove::protobuf::encoded_len_varint(*self as u64)
```
defunctzombie, line 81: Why does this cast self to u64?

gasmith:
> https://github.com/foxglove/foxglove-sdk/blob/main/rust/foxglove/src/encode.rs#L31-L36
> 
> The comments here make this optional but it seems like you are saying it should not be optional? or that it is important to provide this?

It should probably be mandatory. The `Channel::log` implementation [panics](https://github.com/foxglove/foxglove-sdk/blob/3f2fcd2e84f9d32ea37ae549e4eba780ef96d9e7/rust/foxglove/src/channel.rs#L132-L138) for any message that requires more than 256k bytes, but does not implement `Encode::encoded_len()`.

eloff:
> > https://github.com/foxglove/foxglove-sdk/blob/main/rust/foxglove/src/encode.rs#L31-L36
> 
> > 
> 
> > The comments here make this optional but it seems like you are saying it should not be optional? or that it is important to provide this?
> 
> 
> 
> It probably should be mandatory, or at least the documentation should be clear about the situations in which it is mandatory. The `Channel::log` implementation [panics](https://github.com/foxglove/foxglove-sdk/blob/3f2fcd2e84f9d32ea37ae549e4eba780ef96d9e7/rust/foxglove/src/channel.rs#L132-L138) for any message that requires more than 256k bytes, but does not implement `Encode::encoded_len()`.

That's probably a limitation that could be lifted.

gasmith:
> > It probably should be mandatory, or at least the documentation should be clear about the situations in which it is mandatory. The `Channel::log` implementation [panics](https://github.com/foxglove/foxglove-sdk/blob/3f2fcd2e84f9d32ea37ae549e4eba780ef96d9e7/rust/foxglove/src/channel.rs#L132-L138) for any message that requires more than 256k bytes, but does not implement `Encode::encoded_len()`.
> 
> That's probably a limitation that could be lifted. I could reuse the Arena I implemented for the c++ schemas PR, I'll make the change there.

I misspoke, actually. It looks like `SmallBytes` implementation of `BufMut` does resize the buffer dynamically. I don't know why I thought otherwise.

eloff:
> > > It probably should be mandatory, or at least the documentation should be clear about the situations in which it is mandatory. The `Channel::log` implementation [panics](https://github.com/foxglove/foxglove-sdk/blob/3f2fcd2e84f9d32ea37ae549e4eba780ef96d9e7/rust/foxglove/src/channel.rs#L132-L138) for any message that requires more than 256k bytes, but does not implement `Encode::encoded_len()`.
> > 
> > 
> > That's probably a limitation that could be lifted. I could reuse the Arena I implemented for the c++ schemas PR, I'll make the change there.
> 
> I misspoke, actually. It looks like `SmallBytes` implementation of `BufMut` does resize the buffer dynamically. I don't know why I thought otherwise.

I came here to say this, I looked into the code, and it handles the case just fine, using encoded_len as an optimization hint, but not requiring that it's accurate.

So all we need to do is update the comment to specify that it's optional, but providing a good estimate, and erring on the side of over-estimating, is important to serialization performance.

rust/foxglove/src/protobuf.rs
bryfox: Yes, and anything above 127 (highest bit set) will use two bytes. I added a test.

rust/foxglove/src/protobuf.rs
```diff
@@ -219,6 +262,12 @@ impl ProtobufField for i8 {
         let encoded = ((n << 1) ^ (n >> 7)) as u64;
         encode_varint(encoded, buf);
     }
+
+    fn encoded_len(&self) -> usize {
+        let n = *self;
+        let encoded = ((n << 1) ^ (n >> 7)) as u64;
+        prost::encoding::encoded_len_varint(encoded)
```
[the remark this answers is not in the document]
bryfox: Yes, up to 10 bytes. This is currently using [zig-zag encoding](https://protobuf.dev/programming-guides/encoding/#signed-ints) (documented here as "Signed integers are encoded using sint32 or sint64."). An alternative is the standard/two's complement encoding, where every negative number takes up 10 bytes. I added a test here too.

rust/foxglove/src/protobuf.rs
bryfox: Yes, updated 👍

rust/foxglove-derive/src/lib.rs
bryfox: The cast came from the original implementation of `write` in the base branch, which was basically doing the same thing as prost, and prost's interface accepts u64. It works for the standard varint encoding because both [protobuf](https://protobuf.dev/programming-guides/encoding/#signed-ints) and [rust](https://doc.rust-lang.org/book/ch03-02-data-types.html#integer-types) use two's complement for signed ints. Should we do something more explicit here?

bryfox:
[d03e1ea](https://github.com/foxglove/foxglove-sdk/pull/408/commits/d03e1ea8e11cf7454142b48ffc7cbaf51fea01d3) fixes the tag encoding (see updated description). I think I've addressed feedback above. The remaining issue is to solve the crate deployment, which I'll deal with in a separate PR.

rust/foxglove/src/protobuf.rs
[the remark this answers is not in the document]
eloff: Shouldn't a zig-zag encoded i8 be at most 2 bytes? (only when you need 8 bits to represent the encoded integer)

gasmith (approved):

rust/foxglove/src/protobuf.rs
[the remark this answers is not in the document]
eloff:
```suggestion
        if encoded < 128 { 1 } else { 2 }
```

I think you could even avoid doing the zig-zag encoding by checking n >= -64 && n < 64 (double check I don't have an off-by-one error there)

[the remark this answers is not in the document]
bryfox: > Shouldn't a zig-zag encoded i8 be at most 2 bytes?

Yes, thank you! I had confused myself in some way. The encoding was wrong here, as was my subsequent test, since the ints weren't wide enough for the shift. Fixed in 215345c and added more tests.

> having a consistent, readable implementation is probably better

I think I agree with this, but I'm happy to file an issue to benchmark and implement if you think that's worthwhile.

PR #409, opened 2025-05-07
rust: Add shutdown handle for websocket server

gasmith:
### Changelog
- Breaking: rust: Add shutdown handle for websocket server

### Description
We need a way to wait for the websocket server to gracefully shutdown. This is generally useful, but particularly important in C & C++ where we need to ensure that there will be no future callbacks, so that we can clean up resources.

This change accomplishes that goal by spawning client poll tasks within a `JoinSet` on the server. When stop is called, we extract the `JoinSet` into a `ShutdownHandle` and return it to the user. The user can choose to wait for the tasks to complete, detach them to finish in the background, or drop the handle to abort them immediately.

The C and python implementations now block waiting for the server to shutdown gracefully.


bryfox (approved):
👍

PR #410, opened 2025-05-08
Generate C++ documentation in CI

bryfox:
This adds Doxygen generation to the  C++ CI workflow. We may build documentation around this rather than using the HTML output directly, but it will fail CI if we're missing any public documentation.

Supports FG-11290.

### DIFF
--- a/.github/workflows/c_cpp.yml
+++ b/.github/workflows/c_cpp.yml
@@ -26,6 +26,16 @@ jobs:
       - run: make CLANG_TIDY=true build
         working-directory: cpp
 
+  docs:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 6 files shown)


.github/workflows/c_cpp.yml
```diff
@@ -26,6 +26,16 @@ jobs:
       - run: make CLANG_TIDY=true build
         working-directory: cpp
 
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ssciwr/doxygen-install@v1
```
bryfox, line 33: https://github.com/ssciwr

gasmith (approved):

PR #411, opened 2025-05-08
C++: Use ccache, where available

gasmith:
Not sure why I didn't do this _before_ working on a giant stack of C++ PRs.


eloff (approved):
Yeah, we should have thought of this sooner, oh well. LGTM

PR #413, opened 2025-05-08
Add foxglove schemas for C++

eloff:
This PR combines the Rust and C types from the earlier PR with the C++ code.
FG-11372 and FG-11373. It adds generated foxglove schema and channel types for C++.

This PR is not as scary as it seems, half the file changes are generated files changing because I added descriptions to the the enum fields. A lot of the C++ changes are just renaming Channel to RawChannel. Most of the lines added are generated, understand part of it and you understand it all.

I would start with channels.hpp to remind yourself what Channel looks like in C++ (now renamed to RawChannel) schemas.hpp, then schemas.cpp, then generated_types.rs. After that generateSdkRustCTypes.ts and generateSdkCpp.ts, which generate those files. That's the meat of it.

By having a dedicated channel type per schema type, we avoid the overhead of dynamic dispatch (e.g. based on an enum with a giant switch statement or virtual function table or similar).

It adds conversion code from C++ -> C. This may need to allocate in cases where the C type has a different layout, so it also adds a stack-based arena allocator used for those temporary allocations. The arena is simply discarded when the function returns, as the converted type doesn't outlive the function call.

The C -> Rust conversion code works the same way, but with lots of unsafe code and ManuallyDrop to shoehorn borrowed data into owned types without dropping it. Again a stack based arena is used (which is pinned so it isn't accidentally moved). It should be possible to optimize this in the future for types where the C layout of the protobuf struct could be the same if we add repr(c) to it, we could avoid having to do this conversion for those types (e.g. Color). The C++ code already does this optimization, see isSameAsCType in generateSdkCpp.ts.

One gotcha is the Rust conversions can fail (e.g. invalid utf-8) so we have to do all fallable conversions first, keeping the results inside MnaullyDrop, before constructing the final ManuallyDrop<T> result.

We return the typed channel to C as a raw channel pointer, and then cast it back to the appropriate typed channel in the log call.

The stack based allocators have a fixed buffer size of 128kb. They fall back to heap allocation if they run out of space, tracking and freeing the allocation when the arena is cleaned up. This seems extremely unlikely as they're used only for small vectors and structs. I've implemented many memory allocators in the past, but this is the first time I've been able to include any in a PR at work!

The gist of this is we do some copying/transforming while other fields are borrowed as-is (e.g. string, bytes, some vectors), but no heap allocations and freeing on the logging path, apart from what is needed to build the message schema object itself in C++.

I added a couple cherry-picked tests - it's not quite as good as generating comprehensive tests, but this has dragged on for so long I don't feel I have the luxury of time to do that.

Issue for more tests: FG-11383
Issue for example(s): FG-11374

Ideas welcome!


cpp/foxglove/include/foxglove/schemas.hpp
```diff
@@ -0,0 +1,2081 @@
...
+  /// @brief y coordinate length
+  double y = 1;
+};
+
+/// @brief A field present within each element in a byte array of packed elements.
+struct PackedElementField {
+  /// @brief Numeric type
+  enum class NumericType : uint8_t {
+    /// @brief Unknown numeric type
+    UNKNOWN = 0,
```
gasmith, line 374: These should be CamelCase. Not sure why clang-tidy isn't catching that.

cpp/foxglove/include/foxglove/schemas.hpp
```diff
@@ -0,0 +1,2081 @@
...
+  ///
+  /// @return The ID of the channel.
+  [[nodiscard]] uint64_t id() const;
+
+  CameraCalibrationChannel(const CameraCalibrationChannel& other) noexcept = delete;
+  CameraCalibrationChannel& operator=(const CameraCalibrationChannel& other) noexcept = delete;
+  /// @brief Default move constructor.
+  CameraCalibrationChannel(CameraCalibrationChannel&& other) noexcept = default;
+  /// @brief Default destructor.
+  ~CameraCalibrationChannel() = default;
```
gasmith, line 966: Needs move assignment

cpp/foxglove/include/foxglove/schemas.hpp
```diff
@@ -0,0 +1,2081 @@
...
+    const std::string_view& topic, const Context& context = Context()
+  );
+
+  /// @brief Log a message to the channel.
+  ///
+  /// @param value The CameraCalibration message to log.
+  /// @param log_time The timestamp of the message. If omitted, the current time is used.
+  FoxgloveError log(
+    const CameraCalibration& value, std::optional<uint64_t> log_time = std::nullopt
+  );
```
gasmith, line 954: cons (but not noexcept, since Arena::alloc can throw std::bad_alloc).

cpp/foxglove/include/foxglove/schemas.hpp
```diff
@@ -0,0 +1,2081 @@
...
+
+  /// @brief Uniquely identifies a channel in the context of this program.
+  ///
+  /// @return The ID of the channel.
+  [[nodiscard]] uint64_t id() const;
```
gasmith, line 959: noexcept

cpp/foxglove/include/foxglove/schemas.hpp
```diff
@@ -0,0 +1,2081 @@
...
+#include <optional>
+#include <string>
+#include <type_traits>
+#include <vector>
+
+struct foxglove_channel;
+
+namespace foxglove::schemas {
+
+/// @brief A functor for freeing a channel. Used by ChannelUniquePtr. For internal use only.
```
gasmith, line 21: Use `/// @cond foxglove_internal` to omit this from the docs. See `WarnStream` in `error.hpp` for an example.

cpp/foxglove/include/foxglove/arena.hpp
```diff
@@ -0,0 +1,116 @@
...
+    typename T, typename S, typename Fn,
+    typename = std::enable_if_t<std::is_pod_v<T> && std::is_invocable_v<Fn, T&, const S&, Arena&>>>
+  T* map(const std::vector<S>& src, Fn&& map_fn) {
+    const size_t elements = src.size();
+    T* result = alloc<T>(elements);
+    T* current = result;
+
+    // Convert the elements from S to T, placing them in the result array
+    for (auto it = src.begin(); it != src.end(); ++it) {
+      map_fn(*current++, *it, *this);
```
gasmith, line 42: std::forward(map_fn)

cpp/foxglove/include/foxglove/arena.hpp
```diff
@@ -0,0 +1,116 @@
...
+#include <cstddef>
+#include <cstdint>
+#include <memory>
+#include <new>
+#include <type_traits>
+#include <vector>
+
+namespace foxglove {
+
+/**
```
gasmith, line 12: Hide the Arena type from the docs with `/// @cond foxglove_internal`.

gasmith (commented):
Took a first pass, mostly on the C++ side Mostly minor nits. Will continue reviewing after lunch.

cpp/foxglove/include/foxglove/arena.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+    // Calculate aligned offset
+    const size_t alignment = alignof(T);
+    const size_t misalignment = _offset % alignment;
+    const size_t alignment_padding = misalignment > 0 ? alignment - misalignment : 0;
+    const size_t aligned_offset = _offset + alignment_padding;
+
+    // Check if we have enough space
+    if (aligned_offset + bytes_needed > Size) {
+      throw std::bad_alloc();
```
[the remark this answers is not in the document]
eloff: I was hoping to avoid it on the first pass, but yes, we can add that

cpp/foxglove/include/foxglove/channel.hpp
```diff
@@ -63,23 +64,17 @@ class Channel final {
...
 
-  Channel(const Channel&) = delete;
-  Channel& operator=(const Channel&) = delete;
-
+  RawChannel(const RawChannel&) = delete;
+  RawChannel& operator=(const RawChannel&) = delete;
   /// @brief Default move constructor.
-  Channel(Channel&& other) noexcept = default;
-  /// @brief Default move assignment.
-  Channel& operator=(Channel&& other) noexcept = default;
```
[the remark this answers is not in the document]
eloff: I thought so too, but I was having issues with it. I'll revisit and see if it's still a problem.

cpp/foxglove/include/foxglove/arena.hpp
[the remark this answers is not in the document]
gasmith: Yeah, it's fine to defer that for a separate PR. This one's already a monster. :)

c/include/foxglove-c/foxglove-c.h
```diff
@@ -76,19 +78,48 @@ enum foxglove_error
...
-  FOXGLOVE_LOG_LEVEL_WARN = 3,
+  FOXGLOVE_LOG_LEVEL_WARNING = 3,
   FOXGLOVE_LOG_LEVEL_ERROR = 4,
+  FOXGLOVE_LOG_LEVEL_FATAL = 5,
+};
+#ifndef __cplusplus
+typedef int32_t foxglove_log_level;
+#endif // __cplusplus
+
+enum foxglove_logging_level
```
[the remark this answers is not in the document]
eloff: I didn't add this one, but I can add a docstring to it.

c/include/foxglove-c/foxglove-c.h
```diff
@@ -49,6 +49,8 @@
  */
 #define FOXGLOVE_SERVER_CAPABILITY_SERVICES (1 << 4)
 
+#define Arena_SIZE (256 * 1024)
```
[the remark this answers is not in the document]
eloff: It shouldn't be exported at all

c/src/generated_types.rs
```diff
@@ -0,0 +1,4298 @@
...
+    }
+
+    /// Create a new typed channel, and return an owned raw channel pointer to it.
+    ///
+    /// # Safety
+    /// This is intended for internal use only.
+    /// We're trusting the caller that the channel will only be used with this type T.
+    #[doc(hidden)]
+    #[unsafe(no_mangle)]
+    pub unsafe extern "C" fn foxglove_channel_create_camera_calibration(
```
gasmith, line 232: Nit: Did you intend for this to be an associated item, rather than a plain function like `foxglove_channel_log_camera_calibration`?

c/src/arena.rs
```diff
@@ -0,0 +1,108 @@
...
+        src: *const S,
+        len: usize,
+    ) -> Result<ManuallyDrop<Vec<S::NativeType>>, FoxgloveError> {
+        if len == 0 {
+            return Ok(ManuallyDrop::new(Vec::new()));
+        }
+
+        let result = self.as_mut().alloc::<S::NativeType>(len);
+
+        // Convert the elements from S to T, placing them in the result array
```
gasmith, line 80: Nit: "S to S::NativeType"

c/src/arena.rs
```diff
@@ -0,0 +1,108 @@
...
+use std::mem::{align_of, size_of, MaybeUninit};
+use std::pin::Pin;
+
+use foxglove::FoxgloveError;
+
+/// A trait for converting a Foxglove C type to its native Rust () representation
+pub trait BorrowToNative {
+    type NativeType;
+
+    unsafe fn borrow_to_native(
```
gasmith, line 13: Nit: Instead of copying the same docstring on all of the implementations, let's put it here.

gasmith (commented):
Just a few more nits. I think this generally looks good.

typescript/schemas/src/internal/generateSdkCpp.ts
```diff
@@ -0,0 +1,410 @@
...
+          return `${formatComment(field.description, 2)}\n  ${fieldType} ${toSnakeCase(field.name)}${defaultStr};`;
+        })
+        .join("\n\n"),
+      `};`,
+    ].join("\n");
+  });
+
+  const channelClasses = schemas.filter(hasChannelType).map(
+    (schema) => `/// @brief A channel for logging ${schema.name} messages to a topic.
+      class ${schema.name}Channel {
```
jtbandes, line 173: The PR description says:

> By having a dedicated channel type per schema type, we avoid the overhead of dynamic dispatch (e.g. based on an enum with a giant switch statement or virtual function table or similar).

This makes sense on the C side of things, but on the C++ side doing `Channel<T>` wouldn't necessarily mean dynamic dispatch has to be involved, I think.

Basically, I just want to point out that I think we can decide which ergonomics we want for C++ users, and make it work equally efficiently under the hood either way. Maybe that's already happened – curious what was behind the choice to avoid the `Channel<T>` in my draft PR and use `TChannel` instead?  My first thought would be that `Channel<T>` could be a little nicer as a user, but I didn't really push that assumption very far so maybe it can be disproven.

jtbandes (commented):
🌈 🙏 📿 🫸🪨⛰️

cpp/foxglove/include/foxglove/channel.hpp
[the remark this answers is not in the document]
eloff: It works now, I don't remember why I had to remove it earlier :man_shrugging:

cpp/foxglove/include/foxglove/arena.hpp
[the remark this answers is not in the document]
eloff: I went ahead and implemented it for both arenas, so this should be covered now.

c/src/generated_types.rs
```diff
@@ -0,0 +1,4298 @@
...
+        };
+        unsafe { msg.borrow_to_native(arena) }
+    }
+
+    /// Create a new typed channel, and return an owned raw channel pointer to it.
+    ///
+    /// # Safety
+    /// This is intended for internal use only.
+    /// We're trusting the caller that the channel will only be used with this type T.
+    #[doc(hidden)]
```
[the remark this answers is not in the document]
eloff: It doesn't, I'll remove this

c/src/generated_types.rs
eloff: I wanted a plain function

cpp/foxglove/include/foxglove/channel.hpp
```diff
@@ -34,7 +35,7 @@ struct Schema {
 };
 
 /// @brief A channel for messages logged to a topic.
-class Channel final {
+class RawChannel final {
```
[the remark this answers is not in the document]
eloff: That one really came back to bite you :)

typescript/schemas/src/internal/generateSdkCpp.ts
eloff: It came down to the implementation  for Channel<T> needing to be in the header, so either we pull in the C header into the global namespace, which we went to great lengths to avoid, or we use some kind of dynamic dispatch (e.g. a pointer to implementation).

This option was the simplest to implement, I think it's about the same for the user, two less angle brackets to type, otherwise not much difference.

cpp/foxglove/tests/test_mcap.cpp
```diff
@@ -285,3 +287,144 @@ TEST_CASE("Channel can outlive Schema") {
...
+  REQUIRE(c_msg.texts[0].position->x == msg.texts[0].position->x);
+  REQUIRE(c_msg.texts[0].position->y == msg.texts[0].position->y);
+  REQUIRE(c_msg.texts[0].text.data == msg.texts[0].text.data());
+  REQUIRE(c_msg.texts[0].text.len == msg.texts[0].text.size());
+  REQUIRE(c_msg.texts[0].font_size == msg.texts[0].font_size);
+  REQUIRE(c_msg.texts[0].text_color->r == msg.texts[0].text_color->r);
+  REQUIRE(c_msg.texts[0].text_color->g == msg.texts[0].text_color->g);
+  REQUIRE(c_msg.texts[0].text_color->b == msg.texts[0].text_color->b);
+  REQUIRE(c_msg.texts[0].text_color->a == msg.texts[0].text_color->a);
+  REQUIRE(c_msg.texts[0].background_color->r == msg.texts[0].background_color->r);
```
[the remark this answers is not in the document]
eloff: I really didn't want to delve into custom matchers in addition to everything else, but yes, that would work

cpp/foxglove/include/foxglove/arena.hpp
```diff
@@ -0,0 +1,116 @@
...
+   */
+  template<typename T>
+  T* alloc(size_t elements) {
+    const size_t bytes_needed = elements * sizeof(T);
+
+    // Calculate aligned offset
+    const size_t alignment = alignof(T);
+    const size_t misalignment = _offset % alignment;
+    const size_t alignment_padding = misalignment > 0 ? alignment - misalignment : 0;
+    const size_t aligned_offset = _offset + alignment_padding;
```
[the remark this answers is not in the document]
eloff: Nice, didn't know that existed

typescript/schemas/src/internal/schemas.ts
```diff
@@ -1304,10 +1304,10 @@ const PositionCovarianceType: FoxgloveEnumSchema = {
...
   protobufEnumName: "PositionCovarianceType",
   values: [
-    { name: "UNKNOWN", value: 0 },
-    { name: "APPROXIMATED", value: 1 },
-    { name: "DIAGONAL_KNOWN", value: 2 },
-    { name: "KNOWN", value: 3 },
+    { name: "UNKNOWN", value: 0, description: "Unknown position covariance type" },
+    { name: "APPROXIMATED", value: 1, description: "Position covariance is approximated" },
+    { name: "DIAGONAL_KNOWN", value: 2, description: "Position covariance is diagonal" },
+    { name: "KNOWN", value: 3, description: "Position covariance is known" },
```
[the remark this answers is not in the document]
eloff: I did as best I could from the source you linked. It will do for now, if we get feedback we can always improve it later

typescript/schemas/src/internal/generateSdkRustCTypes.ts
```diff
@@ -0,0 +1,287 @@
...
+          lines.push(`pub ${identName}_len: usize,`);
+        }
+      }
+      return lines.join("\n");
+    })
+    .join("\n\n")}
+}
+
+${
+  name.endsWith("Primitive")
```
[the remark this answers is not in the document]
eloff: We don't have a ticket for that, but we probably should and it affects the whole sdk. The color channel we support is kind of silly (but it doesn't end in primitive).

eloff:
This should be ready for final review.

c/src/arena.rs
```diff
@@ -0,0 +1,248 @@
...
+        if aligned_offset + bytes_needed > Self::SIZE {
+            let layout = std::alloc::Layout::array::<T>(n).unwrap();
+            // SAFETY: layout is valid and non-zero and we don't assume the memory is initialized
+            let ptr = unsafe { std::alloc::alloc(layout) };
+            self.overflow.borrow_mut().push((ptr, layout));
+            return ptr as *mut T;
+        }
+
+        // SAFETY: [result, result+n) is properly aligned and within the bounds of buffer
+        let result = unsafe { self.buffer.as_ptr().add(aligned_offset) as *mut T };
```
gasmith, line 96: I ran this past miri, and it doesn't like the fact that we're using `as_ptr()` to get a an immutable raw pointer, which we subsequently write through as a mutable reference:

```
error: Undefined Behavior: attempting a write access using <283443> at alloc124434[0x0], but that tag only grants SharedReadOnly permission for this location
   --> c/src/arena.rs:181:17
    |
181 |                 *int_ptr.add(i) = i as i32;
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                 |
    |                 attempting a write access using <283443> at alloc124434[0x0], but that tag only grants SharedReadOnly permission for this location
    |                 this error occurs as part of an access at alloc124434[0x0..0x4]
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <283443> was created by a SharedReadOnly retag at offsets [0x0..0x20000]
```

So the first thing I tried was to wrap the buffer in a `Cell`, so that `as_ptr()` gives us a `*mut [MaybeUninit<u8>; Arena::SIZE]`. That helps a little, but then miri complains that `.add(aligned_offset)` is undefined pointer arithmetic, because the resulting pointer does not point to an allocated buffer of at least N bytes:

```
error: Undefined Behavior: out-of-bounds pointer arithmetic: expected a pointer to 5242880 bytes of memory, but got alloc124434 which is only 131112 bytes from the end of the allocation
   --> c/src/arena.rs:96:31
    |
96  |         let result = unsafe { self.buffer.as_ptr().add(aligned_offset) as *mut T };
    |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ out-of-bounds pointer arithmetic: expected a pointer to 5242880 bytes of memory, but got alloc124434 which is only 131112 bytes from the end of the allocation
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
```

To work around that, we can erase our knowledge of the array length by getting a raw pointer to the first element and using that as the basis for our arithmetic:

```rust
let buffer = self.buffer.as_ptr();
let base = unsafe { &raw mut (*buffer)[0] };
let result = unsafe { base.add(aligned_offset) as *mut T };
```

That seems to be enough to satisfy miri.

cpp/foxglove/include/foxglove/arena.hpp
```diff
@@ -0,0 +1,136 @@
...
+#include <cstdlib>
+#include <memory>
+#include <new>
+#include <type_traits>
+#include <vector>
+
+namespace foxglove {
+
+/**
+ * @private
```
gasmith, line 16: When I generate docs, the `@private` tag does not seem to have any effect. Use `@cond foxglove_internal` + `@endcond`.

cpp/foxglove/include/foxglove/arena.hpp
```diff
@@ -0,0 +1,116 @@
...
+   * @param src The source vector containing elements to map
+   * @param map_fn Function taking (T& dest, const S& src) to map elements.
+   * T must be a POD type, without a custom constructor or destructor.
+   * @return Pointer to the beginning of the allocated array of src.size() T elements
+   * @throws std::bad_alloc if the arena doesn't have enough space
+   */
+  template<
+    typename T, typename S, typename Fn,
+    typename = std::enable_if_t<std::is_pod_v<T> && std::is_invocable_v<Fn, T&, const S&, Arena&>>>
+  T* map(const std::vector<S>& src, Fn&& map_fn) {
```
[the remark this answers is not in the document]
gasmith: Bump

cpp/foxglove/include/foxglove/arena.hpp
gasmith: Bump

cpp/foxglove/include/foxglove/arena.hpp
```diff
@@ -0,0 +1,116 @@
...
+   * @param src The source vector containing elements to map
+   * @param map_fn Function taking (T& dest, const S& src) to map elements.
+   * T must be a POD type, without a custom constructor or destructor.
+   * @return Pointer to the beginning of the allocated array of src.size() T elements
+   * @throws std::bad_alloc if the arena doesn't have enough space
+   */
+  template<
+    typename T, typename S, typename Fn,
+    typename = std::enable_if_t<std::is_pod_v<T> && std::is_invocable_v<Fn, T&, const S&, Arena&>>>
+  T* map_one(const S& src, Fn&& map_fn) {
```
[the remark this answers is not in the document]
gasmith: Bump

cpp/foxglove/include/foxglove/arena.hpp
```diff
@@ -0,0 +1,116 @@
...
+    T* result = alloc<T>(1);
+    map_fn(*result, src, *this);
```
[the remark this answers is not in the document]
gasmith: Bump

cpp/foxglove/include/foxglove/arena.hpp
```diff
@@ -0,0 +1,136 @@
...
+#include <cstdint>
+#include <cstdlib>
+#include <memory>
+#include <new>
+#include <type_traits>
+#include <vector>
+
+namespace foxglove {
+
+/**
```
gasmith, line 15: These comments should be `///` to match the style adopted elsewhere.

cpp/foxglove/include/foxglove/arena.hpp
gasmith: Also need to revert the changes to `error.hpp` and `context.hpp`.

c/src/arena.rs
eloff: I think what I'll do is put it in an UnsafeCell, it will fix the miri issue while also giving us the strongest "don't make assumptions" hint to the Rust compiler

cpp/foxglove/include/foxglove/arena.hpp
eloff: I was pretty proud of myself for finding a cleaner way of excluding it from the docs, weird that it doesn't work, oh well

gasmith (approved):

PR #414, opened 2025-05-09
Add instances for prost_reflect::ReflectMessage

dspyz-matician:
### Changelog
Add instances for prost_reflect::ReflectMessage

### Docs

None

### Description

It's often useful to reflectively use properties of protobuf message types (eg to get the fully qualified path name of a particular message). The `prost_reflect` crate provides this for types which derive `prost_reflect::ReflectMessage`

Types do not implement `prost_reflect::ReflectMessage`

</td><td>

Types do implement `prost_reflect::ReflectMessage`

</td></tr></table>


dspyz-matician:
It looks like I have the wrong version of the protobuf compiler. Any idea which version CI is running?

defunctzombie:
@dspyz-matician https://github.com/foxglove/foxglove-sdk/blob/main/.github/workflows/ci.yml#L21

looks like 29.x

dspyz-matician:
I don't know what `29.x` means? Are there multiple ways that `protoc` can be versioned?

```
dspyz@dspyz-desktop:~/foxglove-sdk$ protoc --version
libprotoc 3.21.12
```

defunctzombie:
> I'm confused about what `29.x` means. Are there multiple ways that `protoc` can be versioned?

My guess is it refers to the "protobuf release" version: https://github.com/protocolbuffers/protobuf/releases

dspyz-matician:
I definitely used 29.4 this time. Could you regenerate it for me?
Thank you

dspyz-matician:
I tried 29.2 as well, but that just gave me the same thing I started with

gasmith:
Thanks for the contribution! Could you provide us with a little more information about what you're trying to do? There might be other ways to get at the information you want, without adding a new dependency to the SDK.

Right now we use a protobuf encoding for the foxglove schema messages, but that could change in the future, so we want to be careful about how much protobuf-specific stuff we put into the public API.

Each of the foxglove schema messages implements the [`Encode` trait](https://docs.rs/foxglove/latest/foxglove/trait.Encode.html), in such a way that `Encode::get_schema()` is guaranteed to return a [Schema](https://docs.rs/foxglove/latest/foxglove/struct.Schema.html).

The schema is the intended abstraction layer for introspection. For a protobuf message, the schema contains:

- `name`: The fully-qualified message name, e.g. `foxglove.FrameTransform`.
- `encoding`: `protobuf`
- `data`: The binary file descriptor set for this message and its dependencies.

So if you really want a `prost_reflect::MessageDescriptor`, you can construct one from the schema like so:

```rust
use foxglove::schemas::FrameTransform;
use foxglove::Encode;
use prost::Message;
use prost_reflect::DescriptorPool;
use prost_types::FileDescriptorSet;

let schema = FrameTransform::get_schema().unwrap();
assert_eq!(schema.encoding, "protobuf");
let fds = FileDescriptorSet::decode(&*schema.data).unwrap();
let dp = DescriptorPool::from_file_descriptor_set(fds).unwrap();
let md = dp.get_message_by_name(&schema.name).unwrap();
println!("{md:?}");
```

dspyz-matician:
I'm trying to read prost values from a wrapper around an mcap channel and want to add an assertion at construction-time that the type I'm reading corresponds to the protobuf type that the mcap channel says it contains. This involves doing something like:

```
pub struct ProstChannel<T> {
    channel: Channel,
    _phantom: PhantomData<fn() -> T>,
}

impl<T: Default + ReflectMessage> ProstChannel<T> {
    pub fn new(channel: Channel) -> Result<Self> {
        let schema = &channel.stream.mcap.summary.schemas[&channel.idx];
        if schema.encoding != "protobuf" {
            bail!("Channel {} is not protobuf encoded", channel.idx);
        }
        let message_descriptor = T::default().descriptor();
        let type_name = message_descriptor.full_name();
        let schema_name = &schema.name;
        ensure!(
            schema_name == type_name,
            "Channel {} has schema {}, but expected {}",
            channel.idx,
            schema_name,
            type_name
        );
        Ok(ProstChannel { channel, _phantom: PhantomData })
    }
}
```

The `descriptor` function here comes from the `ReflectMessage` trait.

I want to be able to generically use this `ProstChannel` struct filling in my custom protobuf type for `T` _as well as_ with filling in prost types defined in the foxglove crate such as `foxglove::schemas::CompressedVideo`. For that, `CompressedVideo` needs to _also_ implement `ReflectMessage`.

eloff:
I think I see where you're going with this.

The trouble is the protobuf implementation is really an implementation detail, we don't want to make it part of the public interface, because it could change in the future. If we merged this PR, it would fence us in. We're not currently wanting to commit to making protobuf part of the public API.

I think @gasmith's recommendation and code example would be a better way to go, add your own trait and implement it automatically for your types using a blanket implementation with prost_reflect and explicitly for our types (or visa versa).

eloff:
Thanks for going to the trouble to open this pull request and explain the problem you're trying to solve. I'm closing this because it's not compatible with the current direction of the SDK, for the reasons discussed. I believe the workaround offered by @gasmith is a good alternative for now.

PR #415, opened 2025-05-09
Disable unused mcap features

dspyz-matician:
### Changelog
None

### Docs
None

### Description
Depending on lz4 and zstd can be a nuisance for downstream code especially if they're unused

Rust crate depends on lz4 and zstd

</td><td>

Rust crate doesn't depend on lz4 and zstd

</td></tr></table>


dspyz-matician:
In some contexts we're forced to work with bazel rather than Cargo which makes adding C dependencies a nuisance. But also, if you have unused dependencies wouldn't you want to remove them just to reduce the binary size anyway?

gasmith (commented):
Adding the features makes sense.

Regarding the default features: I'm OK with removing `lz4` by default, but I'm less sure about `zstd`, because that's the default compression used by the mcap writer when the feature is present. Users might not realize they have to explicitly enable the feature to get compression, especially if they're familiar with the underlying mcap crate (which includes zstd by default).

You should also update the crate's top-level rustdoc:
- Feature flags are documented [here](https://github.com/foxglove/foxglove-sdk/blob/cd14fd363eaa32cb147ce00673f6ece05abeb360/rust/foxglove/src/lib.rs#L247-L259).
- It has [an example that uses lz4](https://github.com/foxglove/foxglove-sdk/blob/cd14fd363eaa32cb147ce00673f6ece05abeb360/rust/foxglove/src/lib.rs#L201-L203). Maybe just update the example to do `.compression(None)` instead.

I think you'll also need to update the `example-mcap` crate, which disables foxglove's default features, but definitely depends on lz4 and zstd. (@eloff we might consider enhancing CI to build workspace packages individually, so that we can catch breakages like that.)

eloff:
I think we should keep both enabled by default, but have features to disable them. This mirrors how the mcap library itself works.

dspyz-matician:
Okay, sounds good

gasmith:
Merged as part of #433. Thanks again for the contribution!

PR #417, opened 2025-05-09
C: Make context the first callback argument

gasmith:


eloff (approved):
LGTM, thanks for spotting that and cleaning it up

PR #418, opened 2025-05-12
Improve C++ documentation; rework docs publishing

bryfox:
### Changelog
None

### Description

This replaces the generated Doxygen HTML pages with Sphinx docs, using the breathe & exhale extensions (BSD licensed). Doxygen is still used to generate the XML source for these pages. I used a similar layout to the python docs, with examples next to the API reference. I chose a different theme, since I don't think any of these are 'standard' for c++, and which I thought looked clean.

This also adds a new "Docs" workflow, which deploys both the C++ and Python documentation to gh-pages. We build both c++ and python docs, then upload in a single job. This has no dependencies on the deployment job; we could introduce a callable workflow, but I don't think it's needed.

I deployed C++ documentation from this branch: https://foxglove.github.io/foxglove-sdk/cpp/


sofuture (approved):

sofuture:
LGTM the single workflow makes more sense

cpp/foxglove/include/foxglove/server/parameter.hpp
```diff
@@ -208,6 +208,9 @@ class ParameterView final {
     return value.has_value() && value->is<T>();
   }
 
+  /// @cond duplicate-ids-bug
+  // https://github.com/doxygen/doxygen/issues/9468
```
gasmith, line 212: :sob:

gasmith (approved):
LG. For the record, I like the theme.

cpp/foxglove/include/foxglove/server/parameter.hpp
bryfox: Yeah, overall I like the produced documentation, but output of templated types is worse (less usable) than Doxygen, even when this duplicate ID bug doesn't occur. I think we can probably improve it in the future.

bryfox:
I'm going to merge this now; will update links from our main docs site to point to this once we do another release (since right now the C++ docs contain some things added since the 0.7.1 release).

PR #419, opened 2025-05-12
ci: Build rust examples individually

gasmith:
I noticed with #415 breaks the `example-mcap` build, but it isn't caught in CI, because we're building the entire workspace all at once, which means we're building dependencies with the union of all features specified by all target packages.

This change adds a step to CI to build each example individually, to ensure that we don't break them when we introduce new feature flags. We only need to do this for examples, because we already build the other targets individually.


bryfox (approved):
Good catch

PR #421, opened 2025-05-13
C/C++: Add websocket clear session and publish status

gasmith:
Adding a few last missing APIs for the websocket server.


cpp/foxglove/tests/test_server.cpp
```diff
@@ -33,6 +33,20 @@ constexpr std::underlying_type_t<T> toUnderlying(T e) noexcept {
   return static_cast<std::underlying_type_t<T>>(e);
 }
 
+// Helper function to wait for and retrieve a message from the client queue
+std::string waitForMessage(
```
bryfox, line 37: 👍

bryfox (approved):

cpp/foxglove/tests/test_server.cpp
gasmith: Wait, wait, we can do more!

PR #422, opened 2025-05-14
rust: make toplevel example snippet pastable into any function

james-rms:
### Changelog
None.
### Docs

Is a docs PR.
### Description

Today there is a code snippet showing the user how to use the foxglove SDK, but to use it they must wrap it in a function that returns `Result<T, E> where E: From<FoxgloveError>`. This means they'd need to write non-trivial extra code to use this code as written. This PR updates the example to use `expect`, which allows a user to get started immediately from `cargo init .; cargo add foxglove`.

</td><td>

</td></tr></table>

### DIFF
--- a/rust/foxglove/src/lib.rs
+++ b/rust/foxglove/src/lib.rs
@@ -17,7 +17,6 @@
 //! use foxglove::{McapWriter, log};
 //! use foxglove::schemas::Log;
 //!
-//! # fn func() {
 //! // Create a new MCAP file named 'test.mcap'.
 //! let mcap = McapWriter::new().create_new_buffered_file("test.mcap").expect("create failed");
 //!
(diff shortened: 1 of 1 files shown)


bryfox (approved):

rust/foxglove/src/lib.rs
```diff
@@ -17,18 +17,18 @@
 //! use foxglove::{McapWriter, log};
 //! use foxglove::schemas::Log;
 //!
-//! # fn func() -> Result<(), foxglove::FoxgloveError> {
+//! # fn func() {
```
gasmith, line 20: I think you can get rid of this line, and the closing `# }` on L31. That scaffolding was only required to support the `?`.

gasmith (approved):

gasmith:
Worth mentioning that this is only one of several examples in that top-level rustdoc that use `?`. Perhaps we should update all of them?

james-rms:
I think that's worthwhile, i'll do a broader pass.

PR #423, opened 2025-05-14
rust: Format doc snippets with nightly

gasmith:
TIL that there's a nightly rustfmt option you can use to format snippets in doc comments.


bryfox (approved):
Should we run this in CI?

gasmith:
> Should we run this in CI?

I'd say yes if it wasn't an unstable feature. It'd be a bit of pain to get it working, and I'd be a little worried about churn as the feature stabilizes. Considering it's purely cosmetic, I think we can just run it manually from time to time.

PR #425, opened 2025-05-15
Add deployment steps for derive crate

bryfox:
This builds on #324 to support a derive macro for `Encode`.

This PR sets up publishing for the `foxglove-derive` crate. 

The `bump-sdk-version` task will now keep the foxglove-derive dependency pinned to the same version. This isn't strictly required, but seems cleaner and avoids needing a manual bump (which is easy to forget since we reference a path locally).

I moved the CI publish step to its own job. It now publishes the derive crate, waits for the latest version to be available on crates.io, and then publishes the foxglove crate as before.

As of right now, the dry-run CI check will still fail since we haven't published the derive crate at all. I think once we publish a v0.7.1 version, that check will pass, and that's the easiest way forward. I tested this with a local registry. We can publish that manually, or I can publish derive 0.7.1 from a branch.

Lastly, I updated some of the derive crate metadata — notably, adding the MIT license — and I moved some shared package config to the workspace for clarity.

There are some third-party crates that can deal with this, but they appear to address a lot of problems we have already solved, and at the core still seem to do a sleep/poll for the dependent crate publishing.


.github/workflows/ci.yml
```diff
@@ -38,9 +38,28 @@ jobs:
...
+        name: Detect local Cargo version
+        run: |
+          version=$(cargo metadata --no-deps --format-version 1 | jq -r '.packages[] | select(.name=="'foxglove'").version')
+          echo "version=$version" >> $GITHUB_OUTPUT
+      - name: Publish foxglove-derive
+        run: cargo publish --package foxglove-derive
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+      - name: Wait for foxglove-derive to be published
+        run: until cargo info foxglove-derive@${{ steps.detect-version.outputs.version }}; do sleep 10; done
```
gasmith, line 60: If we fail here (e.g., due to a timeout), will we be able to re-run the workflow? Or would the attempt to publish a duplicate `foxglove-derive` version fail?

gasmith (approved):
Looks reasonable to me.

gasmith:
> As of right now, the dry-run CI check will still fail since we haven't published the derive crate at all. I think once we publish a v0.7.1 version, that check will pass, and that's the easiest way forward. I tested this with a local registry.

Will this also be a problem when running the dry-run after a version bump?

bryfox:
> Will this also be a problem when running the dry-run after a version bump?

We don't run the check when drafting the release, but (oops, yes) we need to skip the dry run just before publishing. [e1d0c4a](https://github.com/foxglove/foxglove-sdk/pull/425/commits/e1d0c4aaab474cd437a3b1ec184741b2a95db9a9) does that. I don't see a reason to run the dry run for either crate if we're about to publish.

.github/workflows/ci.yml
bryfox: I wasn't sure about that, and I don't see a way to test. Certainly the dry-run doesn't warn us that there's a problem.

FWIW, I did some digging around to see how other people do this, and sleeping/polling seems to be the approach — even with other tools like cargo-release (which seemed much more than we needed).

gasmith: That's fair. I suppose we'll figure out how to cross that bridge if we ever get there.

defunctzombie:
👍 I am somewhat surprised there's not a more sanctioned way to handle all this given that a separate create is a requirement for making a derive macro.

scripts/bumpSdkVersion.ts
```diff
@@ -5,6 +5,7 @@ import path from "path";
 import semver from "semver";
```
defunctzombie, line 5: I am amused that our bump version script is written in a language that doesn't exist for any sdk :D

PR #430, opened 2025-05-20
TS: move rosmsg-msgs-common back to dependencies, bump to 1.7.1

jtbandes:
### Changelog
TS: fixed missing dependency on `@foxglove/rosmsg-msgs-common`

### Docs

None

### Description

This was moved to devDependencies in https://github.com/foxglove/foxglove-sdk/pull/170, however there are some exported functions whose type definitions depend on it (admittedly they are under `internal`, but we use them for generating docs). The missing dependency caused TS errors downstream.

### DIFF
--- a/typescript/schemas/package.json
+++ b/typescript/schemas/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@foxglove/schemas",
-  "version": "1.7.0",
+  "version": "1.7.1",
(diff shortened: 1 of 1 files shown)


defunctzombie (approved):

gasmith (approved):
Should we also add a CI step to catch this kind of breakage?

jtbandes:
It would be an...interesting kind of CI step that would have to create a new npm project, add this one as a dependency, and then run typescript... I'm not sure it's worth it

PR #431, opened 2025-05-21
Add simple example of writing RawAudio messages to an mcap.

nidanin:
### Changelog
Add a RawAudio example that writes out a mono sinewave to an mcap.

### Docs

None

### Description
Adds a simple python example of writing RawAudio messages to an mcap file. This just writes a single channel sine wave to keep things as simple as possible. We could consider adding an example that writes a stereo file to help people understand how to write interleaved samples if we think that would be useful but it would increase the complexity slightly.


python/foxglove-sdk-examples/write-mcap-rawaudio-file/main.py
```diff
@@ -0,0 +1,80 @@
...
+    total_samples = int(sample_rate * duration_seconds)
+    samples = [
+        int(amplitude * math.sin(2 * math.pi * frequency * t / sample_rate))
+        for t in range(total_samples)
+    ]
+
+    # Open the MCAP file for writing
+    with foxglove.open_mcap(args.path):
+        # Create a channel for RawAudio messages
+        audio_channel = RawAudioChannel(topic="audio")
```
banisadr, line 51: should the topic be `"/audio"`?

nidanin: Good question. I'm not sure.

banisadr (approved):
Looks good to me & works as expected.

PR #432, opened 2025-05-21
Add docs about thread-safety to C++ API

eloff:
What it says on the tin.

I think we're also going to need thread-safety docs for Rust and Python.

### DIFF
--- a/c/include/foxglove-c/foxglove-c.h
+++ b/c/include/foxglove-c/foxglove-c.h
@@ -2734,6 +2734,8 @@ void foxglove_fetch_asset_respond_error(struct foxglove_fetch_asset_responder *r
  *
  * Log styles (colors) may be configured with the FOXGLOVE_LOG_STYLE environment variable "never",
  * "always", or "auto" (default).
+ *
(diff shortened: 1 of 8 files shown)


eloff:
> As a user trying to figure out if my program is correct, I think I would still be wondering if it's safe to call methods from different threads, e.g. just because McapWriter says "calls to create from multiple threads are safe", does that mean I can call close() on a different thread from create()? Same with "creating channels and logging on them is thread-safe" – does it matter which thread these happen from? Do I need to be careful about which thread destructors are called on? (for example, if you put one of these in a shared_ptr then the destructor could easily get called on whichever thread is the last one standing). If there are no restrictions that's also good to mention.

Yeah, these are safe. I'll see if I can clarify that a little. Really the only exceptions to thread-safety for these types is the move assignment operator and move constructor. But I'm not sure why anyone would expect that you could safely access a class while moving out of it on another thread.

> 
> Definitely would be nice to have some overview about thread-safety somewhere. Maybe in the C++ generated docs if that has an overview of the C++ SDK?

I'll add a section on thread-safety to the overview.

jtbandes (approved):

PR #433, opened 2025-05-21
rust: Add features for derive, lz4, schemars, zstd

gasmith:
### Changelog
- rust: Add features for derive, lz4, schemars, zstd

### Docs
None

### Description
This change builds on #415, and adds two additional features:
- `derive` for the new derive macro
- `schemars` for the blanket implementation of `Encode` for `T: Serialize + JsonSchema`

I've updated the docs and examples accordingly.

I also feature-flagged tests & docs so that we can run unit tests with `--no-default-features`, and added that step to CI.


eloff (approved):
This makes sense to me, and it looks good. The CI test with and without features was a nice touch

PR #434, opened 2025-05-21
rust: Enable derive feature by default

gasmith:
### Changelog
None

### Docs
None

### Description
As @defunctzombie points out, the `derive` feature is a core component of the SDK. Let's enable it by default.

### DIFF
--- a/c/Cargo.toml
+++ b/c/Cargo.toml
@@ -15,7 +15,11 @@ workspace = true
 base64 = "0.22.1"
 bitflags = "2.9.0"
 env_logger = "0.11.5"
-foxglove = { path = "../rust/foxglove", features = ["unstable"] }
+foxglove = { path = "../rust/foxglove", default-features = false, features = [
+  "live_visualization",
+  "lz4",
+  "zstd",
+] }
 log.workspace = true
 mcap.workspace = true
 tracing.workspace = true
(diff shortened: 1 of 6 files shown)


c/Cargo.toml
```diff
@@ -15,7 +15,11 @@ workspace = true
 base64 = "0.22.1"
 bitflags = "2.9.0"
 env_logger = "0.11.5"
-foxglove = { path = "../rust/foxglove", features = ["unstable"] }
+foxglove = { path = "../rust/foxglove", default-features = false, features = [
```
defunctzombie, line 18: Why does this crate disable the default features?

defunctzombie (approved):
One day we should find a better name for "live_visualization". It's more about offering connectivity I think? maybe not.

c/Cargo.toml
gasmith: This is cbindgen wrapper that we use for C & C++. The derive feature can't be exposed through that interface.

gasmith:
> One day we should find a better name for "live_visualization". It's more about offering connectivity I think? maybe not.

I'm inclined to agree. If we want to change it, we should do so before 1.0. I still think we should call it `websocket`. We're not hiding the fact that it's a websocket elsewhere in the SDK. And when we replace the transport with WebRTC or whatever, we'll probably introduce a new abstraction for that, which is less tightly-bound to ws-protocol.

PR #435, opened 2025-05-21
Accept u64, Timestamp, SystemTime, and DateTime as log timestamps

eloff:
Adds a new trait `ToUnixNanos` to unify the different kinds of timestamps we can accept.
Adds a constructor `with_log_time` to `PartialMetadata` accepting an impl `ToUnixNanos` arg
Adds a new log method `log_with_time` that calls `log_with_meta(msg, PartialMetadata::with_log_time(timestamp))`
Modifies log! macro to accept any impl `ToUnixNanos` arg for log_time.

Discussion on the story here: FG-11537


rust/foxglove/src/metadata.rs
```diff
@@ -1,3 +1,38 @@
...
+
+impl ToUnixNanos for Timestamp {
+    fn to_unix_nanos(&self) -> u64 {
+        self.total_nanos()
+    }
+}
+
+impl ToUnixNanos for SystemTime {
+    fn to_unix_nanos(&self) -> u64 {
+        self.duration_since(UNIX_EPOCH).unwrap().as_nanos() as u64
```
gasmith, line 25: Use `expect()` for consistency with the chrono implementation?

Or perhaps we should saturate instead of panicking? Probably not worth the effort.

rust/foxglove/src/log_macro.rs
```diff
@@ -70,9 +70,7 @@ macro_rules! log {
         $crate::log_with_meta!(
             $topic,
             $msg,
-            $crate::PartialMetadata {
-                log_time: Some($log_time),
-            }
+            $crate::PartialMetadata::with_log_time($log_time)
```
gasmith, line 73: Should we update the doc to describe the types accepted for `log_time`?

gasmith (approved):
LG, just minor comments.

PR #437, opened 2025-05-26
Add note to docs that schema classes are immutable and don't provide getters and setters

eloff:
Our implementation converts from Python types to Rust types eagerly in the constructor. Stacking getters on top of that requires converting back from Rust -> Python, and then a setter would do the conversion again Python -> Rust. For types like strings, bytes, lists, nested structs, lists of nested structs, each of these conversions requires memory allocations and copying.

We didn't provide getters and setters to begin with because we wanted to discourage inefficient usage.

I think the right way to do this is keep the data as Python types until the call to log, and do the conversion inside the log method as a temporary conversion and then discard the temporary before returning. This is how I implemented the C++ conversion and serialization efficiently. This would also let us do things like log a large byte array without allocation and copying (e.g. from image, audio, and video types.) Then getters and setters simply access the Python type and are cheap and natural in Python, but I don't think we have time for that now.

Given the above constraint, we can choose to provide the interface now, but inefficiently, or we can keep the status quo and not facilitate the API to be used in a way that will perform very poorly. I argue that we should keep the status quo until we have the time to implement it correctly. It will force users to write their code in a way that works reasonably well today, and when we relax that constraint in the future, it won't break existing code, but the API will get friendlier. At most we could think of providing getters only now, keeping the schema classes immutable. But the user could work around the lack of getters by declaring intermediate variables and using those instead - which avoid the additional Rust -> Python conversion.

I think we can just call this out clearly in the documentation now, and that's what this PR does.


amacneil (approved):

eloff:
> Documenting the current behavior is good.
> 
> We should definitely implement getters/setters though, this was a confusing/surprising limitation.

Yes. I think it will take 1-2 weeks of a dev working on it full time to implement it correctly, judging by how long the C++ version took me. As soon as we can prioritize that effort, let's implement it.

PR #438, opened 2025-05-26
Add a c++ example using the typed channels and foxglove

eloff:
Add an example of using the C++ typed channels, similar to the box and frame transform example in Rust.

The code works and shows more or less how to use the channels, but opening the mcap in the foxglove app is very much underwhelming. I think that cold be improved on, and I lack the domain specific knowledge on how to do that.


eloff:
> > The code works and shows more or less how to use the channels, but opening the mcap in the foxglove app is very much underwhelming. I think that cold be improved on, and I lack the domain specific knowledge on how to do that.
> 
> Maybe hosting a websocket server would be a slightly nicer demo?

I thought about that, but it shows up about the same in the app and the code is much more verbose which kind of buries the lede about what's different with the typed channels.

PR #439, opened 2025-05-27
Add websocket server method to get app URL

gasmith:
### Changelog
- rust, python: Add websocket server method to get app URL

### Docs
None

### Description
This change adds a method to the websocket server handle to return an app URL that you can use to open the websocket as a data source, with an optional layout.

### DIFF
--- a/rust/examples/ws_server/src/main.rs
+++ b/rust/examples/ws_server/src/main.rs
@@ -113,6 +113,10 @@ async fn main() {
         .await
         .expect("Server failed to start");
 
+    let app_url = server.app_url();
+    println!("View in browser: {app_url}");
+    println!("View in desktop: {}", app_url.with_open_in_desktop());
+
     tokio::task::spawn(log_forever(args.fps));
     tokio::signal::ctrl_c().await.ok();
     server.stop().wait().await;
--- a/rust/foxglove/Cargo.toml
+++ b/rust/foxglove/Cargo.toml
@@ -55,6 +55,7 @@ tokio-tungstenite = { workspace = true, optional = true }
 tokio-util = { workspace = true, optional = true }
 tokio = { workspace = true, optional = true }
 tracing.workspace = true
+urlencoding = "2.1.3"
 
 [dev-dependencies]
 assert_matches = "1.5.0"
--- a/rust/foxglove/src/app_url.rs
+++ b/rust/foxglove/src/app_url.rs
@@ -1,14 +1,12 @@
 //! App URL builder
 
-use std::borrow::Cow;
 use std::fmt::Display;
 
 static BASE_URL: &str = "https://app.foxglove.dev";
 
 #[derive(Debug, Clone)]
 enum DataSource {
-    #[allow(dead_code)] // until `AppUrl::with_websocket` is public.
-    WebSocket(String, u16),
+    WebSocket(String),
 }
 
 /// Foxglove app URL.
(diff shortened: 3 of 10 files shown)


rust/examples/ws_server/src/main.rs
```diff
@@ -113,6 +113,10 @@ async fn main() {
         .await
         .expect("Server failed to start");
 
+    let app_url = server.app_url();
```
eloff, line 116: Outside of the scope of this PR, but what if we just opened the url here in the browser?

rust/foxglove/Cargo.toml
```diff
@@ -55,6 +55,7 @@ tokio-tungstenite = { workspace = true, optional = true }
 tokio-util = { workspace = true, optional = true }
 tokio = { workspace = true, optional = true }
 tracing.workspace = true
+urlencoding = "2.1.3"
```
eloff, line 58: There's also this crate: https://crates.io/crates/percent-encoding

They both seem popular and neither have dependencies. I think there's no reason to change this choice.

rust/foxglove/src/app_url.rs
```diff
@@ -0,0 +1,114 @@
...
+pub struct AppUrl {
+    data_source: Option<DataSource>,
+    layout_id: Option<String>,
+    open_in_desktop: bool,
+}
+
+impl Display for AppUrl {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(BASE_URL)?;
+        for (ii, (k, v)) in self.query_params().into_iter().enumerate() {
```
eloff, line 25: There is also the url crate we could have used for this, but I think given our needs at present are very minimal, this is better, as it doesn't bring additional dependencies.

rust/foxglove/src/app_url.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// Returns a vector of URL query parameters.
+    ///
+    /// The parameter values are not URL-encoded.
+    fn query_params(&self) -> Vec<(&str, Cow<str>)> {
+        let mut params = vec![];
+        if let Some(ds) = &self.data_source {
+            match ds {
+                DataSource::WebSocket(host, port) => params.extend([
+                    ("ds", "foxglove-websocket".into()),
+                    ("ds.url", format!("ws://{host}:{port}").into()),
```
eloff, line 49: That reminds me we need to add SSL support sometime

rust/examples/ws_server/src/main.rs
gasmith: Not a bad idea. I guess we'd need a little platform-specific goo (or the [open](https://docs.rs/open/5.3.2/open/) crate) to make that work, but we could do it.

rust/foxglove/src/app_url.rs
gasmith: Haha, yep.

gasmith: Actually, maybe we should keep the `with_websocket` method `pub(crate)` for now until we get that figured out?

gasmith: > Actually, maybe we should keep the `with_websocket` method `pub(crate)` for now until we get that figured out?

Nah, the right way is to just have `with_websocket` just take a URL. That ought to be sufficiently future-proof to make it public.

rust/examples/ws_server/src/main.rs
eloff: Yeah, it's an idea for the future

eloff (approved):
LGTM

PR #442, opened 2025-05-29
foxglove_proto_gen: Remove Primitive suffix filter

gasmith:


eloff (approved):
I was against this idea when we discussed it last: https://foxglove.slack.com/archives/C07RB3SSJE5/p1738695908534439

I guess we can't forsee what users will and won't do with these types. I don't see a point for an ArrowPrimitive channel or a Color channel, but it doesn't do much harm to support it either. :man_shrugging: 

I would say discuss it on engineering, but we did that last time and nobody was interested in it, which sums up my level of enthusiasm about the change as well.

I'm going to approve it, if you think it's the right call, I'll defer to your judgement.

PR #458, opened 2025-06-02
MP3 to MCAP convert example

banisadr:
### Changelog
None

### Docs

None
### Description
Open to debate if this belongs. Vibe coded a mp3 to mcap converter so I could test some stuff out.


nidanin:
My concern with this is that it might be seen as encouraging users to use mp3 and it seems likely that we are going to steer them away from that for our CompressedAudio schema.

banisadr:
> My concern with this is that it might be seen as encouraging users to use mp3 and it seems likely that we are going to steer them away from that for our CompressedAudio schema.

Interesting, yeah I don't have any context on that. Is there a slack thread/doc/linear somewhere with thoughts on it that I can read?

nidanin:
https://linear.app/foxglove/issue/FG-11327/decide-on-compressedaudio-schemasupport

https://github.com/foxglove/foxglove-sdk/pull/428

Nothing's been decided yet though.

PR #459, opened 2025-06-02
build: containerize everything

sofuture:
### Changelog
None

### Docs
None

### Description
Move task execution to a makefile, and declare the (non-language-toolchain) ecosystem inside of a docker container.

### DIFF
--- a/Container.mk
+++ b/Container.mk
@@ -1,6 +1,7 @@
 .PHONY: generate
 generate:
 	poetry install
+	yarn install
 	yarn generate
 
 .PHONY: lint-python
--- a/Container.mk
+++ b/Container.mk
@@ -4,6 +4,12 @@ generate:
 	yarn install
 	yarn generate
 
+.PHONY: build-python
+build-python:
+	poetry -C python/foxglove-sdk check --strict
+	poetry -C python/foxglove-sdk install
+	poetry -C python/foxglove-sdk run maturin develop
+
 .PHONY: lint-python
 lint-python:
 	poetry check --strict
@@ -48,3 +54,32 @@ test-rust-foxglove-no-default-features:
 .PHONY: docs-rust
 docs-rust:
 	cargo +nightly rustdoc -p foxglove --all-features -- -D warnings --cfg docsrs
+
+.PHONY: clean-cpp
+clean-cpp:
+	rm -rf cpp/build*
+
+.PHONY: docs-cpp
+docs-cpp:
+	poetry install -C cpp/foxglove/docs
+	make -C cpp docs
+
+.PHONY: build-cpp
+build-cpp:
+	make -C cpp build
+
+.PHONY: build-cpp-tidy
+build-cpp-tidy:
+	make -C cpp CLANG_TIDY=true build
+
+.PHONY: lint-cpp
+lint-cpp:
+	make -C cpp lint
+
+.PHONY: test-cpp
+test-cpp:
+	make -C cpp test
+
+.PHONY: test-cpp-sanitize
+test-cpp-sanitize:
+	make -C cpp SANITIZE=address,undefined test
--- a/Container.mk
+++ b/Container.mk
@@ -1,0 +1,25 @@
+.PHONY: generate
+generate:
+	poetry install
+	yarn generate
+
+.PHONY: build-rust
+build-rust:
+	cargo build --all-targets
+
+.PHONY: lint-rust
+lint-rust:
+	cargo fmt --all --check
+	cargo clippy --no-deps --all-targets --tests -- -D warnings
+
+.PHONY: test-rust
+test-rust:
+	cargo test --all-features
+
+.PHONY: test-rust-foxglove-no-default-features
+test-rust-foxglove-no-default-features:
+	cargo test -p foxglove --no-default-features
+
+.PHONY: rust-docs
+rust-docs:
+	cargo +nightly rustdoc -p foxglove --all-features -- -D warnings --cfg docsrs
--- a/Container.mk
+++ b/Container.mk
@@ -1,5 +1,3 @@
-MSRV_RUST_VERSION=1.83.0
-
 .PHONY: generate
 generate:
 	poetry install
--- a/Container.mk
+++ b/Container.mk
@@ -45,6 +45,6 @@ test-rust:
 test-rust-foxglove-no-default-features:
 	cargo test -p foxglove --no-default-features
 
-.PHONY: rust-docs
-rust-docs:
+.PHONY: docs-rust
+docs-rust:
 	cargo +nightly rustdoc -p foxglove --all-features -- -D warnings --cfg docsrs
--- a/Container.mk
+++ b/Container.mk
@@ -1,3 +1,5 @@
+MSRV_RUST_VERSION=1.83.0
+
 .PHONY: generate
 generate:
 	poetry install
@@ -34,7 +36,7 @@ build-rust:
 
 .PHONY: build-rust-foxglove-msrv
 build-rust-foxglove-msrv:
-	cargo +1.83.0 build -p foxglove --all-features
+	cargo +$(MSRV_RUST_VERSION) build -p foxglove --all-features
 
 .PHONY: test-rust
 test-rust:
--- a/Container.mk
+++ b/Container.mk
@@ -3,14 +3,30 @@ generate:
 	poetry install
 	yarn generate
 
-.PHONY: build-rust
-build-rust:
-	cargo build --all-targets
+.PHONY: lint-python
+lint-python:
+	poetry check --strict
+	poetry install
+	poetry run black . --check
+	poetry run isort . --check
+	poetry run flake8 .
+
+.PHONY: test-python
+test-python:
+	poetry check --strict
+	poetry install
+	poetry run maturin develop
+	poetry run mypy .
+	poetry run pytest --with-benchmarks
 
 .PHONY: lint-rust
 lint-rust:
 	cargo fmt --all --check
 	cargo clippy --no-deps --all-targets --tests -- -D warnings
+
+.PHONY: build-rust
+build-rust:
+	cargo build --all-targets
 
 .PHONY: test-rust
 test-rust:
--- a/Container.mk
+++ b/Container.mk
@@ -7,9 +7,9 @@ generate:
 lint-python:
 	poetry check --strict
 	poetry install
-	poetry run black . --check
-	poetry run isort . --check
-	poetry run flake8 .
+	poetry run black python --check
+	poetry run isort python --check
+	poetry run flake8 python
 
 .PHONY: test-python
 test-python:
--- a/Container.mk
+++ b/Container.mk
@@ -13,11 +13,15 @@ lint-python:
 
 .PHONY: test-python
 test-python:
-	poetry check --strict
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -0,0 +1,47 @@
+IMAGE_NAME=foxglove-sdk
+STABLE_RUST_VERSION=1.83.0
+
+.PHONY: default
+default: build
```
gasmith: Nit: There's no `build` target

Makefile
```diff
@@ -0,0 +1,47 @@
...
+			echo "Building $$package" && \
+			cargo build -p "$$package"; \
+		done
+	cargo build -p foxglove --verbose --no-default-features
+	# Validate that we can build against the MSRV (minimum specified rust version).
+	cargo +$(STABLE_RUST_VERSION) build -p foxglove --verbose
+	cargo clippy --no-deps --all-targets --tests -- -D warnings
+	cargo +nightly rustdoc -p foxglove --all-features -- -D warnings --cfg docsrs
+	cargo test --all-features --verbose
+	cargo test -p foxglove --no-default-features --verbose
```
gasmith: - We should bust this up into individual targets. We'll want to run these from CI, but I don't think devs would typically want to run it as a monolith. I'd suggest:
  - build-rust: just `cargo build --all-targets`
  - build-rust-examples: i don't think that loop is going to work well without pipefail; we should probably pull that out into a bash script.
  - build-rust-foxglove-stable
  - lint-rust: fmt & clippy
  - test-rust: just `cargo test --all-features`
  - test-rust-foxglove-no-default-features
  - rust-docs
- We should drop `--verbose`, it's practically never useful.
- `foxglove_proto_gen` is part of generate and doesn't belong here.

Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM rust:1.83.0 AS builder
```
gasmith: Will dependabot bump this for us? If not, I'm mildly inclined towards `rust:latest`.

We should also `rustup toolchain install` the additional toolchains that we require:
- 1.83.0 (MSRV)
- nightly

For C++ build reproducibility, we might add one more version in there. Or perhaps we won't bother trying to containerize the C++ build, since we need to build targets for Mac OS and Windows...

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+WORKDIR /app
+
+RUN rustup component add rustfmt clippy
+
+RUN curl -fsSL https://deb.nodesource.com/setup_23.x -o nodesource_setup.sh
+RUN bash nodesource_setup.sh
+RUN apt install -y nodejs
```
gasmith: Prefer `apt-get` over `apt` in scripts, for compatibility.

We should do the standard `apt-get update && apt-get install -y foo && rm -rf /var/lib/apt/lists/*` dance to keep the image size from blowing up.

Makefile
```diff
@@ -0,0 +1,47 @@
...
+build-rust: image
+	docker run -v $(shell pwd):/app $(IMAGE_NAME) make build-rust-inside
+
+.PHONY: generate
+generate: image
+	docker run -v $(shell pwd):/app $(IMAGE_NAME) make generate-inside
+
+.PHONY: generate-inside
+generate-inside:
+	poetry install
```
gasmith: Do we also need a `yarn install`?

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN corepack enable yarn
+
+RUN apt-get update && \
+    apt install -y protobuf-compiler python3.11-dev clang-format
```
gasmith: We definitely want to be more specific about protoc and clang versions.

Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM rust:1.83.0 AS builder
+
+WORKDIR /app
```
gasmith, line 5: Nit: App seems weird. `/sdk`? `/src`?

Makefile
```diff
@@ -0,0 +1,47 @@
...
+	yarn generate
```
gasmith: What's the best way to ensure that the files we write are owned by the current user, for poor folks that are still on root-ful container runtimes? Userns? Bake the user's UID into the container image and create user account there?

Makefile
sofuture: Agree with all of these, my initial approach was 'copy github flow in here' -- should definitely decompose to useful targets.

ritwik:
As a general thought – should we split the container by language being generated so that we can simplify some of the setup steps by using a prebuilt language container + the dependency installation command for that language (e.g. the node container, the python containers)?

gasmith:
> As a general thought – should we split the container by language being generated so that we can simplify some of the setup steps by using a prebuilt language container + the dependency installation command for that language (e.g. the node container, the python containers)?

Unfortunately, the codegen step needs rust, node+webpack, python+poetry, and clang-format.

Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+RUN rustup toolchain install nightly
+
+RUN curl -fsSL https://deb.nodesource.com/setup_23.x -o nodesource_setup.sh
+RUN bash nodesource_setup.sh
+
+RUN apt-get update \
+    && apt-get install -y \
+        protobuf-compiler=3.21.12-3 \
+        python3.11-dev=3.11.2-6+deb12u6 \
+        clang-format=1:14.0-55.7~deb12u1 \
```
bryfox: We should try to keep these in sync with whatever CI uses — at least clang-format and protobuf-compiler.

Right now CI uses clang-format 18.1.3, but I don't think we specify it anywhere.

sofuture:
I think the rustfmt error was from ordering, possibly.

Dockerfile
sofuture: Ideally we can move CI to use this same tooling, or if not exactly that have it use the same declarations somehow.

bryfox:
> I think the rustfmt error was from ordering, possibly.

I'm getting the same error with the latest updates. The following patch fixes it for me.

```diff
-RUN rustup toolchain install nightly
-RUN rustup toolchain install stable
-RUN rustup component add rustfmt clippy
+RUN rustup toolchain install nightly --component rustfmt --component clippy
+RUN rustup toolchain install stable --component rustfmt --component clippy
```

sofuture:
Interestingly that patch now causes me the get the error, seems like cargo toolchain selection nonsense.

Container.mk
```diff
@@ -0,0 +1,41 @@
...
+.PHONY: lint-python
+lint-python:
+	poetry check --strict
+	poetry install
+	poetry run black . --check
+	poetry run isort . --check
+	poetry run flake8 .
+
+.PHONY: test-python
+test-python:
```
gasmith, line 22: This need to run against python/foxglove-sdk/. Looks like poetry takes a `-C <dir>` argument.

Container.mk
```diff
@@ -0,0 +1,41 @@
...
+	poetry check --strict
+	poetry install
+	poetry run maturin develop
+	poetry run mypy .
+	poetry run pytest --with-benchmarks
```
gasmith: Let's separate the benchmarks out into their own target. We should still keep a `pytest` here, too.

Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+.PHONY: shell
+shell: image
+	docker run -v $(shell pwd):/app -it $(IMAGE_NAME) bash
+
+TARGETS := $(shell awk '/^\.PHONY:/ {for(i=2;i<=NF;i++) print $$i}' $(CONTAINER_MAKEFILE))
+
+.PHONY: $(TARGETS)
+$(TARGETS): image
+	docker run -v $(shell pwd):/app -it $(IMAGE_NAME) make -f $(CONTAINER_MAKEFILE) $@
```
gasmith: Let's add `-e CARGO_HOME=/app/.cargo` and `-e POETRY_VIRTUALENVS_PATH=/app/.virtualenvs` (also update `.gitignore` to exclude those directories), so that rust sources and python virtualenvs are cached across invocations.

Container.mk
```diff
@@ -0,0 +1,41 @@
...
+
+.PHONY: lint-rust
+lint-rust:
+	cargo fmt --all --check
+	cargo clippy --no-deps --all-targets --tests -- -D warnings
+
+.PHONY: build-rust
+build-rust:
```
gasmith, line 39: Add a `build-rust-foxglove-msrv` target which does `cargo +1.83.0 -p foxglove build --all-features`. I realize the `--all-features` bit is currently missing from CI.

Dockerfile
bryfox: Discussed out-of-band: the clang-format version above isn't available in deb12, so let's explicitly install a version in CI to match (at least the major version).

Dockerfile
gasmith: I'm pretty sure we should base on `rust:latest`, and only use MSRV for validating the foxglove-sdk build.

gasmith (commented):
Took it for a spin this morning, and most things are working. A few adjustments to the python targets are required.

Dockerfile
gasmith: v14 is pretty old. Could we do [v19](https://packages.debian.org/bookworm/devel/clang-format-19)?

pyproject.toml
```diff
@@ -11,6 +11,7 @@ package-mode = false
 flake8 = "^7.1.1"
 black = "^25.1.0"
 isort = "^6.0.0"
+maturin = "^1.8.6"
```
gasmith: Not required if you take my advice to run `poetry -C python/foxglove-sdk` for the `test-python` target.

Makefile
```diff
@@ -0,0 +1,28 @@
...
+.PHONY: shell
+shell: image
+	docker run -v $(shell pwd):/app -it $(IMAGE_NAME) bash
+
+TARGETS := $(shell awk '/^\.PHONY:/ {for(i=2;i<=NF;i++) print $$i}' $(CONTAINER_MAKEFILE))
+
+.PHONY: $(TARGETS)
+$(TARGETS): image
+	docker run -v $(shell pwd):/app \
+		-e CARGO_HOME=/app/.cargo \
```
gasmith, line 26: Ditto for the `shell` target.

Makefile
```diff
@@ -0,0 +1,28 @@
+IMAGE_NAME=foxglove-sdk
+STABLE_RUST_VERSION=1.83.0
```
gasmith: This might need to move into `Container.mk`, where we can use it for the msrv target? Maybe also rename it: MSRV is a different concept from stable.

Dockerfile
sofuture: yeah, we can get 19, one of those dumb <pkgname>-<version> packages

Makefile
sofuture: no (I don't think)

gasmith:
@bryfox @sofuture: I've pushed a bunch of little tweaks... could you give this another try in your environments?

sofuture:
linting, testing and generating all work for me. Let's ship this when we're relatively good, and iterate new targets afterwards.

gasmith (approved):
:ship:

Makefile
gasmith: it is indeed. i nuked node_modules, and it broke `make generate`. i added a `yarn install`.

PR #460, opened 2025-06-03
rust data loaders

nthiad:
### Changelog

Adds a `foxglove_data_loader` crate for writing foxglove data loader extensions in rust that compile to wasm.

### Docs

[FG-11831](https://linear.app/foxglove/issue/FG-11831/document-writing-a-data-loader-extension-with-the-sdk)

### Description

This PR adds a `foxglove_data_loader` crate with `DataLoader` and `MessageIterator` traits for user code to implement plus a `foxglove_data_loader::export!()` macro which exports user data loader implementation to the wasm component build.

The wasm output of builds made with this interface can be passed as `wasmUrl` to `extensionContext.registerDataLoader()`.

Usage looks like:

``` rs
use foxglove_data_loader::{
    reader, console, BackfillArgs, Channel, DataLoader, DataLoaderArgs, Initialization,
    Message, MessageIterator, MessageIteratorArgs, Schema, TimeRange,
};
foxglove_data_loader::export!(MyLoader);

struct MyLoader { /* ... */ }

impl DataLoader for MyLoader {
    type MessageIterator = MyIterator;
    type Error = anyhow::Error;

    fn new(args: DataLoaderArgs) -> Self {
        // args.paths is a Vec<String> of file paths you can reader::open()
        unimplemented![]
    }

    fn initialize(&mut self) -> Result<Initialization, Self::Error> {
        // the Initialization contains the channels, schemas, time range, and any problems
        unimplemented![]
    }

    fn create_iter(&mut self, args: MessageIteratorArgs) -> Result<Self::MessageIterator, Self::Error> {
        unimplemented![]
    }

    fn get_backfill(&mut self, args: BackfillArgs) -> Result<Vec<Message>, Self::Error> {
        unimplemented![]
    }
}

struct MyIterator { /* ... */ }

impl MessageIterator for MyIterator {
    type Error = anyhow::Error;

    fn next(&mut self) -> Option<Result<Message, Self::Error>> {
        unimplemented![]
    }
}
```

There is a `data_loader::reader::open(&path)` function that returns a Reader from the internal WIT. The `inputs` vec contains a list of path strings you can `reader::open()`. `std::io::Read` and `std::io::Seek` are implemented for the WIT Reader to make it easier to use with the rest of the rust ecosystem. For example, you can do things like:

``` rs
let reader = foxglove_data_loader::reader::open(&inputs[0]);
let mut lines = BufReader::new(reader).lines();
```

Depends on https://github.com/foxglove/app/pull/9659

Fixes: FG-11694


defunctzombie:
Once I have this in a source file - what do I do to build it?

defunctzombie:
> The ::foxglove is to disambiguate an internally generated foxglove module derived from the WIT package name from the sdk itself. There might be a work-around for this.

My editor keeps removing the leading `::` so yea would be nice to not have this leak into the user's code

nthiad:
> Once I have this in a source file - what do I do to build it?

```
cargo add --path ~/dev/foxglove/sdk --no-default-features --features data_loader,derive
cargo build --release --target wasm32-unknown-unknown
```

> My editor keeps removing the leading :: so yea would be nice to not have this leak into the user's code

I agree. The issue is that wit-bindgen defines the `foxglove` from the WIT line `package foxglove:loader@0.1.0` in the current scope so I will need to do more hacks or patch upstream to get around this.

nthiad:
> My editor keeps removing the leading :: so yea would be nice to not have this leak into the user's code

The `::foxglove` disambiguation is no longer an issue on the latest commit.

nthiad:
I hid all the Guest traits and made `DataLoader` and `MessageIterator` internal details. Now you write your loader and iterator with `impl` and so long as it has the correct methods with the correct signatures, things will work.

rust/foxglove/src/data_loader.rs
```diff
@@ -0,0 +1,170 @@
...
+                    std::io::SeekFrom::Start(offset) => {
+                        reader::Reader::seek(&self, offset);
+                    },
+                    std::io::SeekFrom::End(offset) => {
+                        let end = reader::Reader::size(&self) as i64;
+                        reader::Reader::seek(&self, (end - offset) as u64);
+                    },
+                    std::io::SeekFrom::Current(offset) => {
+                        let pos = reader::Reader::position(&self) as i64;
+                        reader::Reader::seek(&self, (pos - offset) as u64);
```
defunctzombie:
```suggestion
                        reader::Reader::seek(&self, (pos + offset) as u64);
```

nthiad:
If we have an example, we should include an example of doing mutable state with RefCell because the `&self` makes it look like you can't have mutable state, but that is not true, it just requires a bit more work. The `&self` refs we get from WIT directly and probably should keep that since there is a perf penalty for using mutability containers like RefCell or Mutex or RwLock.

defunctzombie:
Gave this a spin writing a basic mp4 data loader using this branch paired with https://github.com/foxglove/app/pull/9659. I've ticketed my feedback here:

- https://linear.app/foxglove/issue/FG-11903/roman-feedback-no-ability-to-have-some-channels-and-some-problems
- https://linear.app/foxglove/issue/FG-11904/roman-feedback-the-result-type-returns-string-as-errors
- https://linear.app/foxglove/issue/FG-11905/roman-feedback-need-a-nicer-way-to-create-channels-for-schemas-that

Generally my vibe is this is a great pairing with the app-side experimental PR and something we can get folks tire kicking with.

As for what might block this PR merging I have the following thoughts:

- In the app side we clearly mark these apis as `experimental`, is there some equivalent thing we can do here? We use the `data_loader` feature so maybe that's sufficient or is there some other idiomatic way to indicate this isn't a stable feature?
- It would be nice to have some comments on the trail members. Stuff like ` fn time_range(&self) -> Result<loader::TimeRange, String>;` what should I return?
- `/// And don't forget to call `foxglove::data_loader_export()` on your loader.` nit: generally not a fan of "don't forget" language. Tell the user what to do.
- nit: drop the "custom" everywhere it exists - what's custom about them?

nthiad:
> In the app side we clearly mark these apis as experimental, is there some equivalent thing we can do here? We use the data_loader feature so maybe that's sufficient or is there some other idiomatic way to indicate this isn't a stable feature?

I haven't seen this as much in crates, mostly in rust itself, but here is one example:
https://github.com/async-rs/async-std/blob/v1.8.0/Cargo.toml#L38-L42

It's more complicated in our case because we are using features to gate against wasm-specific functionality already to to also gate for stability we could rename the feature `data_loaders_unstable` or `data_loaders_experimental`. Or another option could be to keep the name `data_loaders` but gate that behind something like `wasm_unstable`. I don't have a good handle of what naming scheme would be more common for this sort of thing.

nthiad:
I still need to figure out why errors in the rust code were reporting "unreachable" in the problems panel without the msg itself. `.expect()/.unwrap()` and `panic![]` were reporting errors to the problems panel correctly, but those might give us issues with proper cleanup and error recovery.

nthiad:
I'm not sure what I did but errors are reported correctly in the problems area now:

error

I raised that error with:

``` rs
return Err(anyhow!["hey ok failure nice"]);
```

in my `create()` function.

amacneil:
> In the app side we clearly mark these apis as experimental, is there some equivalent thing we can do here? We use the data_loader feature so maybe that's sufficient or is there some other idiomatic way to indicate this isn't a stable feature?

Since this is a separate crate, one way we can do this is just not publish the crate. Add a readme saying its experimental and must be imported as a git crate.

rust/foxglove_data_loader/examples/ndjson.rs
```diff
@@ -0,0 +1,374 @@
...
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+struct Sphere {
+    id: String,
+    time: f64,
+    x: f64,
+    y: f64,
+    z: f64,
+}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
```
achim-k: You don't need to define the schema below If you derive from `foxglove::Encode` here. This generates a schema for you and handles encoding into protobuf. See also https://docs.rs/foxglove/latest/foxglove/#custom-data

Example where this is used: 
https://github.com/foxglove/mp4-data-loader/blob/49c27d5707cb5bb6eb5f4b152d8886e25e00f99b/data-loader/src/lib.rs#L16-L17 (see that same file for how to get the schema and encode messages)

rust/foxglove_data_loader/examples/ndjson.rs
```diff
@@ -0,0 +1,374 @@
+//! This example data loader parses newline-separated lines of JSON of 3 forms:
+//!
+//! {"type":"info","start":0,"end":30,"sphere_count":360,"temperature_count":120}
```
achim-k: I would probably move the `info` record to the end of the file. If you imagine that the file is written by appending new temperature readings, it is somewhat unexpected to have the summary record at the start of the file as the end time is not known yet.

Do we actually need that record? You currently read the whole file to memory, so you could also derive start and end time from the first and last line?

rust/foxglove_data_loader/examples/ndjson.rs
```diff
@@ -0,0 +1,374 @@
...
+//! {"type":"temperature","time":0,"ambient":21,"cpu0":70,"cpu1":65,"cpu2":68,"cpu3":72}
+//! {"type":"sphere","id":"A","time":0,"x":0,"y":2,"z":1.5}
```
achim-k: The sphere messages seem a bit arbitrary and don't really fit to the temperature messages. Maybe we could do "log" messages instead? Would also help keeping this example brief.

rust/foxglove_data_loader/examples/ndjson.rs
```diff
@@ -0,0 +1,374 @@
...
+//!
+//! The loader stores the records in memory and puts the sphere data into a SceneUpdate topic called
+//! scene and the temperature rows go into a topic called /temperature.
+
+#[cfg(not(target_family = "wasm"))]
+fn main() {}
```
achim-k: this can be removed?

rust/foxglove_data_loader/examples/ndjson.rs
```diff
@@ -0,0 +1,374 @@
...
+        else {
+            bail!["did not find type=\"info\" row"]
+        };
+        self.rows.replace(rows);
+        console::log(&format!["info={info:?}"]);
+
+        let sphere_channel = Channel {
+            id: SPHERE_CHANNEL_ID,
+            schema_id: Some(SPHERE_SCHEMA_ID),
+            topic_name: "/scene".to_string(),
```
achim-k: is the `to_string()` necessary?

rust/foxglove_data_loader/examples/ndjson.rs
```diff
@@ -0,0 +1,374 @@
...
+    id: String,
+    time: f64,
+    x: f64,
+    y: f64,
+    z: f64,
+}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+struct Temperature {
+    time: f64,
```
achim-k: A bit unclear what unit the time is in. Maybe rename it to `time_seconds`? Alternatively add a comment. Also applies for the temperature fields below

rust/foxglove_data_loader/examples/ndjson.rs
```diff
@@ -0,0 +1,374 @@
...
+            message_count: Some(info.sphere_count),
+        };
+        let sphere_schema = Schema::from_id_sdk(
+            SPHERE_SCHEMA_ID,
+            SceneUpdate::get_schema().ok_or(anyhow!["failed to get SceneUpdate schema"])?,
+        );
+
+        let temperature_channel = Channel {
+            id: TEMPERATURE_CHANNEL_ID,
+            schema_id: Some(TEMPERATURE_SCHEMA_ID),
```
achim-k: it would be nice if the author could just specify the schema here and the schema id assignment happens in the background.

rust/foxglove_data_loader/examples/ndjson.rs
nthiad: Yes since string literals are `&str` and `topic_name` needs to be an owned `String`. If this were a reference then the struct would need a lifetime type parameter.

rust/foxglove_data_loader/examples/ndjson.rs
nthiad: This is required so that `cargo test` which is run by CI won't fail. `cargo test` compiles examples and will otherwise complain that there is no `main()`.

nthiad:
I've removed the ndjson example from this PR in favor of the location in create-foxglove-extension with my other PR so that we only have one copy of this example and so that they don't get out of sync. I'll add a follow-up PR to this crate with a different example and I'll apply all the example feedback here to the ndjson rust example in create-foxglove-extension.

rust/foxglove_data_loader/examples/ndjson.rs
nthiad: i swapped these out for an accelerometer message using a Vector3. it feels less contrived and the example got simpler

rust/foxglove_data_loader/examples/ndjson.rs
nthiad: I removed the record and derive these instead.

rust/foxglove_data_loader/examples/ndjson.rs
nthiad: i added a builder interface for `Initialization`:

``` rs
        let mut init = Initialization::builder()
            .start_time(seconds_to_nanos(start_seconds))
            .end_time(seconds_to_nanos(end_seconds));

        let mut accelerometer_schema = init.add_schema::<Vector3>()?;
        let accelerometer_channel = accelerometer_schema
            .add_channel("/accelerometer")
            .message_count(accelerometer_count as u64);
        self.accelerometer_channel_id.replace(accelerometer_channel.id);
            
        let mut temperature_schema = init.add_schema::<Temperature>()?;
        let temperature_channel = temperature_schema
            .add_channel("/temperature")
            .message_count(temperature_count as u64);
        self.temperature_channel_id.replace(temperature_channel.id);

        Ok(init.build())
```

Still need to save the channel ids to pass them around to different places, but I think this is an improvement to the ergonomics. And the previous interface works the same as before, if users prefer that level of direct control.

nthiad:
I've been thinking `iter()` might be better than `create_iter()` in the trait to better match rust namings. And the MessageIterator could be impl Iterator so it's easier to whip up something without having to necessarily implement the trait directly.

bennetthardwick:
> I've been thinking `iter()` might be better than `create_iter()` in the trait to better match rust namings. And the MessageIterator could be impl Iterator so it's easier to whip up something without having to necessarily implement the trait directly.

I feel like `iter()` implies a cheap iteration over data without copying etc. It also doesn't normally take any arguments.

nthiad:
I moved things around for inner mutability (a constraint that WIT gives us) so the trait methods are all now `&mut self`. This way implementations don't need to wrap all mutable state in `RefCell` and we don't need to explain why mutable state is complicated.

bennetthardwick (approved):

PR #463, opened 2025-06-05
build: Containerize C++ builds too

gasmith:
Let's support C++ too, why not.


cpp/CMakeLists.txt
```diff
@@ -24,7 +24,6 @@ FetchContent_Declare(
   Catch2
   GIT_REPOSITORY https://github.com/catchorg/Catch2.git
   GIT_TAG        v3.8.0
-  SYSTEM
```
gasmith, line 27: I'm not sure why this was here, but it was breaking git fetch in an inscrutable way for cmake 3.25, and things seem to work just fine without it.

PR #464, opened 2025-06-06
Log struct with jsonschema & msgpack

bryfox:
### Changelog
None

### Docs
None

### Description

This adds a C++ example which illustrates logging of custom data without manual serialization.

The example sets up two channels; one with a json schema derived from the struct, and one using [msgpack](https://msgpack.org/). The latter is schemaless, but provides a binary option. The Foxglove app handles this in the same way as schemaless JSON. A handful of messages are written to an MCAP file. Both techniques rely on https://json.nlohmann.me/, which is also used by the ws-services example.

This provides a couple of options for users who want to put together logging of custom schemas quickly, without caring about the serialization format. In python, we accomplish this by logging native types (dict, etc); in rust, we provide derive macros for `Encode`, or illustrate usage with `schemars::JsonSchema`.


cpp/examples/auto-serialize/src/main.cpp
```diff
@@ -0,0 +1,107 @@
...
+    std::cerr << "Failed to create MsgPack channel: " << foxglove::strerror(chan1_result.error()) << '\n';
+    return 1;
+  }
+  auto ch1 = std::move(chan1_result.value());
+
+  // 2: Channel with MsgPack
+  auto chan2_result = foxglove::RawChannel::create("/msgpack", "msgpack");
+  if (!chan2_result.has_value())
+  {
+    std::cerr << "Failed to create JSON channel: " << foxglove::strerror(chan2_result.error()) << '\n';
```
defunctzombie: msgpack channel?

cpp/examples/auto-serialize/src/jsonschema.hpp
```diff
@@ -0,0 +1,58 @@
...
+template<typename T, typename U>
+struct field_type<T U::*, U> {
+  using type = T;
+};
+
+/**
+ * Generate a simple JSON schema for object types, iterating over fields. Nested objects and
+ * arrays are not described, and enum values are treated as their JSON representation.
+ */
+template<typename T>
```
eloff, line 20: It's pretty amazing how this works

cpp/examples/auto-serialize/src/main.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+NLOHMANN_JSON_SERIALIZE_ENUM(MessageLevel, {{DEBUG, "debug"}, {INFO, "info"}})
+
+struct Message {
+  MessageLevel level;
+  std::string msg;
+  int count;
+};
+
+NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(Message, level, msg, count)
```
eloff, line 36: This is how the to_json/from_json get defined, I thought it was impossible for a moment until I realized you're passing the struct and the fields to serialize, that makes sense

bryfox: Yeah, this is straight from their documentation, so I've added a link to that.

PR #466, opened 2025-06-07
python: Add missing stub for Timestamp.now()

bryfox:
### Changelog
Fix: python: Add missing stub for Timestamp.now()

### Docs
None

### Description

We have a Timestamp.now() implementation, but were missing the stub interface.


eloff (approved):
Good catch

PR #472, opened 2025-06-10
C++: implement additional channel getters

bryfox:
### Changelog
C++: implement additional channel getters

### Description

This adds several getters on `Channel` in C++ which are present in Rust; for topic, message_encoding, has_sinks, and schema.

### DIFF
--- a/cpp/foxglove/tests/test_channel.cpp
+++ b/cpp/foxglove/tests/test_channel.cpp
@@ -47,7 +47,8 @@ TEST_CASE("channel.message_encoding()") {
 }
 
 TEST_CASE("channel.has_sinks()") {
-  FileCleanup cleanup("test.mcap");
+  auto fname = "test-channel-has-sinks.mcap";
+  FileCleanup cleanup(fname);
 
   auto context = foxglove::Context::create();
   auto channel = foxglove::RawChannel::create("test", "json", std::nullopt, context);
@@ -56,7 +57,7 @@ TEST_CASE("channel.has_sinks()") {
 
   foxglove::McapWriterOptions mcap_options = {};
   mcap_options.context = context;
-  mcap_options.path = "test.mcap";
+  mcap_options.path = fname;
   auto writer = foxglove::McapWriter::create(mcap_options);
   REQUIRE(writer.has_value());
 


c/src/lib.rs
```diff
@@ -1032,6 +1041,94 @@ pub extern "C" fn foxglove_channel_get_id(channel: Option<&FoxgloveChannel>) ->
     u64::from(channel.0.id())
 }
 
+/// Get the topic of a channel.
+///
+/// # Safety
+/// `channel` must be a valid pointer to a `foxglove_channel` created via `foxglove_channel_create`.
+///
+/// If the passed channel is null, an empty value is returned.
```
gasmith, line 1049: We may want to indicate that the returned value remains valid only until the channel is freed.

cpp/foxglove/include/foxglove/channel.hpp
```diff
@@ -71,6 +71,26 @@ class RawChannel final {
   /// @return The ID of the channel.
   [[nodiscard]] uint64_t id() const noexcept;
 
+  /// @brief Get the topic of the channel.
```
gasmith, line 74: These should also indicate the lifetime of the returned value. The `string_view`s are pretty obviously tied to the lifetime of the channel, but users might not realize that the same is true of the `Schema`'s data pointer.

cpp/foxglove/tests/test_channel.cpp
```diff
@@ -27,3 +31,68 @@ TEST_CASE("duplicate topic") {
...
+
+TEST_CASE("channel.message_encoding()") {
+  auto context = foxglove::Context::create();
+  auto channel = foxglove::RawChannel::create("test", "json", std::nullopt, context);
+  REQUIRE(channel.has_value());
+  REQUIRE(channel.value().message_encoding() == "json");
+}
+
+TEST_CASE("channel.has_sinks()") {
+  FileCleanup cleanup("test.mcap");
```
gasmith: If I'm not mistaken, all these tests run in the same working directory.

Maybe name the file after the test case to avoid conflicts. Or better yet, put it in a randomly-named file/dir under /tmp. Looks like there's a `std::filesystem::temp_directory_path` and `std::tmpnam`.

I was seriously surprised that catch2 doesn't seem to provide any support for tempdir/tempfile fixtures out of the box.

gasmith (approved):

cpp/foxglove/tests/test_channel.cpp
bryfox: > Maybe name the file after the test case to avoid conflicts. Or better yet, put it in a randomly-named file/dir under /tmp. Looks like there's a std::filesystem::temp_directory_path and std::tmpnam.

In earlier commits above, I started with `std::tmpnm`, but decided not to ignore the deprecation warnings about it being unsafe, even though it's in a test context. Then I went with `std::filesystem::temp_directory_path` files named with millis since epoch. Then I realized we had this file cleanup approach in MCAP tests, so I just went with that. Presumably we're running serially everywhere.

I like the idea to name after the test case, and I think that's reasonable for now. I'll do that here and in #475. We can revisit in the future, if/when it's causing us headaches.

gasmith: > Presumably we're running serially everywhere.

Oh right! I guess that really makes the whole thing moot.

PR #475, opened 2025-06-11
c++: Add channel.close() method

bryfox:
### Changelog
c++: Add channel.close() method

### Description

This adds a close() method to channels in C++.

This is based on #472 because the test makes use of a getter from there.


c/src/lib.rs
```diff
@@ -1014,6 +1014,19 @@ pub(crate) unsafe fn do_foxglove_channel_create<T: foxglove::Encode>(
     Ok(Arc::into_raw(builder.build::<T>().into_inner()) as *const FoxgloveChannel)
 }
 
+/// Close a channel.
+///
```
gasmith, line 1018: Might add the rest of the docs ("You can use this to explicitly unadvertise...").

Also, we should mention that calling close _does not_ free the channel.

gasmith (approved):

PR #479, opened 2025-06-11
C++: Support channel metadata

bryfox:
### Changelog
C++: Support channel metadata

### Docs
None

### Description

This adds support for channel metadata in C++. A user may supply key/value pairs when constructing the channel.

The C library provides an iterator on top of the underlying BTreeMap, which the C++ method uses to build up its own map. I expect metadata to be typically small, and this getter to be called infrequently.

**Open question:** what parameter order do we want for `::create`? I added metadata last; I think it's possibly less common to use this than Context. If you want to supply metadata, but no Context, you have to provide the default `foxglove::Context()`. Longer-term, we could provide a builder interface around this if needed, but it was more than I wanted to tackle here.

The new struct names are also up for debate. I chose the generic "FoxgloveKeyValue", thinking we might have other uses for that (e.g., supplying metadata for an MCAP file).

Finally, I'd like to break apart lib.rs at some point, but it's not going to happen here.


c/src/lib.rs
```diff
@@ -1155,6 +1191,92 @@ pub extern "C" fn foxglove_channel_has_sinks(channel: Option<&FoxgloveChannel>)
...
+        channel: channel as *const _,
+        index: 0,
+    }))
+}
+
+/// Get the next key-value pair from the metadata iterator.
+///
+/// Returns true if a pair was found and stored in `key_value`, false if the iterator is exhausted.
+///
+/// # Safety
```
gasmith, line 1229: Nit: I'd add that the channel itself must still be valid.

gasmith:
> **Open question:** what parameter order do we want for `::create`? I added metadata last; I think it's possibly less common to use this than Context. If you want to supply metadata, but no Context, you have to provide the default `foxglove::Context()`. Longer-term, we could provide a builder interface around this if needed, but it was more than I wanted to tackle here.

I think the ordering you have is reasonable. A builder feels like the right endgame, but I agree we shouldn't tackle that here.

> The new struct names are also up for debate. I chose the generic "FoxgloveKeyValue", thinking we might have other uses for that (e.g., supplying metadata for an MCAP file).

Sounds fine. Its a bummer that properties use owned strings, such that we can't reuse it there.

PR #480, opened 2025-06-12
Add rust example showing different ways to log data

eloff:
Add some examples of logging different types of structured data in Rust:

1. Logging Foxglove schemas
2. Logging custom types as protobuf
3. Logging custom types as json with schemars to generate the schema
4. Logging json with a custom schema
5. Logging schemless data


rust/examples/structured-logging/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+#[derive(Encode, Debug)]
+struct Banana {
+    length: f64,
+    ripeness: f64,
+}
+
+// This channel logs images using Foxglove's image schema
+static IMG_CHANNEL: LazyChannel<schemas::CompressedImage> = LazyChannel::new("/image");
+// This channel logs schemaless JSON
+static SCHEMALESS_CHANNEL: LazyRawChannel = LazyRawChannel::new("/schemaless", "json");
```
bryfox: Do you think "schemaless" json needs its own example in addition to json, or is it enough to point out that schema is optional?

bryfox (commented):
The quickstart example currently covers a lot of these cases, and this doesn't cover protobuf. I like the idea of one example covering all the options. Maybe we should remove some of that scope from the quickstart in favor of this one, and consolidate the protobuf example into this too?

amacneil:
let's add `*.webp` and other binary types to lfs, similar to mcap repo: https://github.com/foxglove/mcap/blob/main/.gitattributes

rust/examples/structured-logging/src/main.rs
eloff: It could be a comment, but people tend to gloss over comments. I think this illustrates it better

rust/examples/structured-logging/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+// This channel logs JSON with a jsonschema
+static POINTS_CHANNEL: LazyRawChannel = LazyRawChannel::new("/points", "json").schema(
+    "point",
+    "jsonschema",
+    r#"{"type": "object", "properties": {"x": {"type": "number"}, "y": {"type": "number"}}}"#
+        .as_bytes(),
+);
+// This channel logs a custom type as JSON using the schemars feature to generate the jsonschema automatically
+static ORANGE_CHANNEL: LazyChannel<Apple> = LazyChannel::new("/apple");
```
gasmith: `APPLE_CHANNEL`?

rust/examples/structured-logging/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+    );
+
+    // For a typed JSON channel, we can just pass the struct directly.
+    // The type system ensures conformance with the schema.
+    ORANGE_CHANNEL.log(&Apple {
+        color: "red".to_string(),
+        diameter: 10.0,
+    });
+
+    // Or if we just want to serialize a struct without caring about the format,
```
gasmith: nit: s/format/encoding/?

rust/examples/structured-logging/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[[bin]]
+name = "structured-logging"
+path = "src/main.rs"
+
+[dependencies]
+bytes.workspace = true
+env_logger = "0.11.8"
+foxglove = { version = "0.8.1", path = "../../foxglove", features = ["derive", "schemars"] }
+foxglove_derive = { version = "0.8.1", path = "../../foxglove_derive" }
```
gasmith: Shouldn't be necessary to include this as a direct dependency, right? Just import `Encode` from the `foxglove` crate.

rust/examples/structured-logging/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "structured-logging"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+repository.workspace = true
+license.workspace = true
```
gasmith: For other examples, we chose not to rely on workspace dependencies, so that you could just copy the example out of the workspace and build from it directly. There are two exceptions: one for lints, and one for the foxglove crate itself, which we typically reference by path to ensure that we don't break compatibility in CI.

In any case, we don't want to set the version on example crates.

rust/examples/structured-logging/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "structured-logging"
```
gasmith: [Adrian says](https://github.com/foxglove/foxglove-sdk/pull/429) thou shalt not kebab-case your crates or the directories containing them.

Also please prefix the crate name (but not the directory) with `example_` for uniformity with other examples and [compatibility with CI](https://github.com/foxglove/foxglove-sdk/blob/291dc7d1c13e2bc8ee2e48ca1c93122b3aebd761/.github/workflows/ci.yml#L33-L43).

rust/examples/structured-logging/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+repository.workspace = true
+license.workspace = true
+
+[[bin]]
+name = "structured-logging"
+path = "src/main.rs"
```
gasmith: Nit: not required.

PR #481, opened 2025-06-12
Add example showing how to use prost library and generating protobuf types

eloff:
This basically mirrors how we do this for our own foxglove types, but without foxglove specific things, so users could copy and paste it into their own project to get started.

It's quite involved, but I don't see how to avoid that. On the plus side, I put in the work setting it up, so hopefully it's a lot easier for others who want to do this.


rust/examples/prost-protobuf/build.rs
```diff
@@ -0,0 +1,247 @@
...
+    path::{Path, PathBuf},
+};
+
+use anyhow::Context;
+use prost::Message;
+use prost_types::{FileDescriptorProto, FileDescriptorSet};
+use walkdir::WalkDir;
+
+// This is a simplfied version of foxglove_proto_gen,
+// which we use internally to generate prost types for the Foxglove schemas in the sdk.
```
bryfox, line 14: It seems a little strange to talk about internal SDK details (and to use "we" meaning contributors) in an example. If foxglove_proto_gen is useful to me as someone following the example, maybe we could add top-level documentation about what this does, and something like "see foxglove_proto_gen for more details".

I think top-level docs would be useful regardless — what this build script does, and why I might want to use something like it.

gasmith (commented):
This example is really complicated. I don't think many users want to get this far into the weeds of protobuf codegen.

Perhaps there's a nice way to integrate with `prost_reflect_build`, such that the `build.rs` is just [the standard boilerplate from the docs](https://docs.rs/prost-reflect-build/latest/prost_reflect_build/), and we do the "heavy-lifting" to serialize a FileDescriptorSet at runtime (starting from [ReflectMessage::descriptor](https://docs.rs/prost-reflect/latest/prost_reflect/trait.ReflectMessage.html#tymethod.descriptor)). It means manually writing a `impl Encode for Apple`, but that should be trivial, apart from building the schema.

defunctzombie:
This example seems crazy hard to me compared to https://github.com/foxglove/foxglove-sdk/pull/482. What is the gap of 483 that causes the build script in this example to be so much more?

eloff:
Closing in favor of https://github.com/foxglove/foxglove-sdk/pull/482

We can always revive this in the future if we feel it would be more helpful than confusing

PR #482, opened 2025-06-12
Add a simple prost example

bryfox:
This example ties together the basic concepts documented by the prost-build, prost, and foxglove crates. It's less flexible than #481, but is a quicker way to get started with custom serialization and protobuf. I think both examples are useful.

### DIFF
--- a/rust/examples/prost/Cargo.toml
+++ b/rust/examples/prost/Cargo.toml
@@ -7,7 +7,7 @@ publish = false
 workspace = true
 
 [dependencies]
-foxglove = { path = "../../foxglove", features = [] }
+foxglove = { path = "../../foxglove", default-features = false }
 prost = "0.13.5"
 
 [build-dependencies]
--- a/rust/examples/prost/Cargo.toml
+++ b/rust/examples/prost/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "example_prost"
+edition = "2021"
+publish = false
+
+[lints]
+workspace = true
+
+[dependencies]
+foxglove = { path = "../../foxglove", features = [] }
+prost = "0.13.5"
+
+[build-dependencies]
+prost-build = "0.13.5"
--- a/rust/examples/prost/src/main.rs
+++ b/rust/examples/prost/src/main.rs
@@ -1,4 +1,4 @@
-use foxglove::{Channel, ChannelBuilder, McapWriter, Schema};
+use foxglove::{ChannelBuilder, McapWriter, Schema};
 use prost::Message;
 
 pub mod fruit {
@@ -44,7 +44,7 @@ impl foxglove::Encode for fruit::Apple {
     }
 
     fn encode(&self, buf: &mut impl prost::bytes::BufMut) -> Result<(), Self::Error> {
-        prost::Message::encode(self, buf)?;
+        Message::encode(self, buf)?;
         Ok(())
     }
 }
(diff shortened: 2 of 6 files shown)


eloff:
This is a lot simpler than the approach I took, with the limitation that you have to manually serialize the message.

I wonder maybe I can simplify the example to be more like this, but with the impl Encode generation added. I think it's possible.

rust/examples/prost/src/main.rs
```diff
@@ -0,0 +1,39 @@
...
+///
+/// To run this example, in addition to the `prost` and `prost-build` crates, you must install a
+/// [protobuf compiler](https://github.com/protocolbuffers/protobuf#protobuf-compiler-installation).
+fn main() {
+    let writer = McapWriter::new()
+        .create_new_buffered_file("fruit.mcap")
+        .expect("failed to create writer");
+
+    // Set up a channel for our protobuf messages
+    let schema = Schema::new("apple", "protobuf", APPLE_SCHEMA);
```
gasmith: Should the name be `fruit.Apple`?

rust/examples/prost/src/main.rs
```diff
@@ -0,0 +1,39 @@
+use foxglove::{ChannelBuilder, McapWriter, Schema};
+use prost::Message;
+
+pub mod fruit {
+    include!("../generated/fruit.rs");
+}
+
+const APPLE_SCHEMA: &[u8] = include_bytes!("../generated/apple.fdset");
```
gasmith, line 8: I don't think this fdset includes transitive dependencies from other proto files, right? I guess we leave that as an exercise for the reader? An alternative would be to use a crate like `prost_reflect` to build a [DescriptorPool](https://docs.rs/prost-reflect/latest/prost_reflect/struct.DescriptorPool.html), and then construct the schema for each message type somewhat more dynamically.

gasmith:
> This is a lot simpler than the approach I took, with the limitation that you have to manually serialize the message.
> 
> I wonder maybe I can simplify the example to be more like this, but with the impl Encode generation added. I think it's possible.

I think adding `impl Encode` would be nice, for supporting typed channels, but I'm not sure it's important to show users how to do that in codegen. If they really want to do it with codegen, they can figure that out. I think manually writing out an `impl Encode` in the example wouldn't be so bad.

rust/examples/prost/src/main.rs
bryfox: It seems to. To test, I created `stem.proto`, imported it and added a Stem field on my apple. And the resulting output from `mcap list schemas` included both `fruit.Stem` and `fruit.Apple`.

If there's another reason to use the descriptor pool, I'm happy to use it; otherwise I think I'd prefer to leave the additional crate out of the picture for simplicity. Thoughts?

bryfox:
👍 Thanks, I like the manual Encode better as an example. I also fixed the schema name.

bryfox:
We had feedback elsewhere that smaller individual examples are preferable, so I'm going to mark this ready for review, rather than suggesting we merge this with the other protobuf or 'logging approach' PRs.

rust/examples/prost/src/main.rs
gasmith: Fabulous. In that case, I fully retract my suggestion to use `prost_reflect`. What you have is the way to go.

rust/examples/prost/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "example_prost"
+edition = "2021"
+publish = false
+
+[lints]
+workspace = true
+
+[dependencies]
+foxglove = { path = "../../foxglove", features = [] }
```
gasmith: Nit: I think `features = []` is meaningless; perhaps you meant `default-features = false`?

bryfox: Yep, thank you

PR #483, opened 2025-06-12
Add C++ example showing logging custom generated protobuf

eloff:
This adds a C++ example using a custom protobuf type generated with google's C++ protobuf library.

It's painful, I think we could improve this story a lot by moving the boilerplate for the schema and channel creation and serialization into the Foxglove SDK.

The CMakeLists.txt needs some cleanup, it's probably not the minimal version.

Maybe it also needs to be dockerized? It's not trivial to get this working on Linux, nevermind another platform.

### DIFF
--- a/cpp/examples/custom-protobuf/CMakeLists.txt
+++ b/cpp/examples/custom-protobuf/CMakeLists.txt
@@ -1,0 +1,88 @@
+cmake_minimum_required(VERSION 3.14)
+project(quickstart LANGUAGES CXX)
+
+# Set C++ standard
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+# Find Protocol Buffers
+set(Protobuf_DEBUG ON)  # Enable debug output
+set(Protobuf_USE_STATIC_LIBS ON)  # Force static libraries
+set(Protobuf_VERSION 3.21.12)  # Explicitly set version to match system library
+find_package(Protobuf 3.21.12 REQUIRED)
+
+# Print debug information
+message(STATUS "Protobuf_VERSION: ${Protobuf_VERSION}")
+message(STATUS "Protobuf_PROTOC_EXECUTABLE: ${Protobuf_PROTOC_EXECUTABLE}")
+message(STATUS "Protobuf_INCLUDE_DIR: ${Protobuf_INCLUDE_DIR}")
+message(STATUS "Protobuf_LIBRARIES: ${Protobuf_LIBRARIES}")
+
+# Ensure we use the protoc found by FindProtobuf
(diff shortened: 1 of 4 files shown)


cpp/examples/custom-protobuf/CMakeLists.txt
```diff
@@ -0,0 +1,88 @@
...
+project(quickstart LANGUAGES CXX)
+
+# Set C++ standard
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+# Find Protocol Buffers
+set(Protobuf_DEBUG ON)  # Enable debug output
+set(Protobuf_USE_STATIC_LIBS ON)  # Force static libraries
+set(Protobuf_VERSION 3.21.12)  # Explicitly set version to match system library
```
bryfox, line 11: Maybe we should containerize this example in particular? I think we could probably install protobuf as part of the build, but keeping the configuration smaller also seems nice.

cpp/examples/custom-protobuf/CMakeLists.txt
```diff
@@ -0,0 +1,88 @@
...
+
+# Link Protocol Buffers to fruit_proto
+target_link_libraries(fruit_proto PUBLIC ${Protobuf_LIBRARIES})
+
+# Fetch Foxglove SDK
+include(FetchContent)
+FetchContent_Declare(
+    foxglove
+    # See available releases and builds here: https://github.com/foxglove/foxglove-sdk/releases
+    URL https://github.com/foxglove/foxglove-sdk/releases/download/sdk%2Fv0.9.0/foxglove-v0.9.0-cpp-x86_64-unknown-linux-gnu.zip
```
bryfox: If we don't want to containerize, can we reference the existing source, so users don't need to mess with this?

bryfox (commented):
The example itself looks good to me. The setup seems to assume a particular OS/version. It'd be great to make this more resilient — either with a container or removing reliance on system dependencies. Maybe @sofuture has thoughts here too.

bryfox:
> The setup seems to assume a particular OS/version

We could also document the assumptions/requirements here for now, and let users adjust as needed. We can always improve the example down the line. This is already a big step forward from no example...

cpp/examples/custom-protobuf/CMakeLists.txt
eloff: I worked on this, but I don't think that's a good idea now.

This doesn't need to be modified to run, and it can be copy-pasted in the user's CMakeLists.txt file, which is important.

If we use the adjacent code in the repo, we have to include this with comments explaining to the user to use this in their own make file, but then add lots of custom build stuff to pull in the C and CPP libraries and build them, which otherwise isn't needed.

bryfox: 👍 Yeah, that makes sense. I added bare-bones containerization in #494; I think that's a reasonable starting point.

bryfox (approved):

PR #494, opened 2025-06-16
Containerize the C++ protobuf example

bryfox:
This is based on #483.

It adds a Dockerfile and updates the Readme for running the C++ protobuf example. This example in particular takes some more effort to set up (notably, the protobuf compiler), so this is the easiest way to see it run. The example specifies running on linux/amd64 because it doesn't yet support multi-arch builds; I think that's fine for seeing the example work.

Tested on macOS (apple silicon).


eloff (approved):
I tried it out on Linux and it built and ran following the instructions in the readme. Looks great, thanks for doing this.

PR #500, opened 2025-06-18
Rust: support filtering logging sinks by topic

bryfox:
### Changelog
Rust: support filtering logging sinks by topic

### Docs
TBD.

### Description

This implements sink filtering, so that a user can (for example) log "heavy" topics to a different MCAP than the rest of their data, but still log all messages to a WS server in the same context. Python and C++ will follow, but released at the same time.

The example and tests (which make up most of the changeset here) illustrate the usage. I started with a version that also allowed you to change filtering on the sink after the fact, but decided to keep it simpler for now.

### DIFF
--- a/rust/examples/message-filtering/src/main.rs
+++ b/rust/examples/message-filtering/src/main.rs
@@ -1,0 +1,205 @@
+//! This example demonstrates how to use the Foxglove SDK to filter messages when logging to an MCAP
+//! file and/or a WebSocket server.
+//!
+//! Oftentimes, you may want to split "heavy" topics out into separate MCAP recordings, but still
+//! log everything for live visualization. Splitting on topic in this way can be useful for
+//! selectively retrieving data from bandwidth-constrained environments, such as with the Foxglove
+//! Agent.
+//!
+//! In this example, we log some point cloud data to one MCAP file, and some minimal metadata to
+//! another.
+use foxglove::schemas::{
+    packed_element_field::NumericType, PackedElementField, PointCloud, Pose, Quaternion, Vector3,
+};
+use foxglove::schemas::{FrameTransform, FrameTransforms};
+use foxglove::{Encode, FilterableChannel, LazyChannel, McapWriter, SinkChannelFilter};
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use std::time::{Duration, SystemTime};
+
+const SMALL_MCAP_FILE: &str = "example-topic-splitting-small.mcap";
+const LARGE_MCAP_FILE: &str = "example-topic-splitting-large.mcap";
+
+#[derive(Encode)]
+struct Message {
+    state: String,
+}
+
+static INFO_CHANNEL: LazyChannel<Message> = LazyChannel::new("/info");
+static POINT_CLOUD_CHANNEL: LazyChannel<PointCloud> = LazyChannel::new("/point_cloud");
+static POINT_CLOUD_TF_CHANNEL: LazyChannel<FrameTransforms> = LazyChannel::new("/point_cloud_tf");
+
+/// A filter which will drop all of our point_cloud messages
+struct SmallTopicFilter;
+impl SinkChannelFilter for SmallTopicFilter {
+    fn should_subscribe(&self, channel: &dyn FilterableChannel) -> bool {
+        !channel.topic().starts_with("/point_cloud")
+    }
+}
+
+/// A filter which keeps _only_ our point_cloud messages
+struct LargeTopicFilter;
+impl SinkChannelFilter for LargeTopicFilter {
+    fn should_subscribe(&self, channel: &dyn FilterableChannel) -> bool {
+        channel.topic().starts_with("/point_cloud")
+    }
+}
+
+/// We'll send all messages to the Foxglove app. We don't need a filter for this, since its the same
+/// as having no filter applied, but this demonstrates how to apply a filter to the WS server.
+struct LiveVizFilter;
+impl SinkChannelFilter for LiveVizFilter {
+    fn should_subscribe(&self, _channel: &dyn FilterableChannel) -> bool {
+        true
+    }
+}
+
+fn main() {
+    let env = env_logger::Env::default().default_filter_or("debug");
+    env_logger::init_from_env(env);
+
+    let done = Arc::new(AtomicBool::default());
+    ctrlc::set_handler({
+        let done = done.clone();
+        move || {
+            done.store(true, Ordering::Relaxed);
+        }
+    })
+    .expect("Failed to set SIGINT handler");
+
+    // We'll log to both an MCAP file, and to a running Foxglove app via a server.
+    let small_mcap = McapWriter::new()
+        .with_channel_filter(Arc::new(SmallTopicFilter))
+        .create_new_buffered_file(SMALL_MCAP_FILE)
+        .expect("Failed to create mcap writer");
+
+    let large_mcap = McapWriter::new()
+        .with_channel_filter(Arc::new(LargeTopicFilter))
+        .create_new_buffered_file(LARGE_MCAP_FILE)
+        .expect("Failed to create mcap writer");
+
+    foxglove::WebSocketServer::new()
+        .channel_filter(Arc::new(LiveVizFilter))
+        .start_blocking()
+        .expect("Server failed to start");
+
+    let start = SystemTime::now();
+    let cloud_tf = FrameTransforms {
+        transforms: vec![FrameTransform {
+            parent_frame_id: "world".to_string(),
+            child_frame_id: "points".to_string(),
+            translation: Some(Vector3 {
+                x: -10.0,
+                y: -10.0,
+                z: 0.0,
+            }),
+            ..Default::default()
+        }],
+    };
+
+    while !done.load(Ordering::Relaxed) {
+        let elapsed = SystemTime::now()
+            .duration_since(start)
+            .expect("Time went backwards");
+
+        let state = get_state(elapsed);
+        INFO_CHANNEL.log(&Message { state });
+
+        let point_cloud = make_point_cloud(elapsed);
+        POINT_CLOUD_CHANNEL.log(&point_cloud);
+        POINT_CLOUD_TF_CHANNEL.log(&cloud_tf);
+
+        std::thread::sleep(Duration::from_millis(33));
+    }
+
+    small_mcap.close().expect("Failed to close mcap");
+    large_mcap.close().expect("Failed to close mcap");
+}
+
+fn get_state(elapsed: Duration) -> String {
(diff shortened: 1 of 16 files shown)


rust/foxglove/src/mcap_writer.rs
```diff
@@ -28,6 +28,13 @@ use mcap_sink::McapSink;
 pub struct McapWriter {
     options: McapWriteOptions,
     context: Arc<Context>,
+    channel_filter: Option<Arc<dyn SinkChannelFilter>>,
+}
+
+impl Debug for dyn SinkChannelFilter {
```
bryfox: This was needed for MCAPSink, which implements Debug. I'm not sure how useful that is, and this seems mildly annoying — though maybe there's an easier way?

rust/foxglove/src/mcap_writer/mcap_sink.rs
```diff
@@ -13,20 +13,26 @@ type McapChannelId = u16;
...
     writer: mcap::Writer<W>,
     // ChannelId -> mcap file channel id.
+    //
+    // If the value is `None`, then we do not log to this channel; for example, if the channel is
+    // filtered out by the `SinkChannelFilter`.
+    //
     // Note that the underlying writer may re-use channel_ids based on the metadata of the channel,
     // so multiple `ChannelIds` may map to the same `McapChannelId`.
-    channel_map: HashMap<ChannelId, McapChannelId>,
+    channel_map: HashMap<ChannelId, Option<McapChannelId>>,
```
bryfox, line 22: I changed the semantics as above; if we're filtering the channel, then we put a `None` in the slot to indicate there is no channel to log to. I think it might be clearer to have a separate channel_filter map — we want to cache this information in some way — but went with this approach for now. Open to suggestions.

rust/foxglove/src/sink_channel_filter.rs
```diff
@@ -0,0 +1,36 @@
+use std::collections::BTreeMap;
+
+use crate::{ChannelId, RawChannel};
+
+/// Information about a channel, which is passed to a [`SinkChannelFilter`].
+pub trait FilterableChannel {
```
bryfox: I don't love this name, but I need some sort of view of a channel here. The websocket module already has a `ChannelView`, but that doesn't include metadata. I think it shouldn't; and I think metadata is useful here, so one doesn't need to hardcode topic names or reference IDs.

rust/examples/message-filtering/src/main.rs
```diff
@@ -0,0 +1,205 @@
...
+        let done = done.clone();
+        move || {
+            done.store(true, Ordering::Relaxed);
+        }
+    })
+    .expect("Failed to set SIGINT handler");
+
+    // We'll log to both an MCAP file, and to a running Foxglove app via a server.
+    let small_mcap = McapWriter::new()
+        .with_channel_filter(Arc::new(SmallTopicFilter))
```
gasmith: We can offer the full "pass an Arc<dyn Trait>" method for flexibility, but I think it'd also be worthwhile to offer a version that takes a closure for inline brevity. Compare:

```rust
let small_mcap = McapWriter::new()
    .with_channel_filter_fn(|c| !c.topic().starts_with("/point_cloud"))
```

rust/foxglove/src/mcap_writer.rs
gasmith: Probably best to manually implement `Debug` for `McapSink`, and just omit the filter from the output, finishing with `finish_non_exhaustive()`, as we do for `RawChannel` and others.

rust/foxglove/src/mcap_writer/mcap_sink.rs
gasmith: Looks reasonable to me.

rust/foxglove/src/sink_channel_filter.rs
gasmith: I'm not sure I see the value in using a trait here, because I don't think we're actually abstracting over distinct concrete types. If we do have a trait, I think it should be a common one. Restricting `ChannelView` to exclude `metadata` seems overly prescriptive, in my opinion.

But as an alternative to the trait, consider separating out a `channel::Descriptor` struct from `RawChannel`. We could just pass a `&Descriptor`/`&Arc<Descriptor>` in most places where we're currently passing the `&RawChannel` or `ChannelView`. Bonuses include: 
 - Passing a `&Descriptor` means that we don't need a vtable lookups for inspecting the channel.
 - A `RawChannel` with a `ChannelId` and an `Arc<Descriptor>` fits neatly within a cache line, and the `log()` implementations don't need to access the descriptor in the common case.

bryfox: I like it, thanks. I haven't replaced all usage yet, but the trait is gone. LMK if that's not what you had in mind.

rust/examples/message-filtering/src/main.rs
bryfox: Yes, I like it and made the change. I'm actually tempted to remove the other interface, at least for now.

rust/foxglove/src/channel/raw_channel.rs
```diff
@@ -34,12 +70,9 @@ static WARN_THROTTLER_INTERVAL: Duration = Duration::from_secs(10);
 ///
 /// You should choose a unique topic name per channel for compatibility with the Foxglove app.
 pub struct RawChannel {
-    id: ChannelId,
-    context: Weak<Context>,
-    topic: String,
-    message_encoding: String,
+    descriptor: ChannelDescriptor,
     schema: Option<Schema>,
```
gasmith: Assuming we eventually update the `Sink` trait methods to take a `&ChannelDescriptor`, we should also put the schema in there. While we're at it, I think it's worth making it an `Arc<ChannelDescriptor>`, so that it can be cheaply cloned (e.g., by the websocket sink), and so that the `RawChannel` struct fits within a cacheline. If we do that, that we should also keep a copy of the `ChannelId` on the `RawChannel` itself, so that we (typically) don't need to access the descriptor on the logging path.

gasmith (approved):
LG, with a minor layout nit.

rust/examples/message-filtering/src/main.rs
gasmith: I'm fine with removing the other interface, sure. If we make the trait more complex in the future, then perhaps it'll come back.

bryfox:
OK, I've queued up the Python (#504) & C++ (#519) PRs on top of this.

This PR added an `Arc` around the descriptor in the RawChannel. #504 also makes the callback take an `Arc`; I had trouble implementing it without.

Note I have not put `schema` in the channel descriptor, since schema data can be rather large and for those PRs above, I am allocating new descriptors to call these functions.

gasmith:
> This PR added an `Arc` around the descriptor in the RawChannel. #504 also makes the callback take an `Arc`; I had trouble implementing it without.

Maybe we can push that arc down a little further (e.g., `ChannelDescriptor(Arc<Inner>)`), since it's something that will be broadly cloned.
 
> Note I have not put `schema` in the channel descriptor, since schema data can be rather large and for those PRs above, I am allocating new descriptors to call these functions.

Not sure I understand - why wouldn't this just be a cheap clone of the Arc?

rust/examples/message-filtering/src/main.rs
```diff
@@ -0,0 +1,183 @@
+//! This example demonstrates how to use the Foxglove SDK to filter messages when logging to an MCAP
+//! file and/or a WebSocket server.
+//!
+//! Oftentimes, you may want to split "heavy" topics out into separate MCAP recordings, but still
+//! log everything for live visualization. Splitting on topic in this way can be useful for
+//! selectively retrieving data from bandwidth-constrained environments, such as with the Foxglove
+//! Agent.
```
eloff, line 7: Maybe add that if you just want to partition topics into different groups of sinks (eg different mcap files) then multiple contexts is the way to go.

rust/foxglove/src/channel/raw_channel.rs
```diff
@@ -54,26 +87,32 @@ impl RawChannel {
         metadata: BTreeMap<String, String>,
     ) -> Arc<Self> {
         Arc::new(Self {
-            id: ChannelId::next(),
+            descriptor: Arc::new(ChannelDescriptor {
```
eloff: Do we really need an arc inside an arc? I feel like we could avoid the extra heap allocation and indirection.

rust/foxglove/src/mcap_writer.rs
```diff
@@ -24,10 +25,20 @@ use mcap_sink::McapSink;
 /// Logged messages are buffered in a [`BufWriter`]. When the writer is dropped, the buffered
 /// messages are flushed to the writer and the writer is closed.
 #[must_use]
-#[derive(Debug, Clone)]
+#[derive(Clone)]
 pub struct McapWriter {
     options: McapWriteOptions,
     context: Arc<Context>,
+    channel_filter: Option<Arc<dyn SinkChannelFilter>>,
```
eloff, line 32: Could this be a Box instead of an Arc? Do we ever clone it?

rust/foxglove/src/mcap_writer/mcap_sink.rs
eloff: I guess a separate hashset of filtered channel ids is more efficient, but this works.

bryfox:
I made ChannelDescriptor wrap an Arc<Inner> as suggested; thanks.

I don't think we have a zero-cost way to get the schema in python today (#504). Not that it's not possible, but I we're still doing a conversion there today. That said, I don't think that optimization is closed to us, and it's probably not a big deal at all right now.

rust/foxglove/src/channel/channel_descriptor.rs
```diff
@@ -0,0 +1,64 @@
...
+
+impl ChannelDescriptor {
+    pub(crate) fn new(
+        id: ChannelId,
+        topic: String,
+        message_encoding: String,
+        metadata: BTreeMap<String, String>,
+        schema: Option<Schema>,
+    ) -> Arc<Self> {
+        Arc::new(Self(Arc::new(Inner {
```
gasmith: No need to double-wrap with the arcs here. Just `#[derive(Clone)]` on `ChannelDescriptor`, and document that it's a cheap refcounted clone.

eloff (approved):
LGTM, a couple nits but nothing that needs to hold it back

rust/foxglove/src/channel/raw_channel.rs
bryfox: Missed this in yesterday's refactor; thanks

rust/foxglove/src/mcap_writer.rs
bryfox: McapWriter derives Clone (and does on main)

PR #504, opened 2025-06-20
Python: support filtering logging sinks by topic

bryfox:
### Changelog
Python: support filtering logging sinks by topic

### Description

This builds on #500 to bring channel filters to Python.

This constructs a new channel descriptor when the filter is called, but this isn't called frequently since we cache the results of this for the MCAP sink.

The example is generally the same as the Rust one; I removed the point cloud part of the live_visualization example in favor of this, since it wasn't demonstrating much on its own there.

### DIFF
--- a/python/foxglove-sdk-examples/message-filtering/main.py
+++ b/python/foxglove-sdk-examples/message-filtering/main.py
@@ -4,8 +4,8 @@ import time
 from math import cos, sin
 
 import foxglove
-from foxglove import ChannelDescriptor
 import foxglove.channels
+from foxglove import ChannelDescriptor
 from foxglove.schemas import (
     FrameTransform,
     FrameTransforms,


python/foxglove-sdk/src/sink_channel_filter.rs
```diff
@@ -0,0 +1,88 @@
...
+    #[getter]
+    fn message_encoding(&self) -> &str {
+        self.0.message_encoding()
+    }
+
+    /// Returns the metadata for this channel.
+    #[getter]
+    fn metadata<'py>(&self, py: Python<'py>) -> PyResult<Bound<'py, PyDict>> {
+        let metadata = self.0.metadata().into_py_dict(py).unwrap_or_else(|err| {
+            tracing::error!("Failed to constrcut channel metadata: {}", err.to_string());
```
gasmith:
```suggestion
            tracing::error!("Failed to construct channel metadata: {}", err.to_string());
```

python/foxglove-sdk/src/sink_channel_filter.rs
```diff
@@ -0,0 +1,88 @@
...
+            PyDict::new(py)
+        });
+        Ok(metadata)
+    }
+
+    /// Returns the schema for this channel.
+    #[getter]
+    fn schema(&self) -> Option<PySchema> {
+        tracing::info!("schema?: {:?}", self.0.schema().is_some());
```
gasmith:
```suggestion
```

python/foxglove-sdk/src/websocket.rs
```diff
@@ -366,7 +366,7 @@ impl foxglove::websocket::service::Handler for ServiceHandler {
 /// To connect to this server: open Foxglove, choose "Open a new connection", and select Foxglove
 /// WebSocket. The default connection string matches the defaults used by the SDK.
 #[pyfunction]
-#[pyo3(signature = (*, name = None, host="127.0.0.1", port=8765, capabilities=None, server_listener=None, supported_encodings=None, services=None, asset_handler=None, context=None))]
+#[pyo3(signature = (*, name = None, host="127.0.0.1", port=8765, capabilities=None, server_listener=None, supported_encodings=None, services=None, asset_handler=None, context=None, channel_filter=None))]
```
gasmith: Do we need to update the docstring here? Maybe not, because we have the python wrapper function?

python/foxglove-sdk/src/sink_channel_filter.rs
```diff
@@ -0,0 +1,88 @@
...
+            let result = handler
+                .bind(py)
+                .call((descriptor,), None)
+                .and_then(|f| f.extract::<bool>());
+
+            match result {
+                Ok(result) => result,
+                Err(err) => {
+                    tracing::error!("Error in SinkChannelFilter: {}", err.to_string());
+                    false
```
gasmith, line 81: Should we document this behavior?

gasmith (approved):

python/foxglove-sdk/src/sink_channel_filter.rs
bryfox: Yes, thanks — done.

python/foxglove-sdk/src/websocket.rs
bryfox: We don't, for that reason. I removed most of the docs here to avoid confusion, and at some point we'll do some further cleanup on these wrappers.

python/foxglove-sdk-examples/message-filtering/main.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def drop_large_topics(channel: ChannelDescriptor) -> bool:
+    return not keep_large_topics(channel)
+
+
+# We'll send all messages to the Foxglove app. We don't need a filter for this, since its the same
+# as having no filter applied, but this demonstrates how to apply a filter to the server.
+def live_viz_filter(ch: ChannelDescriptor) -> bool:
+    print(ch.topic)
```
eloff: Is it supposed to have all these prints, or were they for debugging?

python/foxglove-sdk/src/lib.rs
```diff
@@ -211,6 +217,12 @@ fn open_mcap(
     } else {
         McapWriter::with_options(options)
     };
+    let handle = if let Some(channel_filter) = channel_filter {
+        handle.channel_filter(Arc::new(PySinkChannelFilter(Arc::new(channel_filter))))
```
eloff: Why the double arc?

eloff (commented):
LGTM. It seems like there should be a way to avoid the nested arc, but it's a minor thing.

python/foxglove-sdk-examples/message-filtering/main.py
bryfox: 🔪

python/foxglove-sdk/src/lib.rs
bryfox: I removed the inner Arc in dfdafbb. It looks like the recommended way to clone the handler in this case is to call `clone_ref` once we have the `py` token for the GIL context.

PR #505, opened 2025-06-23
Python: expose metadata in channel constructors

bryfox:
### Changelog
Python: expose metadata in channel constructors

### Description

This allows users to add metadata to channels. The core was already there; this just adds it to the public interfaces.


eloff (approved):
LGTM

PR #509, opened 2025-06-24
Introduce strict build and configure CI to use it

ericmlujan:
### Description

</td><td>

</td></tr></table>


cpp/foxglove/tests/test_server.cpp
```diff
@@ -786,7 +786,7 @@ TEST_CASE("Service callbacks") {
   auto server = startServer(context, foxglove::WebSocketServerCapabilities::Services, {}, {"json"});
 
   // Register an echo service.
-  foxglove::ServiceSchema echo_schema{"echo schema"};
+  foxglove::ServiceSchema echo_schema{"echo schema", std::nullopt, std::nullopt};
```
ericmlujan: Uninitialized struct members

cpp/foxglove/tests/test_server.cpp
```diff
@@ -477,7 +477,7 @@ TEST_CASE("Parameter callbacks") {
     std::vector<foxglove::Parameter> owned_params;
     owned_params.reserve(params.size());
     for (const auto& param : params) {
-      owned_params.emplace_back(std::move(param.clone()));
```
ericmlujan, line 480: Copy-elision

cpp/foxglove/tests/test_parameter.cpp
```diff
@@ -225,7 +225,7 @@ TEST_CASE("Parameter error cases") {
     foxglove::Parameter param(
       "test_param",
       foxglove::ParameterType::ByteArray,
-      std::move(foxglove::ParameterValue("invalid-base64!"))
```
ericmlujan, line 228: Copy-elision won't happen if std::move() called on a temporary

cpp/foxglove/tests/test_mcap.cpp
```diff
@@ -19,6 +19,16 @@ using Catch::Matchers::ContainsSubstring;
 using Catch::Matchers::Equals;
 using foxglove_tests::FileCleanup;
 
+namespace {
```
ericmlujan: A bunch of changes in this file relating to not initializing fields in `McapWriterOptions`. Made a helper to keep the testing code concise.

cpp/foxglove/src/server.cpp
```diff
@@ -252,9 +252,9 @@ WebSocketServer::WebSocketServer(
   foxglove_websocket_server* server, std::unique_ptr<WebSocketServerCallbacks> callbacks,
   std::unique_ptr<FetchAssetHandler> fetch_asset
 )
-    : impl_(server, foxglove_server_stop)
```
ericmlujan, line 255: Squelching `-Wreorder`

cpp/foxglove/include/foxglove/server/parameter.hpp
```diff
@@ -258,7 +258,6 @@ class ParameterView final {
       return false;
     }
     try {
-      auto value = this->value();
```
ericmlujan, line 261: Was unused in fcn body

cpp/foxglove/include/foxglove/mcap.hpp
```diff
@@ -13,7 +13,7 @@ struct foxglove_mcap_writer;
 /// The foxglove namespace.
 namespace foxglove {
 
-struct Context;
+class Context;
```
ericmlujan, line 16: Forward declaration didn't match original ([referenced code](https://github.com/foxglove/foxglove-sdk/blob/53d501a464e3077474709863f58d1c7cb8757eb2/cpp/foxglove/include/foxglove/context.hpp#L27) is declared as a class, not struct)

cpp/examples/ws-services/src/main.cpp
```diff
@@ -82,7 +80,7 @@ std::vector<std::byte> makeBytes(std::string_view sv) {
  * A service that always responds with an empty json object.
  */
 bool registerEmptyService(foxglove::WebSocketServer& server) {
-  foxglove::ServiceSchema empty_schema{"/std_srvs/Empty"};
+  foxglove::ServiceSchema empty_schema{"/std_srvs/Empty", std::nullopt, std::nullopt};
```
ericmlujan: Explicitly initializing optional struct members

cpp/examples/ws-services/src/main.cpp
```diff
@@ -63,10 +63,8 @@ int main(int argc, const char** argv) {
     done = true;
   };
 
-  uint32_t i = 0;
```
ericmlujan, line 66: Loop counter wasn't being read from anywhere

cpp/CMakeLists.txt
```diff
@@ -79,6 +79,16 @@ if(DEFINED SANITIZE)
   set(SANITIZER_CARGO_FLAGS -Zbuild-std)
 endif()
 
+if (STRICT)
+  # TODO: Review the warnings for -Wunused-parameter and see if they can be resolved without affecting public-facing APIs
```
ericmlujan: If reviewers have clear opinions on this, I'm happy to chat/resolve them, I'm mostly just lacking context on whether the unused parameters are important for maintaining API signatures. Most flagged functions were in schemas.cpp and almost all were complaining about an unused `Arena` parameter. You can see [`geoJSONtoC()`](https://github.com/foxglove/foxglove-sdk/blob/53d501a464e3077474709863f58d1c7cb8757eb2/cpp/foxglove/src/schemas.cpp#L1195-L1197) as a canonical example of this. If this is intentional/required, we can discuss the best way to suppress it, I'll file a Linear, and enable this warning down the road..

cpp/CMakeLists.txt
```diff
@@ -79,6 +79,16 @@ if(DEFINED SANITIZE)
...
+  # TODO: -Wfloat-equal is flagging lots of tests in test_arena.cpp. Need to check with the authors to see if they really want
```
ericmlujan: Perhaps @eloff would be the right person to weigh in here

cpp/examples/param-server/src/main.cpp
```diff
@@ -141,9 +141,9 @@ int main(int argc, const char* argv[]) {
     auto elapsed = std::chrono::duration<double>(now - start_time).count();
     auto param = foxglove::Parameter("elapsed", elapsed);
     param_store.insert_or_assign("elapsed", param.clone());
-    std::vector<foxglove::Parameter> params;
```
ericmlujan, line 144: Variable shadow

cpp/foxglove/include/foxglove/server/parameter.hpp
```diff
@@ -413,9 +412,9 @@ class ParameterView final {
     const auto& dict = value->get<ParameterValueView::Dict>();
     std::map<std::string, T> result;
     for (const auto& elem : dict) {
-      std::string key(elem.first);
```
ericmlujan, line 416: Variable shadow

cpp/foxglove/tests/test_mcap.cpp
```diff
@@ -118,8 +118,10 @@ TEST_CASE("different contexts") {
   auto context2 = foxglove::Context::create();
 
   // Create writer on context1
-  foxglove::McapWriterOptions options{context1};
```
ericmlujan, line 121: Initializing `McapWriterOptions` and passing only the context in the initializer caused warnings because not every struct member is being initialized. I experimented with using designated initializers, but that required C++20 support and we're building with C++17.

cpp/foxglove/tests/test_server.cpp
```diff
@@ -748,7 +748,7 @@ std::vector<std::byte> makeServiceRequest(
   buffer.emplace_back(static_cast<std::byte>(2));  // Service call request opcode
   writeUint32LE(buffer, service_id);
   writeUint32LE(buffer, call_id);
-  writeUint32LE(buffer, encoding.size());
+  writeUint32LE(buffer, static_cast<uint32_t>(encoding.size()));
```
ericmlujan, line 755: Implicit precision change from `size_t` to `uint32_t`

cpp/foxglove/tests/test_server.cpp
```diff
@@ -831,11 +831,11 @@ TEST_CASE("Service callbacks") {
   REQUIRE(parsed["op"] == "advertiseServices");
   REQUIRE(parsed.contains("services"));
   std::map<std::string, uint32_t> service_ids;
-  for (const auto& service : parsed["services"]) {
```
ericmlujan, line 834: Variable shadow

cpp/CMakeLists.txt
jtbandes: Maybe you can slap a `[[maybe_unused]]` on the unused variable to suppress it?

cpp/CMakeLists.txt
```diff
@@ -79,6 +79,18 @@ if(DEFINED SANITIZE)
   set(SANITIZER_CARGO_FLAGS -Zbuild-std)
 endif()
 
+if (STRICT)
```
jtbandes, line 83: when would we **not** want this enabled? If I'm remembering right that this cmakelists is purely for internal use (i.e. we're not inflicting these strict flags on all the downstream users of the c++ library) then maybe there's no harm in just having it be always on?

cpp/examples/ws-services/src/main.cpp
jtbandes: this one is a little unfortunate from an ergonomics perspective...but whatever appeases the compiler I guess 🤷   which warning was this?

cpp/foxglove/include/foxglove/server/parameter.hpp
jtbandes: some other possible suggestions here, take your pick (or leave as is): `result.emplace(elem.first, elem.second.get<T>())` or `const auto& [dictKey, dictValue] : ...`

jtbandes (approved):
The uninitialized struct members seem like the most "annoying" of the changes - but I'm not against making the changes, just curious if there are any alternatives that other people use in the wild. I haven't been deep in the C++ mines for years at this point.

I think it could be fine to keep this always on.

cpp/CMakeLists.txt
ericmlujan: @jtbandes I'll do that for now, it feels at least more localized. I had debated doing that earlier, but wanted to have a reviewer chime in on it before doing it

cpp/examples/ws-services/src/main.cpp
ericmlujan: It was `-Wmissing-field-initializers` since the `request` and `response` fields in `ServiceSchema` aren't default initialized

cpp/foxglove/include/foxglove/server/parameter.hpp
ericmlujan: I like the structured binding approach better, will update!

cpp/CMakeLists.txt
ericmlujan: Correct, I think the only way a downstream user would encounter this is if they were choosing to build the SDK themselves from source, and if we're warning-free, they shouldn't be blocked. I'll go through and update this to be default-on, though I'm generally a fan of creating _some_ kind of switch to easily disable it for development purposes/niche cases.

cpp/examples/ws-services/src/main.cpp
eloff: Maybe it should be default initialized to std::nullopt?

cpp/CMakeLists.txt
eloff: These are not incorrect usages of float equals, no math was used here, but the compiler doesn't know that. Floats are not necessary to the tests, any type could be used, so I think just change the tests to use integers and we're fine here.

cpp/CMakeLists.txt
eloff: This was to make the code generation easier (fewer branches). I don't think it's worth complicating the code generation, but maybe we can use Jacob's recommended fix, or disable the warning for just this generated file?

cpp/CMakeLists.txt
eloff: Yeah, I think on by default makes sense

eloff (approved):
I added some recommendations / background. I'll leave it up to you how you want to proceed. The rest of the PR looks good.

cpp/CMakeLists.txt
ericmlujan: 👍 I went through and added unused parameter flags to suppress the warnings in ab213ea

cpp/CMakeLists.txt
ericmlujan: 👍 will do, thanks!

PR #515, opened 2025-07-02
Thread through the client ID to the subscribe/unsubscribe callbacks for WebSocket server

ericmlujan:
### Changelog
- Create a new function, `Channel::log_to_sink`, allowing users to optionally specify an individual Sink to target a logged message to. If left `None`, the message is logged to all sinks associated with the channel's context.
- Update C++ bindings for  `Channel::log` to take a sink ID as an optional argument.
- Update C++ `WebSocketServerCallbacks::onSubscribe` and `WebSocketServerCallbacks::onUnsubscribe` function signatures to take `foxglove::ClientMetadata` as a second parameter, which surfaces information about which client and associated sinks are requesting to subscribe or unsubscribe to a channel. 

### Docs
Doxygen comments are updated.

### Description
These changes are motivated by work to port `ros-foxglove-bridge` to use the SDK. The main problem being solved is ensuring that messages are delivered to late-joining clients. The ability to target `log()` calls to individual sinks gives WebSocketServer users more control over how messages should be delivered in this situation (i.e. by implementing logic that can republish a message to only a newly joining client without sending a duplicate to others). 

Surfacing information about which sink a given client is associated with allows sinks to be used as the semantic for targeting message logging, rather than coupling it to WebSocket client IDs.

Unit tests pass. Added additional tests to verify that messages are only logged to the appropriate sinks if a target sink ID is passed.

All calls to `Channel::log` deliver messages to every sink associated with that channels's context.

</td><td>

Calls to `Channel::log` can optionally be targeted to individual sinks.

</td></tr></table>

Fixes: FG-12285


c/include/foxglove-c/foxglove-c.h
```diff
@@ -301,6 +301,13 @@ typedef struct foxglove_string {
   size_t len;
 } foxglove_string;
 
+typedef uint64_t FoxgloveSinkId;
+
+typedef struct foxglove_client_metadata {
+  uint32_t id;
+  const FoxgloveSinkId *sink_id;
```
eloff: Why is this a pointer, can we use zero as the null value?

c/include/foxglove-c/foxglove-c.h
```diff
@@ -439,8 +446,12 @@ typedef struct foxglove_server_callbacks {
    * A user-defined value that will be passed to callback functions
    */
   const void *context;
-  void (*on_subscribe)(const void *context, uint64_t channel_id);
-  void (*on_unsubscribe)(const void *context, uint64_t channel_id);
+  void (*on_subscribe)(const void *context,
+                       uint64_t channel_id,
+                       const struct foxglove_client_metadata *client);
```
eloff: The struct is small, can we pass it by value? Then we don't have the risk that someone stores the pointer.

c/include/foxglove-c/foxglove-c.h
```diff
@@ -2132,7 +2143,8 @@ void foxglove_channel_metadata_iter_free(struct foxglove_channel_metadata_iterat
 foxglove_error foxglove_channel_log(const struct foxglove_channel *channel,
                                     const uint8_t *data,
                                     size_t data_len,
-                                    const uint64_t *log_time);
+                                    const uint64_t *log_time,
+                                    const FoxgloveSinkId *sink_id);
```
eloff: I prefer we use 0 as the null value, 0 is not a valid sinkid

c/src/lib.rs
```diff
@@ -293,13 +293,66 @@ pub struct FoxgloveClientChannel {
     pub schema_len: usize,
 }
 
+// REVIEW: This struct only holds the subset of the Client struct that is being used for accounting
+// in the ros_foxglove_bridge. If there are other fields that may be useful to expose, I am
+// happy to implement them here.
+#[repr(C)]
+pub struct FoxgloveClientMetadata {
+    pub id: u32,
+    pub sink_id: *const FoxgloveSinkId, // NULL means no sink_id
```
eloff: Ditto

c/src/lib.rs
```diff
@@ -293,13 +293,66 @@ pub struct FoxgloveClientChannel {
...
+    fn from(client: foxglove::websocket::Client) -> Self {
+        // Convert the Rust SinkId to the C SinkId type. If it doesn't have a value,
+        // we use a null pointer.
+        //
+        // REVIEW: Is there a more idiomatic way to do this when binding Rust code?
+        // I'm looking for a way to bind an optional value to pass into a C-bound struct.
+        // I'm currently using a pointer for this since that's what I'm used to,
+        // but could also be convinced creating a new struct that replicates Optional would be better.
+        // Futzing with creating a Box, casting to a raw pointer, then dropping feels like overkill,
+        // but it's the best thing I could think up that isn't either a memory leak or a dangling reference.
```
eloff: I think in general the pointer is good for the C API, although the rust struct could maybe be Optional<Box<...>> if cbindgen converts that to a pointer. But in this case we can just use 0 as the null value and skip all of that complexity.

c/src/lib.rs
```diff
@@ -1288,6 +1343,7 @@ pub unsafe extern "C" fn foxglove_channel_log(
     data: *const u8,
     data_len: usize,
     log_time: Option<&u64>,
+    sink_id: Option<&FoxgloveSinkId>,
```
eloff: By value

cpp/foxglove/include/foxglove/channel.hpp
```diff
@@ -65,8 +65,11 @@ class RawChannel final {
   /// @param data The message data.
   /// @param data_len The length of the message data, in bytes.
   /// @param log_time The timestamp of the message. If omitted, the current time is used.
+  /// @param sink_id The sink ID associated with the message. Can be used to target logging messages
+  /// to a specific client or mcap file.
   FoxgloveError log(
-    const std::byte* data, size_t data_len, std::optional<uint64_t> log_time = std::nullopt
+    const std::byte* data, size_t data_len, std::optional<uint64_t> log_time = std::nullopt,
+    std::optional<uint64_t> sink_id = std::nullopt
```
eloff, line 72: Yes, optional would be the best way to represent this in C++

rust/foxglove/src/sink.rs
```diff
@@ -10,6 +10,11 @@ use crate::{ChannelId, FoxgloveError, RawChannel};
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub struct SinkId(u64);
```
eloff: If we change this to NonZeroU64 it will be 64 bits when wrapped with Optional. It also makes it a little clearer that zero is an invalid value.

Cbindgen will be able to covert a u64 to an optional<sinkid>. So we use u64 on the C side and optional on the rust side.

c/include/foxglove-c/foxglove-c.h
ericmlujan: I wasn't sure that it was safe to assume 0 was invalid/null. That makes life substantially simpler without all the pointer futzing. I'll update this and breathe a sign of relief :)

c/include/foxglove-c/foxglove-c.h
ericmlujan: Yep! Originally I had put more fields in here before thinking twice and stripping them out. Will remove.

rust/foxglove/src/sink.rs
ericmlujan: Perfect, that dramatically simplifies things :)

ericmlujan:
@eloff Thanks so much for the review and feedback. Took a pass at addressing your comments; let me know if I hit the mark

c/src/lib.rs
```diff
@@ -293,13 +293,34 @@ pub struct FoxgloveClientChannel {
     pub schema_len: usize,
 }
 
+// REVIEW: This struct only holds the subset of the Client struct that is being used for accounting
+// in the ros_foxglove_bridge. If there are other fields that may be useful to expose, I am
+// happy to implement them here.
```
eloff: That's good for now, we'll wait until we have a good reason to add something else

eloff (approved):
Excellent work! This looks perfect to me

c/src/lib.rs
ericmlujan: Awesome, will remove this comment and :shipit:

PR #516, opened 2025-07-02
Fix GCC build

### LINKED ISSUE #514: Building C++ SDK with GCC fails due to template specialization
**Description**

Building C++ SDK with GCC fails due to template specialization.

- Version: 0.9.0
- Platform: Linux x86_64 (Ubuntu 22.04), gcc 11.4.0 (also tested with gcc 12.3.0)

**Steps To Reproduce**

```
cmake -DCMAKE_C_COMPILER=/usr/bin/gcc -DCMAKE_CXX_COMPILER=/usr/bin/g++ -S ./cpp -B ./build -G Ninja
ninja -C build
```

This fails with build errors due to template specialization in paramter.cpp

Whereas configuring with a clang toolchain succeeds.

```
cmake -DCMAKE_C_COMPILER=/usr/bin/clang -DCMAKE_CXX_COMPILER=/usr/bin/clang++ -S ./cpp -B ./build -G Ninja
```

jtbandes:
### Changelog
C++: fixed building with GCC

### Docs

None

### Description

Fixes FG-12176
Fixes #514 

It is 2025, and GCC does not support a C++17 feature 🙃 

It seems like maybe we switched to clang for all platforms in #383 because of this issue?

### DIFF
--- a/cpp/Makefile
+++ b/cpp/Makefile
@@ -30,8 +30,6 @@ clean:
 .PHONY: build
 build:
 	cmake \
-		-DCMAKE_C_COMPILER=clang \
-		-DCMAKE_CXX_COMPILER=clang++ \
 		-DCMAKE_BUILD_TYPE=RelWithDebInfo \
 		-DCMAKE_EXPORT_COMPILE_COMMANDS=true \
 		$(if $(SANITIZE),-DSANITIZE=$(SANITIZE) -DRust_TOOLCHAIN=nightly) \
--- a/cpp/Makefile
+++ b/cpp/Makefile
@@ -50,7 +50,7 @@ clean-docs:
 	rm -rf ./foxglove/docs/generated
 	rm -rf $(BUILD_DIR)/docs
 
-$(BUILD_DIR)/docs/html:
+$(BUILD_DIR)/docs/html: clean-docs
 	cmake -B $(BUILD_DIR)
 	cmake --build $(BUILD_DIR) --target doxygen
 	cmake --build $(BUILD_DIR) --target html-docs
--- a/cpp/foxglove/include/foxglove/server/parameter.hpp
+++ b/cpp/foxglove/include/foxglove/server/parameter.hpp
@@ -90,6 +90,21 @@ private:
       : impl_(ptr) {}
 };
 
+// Template specializations for ParameterValueView
+//
+// Documented manually as tparams because specializations are merged:
+// https://github.com/doxygen/doxygen/issues/9468
+
+/// @fn ParameterValueView::is<std::string>()
+/// @tparam std::string Checks whether the parameter value is a string.
+
+/// @fn ParameterValueView::get<ParameterValueView>()
+/// @tparam ParameterValueView Extracts the parameter as itself.
+
+/// @fn ParameterValueView::get<std::string>()
+/// @tparam std::string Extracts the parameter as an owned string.
+
+/// @cond ignore-template-specializations
 /// @brief Checks whether the parameter value is a string.
 template<>
 [[nodiscard]] inline bool ParameterValueView::is<std::string>() const noexcept {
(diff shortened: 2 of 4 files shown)


jtbandes:
Seems to work but I'll need some help from someone who knows more about doxygen

amacneil:
Can we add to CI?

cpp/Makefile
```diff
@@ -30,8 +30,6 @@ clean:
 .PHONY: build
 build:
 	cmake \
-		-DCMAKE_C_COMPILER=clang \
```
ericmlujan, line 33: Should we update the env vars in our build containers to set `CXX` and `CC` to clang, so that our official builds still use it over gcc? I checked that in our docker images, `c++` and `cc` (which CMake invokes by default if the user doesn't specify anything) are aliases for gcc.

ericmlujan: Separately, agree we should have a GCC test for CI, as @amacneil suggested

cpp/foxglove/include/foxglove/server/parameter.hpp
```diff
@@ -300,18 +257,8 @@ class ParameterView final {
...
-  /// @brief Checks whether the paramter value is an dictionary of generic
-  /// values.
-  template<>
-  [[nodiscard]] bool isDict<ParameterValueView>() const noexcept {
-    auto value = this->value();
-    return value.has_value() && value->is<ParameterValueView::Dict>();
-  }
-
   /// @brief Checks whether the parameter value is a byte array.
-  [[nodiscard]] bool isByteArray() const noexcept {
```
ericmlujan, line 312: Was there a particular reason you moved this as well, given that it isn't an explicit initialization? Or just cleaning things up as you went?

ericmlujan (approved):
LGTM

cpp/foxglove/include/foxglove/server/parameter.hpp
jtbandes: The use of `this->is<std::vector<std::byte>>()` triggered some error about the specialization being defined after the instantiation, so I had to move the function body down below where the `vector<byte>` instantiation was defined.... at least I think that's what was going on 😓

cpp/Makefile
jtbandes: I guess my thinking was that since we don't actually publish the built **C++** artifacts, I didn't see a reason to keep it set to clang... or what did you mean by "official builds"? are you saying we should keep this because we want to test clang on all platforms as well as testing gcc?  (I think by default we are getting clang for the macos builds, gcc for the linux builds and msvc for the windows builds – but haven't verified this)

jtbandes:
> Can we add to CI?

to clarify, I removed some vars that were forcing clang in CI -- the linux builds in CI are now using GCC. You can see it in the logs here: https://github.com/foxglove/foxglove-sdk/actions/runs/16013991239/job/45177082447?pr=516#step:8:22

cpp/Makefile
ericmlujan: That's fair. I suppose I was suggesting we explicitly test clang on macOS and Linux, and just add an additional linux GCC build, but maybe that's overkill. That said, given that this is a public SDK, I think we should be explicit on which compiler/OS pairs we officially support/test against, and document that somewhere

jtbandes: Added new builds so we are now building both clang & gcc on linux

cpp/foxglove/include/foxglove/server/parameter.hpp
```diff
@@ -435,6 +355,121 @@ class ParameterView final {
...
+// Template specializations for ParameterView
+//
+// Documented manually as tparams because specializations are merged:
+// https://github.com/doxygen/doxygen/issues/9468
+
+/// @fn ParameterView::is<std::string_view>()
+/// @tparam std::string_view Checks whether the parameter value is a string.
+
+/// @fn ParameterView::is<std::string>()
+/// @note foobar
```
bryfox: I'm sorry, I was playing around with different ways to render these docs in #518 and neglected to remove this 😅 

```suggestion
```

PR #518, opened 2025-07-02
Update C++ doc generation for template specialization

bryfox:
This is a PR into the branch for #516, and fixes the doc generation errors there.

We needed to ignore the generation of the `ParameterValueView` specializations in addition to `ParameterView`. The additional `@cond` should solve the bug (see issue noted in comments), and get docs state back to where they were.

I also added some manual documentation of the template parameters, so it's clearer from docs what specializations are available. Given the bug, this is the best way I could find to document them; they're listed as template parameters in addition to `T`, which is a little weird but hopefully less confusing than omitting them.

### DIFF
--- a/Container.mk
+++ b/Container.mk
@@ -59,8 +59,13 @@ docs-rust:
 clean-cpp:
 	rm -rf cpp/build*
 
+.PHONY: clean-docs-cpp
+clean-docs-cpp:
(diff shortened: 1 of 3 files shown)


jtbandes (approved):
Thanks 🙏

PR #519, opened 2025-07-02
Support channel filtering in c/c++

bryfox:
### Changelog
C++: support filtering logging sinks by topic

### Description

Based on #500.

This wires up C++ and C to the sink channel filtering on MCAPs and WS Server. I followed existing patterns set by callbacks and fetch-asset, for the most part. The filter function is called once per channel per sink, and the result is cached.

The example is generally the same as Rust and python, with point clouds logged to one MCAP and some lightweight state to another (and both to the app).

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -233,7 +233,7 @@ jobs:
         with:
           target: ${{ matrix.platform.target }}
           working-directory: python/foxglove-sdk
-          args: --release --out dist --find-interpreter
+          args: --release --out dist --interpreter python
           sccache: "false"
       - name: Upload wheels
         uses: actions/upload-artifact@v4
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Visit [Foxglove SDK Docs](https://docs.foxglove.dev/sdk) to get started.
 [![ROS Humble version](https://img.shields.io/ros/v/humble/foxglove_msgs)](https://index.ros.org/p/foxglove_msgs#humble)<br/>
 [![ROS Jazzy version](https://img.shields.io/ros/v/jazzy/foxglove_msgs)](https://index.ros.org/p/foxglove_msgs#jazzy)<br/>
 [![ROS Kilted version](https://img.shields.io/ros/v/kilted/foxglove_msgs)](https://index.ros.org/p/foxglove_msgs#kilted)<br/>
-[![ROS Rolling version](https://img.shields.io/ros/v/rolling/foxglove_msgs)](https://index.ros.org/p/foxglove_msgs#rolling)
+[![ROS Rolling version](https://img.shields.io/ros/v/rolling/foxglove-sdk)](https://index.ros.org/p/foxglove_msgs#rolling)
 
 </td>
 <td>Foxglove schemas for ROS</td>
@@ -92,7 +92,7 @@ Visit [Foxglove SDK Docs](https://docs.foxglove.dev/sdk) to get started.
 [![ROS Humble version](https://img.shields.io/ros/v/humble/foxglove_bridge)](https://index.ros.org/p/foxglove_bridge#humble)<br/>
 [![ROS Jazzy version](https://img.shields.io/ros/v/jazzy/foxglove_bridge)](https://index.ros.org/p/foxglove_bridge#jazzy)<br/>
 [![ROS Kilted version](https://img.shields.io/ros/v/kilted/foxglove_bridge)](https://index.ros.org/p/foxglove_bridge#kilted)<br/>
-[![ROS Rolling version](https://img.shields.io/ros/v/rolling/foxglove_bridge)](https://index.ros.org/p/foxglove_bridge#rolling)
+[![ROS Rolling version](https://img.shields.io/ros/v/rolling/foxglove-sdk)](https://index.ros.org/p/foxglove_bridge#rolling)
 
 </td>
 <td>Foxglove ROS bridge</td>
--- a/cpp/foxglove/include/foxglove/schemas.hpp
+++ b/cpp/foxglove/include/foxglove/schemas.hpp
@@ -1830,10 +1830,23 @@ public:
     std::optional<uint64_t> sink_id = std::nullopt
   ) noexcept;
 
+  /// @brief Close the channel.
+  ///
+  /// You can use this to explicitly unadvertise the channel to sinks that subscribe to channels
+  /// dynamically, such as the WebSocketServer.
+  ///
+  /// Attempts to log on a closed channel will elicit a throttled warning message.
+  void close() noexcept;
+
   /// @brief Uniquely identifies a channel in the context of this program.
   ///
   /// @return The ID of the channel.
   [[nodiscard]] uint64_t id() const noexcept;
+
+  /// @brief Find out if any sinks have been added to the channel.
+  ///
+  /// @return True if sinks have been added to the channel, false otherwise.
+  [[nodiscard]] bool has_sinks() const noexcept;
 
   ArrowPrimitiveChannel(const ArrowPrimitiveChannel& other) noexcept = delete;
   ArrowPrimitiveChannel& operator=(const ArrowPrimitiveChannel& other) noexcept = delete;
@@ -1878,10 +1891,23 @@ public:
     std::optional<uint64_t> sink_id = std::nullopt
   ) noexcept;
 
+  /// @brief Close the channel.
+  ///
+  /// You can use this to explicitly unadvertise the channel to sinks that subscribe to channels
+  /// dynamically, such as the WebSocketServer.
+  ///
+  /// Attempts to log on a closed channel will elicit a throttled warning message.
+  void close() noexcept;
+
   /// @brief Uniquely identifies a channel in the context of this program.
   ///
   /// @return The ID of the channel.
   [[nodiscard]] uint64_t id() const noexcept;
+
+  /// @brief Find out if any sinks have been added to the channel.
+  ///
+  /// @return True if sinks have been added to the channel, false otherwise.
+  [[nodiscard]] bool has_sinks() const noexcept;
 
   CameraCalibrationChannel(const CameraCalibrationChannel& other) noexcept = delete;
   CameraCalibrationChannel& operator=(const CameraCalibrationChannel& other) noexcept = delete;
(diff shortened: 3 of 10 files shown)


bryfox:
@eloff @ericmlujan I've updated the filter's ChannelDescriptor here to include information about the schema, but right now the handling of schema is different than in Rust & Python.

For schemas today, C and C++ each define their own structs. I didn't see a straightforward way to plumb those through directly. Rather than exposing a C struct in the C++ interfaces (which I don't think we do elsewhere), or constructing a separate Schema object, I included the schema name and encoding as strings directly in the descriptor.

Metadata is still iterated and rebuilt in a C++ map, similar to a channel. This isn't called frequently, but I'm absolutely open to ideas here.

c/src/sink_channel_filter.rs
```diff
@@ -0,0 +1,102 @@
...
+                .map(|(key, value)| FoxgloveKeyValue {
+                    key: FoxgloveString::from(key),
+                    value: FoxgloveString::from(value),
+                })
+                .collect();
+            // Safety: we will call from_raw after the callback returns
+            Some(Box::into_raw(Box::new(metadata_items)))
+        } else {
+            None
+        };
```
gasmith: Another option would be to return the channel descriptor as an opaque object, and offer library functions for unpacking metadata, schema, etc. on demand. This would be more aligned with python's dynamic getters.

c/src/lib.rs
```diff
@@ -828,6 +858,24 @@ pub struct FoxgloveMcapOptions {
...
+    /// A filter for channels that can be used to subscribe to or unsubscribe from channels.
+    ///
+    /// This can be used to omit one or more channels from a sink, but still log all channels to another
+    /// sink in the same context. Return false to disable logging of this channel.
+    ///
+    /// This method is invoked from the client's main poll loop and must not block.
+    ///
+    /// # Safety
+    /// - If provided, the handler callback must be a pointer to the filter callback function,
+    ///   and must remain valid until the server is stopped.
```
ericmlujan: Nit: is "server" the term we typically use with mcap writers?

c/src/lib.rs
```diff
@@ -280,6 +282,26 @@ pub struct FoxgloveServerOptions<'a> {
             responder: *mut FoxgloveFetchAssetResponder,
         ),
     >,
+
+    /// Context provided to the `sink_channel_filter` callback.
+    pub sink_channel_filter_context: *const c_void,
```
ericmlujan: For my understanding: in the context (heh) of a WebSocket server, when would this be a different context from the one used to construct the server?

c/src/sink_channel_filter.rs
```diff
@@ -0,0 +1,102 @@
+use crate::{FoxgloveChannelDescriptor, FoxgloveChannelMetadata, FoxgloveKeyValue, FoxgloveString};
+
+/// A filter for channels that can be used to subscribe to or unsubscribe from channels.
+///
+/// This can be used to omit one or more channels from a sink, but still log all channels to another
+/// sink in the same context. The callback should return false to disable logging of this channel.
+///
+/// This method is invoked from the client's main poll loop and must not block.
+#[derive(Clone)]
+pub(crate) struct SinkChannelFilterHandler {
```
ericmlujan: Naming: I'm a little confused as to why this is named what it is. If I understand correctly, this is a concrete implementation of the SinkChannelFilter trait. I think of a "handler" as something responding to an event. I'm wondering if there's a more concise name here? Maybe the trait could be `ChannelFilter` and this implementation is a `SinkChannelFilter`? Not sure, just thinking out loud.

c/src/sink_channel_filter.rs
ericmlujan: That definitely works. Would the use pattern be something like returning a channel ID as a handle, and then calling other functions like `get_channel_metadata(channel_id)`? 

But I do kinda like that a struct is a struct and what you see is what it has. Alternatively, it may be cleaner to implement `From` on the native type for the `FoxgloveChannelDescriptor`, if you're gonna keep this a full struct to pass

cpp/foxglove/include/foxglove/channel.hpp
```diff
@@ -35,6 +35,55 @@ struct Schema {
...
+  [[nodiscard]] const std::optional<std::string>& schema_name() const noexcept;
+  /// @brief Get the schema encoding of the channel.
+  [[nodiscard]] const std::optional<std::string>& schema_encoding() const noexcept;
+
+  // @cond foxglove_internal
+  ChannelDescriptor(ChannelDescriptor&& other) noexcept = default;
+  ChannelDescriptor& operator=(ChannelDescriptor&& other) noexcept = default;
+  ~ChannelDescriptor() = default;
+  // Disable copy operations
+  ChannelDescriptor(const ChannelDescriptor&) = delete;
```
ericmlujan: Curious as to why copying these around is forbidden. They don't own any resources and boil down to a handful of strings. It's probably not the most efficient thing in the world to copy them if you don't have a good reason, but I don't know if I love globally forcing move constructors.

cpp/foxglove/include/foxglove/mcap.hpp
```diff
@@ -61,6 +63,8 @@ struct McapWriterOptions {
   bool repeat_schemas = true;
   /// @brief Whether to truncate the MCAP file.
   bool truncate = false;
+  /// @brief Optional channel filter to use for the MCAP file.
+  SinkChannelFilterFn sink_channel_filter = nullptr;
```
ericmlujan: This confused me a bit, since I didn't even know that `std::function` (a non-pointer type) could be equal-assigned to `nullptr`. Consider explicitly initializing to `{}` instead.

cpp/foxglove/src/server.cpp
```diff
@@ -239,21 +241,52 @@ FoxgloveResult<WebSocketServer> WebSocketServer::create(
     };
   }
 
+  if (options.sink_channel_filter) {
+    sink_channel_filter = std::make_unique<SinkChannelFilterFn>(options.sink_channel_filter);
+
+    c_options.sink_channel_filter_context = sink_channel_filter.get();
```
ericmlujan, line 272: Is this a context? Isn't this a raw pointer to the filter function?

cpp/foxglove/src/server.cpp
```diff
@@ -239,21 +241,52 @@ FoxgloveResult<WebSocketServer> WebSocketServer::create(
...
+    c_options.sink_channel_filter =
+      [](const void* context, const struct foxglove_channel_descriptor* channel) -> bool {
+      try {
+        if (!context) {
+          return true;  // Default to allowing if no filter
+        }
+        auto* filter_func = static_cast<const SinkChannelFilterFn*>(context);
+        auto metadata = buildMetadata_(channel->metadata);
```
ericmlujan: Similarly wondering if there should be a static class method `ChannelDescriptor::from_raw(foxglove_channel_descriptor*)` or custom constructor that wraps this so it's not being duplicated in both the mcap and server implementation

ericmlujan (commented):
Logic generally looks good, had some questions/suggestions on how to reduce the surface area of the code a bit. Tried to think from the perspective of reducing the number of wrapper classes or places we copy values from one to another

c/src/lib.rs
eloff: I think it might not be, but also it could be. I don't think we should exclude the possibility.

c/src/sink_channel_filter.rs
eloff: +1

c/src/sink_channel_filter.rs
```diff
@@ -0,0 +1,102 @@
...
+    callback_context: *const std::ffi::c_void,
```
eloff, line 11: How long does this live for, and how does the user know when to free it?

It seems to me like the lifetime would be bound to the Context?

c/src/sink_channel_filter.rs
```diff
@@ -0,0 +1,102 @@
...
+            let metadata_items: Vec<FoxgloveKeyValue> = channel
+                .metadata()
+                .iter()
+                .map(|(key, value)| FoxgloveKeyValue {
+                    key: FoxgloveString::from(key),
+                    value: FoxgloveString::from(value),
+                })
+                .collect();
+            // Safety: we will call from_raw after the callback returns
+            Some(Box::into_raw(Box::new(metadata_items)))
```
eloff: "Return" the vec itself here and use vec into raw parts instead of boxing it

c/src/sink_channel_filter.rs
```diff
@@ -0,0 +1,102 @@
...
+                count: unsafe { (*metadata_items_ptr).len() },
+            });
+
+            FoxgloveChannelDescriptor {
+                topic: FoxgloveString::from(channel.topic()),
+                encoding: FoxgloveString::from(channel.message_encoding()),
+                schema_name: FoxgloveString::from(&schema_name),
+                schema_encoding: FoxgloveString::from(&schema_encoding),
+                // Safety: we will call from_raw after the callback returns
+                metadata: Box::into_raw(metadata),
```
eloff: Why box this, just have it be null/0 len when omitted?

c/src/sink_channel_filter.rs
gasmith: I'd like to see the channel descriptor used more broadly, in places where we currently pass around an entire channel. As part of that, I want the construction of an FFI channel descriptor to be very cheap - no heap allocations. I nearly suggested a semi-opaque structure, one that includes id, name, encodings... and a raw pointer (of some sort) that we can use for heavier-weight operations like iterating over metadata.

cpp/foxglove/include/foxglove/channel.hpp
```diff
@@ -35,6 +35,55 @@ struct Schema {
   size_t data_len = 0;
 };
```
eloff, line 36: This looks good, but there seems to be a lifetime issue with the callback context (maybe applies to some of our other C++ callbacks too?)

Doesn't seem to be a way to free it. What do you think?

eloff:
This looks good, but there seems to be a lifetime issue with the callback context (maybe applies to some of our other C++ callbacks too?)

Doesn't seem to be a way to free it. What do you think?

c/src/lib.rs
bryfox: Nope, just a lazy copy/paste from above.

c/src/sink_channel_filter.rs
bryfox: I like the semi-opaque structure idea. I started this by trying the pointer + function approach but got lost in pointer management with channels and descriptors. I'll revisit this with fresh eyes.

cpp/foxglove/include/foxglove/channel.hpp
bryfox: Makes sense, thanks. I followed this pattern from elsewhere without putting much thought into it.

c/src/sink_channel_filter.rs
bryfox: I pushed one alternative to the branch [bryan/fg-11732-cpp-sink-filters-opaque](https://github.com/foxglove/foxglove-sdk/compare/bryan/fg-11732-cpp-sink-filters...bryan/fg-11732-cpp-sink-filters-opaque); basically the opaque pointer with corresponding C functions.

bryfox: @gasmith any thoughts on the above alternative?

gasmith: The opaque structure you have in that branch looks good to me. I think that's straightforward.

bryfox:
OK, this has been updated (finally!) with an opaque pointer to a descriptor [discussed in this thread](https://github.com/foxglove/foxglove-sdk/pull/519#discussion_r2208019489).

PR #520, opened 2025-07-02
Don't unadvertise a channel if an error prevented it from being advertised

eloff:
### Changelog

Fixed a bug where a channel that fails to be advertised to a websocket client would still be unadvertised. Now unadvertise is only sent if advertise succeeded for a channel.

### Description

https://linear.app/foxglove/issue/FG-12137/sdk-websocket-server-should-not-unadvertise-channels-that-it-failed-to

I modified one of our tests to identify this issue and fail, then fixed the bug and verified the test passes now.

### DIFF
--- a/rust/foxglove/src/websocket/connected_client.rs
+++ b/rust/foxglove/src/websocket/connected_client.rs
@@ -710,11 +710,6 @@ impl ConnectedClient {
     /// Unadvertises a channel to the client.
     fn unadvertise_channel(&self, channel_id: ChannelId) {
         if self.channels.write().remove(&channel_id).is_none() {
-            tracing::debug!(
-                "Client {} is not advertising channel: {}",
-                self.addr,
-                channel_id
-            );
             return;
         }
 
--- a/rust/foxglove/src/websocket/connected_client.rs
+++ b/rust/foxglove/src/websocket/connected_client.rs
@@ -709,7 +709,14 @@ impl ConnectedClient {
 
     /// Unadvertises a channel to the client.
     fn unadvertise_channel(&self, channel_id: ChannelId) {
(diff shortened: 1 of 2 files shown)


rust/foxglove/src/websocket/connected_client.rs
```diff
@@ -709,7 +709,14 @@ impl ConnectedClient {
 
     /// Unadvertises a channel to the client.
     fn unadvertise_channel(&self, channel_id: ChannelId) {
-        self.channels.write().remove(&channel_id);
+        if self.channels.write().remove(&channel_id).is_none() {
+            tracing::debug!(
+                "Client {} is not advertising channel: {}",
```
bryfox: The wording here is a little strange since these aren't client advertisements. Maybe "Client {} was not advertised channel: {}"?

Or consider removing this entirely — it seems like an implementation detail that we even store the channel for this long? If we don't already, maybe warning on the failed advertisement in the first place makes sense.

bryfox (approved):
Minor thoughts on the debug log, but the fix looks good.

gasmith (commented):
Note that `advertse::advertise_channels()` [filters out channels](https://github.com/foxglove/foxglove-sdk/blob/53a7b23c62f9b87df3bc095fc1a2bbdd7bb0966e/rust/foxglove/src/websocket/advertise.rs#L52) whose descriptors cannot be serialized into the ws-protocol Advertise message. We need to update [the way that `ConnectedClient::advertise_channels` populates its `channels` hashmap](https://github.com/foxglove/foxglove-sdk/blob/53a7b23c62f9b87df3bc095fc1a2bbdd7bb0966e/rust/foxglove/src/websocket/connected_client.rs#L689-L696) to take this filtering into account.

PR #521, opened 2025-07-03
data-loaders: support full frontend problem API

bennetthardwick:
### Changelog

None

### Docs

None

### Description

This change adds support for problem severity and tips. This allows users to write data loaders that can show warnings with tips.

Fix FG-12275.

### DIFF
--- a/rust/foxglove_data_loader/src/lib.rs
+++ b/rust/foxglove_data_loader/src/lib.rs
@@ -81,12 +81,14 @@ macro_rules! export {
 
 use anyhow::anyhow;
 use std::collections::HashMap;
+use std::fmt::{Display, Formatter};
 use std::{cell::RefCell, rc::Rc};
 
 pub use generated::exports::foxglove::loader::loader::{
     self, BackfillArgs, Channel, ChannelId, DataLoaderArgs, Message, MessageIteratorArgs, Schema,
     SchemaId, TimeRange,
 };
+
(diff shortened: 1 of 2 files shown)


rust/foxglove_data_loader/src/lib.rs
```diff
@@ -115,14 +117,78 @@ impl std::io::Seek for reader::Reader {
...
+    pub fn new(severity: Severity, message: impl Into<String>) -> Self {
+        Self(loader::Problem {
+            severity: severity.to_string(),
+            message: message.into(),
+            tip: None,
+        })
+    }
+
+    /// Add additional context to the problem.
+    pub fn with_tip(mut self, tip: impl Into<String>) -> Self {
```
nthiad: Why `with_tip` here instead of `tip`?

bennetthardwick: Not sure, I'll change to tip that feels better.

rust/foxglove_data_loader/wit/loader.wit
```diff
@@ -33,11 +33,20 @@ interface loader {
     type schema-id = u16;
     type error = string;
 
+    record problem {
+        // The severity of the problem. Currently "error", "warn" and "info" are supported.
+        severity: string,
```
achim-k: maybe make this an enum instead?

PR #523, opened 2025-07-03
Only track channels that were actually advertised in ConnectedClient::channels

eloff:
Address this comment: https://github.com/foxglove/foxglove-sdk/pull/520#pullrequestreview-2984405393

### DIFF
--- a/rust/foxglove/src/websocket/connected_client.rs
+++ b/rust/foxglove/src/websocket/connected_client.rs
@@ -1,4 +1,5 @@
 use std::collections::hash_map::Entry;
+use std::collections::HashSet;
 use std::sync::Weak;
 use std::{collections::HashMap, net::SocketAddr, sync::Arc};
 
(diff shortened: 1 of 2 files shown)


gasmith (approved):
Fix looks good. It'd be nice to have a unit test to cover this, if you can muster one. Maybe you can borrow from (or amend) `test_advertise_schemaless_channels()`, which includes a protobuf channel that doesn't serialize for lack of a schema.

eloff:
> Fix looks good. It'd be nice to have a unit test to cover this, if you can muster one. Maybe you can borrow from (or amend) `test_advertise_schemaless_channels()`, which includes a protobuf channel that doesn't serialize for lack of a schema.

It does have one (test_advertise_to_client), but it didn't catch this because you have to have at least one channel filtered and one not, in the same call to advertise_channels. I think I can modify it so it adds both channels in the same call to advertise_channels.

PR #525, opened 2025-07-04
Support adding channel and schema by id in data loaders

bennetthardwick:
### Changelog

None

### Docs

None

### Description

When writing data loaders for formats like MCAP or db3, we already have channel and schema IDs. This change allows authors to use those IDs directly, instead of needing to look up by topic string or using some other method.

It also fixes up schemaless channels, supports multiple channels with the same topic name and adds some test coverage.

It also removes some other methods that might not be necessary.


rust/foxglove_data_loader/src/lib.rs
```diff
@@ -148,31 +144,132 @@ impl Initialization {
...
+            id,
+            schema,
+            channels: channels.clone(),
+            message_encoding: String::from(""),
+        };
+
+        self.schemas.insert(id, schema.clone());
 
-    pub fn get_channel(&self, topic_name: &str) -> Option<Rc<loader::Channel>> {
-        self.channels_by_topic.get(topic_name).cloned()
```
bennetthardwick, line 202: @nthiad I removed this method you use to look up channels by topic name in the nd+json example. I feel like storing the initialization is a bit of an anti-pattern, you could now set the channel IDs to be `const` and use `add_channel_with_id`. 

Happy to make this work though if you're set on it.

nthiad: That's how I had it initially, but feedback I received was that managing channel IDs and saving them as const manually was too cumbersome.

rust/foxglove_data_loader/src/lib.rs
```diff
@@ -422,3 +533,230 @@ pub trait MessageIterator: 'static + Sized {
     type Error: Into<Box<dyn std::error::Error>>;
     fn next(&mut self) -> Option<Result<Message, Self::Error>>;
 }
+
+#[cfg(test)]
```
nthiad, line 608: Can we move these tests to a separate file? This file is already getting pretty long.

rust/foxglove_data_loader/src/lib.rs
bennetthardwick: I don't think it is too bad - you have to deal with `channel_id` anyway since you're putting it in the output message. It just obscures things otherwise.

Also it's not ideal to be doing a hash every message to get the correct channel id so it might just be a necessary evil.

PR #528, opened 2025-07-07
Adding 3D equivalent to `Foxglove.Grid`

jeff-foxglove:
### Changelog
None

### Docs

TODO (need to put the links from the auto generated docs from this into the docs repo)

### Description

Adding an extended version of `Foxglove.Grid` that supports an extra dimension for storing 3D data

Fixes: FG-11711


james-rms:
how do we justify this approach vs. adding a `depth` PackedElementField to the existing grid message?

jeff-foxglove:
> how do we justify this approach vs. adding a depth PackedElementField to the existing grid message?

@james-rms Sorry for the slow circling back to this. My concern was backwards compatibility. In addition to the `depth` field we also need to have the `cell_size` go from `vector2` to `vector3` when present as well, which I thought might break things.

james-rms:
> In addition to the depth field we also need to have the cell_size go from vector2 to vector3 when present as well


I don't think this follows. referring to cell_size's documentation:
> Size of single grid cell along x and y axes, relative to pose

i think in this world, cell size still makes sense as a vector2. IMO the only thing missing once you add a `depth` field is some way to scale the depth of the grid in the 3d view. I'm not familiar enough with our schemas to say whether we can backwards-compatibly add fields to schemas, but if we could, I could imagine adding a single `depth_scale` scalar field to the schema.

--- EDIT --- 

my mistake, I misread the intention of this ticket, which is to store arbitrary voxel grids rather than 2.5D. Please disregard this comment, I apologise :/

jeff-foxglove:
No stress! 

Following on the thread of reusing the existing schema though, would it make more sense to either extend the existing `Foxglove.Grid` message and override. the `cell_size` + add the new property or have a shared base class, or do we prefer to keep different schemas separate?

james-rms:
I think if you want to support more than 2.5d, you'd need to add the concept of depth slices to the grid schema, which i think warrants creating a new schema.

c/include/foxglove-c/foxglove-c.h
```diff
@@ -1130,6 +1130,58 @@ typedef struct foxglove_grid {
...
+  /**
+   * Number of grid rows
+   */
+  uint32_t row_count;
+  /**
+   * Number of grid columns
+   */
+  uint32_t column_count;
+  /**
+   * Size of single grid cell along x, y, and z axes, relative to `pose`
```
james-rms, line 1175: not sure what `relative to pose` means here.

jeff-foxglove: I believe it means the x,y,z axes are local to the `const struct foxglove_pose *pose;` orientation

c/include/foxglove-c/foxglove-c.h
```diff
@@ -1130,6 +1130,58 @@ typedef struct foxglove_grid {
...
+   * Number of bytes between cells within a row in `data`
+   */
+  uint32_t cell_stride;
+  /**
+   * Fields in `data`. `red`, `green`, `blue`, and `alpha` are optional for customizing the grid's color.
+   */
+  const struct foxglove_packed_element_field *fields;
+  size_t fields_count;
+  /**
+   * Grid cell data, interpreted using `fields`, in depth-major, row-major (Z-Y-X) order — values fill each row from left to right along the X axis, with rows ordered from top to bottom along the Y axis, starting at the bottom-left corner when viewed from +Z looking towards -Z with identity orientations
```
james-rms: I find this really difficult to figure out - referring to left, right, up, down doesn't help much when it's not clear what coordinates they refer to. Maybe you could eliminate those concepts and just show what you mean in mathematical terms, like:

> Grid cell data, interpreted using `fields` in depth-major, row-major (Z-Y-X) order. For the `data` element starting at byte offset `offset`, the coordinates of its corner closest to the origin will be:
>  x = ((i % row_stride) * cell_size.x // TBD these are totally wrong
> y = ((i / column_count) % row_count) * cell_size.y
>  z = ((i / (column_count * row_count)) % ) * cell_size.z

I can't really figure out exactly what the calculations should be from the docstrings, maybe laying them out in public documentation would help clear that up too?

jeff-foxglove: Yea that makes sense. I'll reflect this back in the `Foxglove.Grid` docstrings this was based on while I'm at it

typescript/schemas/src/internal/__snapshots__/exportTypeScriptSchemas.test.ts.snap
```diff
@@ -360,7 +360,62 @@ export type Grid = {
...
+  /** Number of bytes between cells within a row in \`data\` */
+  cell_stride: number;
+
+  /** Fields in \`data\`. \`red\`, \`green\`, \`blue\`, and \`alpha\` are optional for customizing the grid's color. */
+  fields: PackedElementField[];
+
+  /**
+   * Grid cell data, interpreted using \`fields\`, in depth-major, row-major (Z-Y-X) order.
+   *  For the data element starting at byte offset offset, the coordinates of its corner closest to the origin will be:
+   *  z = (i / (row_stride * cell_stride)) % slice_stride
```
james-rms: this has to be affected by cell size, right? also have you checked that this results in the coordinates of its corner closest to the origin, or the cube centre, or a different corner?

jeff-foxglove: I guess it depends on if we are talking about indexing or actual positioning in the world. The prior no, the later yes

jeff-foxglove: In terms of the coordinates in the corner or not, no visualization exits yet for this. It is dependant on this schema existing beforehand, so we can make it match

james-rms: You'd want it to be somewhat consistent with the existing grid behaviorthough right?

typescript/schemas/src/internal/schemas.ts
```diff
@@ -1122,7 +1122,74 @@ const Grid: FoxgloveMessageSchema = {
...
+      type: { type: "nested", schema: PackedElementField },
+      array: true,
+      description:
+        "Fields in `data`. `red`, `green`, `blue`, and `alpha` are optional for customizing the grid's color.",
+    },
+    {
+      name: "data",
+      type: { type: "primitive", name: "bytes" },
+      description:
+        "Grid cell data, interpreted using `fields`, in depth-major, row-major (Z-Y-X) order.\n For the data element starting at byte offset offset, the coordinates of its corner closest to the origin will be:\n z = (i / (row_stride * cell_stride)) % slice_stride\n y = (i / cell_stride) % row_stride\n x = i % cell_stride",
```
james-rms:
```suggestion
        "Grid cell data, interpreted using `fields`, in depth-major, row-major (Z-Y-X) order.\n For the data element starting at byte offset i, the coordinates of its corner closest to the origin will be:\n z = (i / (row_stride * cell_stride)) % slice_stride\n y = (i / cell_stride) % row_stride\n x = i % cell_stride",
```

typescript/schemas/src/internal/__snapshots__/exportTypeScriptSchemas.test.ts.snap
james-rms: > I guess it depends on if we are talking about indexing or actual positioning in the world. The prior no, the later yes

i mean if your cell_size vector3 is (10.0, 10.0, 10.0), the second cell in the list will have its corner at (10.0, 0, 0)

jeff-foxglove: Updated all of the doc entries to include the cell_size on those indices

james-rms:
Overall this looks good to me. I don't think we should publish this schema until we have a working panel that consumes it, because there's always something that could come up during development that would require us to change this.

c/cbindgen.toml
```diff
@@ -95,6 +95,7 @@ FrameTransform = "foxglove_frame_transform"
 FrameTransforms = "foxglove_frame_transforms"
 GeoJson = "foxglove_geo_json"
 Grid = "foxglove_grid"
+Grid3D = "foxglove_grid3_d"
```
james-rms: consider renaming to voxel grid?

c/include/foxglove-c/foxglove-c.h
```diff
@@ -1138,12 +1138,71 @@ typedef struct foxglove_grid {
   const struct foxglove_packed_element_field *fields;
   size_t fields_count;
   /**
-   * Grid cell data, interpreted using `fields`, in row-major (y-major) order — values fill each row from left to right along the X axis, with rows ordered from top to bottom along the Y axis, starting at the bottom-left corner when viewed from +Z looking towards -Z with identity orientations
+   * Grid cell data, interpreted using `fields`, in row-major (y-major) order.
+   *  For the data element starting at byte offset i, the coordinates of its corner closest to the origin will be:
+   *  y = (i / cell_stride) % row_stride * cell_size.y
```
james-rms, line 1143: how have you validated that this logic matches what the app does?

jeff-foxglove: This was the proof I was working off of, if you agree that aligns with the proposed docs (starts in the corner of the origin and increases along the `x` before the `y`? Note the elevation here is not from the 3D variant, but just displacing by an elevation layer in the 2D

Screenshot 2025-07-23 at 12 04 23 AM

(console output is the `Float32` conversion of the `Uint8`)

jeff-foxglove:
@james-rms I have the [viz side done](https://github.com/foxglove/app/pull/10230) now using this schema so should be good. I think I will make the minor change from `Foxglove.Grid3D` -> `Foxglove.Grid3` as per @amacneil's suggestion to keep it in line with the ` Vector3` etc. naming convention.

c/include/foxglove-c/foxglove-c.h
```diff
@@ -1138,12 +1138,71 @@ typedef struct foxglove_grid {
...
+  /**
+   * Number of bytes between rows in `data`
+   */
+  uint32_t row_stride;
+  /**
+   * Number of bytes between cells within a row in `data`
+   */
+  uint32_t cell_stride;
+  /**
+   * Fields in `data`. `red`, `green`, `blue`, and `alpha` are optional for customizing the grid's color.
```
james-rms, line 1191: is anything required?

jeff-foxglove: No, works pretty much the same for colouring as the 2d Grid. Either you can view the rgba as a set if they are present or you can visualize an arbitrary field.

PR #529, opened 2025-07-07
Fix data loader seek from end

bennetthardwick:
### Changelog

None

### Docs

None

### Description

The conversion to `SeekFrom::End` was around the wrong way. This came up when trying to build an MCAP loader that seeks from the end of the file to read summary info.


nthiad:
This was my mistake.

> End(i64)
> Sets the offset to the size of this object **plus** the specified number of bytes.

https://doc.rust-lang.org/std/io/enum.SeekFrom.html

achim-k (approved):
This caused the seek bug you mentioned in https://github.com/foxglove/foxglove-sdk/pull/525#discussion_r2191232193 ?

PR #532, opened 2025-07-09
Python: add session_id option to start_server

bryfox:
### Changelog
Python: add session_id option to start_server

### Description

This adds the session_id option to start_server. Typically, the default will work just fine, but we provide this option for other languages, and clear_session() also optionally accepts an explicit ID.

### DIFF
--- a/python/foxglove-sdk/src/websocket.rs
+++ b/python/foxglove-sdk/src/websocket.rs
@@ -366,7 +366,7 @@ impl foxglove::websocket::service::Handler for ServiceHandler {
 /// To connect to this server: open Foxglove, choose "Open a new connection", and select Foxglove
 /// WebSocket. The default connection string matches the defaults used by the SDK.
 #[pyfunction]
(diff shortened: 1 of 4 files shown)


ericmlujan (approved):

python/foxglove-sdk/src/websocket.rs
```diff
@@ -379,12 +379,15 @@ pub fn start_server(
...
-    let session_id = time::SystemTime::now()
-        .duration_since(time::UNIX_EPOCH)
-        .expect("Failed to create session ID; invalid system time")
-        .as_millis()
-        .to_string();
+    let session_id = session_id.unwrap_or_else(|| {
+        time::SystemTime::now()
+            .duration_since(time::UNIX_EPOCH)
+            .expect("Failed to create session ID; invalid system time")
+            .as_millis()
```
eloff: Curious why not as nanos? Could replace this with nanoseconds_since_epoch().to_string()?

eloff (approved):
LGTM

python/foxglove-sdk/src/websocket.rs
bryfox: Actually, we don't need this at all, since Rust will default it for us.

PR #533, opened 2025-07-09
Fix race where on_subscribe callback is called before channel is registered with a context

ericmlujan:
### Changelog
None

### Docs
None

### Description
The Rust WebSocketServer implementation invokes its subscription callback asynchronously before the connected client's sink is registered with that channel in the context. This can cause channel::log() to be called in a callback before the sink is ready to accept messages.

This was discovered when working on porting ros-foxglove-bridge to use the SDK as its backend; unit tests for logging latched messages were failing ~2-5% of the time because messages were being dropped.

Verified that these unit tests now pass reliably with a locally-built SDK artifact.

Fixes: FG-12371

### DIFF
--- a/rust/foxglove/src/websocket/connected_client.rs
+++ b/rust/foxglove/src/websocket/connected_client.rs
@@ -424,7 +424,7 @@ impl ConnectedClient {
         }
 
         let mut channel_ids = Vec::with_capacity(subscribed_channels.len());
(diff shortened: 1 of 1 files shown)


ericmlujan:
@gasmith it's a good call out, but I'm also not sure what if anything it could affect. Given that, I'm inclined to leave it as it is.

PR #534, opened 2025-07-10
Fix data loader severity enum ordering

bennetthardwick:
### Description

I'm not sure how this got mixed up, but the app expects the order `info, warn, error`. There was likely a commit I forgot to push when the problems change was added.

### DIFF
--- a/rust/foxglove_data_loader/wit/loader.wit
+++ b/rust/foxglove_data_loader/wit/loader.wit
@@ -35,8 +35,8 @@ interface loader {
 
     enum severity {
         info,
+        warn,
         error,
-        warn
     }
 
     record problem {


rust/foxglove_data_loader/wit/loader.wit
```diff
@@ -35,8 +35,8 @@ interface loader {
 
     enum severity {
         info,
+        warn,
```
achim-k, line 38: Looks like WIT doesn't support assigning values to enums, e.g. `warn = 1,`

achim-k:
Shouldn't the unit test that you added in the app repo have caught thsi?

bennetthardwick:
> Shouldn't the unit test that you added in the app repo have caught thsi?

Yes, but the artifact in there was from a merged branch while the PRs were still open. When I updated the app to be the latest main artifacts it broke. Definitely not the most ideal workflow, it would be worth thinking about how to make sure everything stays up to date.

PR #542, opened 2025-07-14
Introduce explicit integer Parameter type

ericmlujan:
### Changelog
- Introduce an explicit Integer (64-bit signed integer) type for Parameters.
- Rename the Number parameter type to Float64 to make its semantic clearer
- Update Rust/C/C++/Python bindings to allow constructing Integer and IntegerArray parameters, and to rename Number to Float64 for consistency.

### Docs
Documentation comments updated.

### Description
Previously, the Number parameter type was overloaded to represent both integer and floating-point types (this is because when parameters are sent over the wire as JSON, the distinction is lost). SDK users wanting to construct integer parameters needed to cast them to doubles. When these values were round-tripped through the WebSocketServer interface, a fractional part of `.0` was added, which caused some deserialization libraries to treat them as doubles.

While this can be worked around by SDK users by rolling their own serialization and deserialization overloads to infer when a number should be cast to a native integer type or kept as a double, it's more ergonomic for the SDK to allow for constructing parameter values from native integer types, and to make a best-effort attempt to serialize integer parameters without adding a zero fractional part (ie. `"my_param": 42` over `"my_param": 42.0`).

This PR also includes a custom JSON deserializer to more closely align the behavior of parsing params from the wire into the behavior in the protocol spec (namely, more closely respecting and enforcing checks around the `type` tag).

Only a Number parameter type exists, which handles both floating point and integer values.

</td><td>

A new Integer parameter type exists for handling explicit integers. Number is deprecated and replaced with Float to be more explicit.

</td></tr></table>

Fixes: FG-12407

### DIFF
--- a/c/include/foxglove-c/foxglove-c.h
+++ b/c/include/foxglove-c/foxglove-c.h
@@ -205,6 +205,7 @@ enum foxglove_parameter_value_tag
 #endif // __cplusplus
  {
   FOXGLOVE_PARAMETER_VALUE_TAG_NUMBER,
+  FOXGLOVE_PARAMETER_VALUE_TAG_INTEGER,
   FOXGLOVE_PARAMETER_VALUE_TAG_BOOLEAN,
   FOXGLOVE_PARAMETER_VALUE_TAG_STRING,
   FOXGLOVE_PARAMETER_VALUE_TAG_ARRAY,
@@ -379,6 +380,7 @@ typedef struct foxglove_parameter_value_dict {
  */
 typedef union foxglove_parameter_value_data {
   double number;
+  int64_t integer;
   bool boolean;
   struct foxglove_string string;
   struct foxglove_parameter_value_array array;
@@ -3129,6 +3131,20 @@ foxglove_error foxglove_parameter_create_float64(struct foxglove_parameter **par
                                                  double value);
 
 /**
+ * Creates a new integer parameter.
+ *
+ * The value must be freed with `foxglove_parameter_free`, or by passing it to a consuming
+ * function such as `foxglove_parameter_array_push`.
+ *
+ * # Safety
+ * - `param` must be a valid pointer.
+ * - `name` must be a valid `foxglove_string`. This value is copied by this function.
+ */
+foxglove_error foxglove_parameter_create_integer(struct foxglove_parameter **param,
+                                                 struct foxglove_string name,
+                                                 int64_t value);
+
+/**
  * Creates a new boolean parameter.
  *
  * The value must be freed with `foxglove_parameter_free`, or by passing it to a consuming
@@ -3187,6 +3203,14 @@ foxglove_error foxglove_parameter_create_byte_array(struct foxglove_parameter **
 foxglove_error foxglove_parameter_create_float64_array(struct foxglove_parameter **param,
                                                        struct foxglove_string name,
                                                        const double *values,
+                                                       size_t values_len);
+
+/**
+ * Creates a new parameter which is an array of integer values.
+ */
+foxglove_error foxglove_parameter_create_integer_array(struct foxglove_parameter **param,
+                                                       struct foxglove_string name,
+                                                       const int64_t *values,
                                                        size_t values_len);
 
 /**
@@ -3263,6 +3287,14 @@ void foxglove_parameter_free(struct foxglove_parameter *param);
  * function such as `foxglove_parameter_create`.
  */
 struct foxglove_parameter_value *foxglove_parameter_value_create_number(double number);
+
+/**
+ * Creates a new integer parameter value.
+ *
+ * The value must be freed with `foxglove_parameter_value_free`, or by passing it to a consuming
+ * function such as `foxglove_parameter_create`.
+ */
+struct foxglove_parameter_value *foxglove_parameter_value_create_integer(int64_t integer);
 
 /**
  * Creates a new boolean parameter value.
--- a/c/include/foxglove-c/foxglove-c.h
+++ b/c/include/foxglove-c/foxglove-c.h
@@ -3281,7 +3281,7 @@ struct foxglove_parameter *foxglove_parameter_clone(const struct foxglove_parame
 void foxglove_parameter_free(struct foxglove_parameter *param);
 
 /**
- * Creates a new number parameter value.
+ * Creates a newumber float64 parameter value.
  *
  * The value must be freed with `foxglove_parameter_value_free`, or by passing it to a consuming
  * function such as `foxglove_parameter_create`.
@@ -3289,7 +3289,9 @@ void foxglove_parameter_free(struct foxglove_parameter *param);
 struct foxglove_parameter_value *foxglove_parameter_value_create_float64(double number);
 
 /**
- * Creates a new  or by passing it to a consuming
+ * Creates a new integer parameter value.
+ *
+ * The value must be freed with `foxglove_parameter_value_free`, or by passing it to a consuming
  * function such as `foxglove_parameter_create`.
  */
 struct foxglove_parameter_value *foxglove_parameter_value_create_integer(int64_t integer);
--- a/c/include/foxglove-c/foxglove-c.h
+++ b/c/include/foxglove-c/foxglove-c.h
@@ -204,7 +204,7 @@ enum foxglove_parameter_value_tag
   : uint8_t
 #endif // __cplusplus
  {
-  FOXGLOVE_PARAMETER_VALUE_TAG_NUMBER,
+  FOXGLOVE_PARAMETER_VALUE_TAG_FLOAT64,
   FOXGLOVE_PARAMETER_VALUE_TAG_INTEGER,
   FOXGLOVE_PARAMETER_VALUE_TAG_BOOLEAN,
   FOXGLOVE_PARAMETER_VALUE_TAG_STRING,
@@ -379,7 +379,7 @@ typedef struct foxglove_parameter_value_dict {
  * Storage for `FoxgloveParameterValue`.
  */
 typedef union foxglove_parameter_value_data {
-  double number;
+  double float64;
   int64_t integer;
   bool boolean;
   struct foxglove_string string;
@@ -3286,12 +3286,10 @@ void foxglove_parameter_free(struct foxglove_parameter *param);
  * The value must be freed with `foxglove_parameter_value_free`, or by passing it to a consuming
  * function such as `foxglove_parameter_create`.
  */
-struct foxglove_parameter_value *foxglove_parameter_value_create_number(double number);
+struct foxglove_parameter_value *foxglove_parameter_value_create_float64(double number);
 
 /**
- * Creates a new integer parameter value.
- *
- * The value must be freed with `foxglove_parameter_value_free`, or by passing it to a consuming
+ * Creates a new  or by passing it to a consuming
  * function such as `foxglove_parameter_create`.
  */
 struct foxglove_parameter_value *foxglove_parameter_value_create_integer(int64_t integer);
--- a/c/include/foxglove-c/foxglove-c.h
+++ b/c/include/foxglove-c/foxglove-c.h
@@ -3281,7 +3281,7 @@ struct foxglove_parameter *foxglove_parameter_clone(const struct foxglove_parame
 void foxglove_parameter_free(struct foxglove_parameter *param);
 
 /**
- * Creates a newumber float64 parameter value.
+ * Creates a new float64 parameter value.
  *
  * The value must be freed with `foxglove_parameter_value_free`, or by passing it to a consuming
  * function such as `foxglove_parameter_create`.
--- a/c/src/parameter.rs
+++ b/c/src/parameter.rs
@@ -893,7 +893,7 @@ impl Drop for FoxgloveParameterValue {
     }
 }
 
-/// Creates a newumber float64 parameter value.
+/// Creates a new float64 parameter value.
 ///
 /// The value must be freed with `foxglove_parameter_value_free`, or by passing it to a consuming
 /// function such as `foxglove_parameter_create`.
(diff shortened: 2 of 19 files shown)


rust/foxglove/src/websocket/ws_protocol/parameter.rs
```diff
@@ -34,7 +34,12 @@ pub enum ParameterType {
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(untagged)]
 pub enum ParameterValue {
+    /// An integer value.
+    /// NOTE: This needs to be before `Number` to ensure that values without a fractional part
+    /// are preferrentially deserialized as integers, not floats.
```
gasmith: Note that the `///` docstrings on public types are compiled into [the user-facing documentation](https://docs.rs/foxglove/latest/foxglove/websocket/enum.ParameterValue.html). For internal commentary, we should use `//`:

```suggestion
    // NOTE: This needs to be before `Number` to ensure that values without a fractional part
    // are preferrentially deserialized as integers, not floats.
```

(This suggestion is probably moot, though, considering we're going to need to roll our own `Deserialize` implementation.)

rust/foxglove/src/websocket/ws_protocol/parameter.rs
```diff
@@ -34,7 +34,12 @@ pub enum ParameterType {
...
+    Integer(i64),
     /// A decimal or integer value.
+    /// TODO: Rename to indicate that this is a float only
```
gasmith: Let's just do it. I'm fully in favor of renaming this to "Float", to distinguish it from "Integer". I can't think of any reason we'd want to keep an overloaded term. We're changing the interface anyway by introducing the new variant on an enum that isn't marked non-exhaustive.

rust/foxglove/src/websocket/ws_protocol/parameter.rs
```diff
@@ -165,6 +190,16 @@ mod tests {
         insta::assert_json_snapshot!(Parameter::float64_array("f64[]", [1.23, 4.56]));
     }
 
+    #[test]
+    fn test_integer() {
+        insta::assert_json_snapshot!(Parameter::integer("i64", 123));
+    }
+
+    #[test]
+    fn test_integer_array() {
```
gasmith, line 380: In addition to these encoding tests, and the roundtrip tests from `client::set_parameters`, we should add some explicit decoding tests. For example:

- `{"value": 123}`: int
- `{"value": [123, 456]}`: int[]
- `{"value": 123, "type": "float64"}`: float64
- `{"value": [123, 456], "type": "float64_array"}`: float64[]

My guess is that we're not handling the float cases correctly. We can't rely entirely on serde, because we need to look at the `type` discriminant to identify whether what we're able to parse as an integer was actually declared as a float. We'll most likely need a custom `Deserialize` implementation.

There are also some interesting cases that are under-specified:

- `{"value": 123.0}`: float64 (?)
- `{"value": [123.0, 456.0]}`: float64[] (?)
- `{"value": [123, 456.0]}`: float64[] (??)
- `{"value": [123.0, 456]}`: float64[] (??)

For the (?) cases: The spec isn't clear about how to treat a number with a zero-valued fractional component in the absence of a "type" discriminant. I'm inclined to treat such a number as a float.

For the (??) cases: Likewise, but for a heterogeneous array, containing both integers and numbers with fractional components. I'm inclined to treat such an array as float64[].

rust/foxglove/src/websocket/ws_protocol/parameter.rs
ericmlujan: 😬 was hoping to avoid that, but you are right! and thanks, I think my editor just got happy with the triple slashes ahaha

rust/foxglove/src/websocket/ws_protocol/parameter.rs
ericmlujan: Cool, just wanted to be a bit defensive and get a first round of review before implementing that change and further blowing up the diff.

rust/foxglove/src/websocket/ws_protocol/parameter.rs
ericmlujan: I am also similarly inclined to take the tack of, when it's not entirely obvious, default to float. But that decision should probably be reflected in the protocol spec.

rust/foxglove/src/websocket/ws_protocol/parameter.rs
```diff
@@ -62,6 +67,191 @@ pub struct Parameter {
...
+                        }
+                        "type" => {
+                            r#type = Some(map.next_value()?);
+                        }
+                        "value" => {
+                            value = Some(map.next_value()?);
+                        }
+                        _ => {
+                            // Ignore unknown fields
+                            let _: serde_json::Value = map.next_value()?;
```
gasmith: We should probably return `Error::unknown_field(field, ["name", "type", "value"])` here.

rust/foxglove/src/websocket/ws_protocol/parameter.rs
```diff
@@ -62,6 +67,191 @@ pub struct Parameter {
...
+                    }
+                    _ => {
+                        return Err("Non-numeric value in float64 array".to_string());
+                    }
+                }
+            }
+            Ok(ParameterValue::Array(float_values))
+        }
+        _ => {
+            // For non-array values marked as float64, raise an error
```
gasmith:
```suggestion
            // For non-array values marked as float64_array, raise an error
```

rust/foxglove/src/websocket/ws_protocol/parameter.rs
```diff
@@ -62,6 +67,191 @@ pub struct Parameter {
...
+                // Convert value based on type
+                let parameter_value = match (r#type.as_ref(), value) {
+                    (Some(ParameterType::Float64), Some(val)) => {
+                        Some(convert_to_float64_value(val).map_err(de::Error::custom)?)
+                    }
+                    (Some(ParameterType::Float64Array), Some(val)) => {
+                        Some(convert_to_float64_array_value(val).map_err(de::Error::custom)?)
+                    }
+                    (Some(ParameterType::ByteArray), Some(val)) => {
+                        Some(convert_to_byte_array_value(val).map_err(de::Error::custom)?)
```
gasmith, line 121: Hm, now that we're explicitly type-checking byte arrays during deserialization, we could introduce a new `ByteArray(Vec<u8>)` variant and get rid of the fallible `ParameterValue::decode_byte_array()` method. Perhaps a job for another day, though.

rust/foxglove/src/websocket/ws_protocol/parameter.rs
```diff
@@ -62,6 +67,191 @@ pub struct Parameter {
...
+                            } else {
+                                return Err("Invalid number in mixed array".to_string());
+                            }
+                        }
+                        _ => {
+                            // Convert non-numeric values normally
+                            let param_val: ParameterValue =
+                                serde_json::from_value(item).map_err(|e| e.to_string())?;
+                            float_values.push(param_val);
+                        }
```
gasmith: Do we need this case, if we've already validated that the item is either i64/f64?

python/foxglove-sdk/src/websocket.rs
```diff
@@ -1059,10 +1066,12 @@ impl<'py> FromPyObject<'py> for ParameterTypeValueConverter {
             let val = val.0;
             let (typ, val) = match val {
                 // If the value is a number, the type is float64.
-                PyParameterValue::Number(_) => (Some(PyParameterType::Float64), val),
+                PyParameterValue::Float64(_) => (Some(PyParameterType::Float64), val),
                 // If the value is an array of numbers, then the type is float64 array.
```
gasmith:
```suggestion
// If the value is a float, the type is float64.
                PyParameterValue::Float64(_) => (Some(PyParameterType::Float64), val),
// If the value is an array of floats, then the type is float64 array.
```

python/foxglove-sdk/python/foxglove/tests/test_parameters.py
```diff
@@ -55,7 +79,7 @@ def test_heterogeneous_array() -> None:
     assert p.value == ParameterValue.Array(
```
gasmith, line 79: I wonder whether we should forbid the user from constructing parameter values that we can't roundtrip.

c/src/parameter.rs
```diff
@@ -841,10 +898,19 @@ impl Drop for FoxgloveParameterValue {
...
 #[unsafe(no_mangle)]
-pub extern "C" fn foxglove_parameter_value_create_number(
+pub extern "C" fn foxglove_parameter_value_create_float64(
     number: f64,
 ) -> *mut FoxgloveParameterValue {
-    FoxgloveParameterValue::number(number).into_raw()
+    FoxgloveParameterValue::float64(number).into_raw()
+}
+
+/// Creates a new  or by passing it to a consuming
```
gasmith: Looks like this docstring got a bit mangled.

gasmith (approved):
I've got a few miscellaneous comments, but overall this looks good.

rust/foxglove/src/websocket/ws_protocol/parameter.rs
ericmlujan: I'm inclined to keep this as a story for another time, given that the scope of this PR is already growing quite a bit beyond what we originally intended

rust/foxglove/src/websocket/ws_protocol/parameter.rs
ericmlujan: Good call, adding an `unreachable!()`

c/include/foxglove-c/foxglove-c.h
```diff
@@ -3257,12 +3281,20 @@ struct foxglove_parameter *foxglove_parameter_clone(const struct foxglove_parame
 void foxglove_parameter_free(struct foxglove_parameter *param);
 
 /**
- * Creates a new number parameter value.
+ * Creates a newumber float64 parameter value.
```
bryfox:
```suggestion
 * Creates a new float64 parameter value.
```

c/src/parameter.rs
```diff
@@ -836,15 +893,26 @@ impl Drop for FoxgloveParameterValue {
     }
 }
 
-/// Creates a new number parameter value.
+/// Creates a newumber float64 parameter value.
```
bryfox:
```suggestion
/// Creates a new float64 parameter value.
```

python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -40,17 +40,17 @@ Used with the parameter service during live visualization. Requires the :py:data
...
 
    .. py:data:: Float64Array
 
-      An array of decimal or integer values that can be represented as 64-bit floating point numbers.
+      An array of floating-point values that can be represented as 64-bit floating point numbers.
 
 .. autoclass:: foxglove.websocket.ParameterValue
 
-   .. py:class:: Number(value: float)
+      .. py:class:: Float(value: float)
```
bryfox: Looks like accidental indentation

python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -40,17 +40,17 @@ Used with the parameter service during live visualization. Requires the :py:data
...
 
-      A decimal or integer value.
+     A floating-point value.
```
bryfox, line 54: Can you please add the Integer type here? I forget why these are manually documented, but we should keep it current.


```rst
   .. py:class:: Integer(value: int)

     An integer value.
```

python/foxglove-sdk/src/websocket.rs
```diff
@@ -809,9 +809,9 @@ impl PyMessageSchema {
 pub enum PyParameterType {
     /// A byte array.
     ByteArray,
-    /// A decimal or integer value that can be represented as a `float64`.
+    /// A floating-point value that can be represented as a `float64`.
```
bryfox, line 812: I'm probably missing some context, but I don't understand why we have a new integer type in `ParameterValue`, but not in `ParameterType`. Maybe right now we're assuming the client originates these values, and we handle the conversion, but it seems like I can't directly construct an integer `Parameter`, even though the interface suggests that I should be able to.

If that's an explicit choice, it's fine, but I think we should document it. And this docstring should probably remain the same.

```py
```

bryfox (commented):
I'm probably missing some context, but I don't understand why we have a new integer type in `ParameterValue`, but not in `ParameterType`. Maybe right now we're assuming the client originates these values, and we handle the conversion, but it seems like I can't directly construct an integer `Parameter`, even though the interface suggests that I should be able to.

If that's an explicit choice, it's OK, but I think we should document it.

ericmlujan:
@bryfox good question. the `type` field here is only really involved with the wire format. It's more a holdover from when the Number parameter type was overloaded to include both integers and floats, and interpretation when deserializing was ambiguous. Setting to `float64`, for example, hints that no matter how you may want to interpret a number, it should be treated as a float. 

If you sent, say, `{"name": "foo", "value": 123}` over the wire, the receiving end would have to decide how to interpret that into a native type. Many JSON libraries would match that to a native integer type. But if you really meant for it to be interpreted as a floating-point anyways, you would send `{"name": "foo", "value": 123, "type": "float64"}`. Going the other way to force a floating-point number to be truncated to an int isn't really something I think we want to support, hence leaving it out.

> It seems like I can't directly construct an integer Parameter, even though the interface suggests that I should be able to

You can now create an integer parameter in your SDK language binding of choice, and it'll be serialized and deserialized appropriately assuming you're staying within the SDK. Folks crafting their own JSON payloads may run into problems (like if they send a value `123.0` over the wire and want to ensure it's treated as a native i64 type), but I'm not sure if that's a corner case we want to design around.

ericmlujan:
Merging after offline discussion to share context with @bryfox

PR #543, opened 2025-07-15
C++: use original `tl` namespace for expected

### LINKED ISSUE #540: Not possible to use normal tl::expected with Foxglove SDK
**Description**

If an application uses the normal tl::expected header, including foxglove sdk will cause compilation errors.

This seems to be because foxglove modifies the tl::expected header file to use a different namespace for some classes; but does not rename the TL macros/defines -- so these conflict when normal tl::expected is also included.

**Steps To Reproduce**
Include normal tl::expected (https://github.com/TartanLlama/expected/blob/master/include/tl/expected.hpp) and foxglove-sdk into any sample project. This yields many compilation errors.

For example:

```
error: no template named 'enable_if_t' in namespace 'foxglove::detail'
using expected_enable_from_other = detail::enable_if_t
```

(because the macros/defines have already been defined for the normal tl namespace)

**Expected Behavior**
Should be able to include foxglove sdk, and include normal tl::expected and compile.

**Suggested fix**
- Either use tl::expected as is; or:
- Rename all TL_* macros/defines to FOXGLOVE_*, and rename the few tl::detail::* declarations to foxglove::detail::*

bryfox:
### Changelog
C++: use original `tl` namespace for expected

### Description

This updates `expected.hpp` to use the original `expected` header file from https://github.com/TartanLlama/expected, to fix naming conflicts when a project includes the same header. See #540.

Previously, when using our own namespace, we also conditionally (in C++23) used `std::expected`. We can decide to re-introduce that later, but for simplicity, let's use the header directly. I left the header in the existing location, to avoid breaking any build setups (including what we document in the quickstart).

To test this, I added the example here, which (at the first commit) fails to build with the error described, but succeeds with the updated header. I don't plan to merge this example; it's intended to help PR review.

Fixes #540.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -230,3 +230,10 @@ target_compile_options(example_auto_serialize PUBLIC ${SANITIZER_COMPILE_OPTIONS
 target_include_directories(example_auto_serialize SYSTEM PRIVATE ${json_SOURCE_DIR}/include)
 target_link_options(example_auto_serialize PUBLIC ${SANITIZER_LINK_OPTIONS})
 target_link_libraries(example_auto_serialize foxglove_cpp_static)
+
+add_executable(example_tl_expected examples/tl-expected/src/main.cpp)
+set_property(TARGET example_tl_expected PROPERTY CXX_STANDARD 17)
+set_property(TARGET example_tl_expected PROPERTY CXX_STANDARD_REQUIRED True)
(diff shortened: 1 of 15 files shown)


eloff (approved):
LGTM

Don't forget to remove the example

bryfox:
> do we want to exclude the vendored header from lint/clang-format?

It's excluded in the make target (`grep -v '/expected.hpp'`); is there a better way to configure this?

PR #544, opened 2025-07-16
C++: Move Timestamp and Duration to schemas namespace

bryfox:
### Changelog
C++: (breaking) Move Timestamp and Duration to schemas namespace

### Description

This moves `Timestamp` and `Duration` schemas to the `foxglove::schemas` namespace, for consistency with other definitions. This is a breaking change for anyone referencing one of these schemas in the prior namespace.


bryfox:
@eloff I thought I'd already tagged you on this, but FYI

PR #545, opened 2025-07-16
Do not send deleted/non-existent params to clients

ericmlujan:
### Changelog
- WebSocket server will no longer send non-existent or deleted parameters in its response body

### Docs
None

### Description
When publishing parameters to clients in response to get_parameter or set_parameter requests, or updates to subscribed parameters, filter out parameters with None values. This brings the SDK implementation in-line with the same observable behavior as [the previous ws-protocol implementation](https://github.com/foxglove/ws-protocol/blob/305d217eec5e38ff8d6b8e8b4f64d9d087193193/cpp/foxglove-websocket/include/foxglove/websocket/websocket_server.hpp#L882-L887).

Part of the motivation here is that, according to [the protocol spec](https://github.com/foxglove/ws-protocol/blob/305d217eec5e38ff8d6b8e8b4f64d9d087193193/cpp/foxglove-websocket/include/foxglove/websocket/websocket_server.hpp#L882-L887):

>  If the value is not set (undefined), the parameter shall be unset (removed).

I found that, if these values are not filtered out, and the WebSocket server sends them to the Foxglove app, they still show up in the Parameters panel with an empty value column instead of being fully deleted:

image

After this change, the parameter is no longer present in the UI:

image

----

Today, this issue is unlikely to come up, because the app currently (1) occasionally polls get_parameters to determine which parameters exist and only requests those and (2) doesn't expose UI elements for deleting parameters. To demonstrate this issue, I had to modify the param_server example to manually publish an empty Parameter. But should either of these change, if we published empty parameter values, the frontend would exhibit unexpected behavior unless it was also modified to ignore parameters from the server with no `value` field.

Alternatively, to avoid this issue without this change, users wanting to use the SDK WebSocketServer to provide parameter functionality would need to be careful in their `on_set_parameters` callback to monitor for deleted parameters and filter them out of the vector of parameters their callback returns. I felt that was also non-ideal and moving it into the SDK provides that safety for free.

This PR also does some light refactoring to eliminate duplicate codepaths for publishing parameters to clients. All parameter publication logic now goes through `ConnectedClient::update_parameters()` to centralize any filtering/message modification we may wish to do in the future.

The SDK server will publish empty parameters in its JSON response

</td><td>

Parameters without value fields are filtered out any time the server publishes them to a client

</td></tr></table>

### DIFF
--- a/rust/foxglove/src/websocket/connected_client.rs
+++ b/rust/foxglove/src/websocket/connected_client.rs
@@ -518,12 +518,7 @@ impl ConnectedClient {
 
     pub fn update_parameters(&self, parameters: Vec<Parameter>, request_id: Option<String>) {
         // Filter out parameters that are not set
-        let parameters: Vec<_> = parameters
-            .into_iter()
-            .filter(|p| p.value.is_some())
-            .collect();
-
-        let mut msg = ParameterValues::new(parameters);
+        let mut msg = ParameterValues::new(parameters.into_iter().filter(|p| p.value.is_some()));
         if let Some(id) = request_id {
             msg = msg.with_id(id);
         }
--- a/rust/foxglove/src/websocket/connected_client.rs
+++ b/rust/foxglove/src/websocket/connected_client.rs
@@ -482,11 +482,7 @@ impl ConnectedClient {
         if let Some(handler) = server.listener() {
             let parameters =
                 handler.on_get_parameters(Client::new(self), param_names, request_id.as_deref());
-            let mut msg = ParameterValues::new(parameters);
-            if let Some(id) = request_id {
-                msg = msg.with_id(id);
-            }
-            self.send_control_msg(&msg);
+            self.update_parameters(parameters, request_id);
         }
     }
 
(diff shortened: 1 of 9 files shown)


rust/foxglove/src/websocket/connected_client.rs
```diff
@@ -519,8 +516,19 @@ impl ConnectedClient {
...
     }
 
-    pub fn update_parameters(&self, parameters: Vec<Parameter>) {
-        self.send_control_msg(&ParameterValues::new(parameters));
+    pub fn update_parameters(&self, parameters: Vec<Parameter>, request_id: Option<String>) {
+        // Filter out parameters that are not set
+        let parameters: Vec<_> = parameters
+            .into_iter()
+            .filter(|p| p.value.is_some())
+            .collect();
```
gasmith: Nit: Could avoid the `.collect()` since `ParameterValues::new()` accepts an `impl IntoIterator`.

ericmlujan:
@bryfox @gasmith good call on the documentation. Added it in a few places in f661eff.

PR #546, opened 2025-07-17
ROS: Build packages with docker

amacneil:
### Changelog
None

### Docs
None

### Description
In the process of exploring how to import the `foxglove_bridge` package to this repo (https://github.com/foxglove/foxglove-sdk/pull/547), I discovered that this repo doesn't have a good local development workflow for ROS.

This PR doesn't make any changes other than updating our directory structure and moving CI into docker.

**Changes**

- Migrate `foxglove_msgs` ros package under `ros/src` so that the `ros` directory can be used as a ros workspace (with sibling `build`, `install`, `log` subdirectories in gitignore)
  - ROS build farm will search the whole repo for `package.xml` files so it doesn't matter that we moved the directory
- Add Dockerfile for development and CI
- Replace ros-tooling github actions with our Dockerfile
- Removed `noetic` and `foxy` from CI since they are EOL, and add `kilted`

Need to update required status checks before merging.


ros/src/foxglove_msgs/.gitignore
```diff
@@ -0,0 +1 @@
+/msg/
```
ericmlujan, line 1: Just curious as to why these .gitignores are being introduced as separate files, and not lines in the top-level .gitignore?

ericmlujan (approved):
LGTM!

ros/src/foxglove_msgs/.gitignore
amacneil: no real reason, this one was already here I just gave it a `/` prefix. this particular one is going to go away once we drop support for ros1 which I intend to propose next.

PR #549, opened 2025-07-18
ROS: Remove `foxglove_msgs` support for ROS 1

amacneil:
### Changelog
foxglove_msgs: remove support for ROS 1

### Docs
None

### Description
Remove support for ROS 1 from `foxglove_msgs` package.

ROS 1 is EOL so new versions of any ROS 1 packages can no longer be published. The current latest version ([v3.1.0](https://index.ros.org/p/foxglove_msgs/#noetic)) will remain available to ROS 1 users, but we do not need to keep the source around in main.

In theory, we could continue to support an install from source workflow for ROS 1 users if we release new schemas, but (a) we do not create new schemas very often), and (b) this is additional overhead for a completely unsupported framework.

The path forward if users are still on ROS 1 and need access to future schemas we release, is that they can vendor our schemas into their repo. Likely they will already be doing that with many other packages too since all ROS 1 packages are now unmaintained. I have not removed the ROS 1 schemas from the `schemas` directory, so they are still being generated and are available to manually import into a project if needed.


defunctzombie:
I'm generally 👎 on this change for the following:

1. The maintenance burden with this being present seems minimal or zero (that I am aware of) while reviewing this change and responding to any breakage is potential work.
2. Having this as an easy copy/paste folder seems fine for users that are still on ROS 1
3. Having this present is not blocking our ability to add features or change schemas; if that changes then we could decide there is a purpose to removing.

So basically this for me comes down to not doing anything here is the easier path for us for now.

amacneil:
Closing for now; we can revisit if/when there is a breaking change that makes ROS 1 harder to maintain (e.g. https://github.com/foxglove/foxglove-sdk/pull/703#pullrequestreview-3346940587)

Issue #526, opened 2025-07-07
Crashing C++ Examples


laelliott:
Compiling the examples in `/foxglove-sdk/cpp/examples` is leading to rust pointer problems. I've attached the rust stack trace at the end.

`example_mcap` and `example_foxglove_schemas` both exhibit issues.

It seems that adding ` options.profile = std::string("default");`  to the `  foxglove::McapWriterOptions options = {};
` solves the issue.

- Version: Main (last commit 5bc6979ac63341e37e0f323376869c83d3584462)
- Platform: Ubuntu 24.04, g++ (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0

**Steps To Reproduce**
No extra steps beyond c++ compilation needed

**Expected Behavior**
Proper writing of mcap files and/or execution of the examples

```
thread '<unnamed>' panicked at library/core/src/panicking.rs:226:5:
unsafe precondition(s) violated: slice::from_raw_parts requires the pointer to be aligned and non-null, and the total size of the slice not to exceed `isize::MAX`

This indicates a bug in the program. This Undefined Behavior check is optional, and cannot be relied on for safety.
stack backtrace:
   0:     0x57d1493c4cf2 - std::backtrace_rs::backtrace::libunwind::trace::h74680e970b6e0712
                               at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/../../backtrace/src/backtrace/libunwind.rs:117:9
   1:     0x57d1493c4cf2 - std::backtrace_rs::backtrace::trace_unsynchronized::ha3bf590e3565a312
                               at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/../../backtrace/src/backtrace/mod.rs:66:14
   2:     0x57d1493c4cf2 - std::sys::backtrace::_print_fmt::hcf16024cbdd6c458
                               at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/sys/backtrace.rs:66:9
   3:     0x57d1493c4cf2 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h46a716bba2450163
                               at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/sys/backtrace.rs:39:26
   4:     0x57d149411593 - core::fmt::rt::Argument::fmt::ha695e732309707b7
[...]

PR #551, opened 2025-07-21
C FFI: Add explicit null check for FoxgloveString

bryfox:
### Changelog
C++: Fix a precondition assertion from the Rust library which could panic in Debug configurations

### Description

This adds an explicit null check for string data in `as_utf8_str`. Without this, Rust's precondition assertion would panic in Debug targets (`-DCMAKE_BUILD_TYPE=Debug`) as described in #526. See [docs](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#ffi-handling-null-pointers). This also adds similar checks for other public methods using `slice.from_raw_parts`.

This will now also test a Debug build of the C++ SDK in CI.

### DIFF
--- a/cpp/Makefile
+++ b/cpp/Makefile
@@ -11,6 +11,8 @@ BUILD_DIR=build-sanitize-$(subst $(comma),-,$(SANITIZE))
 $(info Using sanitizer: $(SANITIZE))
 endif
 
+CMAKE_BUILD_TYPE?=RelWithDebInfo
+
 .PHONY: lint
 lint:
 	find . -path './build*' -prune -name '*.h' -o -name '*.hpp' -o -name '*.cpp' |\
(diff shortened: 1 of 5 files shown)


peter-lyons-kehl:
If `self.data` is `null`, but `self.len` is non-zero, should this function return an `Err(whatever_appropriate)`?

Otherwise (as-is), we can also minimize `unsafe {...}` surface by returning:
```rust
    if self.data.is_null() {
       Ok("")
    } else {
       std::str::from_utf8(unsafe {
           std::slice::from_raw_parts(self.data.cast(), self.len)
       })
    }
```

bryfox:
> If `self.data` is `null`, but `self.len` is non-zero, should this function return an `Err(whatever_appropriate)`?
> 
> Otherwise (as-is), we can also minimize `unsafe {...}` surface

Returning an error would make it much less convenient to work with empty `string_view`s, so kept the current behavior and applied your second suggestion — thanks!

cpp/Makefile
```diff
@@ -11,6 +11,8 @@ BUILD_DIR=build-sanitize-$(subst $(comma),-,$(SANITIZE))
 $(info Using sanitizer: $(SANITIZE))
 endif
 
+CMAKE_BUILD_TYPE?=RelWithDebInfo
```
eloff, line 14: In the debug build, should it also use a debug rust build with the extra checks and assertions that gives us?

bryfox: I filed a follow-up issue to improve the debug tests in CI; we'll also need to fix the Windows configuration there (FG-12519). I don't want to hold up this fix for that.

PR #552, opened 2025-07-24
data loaders: add C/C++  framework

james-rms:
### Description

Adds a C++ SDK wrapper for custom data loaders. This is structured as an include-only library, which can be copied into a user's C++ build system. The loader must be built using the WASI SDK, the README explains how to install and use it.

Adds an example C++ data loader into the SDK repo as well.

</td><td>

</td></tr></table>


cpp/foxglove-data-loader/cpp/bindings/data_loader.hpp
```diff
@@ -0,0 +1,189 @@
...
+{
+public:
+  SchemaId id;
+  /** A name that identifies the 'type' that this schema describes.  */
+  std::string name;
+  /** The encoding used to encode the schema definition into `data`. Must match
+   * one of the well-known schema encodings here: https://mcap.dev/spec/registry
+   */
+  std::string encoding;
+  void *data;
```
bennetthardwick: Why is this not `uint8_t` like message?

cpp/foxglove-data-loader/cpp/bindings/data_loader.hpp
```diff
@@ -0,0 +1,189 @@
...
+    /** Get the latest message before the requested `time` for each channel. */
+    virtual Result<std::vector<Message>> get_backfill(const BackfillArgs& args) = 0;
+    virtual ~AbstractDataLoader() {}
+};
+
+} // namespace foxglove_data_loader
+
+/** Constructs a new data loader for the given paths.
+ *
+ * Your data loader module must implement this function.
```
james-rms: need to make sure they return a pointer allocated with new() ,note that it will be deleted later.

cpp/foxglove-data-loader/cpp/bindings/data_loader.hpp
```diff
@@ -0,0 +1,189 @@
+
```
achim-k: Could add an include guard here, e.g. `#pragma once`. But I guess this file is included only once, so not critical

cpp/foxglove-data-loader/cpp/bindings/data_loader.hpp
```diff
@@ -0,0 +1,189 @@
...
+typedef uint64_t TimeNanos;
+
+/**
+ * @brief metadata about a channel of messages. Your data loader reads input files
+ * and produces messages on one or more logical channels, which may differ in topic name,
+ * message encoding, or message definition schema information.
+ */
+class Channel
+{
+public:
```
achim-k: If you only have public member, why not make this a struct like most other types? Also applies to `Schema` below

cpp/foxglove-data-loader/cpp/bindings/data_loader.hpp
```diff
@@ -0,0 +1,189 @@
+
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace foxglove_data_loader {
```
achim-k: Should we nest namespaces? 
```suggestion
namespace foxglove::data_loader {
```

cpp/foxglove-data-loader/cpp/bindings/data_loader.hpp
```diff
@@ -0,0 +1,189 @@
...
+    /** Constructs a new error-valued Result with a message. */
+    static Result<T> error_with_message(std::string message) {
+        return Result<T> {
+            .value = std::nullopt,
+            .error = message,
+        };
+    }
+
+    /** Retrieves a reference to the value, returning a null reference if this is an error result. */
+    T& get() {
```
achim-k: Marking these methods as const would allow you to pass the Result type as a const ref
```suggestion
    const T& get() const {
```

cpp/foxglove-data-loader/cpp/bindings/data_loader.hpp
```diff
@@ -0,0 +1,189 @@
...
+        return value.value();
+    }
+
+    /** Returns true if this is an OK result, false if error. */
+    bool ok() {
```
achim-k:
```suggestion
    bool ok() const {
```

cpp/foxglove-data-loader/cpp/bindings/data_loader.hpp
```diff
@@ -0,0 +1,189 @@
...
+  /** The encoding used to encode the schema definition into `data`. Must match
+   * one of the well-known schema encodings here: https://mcap.dev/spec/registry
+   */
+  std::string encoding;
+  void *data;
+  size_t data_len;
+};
+
+/** An inclusive time range. */
+struct TimeRange {
```
achim-k: code style is inconsistent. Here you have the opening bracket on the same line whereas above it is on the next line

cpp/foxglove-data-loader/cpp/bindings/data_loader.hpp
```diff
@@ -0,0 +1,189 @@
...
+};
+
+struct MessageIteratorArgs
+{
+  /** if non-empty, only messages on or after this log time should be yielded. */
+  std::optional<TimeNanos> start_time;
+  /** if non-empty, only messages on or before should be yielded. */
+  std::optional<TimeNanos> end_time;
+  /** Yield only messages with these channel IDs. */
+  std::vector<ChannelId> channel_ids;
```
achim-k: maybe move that to the top? Looks weird that the optional field are listed first

cpp/foxglove-data-loader/cpp/bindings/data_loader.hpp
```diff
@@ -0,0 +1,189 @@
...
+};
+
+struct BackfillArgs
+{
+  /** For every given channel ID, Retrieve the latest message available in the file(s) for that
+   * channel that has log_time before this timestamp.
```
achim-k:
```suggestion
   * channel that has log_time before or at this timestamp.
```

cpp/foxglove-data-loader/cpp/bindings/data_loader.hpp
james-rms: /shrug i don't really want to force a relationship where there isn't one - there's no code shared, I don't see a strong reason to put them together. You could make the case if you wanted, I don't have a strong opinion.

achim-k: Yeah, not really important

cpp/foxglove-data-loader/examples/data_loader.cpp
```diff
@@ -0,0 +1,199 @@
+#define FOXGLOVE_DATA_LOADER_IMPLEMENTATION
+#include "foxglove_data_loader/data_loader.hpp"
+
+using namespace foxglove_data_loader;
```
amacneil: Should top level folder use snake case too? To match the namespace. That’s what packages in the repo do (at least rust packages do, cpp doesn’t have a pattern yet)

james-rms: :+1: why not.

cpp/foxglove-data-loader/include/foxglove_data_loader/data_loader.hpp
```diff
@@ -0,0 +1,222 @@
...
+
+/** Constructs a new data loader for the given arguments. This function should return a pointer
+ * to your derived data loader constructed with the `new` keyword. This object will be cleaned
+ * up with `delete` at the end of its lifetime.
+ *
+ * Your data loader module must implement this function.
+ */
+foxglove_data_loader::AbstractDataLoader* construct_data_loader(
+  const foxglove_data_loader::DataLoaderArgs& args
+);
```
Muon: Why not offer a `destroy_data_loader()` instead of mandating the use of global `new`/`delete`?

cpp/foxglove-data-loader/include/foxglove_data_loader/data_loader.hpp
```diff
@@ -0,0 +1,222 @@
+
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace foxglove_data_loader {
+
+const char* SDK_VERSION = "0.1.0";
```
Muon: `constexpr const char*` or `const char* const` unless this is intended to be reassignable.

cpp/foxglove-data-loader/include/foxglove_data_loader/data_loader.hpp
james-rms: mainly because:
1. I don't really see a concrete use-case for not using new - if the default heap allocation is unacceptable, you can implement new() yourself
2. If we do this for the data loader, we'd also have to do it for the message iterator, does the message iterator destroy() function need to be a method on AbstractDataLoader? how do we make sure you have enough state in scope to do this properly? I just think it's a lot of additional rabbit-hole for a pretty shallow theoretical gain.

Muon: Okay, in which case, why not return `std::unique_ptr<AbstractDataLoader>` instead? That inherently makes clear both how the allocation is to be done and who owns the result.

james-rms: because we leak this value out of the guest entirely across calls. The owning pointer lives with the host, they pass it in to call methods, and they can call the function that deletes it within the guest at any point. To require an unique_ptr from `construct_` would be to _pretend_ that something in the guest owns the allocation when it actually doesn't.

Muon: I'm not sure that's true. All that returning a `unique_ptr` says is "the caller owns this".  The caller is most certainly free to do whatever it likes with the `unique_ptr` afterward.

cpp/foxglove_data_loader/examples/data_loader.cpp
```diff
@@ -0,0 +1,236 @@
...
+  std::vector<std::string> paths;
+  std::vector<std::vector<uint8_t>> files;
+  std::vector<LineIndex> file_line_indexes;
+  std::vector<size_t> file_line_counts;
+
+  TextDataLoader(std::vector<std::string> paths);
+
+  Result<Initialization> initialize() override;
+
+  Result<AbstractMessageIterator*> create_iterator(const MessageIteratorArgs& args) override;
```
Muon: This should probably return `Result<std::unique_ptr<AbstractMessageIterator>>` too.

cpp/foxglove_data_loader/include/foxglove_data_loader/host_internal.inl
```diff
@@ -194,17 +200,23 @@ __wasm_export_exports_foxglove_loader_loader_method_data_loader_get_backfill_pos
       if (len0 > 0) {
         uint8_t* ptr1 = *((uint8_t**)(arg0 + sizeof(void*)));
         for (size_t i2 = 0; i2 < len0; i2++) {
```
Muon: I'm not sure what the generated code is doing here now. Also this loop now has an empty body?

james-rms: get_backfill returns a dynamically sized array of messages, that array has to be freed. The loop has an empty body, true, i'll get rid of that.

cpp/foxglove_data_loader/include/foxglove_data_loader/data_loader.inl
```diff
@@ -0,0 +1,279 @@
...
+    ret->val.ok.channel_id = msg.channel_id;
+    ret->val.ok.log_time = msg.log_time;
+    ret->val.ok.publish_time = msg.publish_time;
+    ret->val.ok.data.len = msg.data.len;
+    // NOTE: normally the WIT-generated wrapper code would require us to copy the message data
+    // to a new allocation, and it would free() that allocation in the post-return hook. We're not
+    // abiding by the WIT component model ABI here, so we can choose to avoid that copy. The
+    // generated code in `host_internal.inl` has been modified to remove the corresponding free()
+    // call.
+    ret->val.ok.data.ptr = msg.data.ptr;
```
Muon, line 99: I'm not 100% sure what the implications are here. Does the ABI matter to anyone other than us, or is this purely internal?

james-rms: The implications are that if we one day turn this into a Component and try to link this with other Components using shared-everything linking, something very bad will happen. https://github.com/WebAssembly/component-model/blob/main/design/mvp/examples/SharedEverythingDynamicLinking.md

As discussed earlier today, I think generally trying to enter into this world of linked WASM components would be bad for our software, and I want to say we're not going to do that.

PR #554, opened 2025-07-24
Add notes & check for cbindgen

bryfox:
### Changelog
None
### Docs
None

### Description

When creating a new schema, a developer has to add an export-rename definition to the C library's cbindgen configuration. If this is missed, the `generate` task will end up failing with C++ compilation errors. (This is only as of #553, which added the build step to `generate`. However, without that, updated C headers won't have been generated at all and CI will fail.)

This PR adds a check and prints a helpful error message about what steps to take.

Note that cbindgen.toml isn't generated, and it contains comments which aren't preserved during a round-trip through toml parsing/serializing.

Alternatives considered:

- generate cbindgen.toml by concatenating hand-written toml (from a separate file) with the generated renames
- modify the existing toml in-place, discarding comments and multi-line string formatting, leaving it a mix of hand-written and generated

I think the first alternative is viable, if potentially confusing; it does save one step when authoring a schema.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,3 +6,4 @@ python/foxglove-sdk/src/generated/* linguist-generated
 rust/foxglove/src/schemas/foxglove.rs linguist-generated
 schemas/** linguist-generated
 typescript/schemas/src/types/* linguist-generated
+c/include/foxglove-c/foxglove-c.h linguist-generated


bryfox:
I think I am going to explore the alternative; the outcomes here are either throwing an error to go hand-edit some toml, or to see C++ build errors, both of which are terrible UX for a generator script.

PR #555, opened 2025-07-24
Update cbindgen for generated types

bryfox:
### Changelog
None

### Docs
None

### Description

This PR makes cbindgen config part of the `generate` script, so that all generated types automatically have their types exposed correctly in the C headers.

Today, when a developer adds a new schema, they'd have to manually add this rename mapping to cbindgen.toml in addition to writing the typescript definition. I considered improving the docs & error messaging (#554) but I think we should just make this the purview of `generate`.

Any manual cbindgen config should now be added to the 'prelude' toml file.

This also removes the full C++ build from the generate script in favor of the C build; if something goes wrong after generation, CI will catch the error during C++ builds as usual.

Added dependency: [smol-toml](https://www.npmjs.com/package/smol-toml)


gasmith (approved):
I love it! Thanks for doing this.

PR #556, opened 2025-07-24
Add `LocationFixes` schema; color `LocationFix`

bryfox:
### Changelog
Add a new `LocationFixes` schema for visualizing a series of fixes on the Map panel, and a `Color` field to `LocationFix`

### Docs
None

### Description

This adds a `Color` field to `LocationFix` and adds a new `LocationFixes` schema. `LocationFixes` contains an array of `LocationFix` messages.

There are a couple of use cases driving this change:

- A system tracking multiple entity locations publishes a `LocationFixes` message containing a location fix for each entity at a given timestamp. Each location's color can be used to differentiate when visualized.
- Render a history of locations (track). Each fix has its own timestamp.

This array schema is reminiscent of the `FrameTransforms` schema.

Alternatives considered: Instead of adding Color, assign some sort of group ID to location fix, which would allow the app to determine a color at render time.


ros/src/foxglove_msgs/ros1/LocationFixes.msg
```diff
@@ -0,0 +1,10 @@
+# foxglove_msgs/LocationFixes
+# A series of LocationFix messages
+
+# Generated by https://github.com/foxglove/foxglove-sdk
+
+# A series of location fixes
+foxglove_msgs/LocationFix[] fixes
+
+# Color used to visualize this series
+foxglove_msgs/Color color
```
amacneil: the color needs to be added to the `LocationFix` message though. because each track in the array will need its own color

bryfox: OK, I understood a series of points ("track") to be represented by `LocationFixes`. To log multiple tracks on a topic, you'd write multiple `LocationFixes` messages, each with its own color.

I think you're saying that there's no need to impose those semantics; a user just dumps everything into the same `LocationFixes` array. That lets you log a single message, but has the overhead of duplicated colors (and other metadata like track IDs if we ever added it). Encapsulating track points within a single message also seems more natural to me, instead of treating color values as discriminators. But maybe a "bucket of points" is more flexible?

jtbandes: If we wanted to draw lines between LocationFixes in a single track/series, seems like dumping them all into one message would make it hard to know which pieces are connected and which aren't..... not necessarily saying you _shouldn't_ be able to put them all in one message if you want to though

typescript/schemas/src/internal/schemas.ts
```diff
@@ -1367,6 +1367,26 @@ const LocationFix: FoxgloveMessageSchema = {
...
+const LocationFixes: FoxgloveMessageSchema = {
+  type: "message",
+  name: "LocationFixes",
+  description: "A group of LocationFix messages",
+  fields: [
+    {
+      name: "fixes",
+      type: { type: "nested", schema: LocationFix },
+      array: true,
+      description: "One or more location fixes",
```
jtbandes: do you feel one or more actually matters? I can tell you for sure that some people are going to try putting zero in 😄

jtbandes:
Change looks fine to me as long as relevant stakeholders are happy with the results of schema bikeshedding.

PR #558, opened 2025-07-26
Move ros2-foxglove-bridge into SDK repo

ericmlujan:
### Changelog
- [foxglove_bridge] A public beta of `foxglove-bridge` built using the Foxglove SDK is now available.

### Docs
- README from foxglove/ros-foxglove-bridge is migrated from foxglove/ros-foxglove-bridge with new build instructions and eliding ROS 1

### Description
Move source code from `ros2_foxglove_bridge_sdk` from foxglove/ros-foxglove-bridge. Some directories renamed to remove `_sdk` suffix. Build files stripped down to remove ROS1 references and conditionals for switching between old foxglove-ws implementation and Foxglove SDK. 

New build targets stood up to simplify both customer and developer build experience (tried to get as close to the "clone repo and run `make`" North Star as possible). Documentation reworded to reflect new build processes, Foxglove SDK as the transport implementation, and add caveats about public beta.

Unit tests pass, and CI runs successfully. Tested that I could follow the instructions in the `README` for building, both in Docker on a Mac and in a fresh ROS installation/SDK repo clone in Ubuntu, and for both prebuilt SDK binaries and in-tree development builds.

Fixes: FG-9777


Dockerfile.ros
```diff
@@ -1,16 +1,52 @@
...
 # create foxglove user
 ARG USERNAME=foxglove
 ARG USER_UID=1005
 ARG USER_GID=$USER_UID
-RUN groupadd --gid $USER_GID $USERNAME \
-    && useradd --uid $USER_UID --gid $USER_GID -m $USERNAME \
-    && echo "$USERNAME ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers.d/$USERNAME
-USER $USERNAME
+
+# If the user ID and group ID already exist, don't create a new user and instead use the existing user
```
amacneil, line 33: when would this be true?

Dockerfile.ros
```diff
@@ -1,16 +1,52 @@
...
+
+COPY ./cpp /sdk/cpp
+COPY ./c /sdk/c
+COPY ./rust /sdk/rust
+# REVIEW: Is it possible to do this without copying over Python?
+COPY ./python /sdk/python
+COPY ./Cargo.toml /sdk/Cargo.toml
+
+WORKDIR /sdk/cpp
+RUN make
```
amacneil, line 21: I'd probably avoid copying all this into the container and building it? the makefile setup was designed that the container is only a build environment and is not where actual building happens

Dockerfile.ros
```diff
@@ -1,16 +1,52 @@
...
+    fi
+
+RUN echo "$(id -nu $USER_UID) ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers.d/$(id -nu $USER_UID)
+
+USER $USER_UID
+
+# Copy Foxglove SDK build artifacts
+COPY --from=build-sdk --chown=$USER_UID:$USER_GID /sdk/cpp /sdk/cpp
+
+RUN . /opt/ros/$ROS_DISTRO/setup.sh
```
amacneil, line 49: I'm pretty sure this does nothing? I don't think you can source bash files in a dockerfile since this is running in a subshell, if it's important it would need to live in an entrypoint script

ros/src/foxglove_bridge/package.xml
```diff
@@ -0,0 +1,57 @@
...
+    <description>ROS Foxglove Bridge</description>
+    <license>MIT</license>
+    <url type="website">https://github.com/foxglove/ros-foxglove-bridge</url>
+    <author email="ros-tooling+foxglove_bridge@foxglove.dev">Foxglove</author>
+    <maintainer email="ros-tooling+foxglove_bridge@foxglove.dev">Foxglove</maintainer>
+
+    <buildtool_depend condition="$ROS_VERSION == 1">catkin</buildtool_depend>
+    <buildtool_depend condition="$ROS_VERSION == 2">ament_cmake</buildtool_depend>
+
+    <!-- ROS 1 runtime dependencies -->
```
amacneil, line 14: some ROS 1 stuff here that could be cleaned up, or it can be done in a follow up PR

ros/Makefile
```diff
@@ -1,7 +1,15 @@
+ROS_DISTRIBUTIONS := humble iron jazzy kilted rolling
```
amacneil, line 1: `iron` is EOL, let's drop that

ros/Makefile
```diff
@@ -1,7 +1,15 @@
...
+USE_LOCAL_PREBUILT_SDK := OFF
+BUILD_SDK := OFF
+
+# Automatically set USE_LOCAL_PREBUILT_SDK=ON if BUILD_SDK=ON
```
amacneil, line 5: I don't really understand why they are separate flags, could they be one flag?

ros/Makefile
```diff
@@ -14,31 +22,75 @@ deps:
...
-
 define generate_ros_targets
-.PHONY: docker-build-$(1)
-docker-build-$(1):
+.PHONY: docker-build-container-$(1)
+docker-build-container-$(1):
 	docker build -t foxglove-sdk-ros-$(1) -f ../Dockerfile.ros \
 		--build-arg ROS_DISTRIBUTION=$(1) \
+		--build-arg USER_UID=$(id -u) \
+		--build-arg USER_GID=$(id -g) \
```
amacneil, line 60: FYI I didn't do this in the Makefile because it doesn't work on mac, where the uid is like 501 which is invalid on linux. Also on mac you don't need to override the uid anyway.

So I guess at the very least I would not pass these params if the environment is mac.

ros/Makefile
```diff
@@ -14,31 +22,75 @@ deps:
...
+endif
+
+ifeq ($(USE_LOCAL_PREBUILT_SDK), ON)
+	$(info Using pre-built Foxglove C++ SDK binaries from local source tree)
+else
+	$(info Build will use pre-built Foxglove C++ SDK binaries)
+endif
+
+.PHONY: build-bridge
+build-bridge: cpp-sdk
```
amacneil, line 46: is there value in separating out the build bridge vs build msgs as separate make targets? people can always run colcon manually if they want, so it might be easier if our helper make targets just build all ros packages in the repo together?

Dockerfile.ros
ericmlujan: This is true if you're building in a Linux environment because:

1. The docker user's UID/GID must be the same as the user running the docker command on the host system. If they don't match, the `colcon` process inside the container will fail to write any files. This is why we had to [add the UID and GID as build args](https://github.com/foxglove/foxglove-sdk/blob/6a4f954108ecba544430eed12d6721f60260b597/.github/workflows/ros.yml#L26) in CI on `main`.
2. Now that the UID/GID in docker == that of the host user, if the host user's UID is 1000 (as is the case on a fresh ubuntu install with a single user), this will fail because the ROS docker image has a user with that UID. 

This worked on my Mac just fine because macOS assigns UIDs starting from 500, but Ubuntu starts from 1000. I only discovered it when doing a manual test of freshly checking out the repo and attempting to build the bridge 😬

Dockerfile.ros
ericmlujan: Ya I've had a few iterations on this, but you're right. My original strategy was to not mount the full repo as a Docker volume, and instead have the ROS Docker container build the SDK and guarantee that it's available. But now that I've walked that back, it's worth revisiting this and seeing if there's a more minimalistic approach.

ros/src/foxglove_bridge/package.xml
ericmlujan: May as well 🔪 now

ros/Makefile
ericmlujan: 👍 I was erring on the side of more granular build targets, but agree that leaving it to power users to target their colcon builds is better from a maintainability standpoint

ros/Makefile
ericmlujan: I'll iterate on this. I think _some_ UID deconfliction logic will be needed to deal with the case that I outlined above with Linux hosts

ericmlujan:
Superseded by #564 and #565

PR #559, opened 2025-07-28
Deploy foxglove data loader crate

bennetthardwick:
### Changelog

None

### Docs

None

### Description

Adds a workflow to deploy the `foxglove_data_loader` crate. This avoids extensions needing to use a git dependency. I've set the version to `0.0.1` but we could use a `-beta` version instead.

This will be released separately on the `rust/foxglove_data_loader/v..` tag.

Fixes FG-12509.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,13 +83,25 @@ jobs:
     runs-on: ubuntu-latest
     needs: rust
     timeout-minutes: 10
+    env:
+      github_ref: ${{ github.ref }}
     steps:
       - uses: actions/checkout@v4
       - id: detect-version
         name: Detect local Cargo version
         run: |
+          set -eu
+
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -78,6 +78,23 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: rust
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+      - id: detect-version
+        name: Detect local Cargo version
+        run: |
+          version=$(cargo metadata --no-deps --format-version 1 | jq -r '.packages[] | select(.name=="'foxglove_data_loader'").version')
+          echo "version=$version" >> $GITHUB_OUTPUT
```
amacneil, line 96: might be good to check and throw an error if the package version doesn't match the git ref

amacneil (approved):
I didn't test the workflow, but agree it makes sense to version this separately from the main sdk, and the tag pattern `rust/foxglove_data_loader/v` makes sense.

PR #560, opened 2025-07-28
Support wss connections for bridge

bryfox:
### Changelog
TBD

### Docs
None

### Description

This adds initial support for secure websocket connections (wss) in a form that supports the ros-foxglove-bridge interface. Support is behind an undocumented "tls" feature. We plan to expose this in the bridge prior to SDK users in general. This does not contain any support for Python.

The server accepts the contents of a PEM-formatted x509 certificate and a PEM-formatted pkcs8 key. I tested this with a 4096-bit RSA key and self-signed certificate, by trusting the cert in my keychain, using the examples (which take file path arguments).

Under the hood, this uses rustls. Currently, it also uses rustls-native-certs; TBD what we need to support here.


gasmith:
For another example, see [reqwest](https://docs.rs/reqwest/latest/reqwest/#optional-features), which offers `rustls-tls` as an alias for `rustls-tls-webpki-roots`.

c/src/lib.rs
```diff
@@ -546,6 +555,16 @@ unsafe fn do_foxglove_server_start(
         let context = ManuallyDrop::new(unsafe { Arc::from_raw(options.context) });
         server = server.context(&context);
     }
+    if !options.tls_cert.is_null() || !options.tls_key.is_null() {
```
ericmlujan, line 558: If I read this correctly, the desired behavior is that TLS will be enabled and inputs enforced if either `tls_cert` or `tls_key` are not null. I'm wondering if we want to instead have an explicit `enable_tls` option rather than this being implicit. Thoughts?

c/src/lib.rs
```diff
@@ -546,6 +555,16 @@ unsafe fn do_foxglove_server_start(
...
+        if options.tls_cert.is_null() || options.tls_key.is_null() {
```
ericmlujan, line 559: Should we also check for 0 length?

cpp/foxglove/include/foxglove/server.hpp
```diff
@@ -185,6 +185,18 @@ struct WebSocketServerCallbacks {
   std::function<void()> onConnectionGraphUnsubscribe;
 };
 
+/// @brief TLS configuration for a WebSocket server.
+struct TlsIdentity {
+  /// @brief PEM-formatted x509 certificate for the server.
+  const std::byte* cert = nullptr;
```
ericmlujan: In C++ land, I wonder if these should be `std::vector<std::byte>` rather than the C-style pointer-and-length representation? In the example, we use vectors when reading from the file, then use `.data()` and `.size()` to go to the C-like version. Could add a bit of boilerplate that may not be necessary. This is also similar in spirit to the Rust API version, which seems to use Vec<u8>.

cpp/foxglove/src/server.cpp
```diff
@@ -257,6 +257,13 @@ FoxgloveResult<WebSocketServer> WebSocketServer::create(
     };
   }
 
+  if (options.tls_identity) {
+    c_options.tls_cert = reinterpret_cast<const uint8_t*>(options.tls_identity->cert);
```
ericmlujan: If using vectors, this is where we'd convert to the C-style representation (e.g. `tls_cert = reinterpret_cast<const uint8_t*>(options.tls_identity->cert.data()`).

rust/foxglove/src/websocket/server_stream.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+impl<S: AsyncRead + AsyncWrite + Unpin> AsyncRead for ServerStream<S> {
+    fn poll_read(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut ReadBuf<'_>,
+    ) -> Poll<std::io::Result<()>> {
+        match self.get_mut() {
+            ServerStream::Plain(ref mut s) => Pin::new(s).poll_read(cx, buf),
```
ericmlujan: Possible dumb rust question, why does one arm match a mutable reference and the other a value type?

rust/foxglove/src/websocket/server.rs
```diff
@@ -155,7 +215,12 @@ impl Server {
             .unwrap_or_default()
     }
 
-    fn new(weak_self: Weak<Self>, ctx: &Arc<Context>, opts: ServerOptions) -> Self {
+    fn new(
+        weak_self: Weak<Self>,
+        ctx: &Arc<Context>,
+        opts: ServerOptions,
+        #[cfg(feature = "tls")] tls_acceptor: Option<TlsAcceptor>,
```
ericmlujan: I'm wondering if there's a way to constrain the number of sites that we have to put in this attribute (not a Rust expert, so I'm maybe less helpful here than I should be :P). 

But general thought process is that if the `tls_acceptor` member is something generic that's isn't directly dependent on `rusttls`, and substituted for a dummy type where TLS isn't available, we could localize the `#cfg[]` guards to just the implementation of the type for when TLS _is_ available. I'd be super interested if someone more experienced with Rust than I am has ideas about the most idiomatic way to accomplish this.

rust/foxglove/src/lib.rs
```diff
@@ -413,6 +413,9 @@ pub enum FoxgloveError {
     /// An error occurred while encoding a message.
     #[error("Encoding error: {0}")]
     EncodeError(String),
+    /// An error related to configuration
```
ericmlujan, line 416: Do we want to make this a generic error type that captures all invalid configurations? What error do we currently raise when config is invalid? Or is it better to introduce a more specific TLSError type? Could see an argument for both, just wanted to get your thoughts

c/src/lib.rs
bryfox: I don't have a strong preference either way. My view is that the C interface is mostly internal (in that it's undocumented), and it's simpler to have one less flag to think about. In the C++ interface, this is encapsulated as a single option, and I think its presence/absence suffices — unless you think there's a case where leaving some configuration in place but having a quick "enabled" toggle is helpful.

cpp/foxglove/include/foxglove/server.hpp
bryfox: Definitely; I guess I was coming from the C side and didn't give it enough thought. Updated.

rust/foxglove/src/lib.rs
bryfox: I started with something like "TLSError". I can't think of a case where you'd want to handle that differently than other configuration errors, and I thought the error messaging should be enough to understand the context. Also, if we someday had an error interface for handshakes, I would probably favor a different error type there; though "TLSConfigurationError" is certainly an option.

I almost want to re-use `ValueError` for this as it's really a problem with the provided key or cert. How would you feel about that? I guess I'm leaning towards less-specific error types until we identify a strong need for them.

rust/foxglove/src/websocket/server.rs
bryfox: Yeah, this is a good question and I explored it a little bit but wound up here. Maybe @eloff has thoughts.

rust/foxglove/src/websocket/server_stream.rs
bryfox: The short answer is that this is implementation comes from [MaybeTlsStream in tokio-tungstenite](https://github.com/snapview/tokio-tungstenite/blob/v0.27.0/src/stream.rs) to provide a convenient encapsulation for TLS/Plain streams. It's technically not required, since the underlying `accept` can work with any `AsyncRead + AsyncWrite + Unpin`. I spent awhile trying to implement something with generics or type aliases, but those ended up bleeding through everywhere and feeling like a complete mess. I meant to comment on this when updating the PR — maybe @eloff has an opinion here too.

rust/foxglove/src/lib.rs
ericmlujan: I take the point of not wanting to be terribly granular. I think `ConfigurationError` is probably about right; wondering if it should be applied to other config checks (not necessarily a now thing). I feel like `ValueError` is a little too far in the unspecific direction though

c/src/lib.rs
ericmlujan: Given that this is largely internal/not publicly facing, I think it's fine then. I just want to make sure that user-facing behavior is obvious (like either you put a fully valid bundle of configuration forward to enable the feature, or you null it out).

eloff:
> C++ side looks good, though I'd want someone with more Rust experience to give the Rust implementation a pass. I had some overall questions about how prevalent the `#[cfg(feature = "tls")` attributes should be, and if there's something we can do to limit their surface area somewhat

It's temporary, so we don't need to be too concerned. Typically one pushes as much as one can to modules and make the modules conditional with the attribute(eg add a tls module.)

cpp/foxglove/src/server.cpp
```diff
@@ -257,6 +257,13 @@ FoxgloveResult<WebSocketServer> WebSocketServer::create(
     };
   }
 
+  if (options.tls_identity) {
+    c_options.tls_cert = reinterpret_cast<const uint8_t*>(options.tls_identity->cert.data());
```
eloff, line 261: A reinterpret cast from byte to uint8_t? Something is odd about that. Maybe use the latter type in the vector instead of byte?

Strange C++ doesn't realize they're the same

rust/foxglove/src/websocket/server.rs
```diff
@@ -105,9 +136,35 @@ impl ShutdownHandle {
...
+    }))
+}
+
+#[cfg(not(feature = "tls"))]
+/// Creates a new server.
+pub(crate) fn create_server(
+    ctx: &Arc<Context>,
+    opts: ServerOptions,
+) -> Result<Arc<Server>, FoxgloveError> {
+    // TLS configuration is fallible, so build it prior to allocating the Arc with the weak ref
```
eloff: Comment seems out of place here?

rust/foxglove/src/websocket/server.rs
eloff: Yeah, typically you can move things out into a module or unify them behind some type abstraction. I'm not too concerned here, because I think the longer term plan is remove the attribute and make it always enabled? Correct me if I'm wrong

rust/foxglove/src/websocket/server_stream.rs
eloff: I think this is a good approach. When I did it, I wrapped it in an enum with branches for the tls stream and regular stream. Which is almost the same solution. Using generic types everywhere makes a big mess - that's the one I would avoid as well.

rust/foxglove/src/websocket_server.rs
```diff
@@ -77,6 +83,19 @@ impl WebSocketServer {
         self
     }
 
+    /// Configure TLS with a PEM-formatted x509 certificate chain and pkcs8 private key.
+    /// If enabled, the server will only accept connections using wss://.
+    /// If TLS configuration fails, starting the server will result in an error.
+    #[doc(hidden)]
+    #[cfg(feature = "tls")]
+    pub fn tls(mut self, cert: &[u8], key: &[u8]) -> Self {
```
eloff: How do you feel about making this directly take a TlsIdentity struct? It would prevent accidentally mixing up the order of the arguments.

eloff (approved):
This looks great

cpp/foxglove/src/server.cpp
ericmlujan: Yeah ahaha this has been a fun thing to work through on C++, they're not the same type in C++'s eyes and implicit casting between them is specifically disallowed. I also don't love that this proliferates, but it's everywhere in bridge code (for example)

bryfox: I guess we could use uint8_t in the vector, but doing so would just push this on to the user. Does that seem better?

rust/foxglove/src/websocket/server.rs
bryfox: That's the plan, but I don't know how long "long" is. This could probably benefit from a tls module anyway; let me see what I can clean up.

rust/foxglove/src/websocket_server.rs
bryfox: Good idea; done

rust/foxglove/src/websocket/server.rs
bryfox: OK, there's now a "streams" module which exposes the TLS stuff from either "rust_tls" or "no_tls" modules; the latter has some no-op wrapping types. `cfg` flags are greatly reduced. I like this better; thanks for the prod.

rust/foxglove/src/websocket/server_stream.rs
bryfox: It turns out tokio_util already has a type for this, and we already depend on that. Under the hood, it's similar to what was there before, and we lose the boilerplate.

> `Enum Either`
> Combines two different futures, streams, or sinks having the same associated types into a single type.

https://docs.rs/tokio-util/latest/tokio_util/either/enum.Either.html

c/Cargo.toml
```diff
@@ -18,6 +18,7 @@ env_logger = "0.11.5"
 foxglove = { path = "../rust/foxglove", default-features = false, features = [
   "live_visualization",
   "lz4",
+  "tls",
```
bryfox, line 21: @ericmlujan the bridge uses the pre-built libraries rather than building from source itself, right? So we need to include this here?

rust/foxglove/Cargo.toml
```diff
@@ -22,6 +22,12 @@ live_visualization = [
   "dep:serde_with",
 ]
 lz4 = ["mcap/lz4"]
+tls = [
```
bryfox, line 25: I expect this feature to be disabled by default. However, if I add a tls() config to one of the rust examples, it builds [with default features] and runs successfully. @eloff can you check me on this?

cpp/foxglove/src/server.cpp
ericmlujan: I personally favor sticking with the established pattern of using `byte` for our vector types

c/Cargo.toml
ericmlujan: Bridge uses pre-built binaries by default, so I'd include this (there is a mode for also using a locally built Rust SDK, which we use in CI and while developing, but our users are highly unlikely to use that)

rust/foxglove/Cargo.toml
eloff: When I add:

```rust
.tls(TlsIdentity {
            cert: vec![],
            key: vec![],
        })
```
        
to ws_server example,  and try to compile, I get:

```
error[E0599]: no method named `tls` found for struct `WebSocketServer` in the current scope
   --> rust/examples/ws_server/src/main.rs:113:10
    |
110 |       let server = foxglove::WebSocketServer::new()
    |  __________________-
111 | |         .name(env!("CARGO_PKG_NAME"))
112 | |         .bind(&args.host, args.port)
113 | |         .tls(TlsIdentity {
    | |         -^^^ method not found in `WebSocketServer`
    | |_________|
    |
```
    
If I remove the features = ["schemars"] for foxglove to use default features in the Cargo.toml, I still get the error.    
    
Which is what I'd expect. Is it possible your IDE is somehow adding the feature to the Cargo.toml for the example?

I can't reproduce.

bryfox: There must have been something locally causing this, but after switching branches back & forth and rebuilding I can no longer reproduce.

PR #561, opened 2025-07-29
Update data loader wit read to use ptr and length

bennetthardwick:
### Changelog

None

### Docs

None

### Description

Tooling that uses WIT doesn't expect you to mutate `list<u8>` parameters. This change instead requires you to pass through the raw pointer and length that will be written into.

### DIFF
--- a/rust/foxglove_data_loader/wit/loader.wit
+++ b/rust/foxglove_data_loader/wit/loader.wit
@@ -11,7 +11,7 @@ interface reader {
         seek: func(pos: u64) -> u64;
         position: func() -> u64;
         // Read data into a slice at the provided pointer with the provided length.
-        // These values must be correct or else memory could be overwritten resulting in unexpected behaviour. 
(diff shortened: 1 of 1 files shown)


rust/foxglove_data_loader/wit/loader.wit
```diff
@@ -10,7 +10,9 @@ interface reader {
     resource reader {
         seek: func(pos: u64) -> u64;
         position: func() -> u64;
-        read: func(target: list<u8>) -> u64;
+        // Read data into a slice at the provided pointer with the provided length.
+        // These values must be correct or else memory could be overwritten resulting in unexpected behaviour.
```
james-rms: No reason to point at vague "unexpected behavior" - just say what the host promises to do and the guest knows how to deal with that.
```suggestion
        // The host implementation will write up to `len` bytes into the address starting at `ptr`. The caller must ensure that this memory region is valid.
```

PR #565, opened 2025-07-30
Set up build for foxglove_bridge

ericmlujan:
### Changelog
- [foxglove_bridge] A public beta of `foxglove-bridge` built using the Foxglove SDK is now available.

### Docs
- README from foxglove/ros-foxglove-bridge is migrated from foxglove/ros-foxglove-bridge with new build instructions and eliding ROS 1

### Description
New build targets stood up to simplify both customer and developer build experience (tried to get as close to the "clone repo and run `make`" North Star as possible). Documentation reworded to reflect new build processes, Foxglove SDK as the underlying implementation, and add caveats about public beta.

Unit tests pass, and CI runs successfully. 

Tested that I could follow the instructions in the `README` for building, both in Docker on a Mac and in a fresh ROS installation/SDK repo clone in Ubuntu, and for both prebuilt SDK binaries and in-tree development builds.

Supersedes: #558

Fixes: FG-12586


Makefile
```diff
@@ -25,7 +25,7 @@ $(TARGETS): image
 	docker run -v $(shell pwd):/app \
 		-e CARGO_HOME=/app/.cargo \
 		-e POETRY_VIRTUALENVS_PATH=/app/.virtualenvs \
-		-it $(IMAGE_NAME) \
```
ericmlujan, line 28: Found this broke if a target is run from CI, since it's a headless and doesn't provide a tty. Since this is used simply for building, I figured it's safe to remove. Note that the `shell` target above still retains `-it`.

ericmlujan:
I may need a repo admin to remove ROS iron as a required check, now that it's no longer being invoked

Makefile
amacneil: interesting, how was it working before?

iirc if you leave `-t` and only remove `-i`, you will still get colors etc from the output

ros/Makefile
```diff
@@ -15,30 +22,44 @@ deps:
...
+
 .PHONY: docker-sh-$(1)
 docker-sh-$(1):
 	docker run --rm -it -v $(PWD)/..:/sdk foxglove-sdk-ros-$(1)
+
 endef
 $(foreach distribution,$(ROS_DISTRIBUTIONS),$(eval $(call generate_ros_targets,$(strip $(distribution)))))
 
+.PHONY: docker-build-container
+docker-build-container: docker-build-container-rolling
```
amacneil: nit: I would probably call this `docker-build-image` or just `docker-image`

technically a container is an instance of an image

ros/Makefile
```diff
@@ -1,7 +1,14 @@
+ROS_DISTRIBUTIONS := humble jazzy kilted rolling
+USE_LOCAL_SDK := OFF
+EXTRA_DOCKER_ARGS :=
+DOCKER_USER_ARGS :=
 
+ifeq ($(shell uname -s), Linux)
+	DOCKER_USER_ARGS = --build-arg USER_UID=$(shell id -u) --build-arg USER_GID=$(shell id -g)
```
amacneil, line 7: 👍 

I guess you could also prepend this to EXTRA_DOCKER_ARGS rather than adding a new var, if you want

ros/src/foxglove_bridge/package.xml
```diff
@@ -1,36 +1,26 @@
 <?xml version="1.0"?>
 <package format="3">
     <name>foxglove_bridge</name>
-    <version>0.8.5</version>
+    <version>1.0.0</version>
```
amacneil: maybe don't bump the version number just yet - or bump it to 0.9.0. we should tag 1.0.0 once we feel its stable

ros/Makefile
ericmlujan: I originally did it that way (e.g. `EXTRA_DOCKER_ARGS += <all the stuff>`, but found that if you supplied `EXTRA_DOCKER_ARGS` from the command line, it'd overwrite this. So I wanted to still provide the facility to set custom args without blowing away the UID.

ros/src/foxglove_bridge/package.xml
ericmlujan: 👍

ros/src/foxglove_bridge/CMakeLists.txt
```diff
@@ -54,40 +51,47 @@ if (ENDIAN)
...
+  )
+else()
+  message(STATUS "Fetching Foxglove SDK artifacts from GitHub")
   if (CMAKE_SYSTEM_NAME STREQUAL "Linux" AND CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64")
-    set(FOXGLOVE_SDK_RELEASES "https://github.com/foxglove/foxglove-sdk/releases/download/sdk%2Fv0.10.0/foxglove-v0.10.0-cpp-aarch64-unknown-linux-gnu.zip" "81e2379e3a08160c023779eab0fe5c5764ace3b30c8727b0030ac7dc8b415016")
+    set(FOXGLOVE_SDK_URL "https://github.com/foxglove/foxglove-sdk/releases/download/sdk%2Fv0.10.0/foxglove-v0.10.0-cpp-aarch64-unknown-linux-gnu.zip")
+    set(FOXGLOVE_SDK_SHA "81e2379e3a08160c023779eab0fe5c5764ace3b30c8727b0030ac7dc8b415016")
   elseif(CMAKE_SYSTEM_NAME STREQUAL "Linux" AND CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
-    set(FOXGLOVE_SDK_RELEASES "https://github.com/foxglove/foxglove-sdk/releases/download/sdk%2Fv0.10.0/foxglove-v0.10.0-cpp-x86_64-unknown-linux-gnu.zip" "209d34a8703d44a129c559493e1a3fb215888b4d5973622750ac28b1c345946c")
+    set(FOXGLOVE_SDK_URL "https://github.com/foxglove/foxglove-sdk/releases/download/sdk%2Fv0.10.0/foxglove-v0.10.0-cpp-x86_64-unknown-linux-gnu.zip")
```
amacneil: could these urls be more formulaic? e.g. one variable for the version to pull, but substitute `CMAKE_SYSTEM_PROCESSOR` and `CMAKE_SYSTEM_NAME` directly into the url

ros/src/foxglove_bridge/CMakeLists.txt
```diff
@@ -24,19 +26,14 @@ macro(enable_strict_compiler_warnings target)
   endif()
 endmacro()
 
-find_package(nlohmann_json QUIET)
 find_package(OpenSSL REQUIRED)
 find_package(Threads REQUIRED)
-find_package(websocketpp REQUIRED)
 find_package(ZLIB REQUIRED)
```
amacneil: we probably don't need these dependencies either? it should all be happening in our rust layer

amacneil (commented):
overall direction looks good. someone else should give it a review too though.

Makefile
ericmlujan: The [CI config for the C++ SDK](https://github.com/foxglove/foxglove-sdk/blob/main/.github/workflows/c_cpp.yml) is setting up a full Rust/C++ environment and doing a build directly from the checkout. I wanted to avoid pulling all that logic in here, and found that invoking `make build-cpp` (which defers all the env setup to the repo's Dockerfile) was the most minimalist approach.

Good to know about `-t` :)

ros/src/foxglove_bridge/CMakeLists.txt
ericmlujan: I'll defer paring these down to a later PR. Since we have a C++ WebSocket client used for testing, we may not be able to fully avoid removing these (other than ZLIB, which I confirmed isn't needed anymore). But in that PR, I'd like to refactor the build file to be a bit more granular and localize these deps to just test targets.

PR #567, opened 2025-07-31
Expose sink_id in channel logging functions

ericmlujan:
### Changelog
- [python] calls to `channel.log()` can now optionally specify a sink ID to target message logging to a particular sink
- [c++] calls to `channel.log()` can now optionally specify a sink ID to target message logging to a particular sink

### Docs
Docstrings for SDK documentation updated

### Description
Exposes an optional parameter for `log()` on channel bindings to target logging messages to a specific sink. This is already being used internally by `foxglove-bridge` to allow for selectively publishing low-frequency/latched topics to late joining clients while avoiding duplicate message delivery to already-connected clients.

Tests added to C++ server unit tests to test this behavior with an actual WebSocket client connected in the test. I added a test to the Python side to make sure that the parameter was wired through without causing a runtime error. I held off on adding deeper behavior testing on the Python side because it'd require a Python implementation of a WebSocket client, and I wanted to keep the LoC count of this PR tractable.

Fixes: FG-12408 FG-12405

### DIFF
--- a/c/include/foxglove-c/foxglove-c.h
+++ b/c/include/foxglove-c/foxglove-c.h
@@ -2553,7 +2553,8 @@ foxglove_error foxglove_channel_create_grid3(struct foxglove_string topic,
  */
 foxglove_error foxglove_channel_log_grid3(const struct foxglove_channel *channel,
                                           const struct foxglove_grid3 *msg,
-                                          const uint64_t *log_time);
+                                          const uint64_t *log_time,
+                                          FoxgloveSinkId sink_id);
 
 /**
  * Create a new typed channel, and return an owned raw channel pointer to it.
--- a/c/src/generated_types.rs
+++ b/c/src/generated_types.rs
@@ -1452,6 +1452,7 @@ pub extern "C" fn foxglove_channel_log_grid3(
     channel: Option<&FoxgloveChannel>,
     msg: Option<&Grid3>,
     log_time: Option<&u64>,
+    sink_id: FoxgloveSinkId,
 ) -> FoxgloveError {
     let mut arena = pin!(Arena::new());
     let arena_pin = arena.as_mut();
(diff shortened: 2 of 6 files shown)


python/foxglove-sdk/python/foxglove/tests/test_channel.py
```diff
@@ -235,3 +235,9 @@ def test_exposes_unique_channel_ids(new_topic: str) -> None:
     assert ch_1.id() > 0
     assert ch_1.id() < ch_2.id()
     assert ch_2.id() < ch_3.id()
+
+
+def test_log_message_to_specific_sink(new_topic: str) -> None:
+    ctx = Context()
+    ch = Channel(new_topic, context=ctx)
+    ch.log("test", sink_id=1)
```
eloff, line 243: The test just verifies the argument exists and doesn't crash, which maybe is the only viable thing to do here, I think that's fine since the behavior is tested on the rust side.

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -862,7 +862,7 @@ void FoxgloveBridge::rosMessageHandler(ChannelId channelId, SinkId sinkId,
 
   auto& channel = _channels.at(channelId);
   // TODO: Change to log() when the released SDK is updated to make this a public overload
```
eloff: Delete the comment

python/foxglove-sdk/python/foxglove/tests/test_channel.py
ericmlujan: Yep, that was the very least I wanted to test. Ideally in python we'd also have some kind of mock sink or client available but didn't want to scope creep this PR too much

PR #273, opened 2025-03-07
RawImage: document encodings in schemas

james-rms:
### Docs

Is docs

### Description

Documents our supported raw image encodings in (slightly more) detail.

<table><tr><th>Before</th><th>After</th></tr><tr><td>

</td><td>

</td></tr></table>

Fixes: FG-6974

PR #571, opened 2025-08-05
Bump @foxglove/schemas to 1.7.3

bryfox:
### Changelog

- Add Timestamp and Duration to JSON schemas (#391)
- Document RawImage encodings (#273)


amacneil (approved):
We should start versioning this with the sdk

bryfox:
> We should start versioning this with the sdk

👍 FG-12653

PR #577, opened 2025-08-06
Adjusting the voxel grid schema

jeff-foxglove:
### Docs

None

### Description

Minor tweaks to `Foxglove.Grid3` to change it to `Foxglove.VoxelGrid` + some field renaming and doc improvements

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -1,0 +1,22 @@
+name: Dependabot
+
+on:
+  pull_request: {}
+
+jobs:
+  review:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    if: github.event.pull_request.user.login == 'dependabot[bot]'
+    steps:
+      - name: Approve Dependabot PRs
+        uses: octokit/request-action@v2.x
+        with:
+          route: POST /repos/{owner}/{repo}/pulls/{pull_number}/reviews
+          owner: ${{ github.event.repository.owner.login }}
+          repo: ${{ github.event.repository.name }}
+          pull_number: ${{ github.event.pull_request.number }}
+          event: APPROVE
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -58,7 +58,7 @@ jobs:
     if: startsWith(github.ref, 'refs/tags/sdk/v')
     steps:
       - name: Download artifacts
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v5
         with:
           path: artifacts
       - run: test -d artifacts/python
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -61,7 +61,7 @@ jobs:
           git push -f origin "${{ env.SDK_BRANCH }}"
 
       - name: Download C/C++ artifacts
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v5
         with:
           pattern: ${{ needs.build-artifacts.outputs.artifact_pattern }}
           path: ci_artifacts/
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -242,7 +242,7 @@ jobs:
       contents: write
       attestations: write
     steps:
-      - uses: actions/download-artifact@v4
+      - uses: actions/download-artifact@v5
         with:
           path: python/foxglove-sdk
       - run: find python/foxglove-sdk/wheels*
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -225,7 +225,7 @@ dependencies = [
  "quote",
  "serde",
  "serde_json",
- "syn 2.0.103",
+ "syn 2.0.104",
  "tempfile",
  "toml",
 ]
@@ -299,7 +299,7 @@ dependencies = [
  "heck",
  "proc-macro2",
  "quote",
- "syn 2.0.103",
+ "syn 2.0.104",
 ]
 
 [[package]]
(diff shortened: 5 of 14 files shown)


typescript/schemas/src/internal/schemas.ts
```diff
@@ -1246,7 +1246,7 @@ const Grid3: FoxgloveMessageSchema = {
       name: "pose",
       type: { type: "nested", schema: Pose },
       description:
-        "Origin of grid's corner relative to frame of reference; grid is positioned in the x-y plane relative to this origin",
+        "Origin of grid's corner relative to frame of reference; grid is positioned in the x-y plane, at z=0, relative to this origin",
```
amacneil:
```suggestion
        "Origin of grid's corner relative to frame of reference",
```

it's not at z=0

typescript/schemas/src/internal/schemas.ts
```diff
@@ -1264,7 +1264,7 @@ const Grid3: FoxgloveMessageSchema = {
       description: "Size of single grid cell along x, y, and z axes, relative to `pose`",
     },
     {
-      name: "slice_stride",
+      name: "depth_stride",
```
amacneil: claude says

> Great improvement with `depth_stride`! Much clearer than `slice_stride`.

typescript/schemas/src/internal/schemas.ts
```diff
@@ -1289,7 +1289,7 @@ const Grid3: FoxgloveMessageSchema = {
       name: "data",
       type: { type: "primitive", name: "bytes" },
       description:
-        "Grid cell data, interpreted using `fields`, in depth-major, row-major (Z-Y-X) order.\n For the data element starting at byte offset i, the coordinates of its corner closest to the origin will be:\n z = (i / (row_stride * cell_stride)) % slice_stride * cell_size.z\n y = (i / cell_stride) % row_stride * cell_size.y\n x = i % cell_stride * cell_size.x",
+        "Grid cell data, interpreted using `fields`, in depth-major, row-major (Z-Y-X) order.\n For the data element starting at byte offset i, the coordinates of its corner closest to the origin will be:\n z = (i / (row_stride * cell_stride)) % depth_stride * cell_size.z\n y = (i / cell_stride) % row_stride * cell_size.y\n x = i % cell_stride * cell_size.x",
```
amacneil: plz review/verify this comment from claude:

> Same issue as before - `depth_stride` is bytes, not a count, so `% depth_stride` is meaningless.

its suggested fix:

```
Grid cell data, interpreted using `fields`, in depth-major, row-major (Z-Y-X) order.
For the data element starting at byte offset i, the voxel indices are:
  depth_index = i / depth_stride
  row_index = (i % depth_stride) / row_stride  
  col_index = (i % row_stride) / cell_stride
And the coordinates of its corner closest to the origin are:
  z = depth_index * cell_size.z
  y = row_index * cell_size.y
  x = col_index * cell_size.x
```

typescript/schemas/src/internal/schemas.ts
jtbandes: Wait what is "depth"? are we talking about z? Should we have x_stride, y_stride, z_stride?

jeff-foxglove: That's why I used slice initially. I don't think we want `x_stride, y_stride, z_stride` as this is supposed to be a logical extension of the `Foxglove.Grid`, so ideally we shouldn't need to rename existing fields

typescript/schemas/src/internal/schemas.ts
jeff-foxglove: >Same issue as before - depth_stride is bytes, not a count, so % depth_stride is meaningless.

Think both work, but this one looks a little cleaner. Not sure why it complains about it being bytes, given all the variables are in the context of bytes...

jtbandes (approved):
I edited the PR title – I don't think "because the AI said so" is a sufficient reason to do things.

I still find some things could be confusing for users, like:
- is it clear how "row"/"col" relate to x/y/z and what is a "slice"? Is this terminology standard for voxel grids?
- What is the purpose of the stride stuff when a simple count could be used – should we ditch these? Do we believe users find value from them (e.g. being able to include extra padding in their buffers but still having the buffers be in a predetermined order?)

typescript/schemas/src/internal/schemas.ts
```diff
@@ -1289,7 +1288,7 @@ const Grid3: FoxgloveMessageSchema = {
       name: "data",
       type: { type: "primitive", name: "bytes" },
       description:
-        "Grid cell data, interpreted using `fields`, in depth-major, row-major (Z-Y-X) order.\n For the data element starting at byte offset i, the coordinates of its corner closest to the origin will be:\n z = (i / (row_stride * cell_stride)) % slice_stride * cell_size.z\n y = (i / cell_stride) % row_stride * cell_size.y\n x = i % cell_stride * cell_size.x",
```
james-rms, line 1292: how have we validated this - we were wrong before, are we going with claude's intuition here or have we checked more thoroughly this time?

jeff-foxglove: I've been running the equations through a JS console to ensure the indexing comes out in the major ordering desired. Both of those worked when I tested them, I am using claude's vs chatgpt's as they are shorter

jeff-foxglove:
@jtbandes 
>is it clear how "row"/"col" relate to x/y/z and what is a "slice"? Is this terminology standard for voxel grids?

I think it's as clear as mud for conventions on what to call the Z. [This reddit discussion](https://www.reddit.com/r/math/comments/1ajrfu/if_a_2d_matrix_has_rows_and_columns_what_is_the/) I think gives a good list of potential ones, where I chose `slice` as I was thinking `depth slice`.

>What is the purpose of the stride stuff when a simple count could be used – should we ditch these?
Do we believe users find value from them (e.g. being able to include extra padding in their buffers but still having the buffers be in a predetermined order?)

I think at minimum we'll need the cell stride as we need to map the byte array to a cell of the fields, unless we're thinking of moving away from a byte array for the data as well

PR #579, opened 2025-08-07
Add rgb camera examples for Python, C++ and Rust

msadowski:
This PR adds examples showing how to stream RGB camera data in real-time. Opening it as a draft to get early feedback on handling dependencies etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,15 +2,14 @@
 
 The Foxglove SDK allows you to log and visualize multimodal data with [Foxglove](https://foxglove.dev).
 
-The SDK is written in Rust, with bindings and packages available for Python and C++.
+The core SDK is written in Rust, with bindings for Python, and C++. We publish prebuilt libraries and Python wheels, so you don’t need a Rust development environment.
 
-For ROS integration, see [ros-foxglove-bridge](https://github.com/foxglove/ros-foxglove-bridge).
+- Stream live data to Foxglove over a local WebSocket
+- Log data to [MCAP](https://mcap.dev/) files for visualization or analysis
+- Leverage built-in [Foxglove schemas](https://docs.foxglove.dev/docs/sdk/schemas) for common visualizations, or your own custom messages using a supported serialization format
+- ROS packages are available for all supported distributions (see our [ROS 2 tutorial](https://docs.foxglove.dev/docs/getting-started/frameworks/ros2))
 
-- Visualize live data using the Foxglove WebSocket protocol
-- Log data to [MCAP](https://mcap.dev/) files for later visualization or analysis
-- Leverage built-in [Foxglove schemas](https://docs.foxglove.dev/docs/visualization/message-schemas/introduction) for common visualizations, or your own custom messages using a supported serialization format
-
-Visit [Foxglove Docs](https://docs.foxglove.dev/) to get started.
+Visit [Foxglove SDK Docs](https://docs.foxglove.dev/sdk) to get started.
 
 ## Packages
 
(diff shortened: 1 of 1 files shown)


.github/actions/common-deps/action.yml
```diff
@@ -21,4 +21,7 @@ runs:
       run: |
         sudo apt-get install -y --no-install-recommends \
           libprotobuf-dev \
-          protobuf-compiler
+          protobuf-compiler \
+          libopencv-dev \
+          clang \
+          libclang-dev
```
bryfox: clang is part of the runner image, and we already build with it; is this needed to get a different version?

cpp/CMakeLists.txt
```diff
@@ -230,3 +230,17 @@ target_compile_options(example_auto_serialize PUBLIC ${SANITIZER_COMPILE_OPTIONS
 target_include_directories(example_auto_serialize SYSTEM PRIVATE ${json_SOURCE_DIR}/include)
 target_link_options(example_auto_serialize PUBLIC ${SANITIZER_LINK_OPTIONS})
 target_link_libraries(example_auto_serialize foxglove_cpp_static)
+
+option(BUILD_OPENCV_EXAMPLE "Build the RGB camera visualization example (requires OpenCV)" OFF)
+
+if(BUILD_OPENCV_EXAMPLE)
```
bryfox, line 236: could we use [EXCLUDE_FROM_ALL](https://cmake.org/cmake/help/v3.31/prop_tgt/EXCLUDE_FROM_ALL.html) for this?

python/foxglove-sdk-examples/rgb-camera-visualization/README.md
```diff
@@ -0,0 +1,35 @@
...
+poetry install
+```
+
+## Running the Example
+
+Navigate to the example directory (`python/foxglove-sdk-examples/rgb-camera-visualization`).
+
+### Basic usage (default camera):
+```bash
+poetry run python3 main.py
```
bryfox: nit: our other examples use `poetry run python ...`

bryfox (commented):
Should these examples have some way to limit the frame rate?

Is it worth mentioning CompressedVideo in the readmes as an alternative?

.github/actions/common-deps/action.yml
msadowski: Indeed, `clang` was redundany here, but `libclang-dev` is still needed by the rust rgb cam example

msadowski:
> Should these examples have some way to limit the frame rate?
> 
> Is it worth mentioning CompressedVideo in the readmes as an alternative?

@bryfox, the OpenCV call to read a camera frame is a blocking call with a default 5s timeout. This way, we default the loop rate to camera FPS and don't have to specify other fields. 

I tried playing with CompressedVideo a little bit, but wasn't able to make it work - will play a bit after this one is merged, and make a follow-up PR

python/foxglove-sdk-examples/rgb-camera-visualization/README.md
msadowski: Fixed, thanks

cpp/CMakeLists.txt
msadowski: I added EXCLUDE_FROM_ALL for this executable. Let me know if you had something else in mind

python/foxglove-sdk-examples/rgb-camera-visualization/minimal.py
bryfox, line 1: This is definitely minimal 😄 — did you mean to check it in?

cpp/CMakeLists.txt
bryfox: I was wondering if we could use EXCLUDE_FROM_ALL instead of the custom BUILD_OPENCV_EXAMPLE var, but I think what you have is fine

python/foxglove-sdk-examples/rgb-camera-visualization/minimal.py
msadowski: Nice catch, I did not mean to check it in!

PR #590, opened 2025-08-12
Add debug param to bridge

ericmlujan:
### Changelog
None

### Docs
None

### Description
Add a new `debug` ROS param to the bridge node that enables the DEBUG logger level in both ROS and the Foxglove SDK.


ericmlujan:
@bryfox Yeah, I wanted to keep the semantic simple at first, before doing something more complex. But in the future if we find it helpful, we can always add `debug:=warn` which defaults to a warn log level, where `debug:=true` aliases to `debug:=debug` if we choose to go down that path later

PR #599, opened 2025-08-15
Bump SDK version in foxglove_bridge

ericmlujan:
### Changelog
None

### Docs
None
### Description
`foxglove_bridge` now uses v0.11 for its prebuilt SDK dep. 

This allows its implementation to use the public version of `RawChannel::log()` now that the sink ID param is exposed and documented.


ros/src/foxglove_bridge/CMakeLists.txt
```diff
@@ -62,14 +62,14 @@ if (USE_LOCAL_SDK)
   )
 else()
   message(STATUS "Fetching Foxglove SDK artifacts from GitHub")
-  set(FOXGLOVE_SDK_VERSION "0.10.0")
+  set(FOXGLOVE_SDK_VERSION "0.11")
```
bryfox: "v0.11.0"?

bryfox (approved):
👍 once shas have been verified with a local test

ros/src/foxglove_bridge/CMakeLists.txt
ericmlujan: Yep

PR #600, opened 2025-08-15
Package foxglove_bridge in docker image

ericmlujan:
### Changelog
- [Foxglove Bridge] The public beta of Foxglove Bridge is now packaged as a Docker image in our public image registry

### Docs
README.md for bridge updated with image pull instructions

### Description
Create GHA workflow for pushing a Docker image containing the Foxglove Bridge to our public image registry. This is in response to customer feedback that providing these images helps them more easily deploy Foxglove Bridge.

Images are built against all currently supported versions of ROS. The GHA workflow is currently configured to build the bridge:
- On a push to `main`, where the version identifier will be the git SHA of the commit on `main`
- On an SDK release tag (i.e. `sdk/vxxx`), where the version identifier will be the SDK version number

The version number of Foxglove Bridge is incremented to equal that of Foxglove SDK, as per our consensus that the versioning/release cadence of these components should be coupled going forward.

Fixes: FLE-1

### DIFF
--- a/.github/workflows/publish-ros-bridge.yml
+++ b/.github/workflows/publish-ros-bridge.yml
@@ -56,13 +56,10 @@ jobs:
       - name: Compute image tags
         id: compute-tags
         run: |
-          set -euo pipefail
           IMAGE_BASE="${DOCKER_REPO}"
           DISTRO="${{ matrix.ros_distribution }}"
           TAG="${IMAGE_BASE}:ros-${DISTRO}-${IMAGE_VERSION}"
-          printf 'tag<<EOF\n' >> "$GITHUB_OUTPUT"
-          printf '%s\n' "${TAG}" >> "$GITHUB_OUTPUT"
-          printf 'EOF\n' >> "$GITHUB_OUTPUT"
+          echo "tag=${TAG}" >> "$GITHUB_OUTPUT"
 
       - name: Build and push
         uses: docker/build-push-action@v6
@@ -72,6 +69,6 @@ jobs:
           push: true
           build-args: |
             ROS_DISTRIBUTION=${{ matrix.ros_distribution }}
-          tags: ${{ steps.compute-tags.outputs.tags }}
+          tags: ${{ steps.compute-tags.outputs.tag }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
--- a/.github/workflows/publish-ros-bridge.yml
+++ b/.github/workflows/publish-ros-bridge.yml
@@ -3,7 +3,6 @@ name: Publish ROS Bridge Docker image
 on:
   workflow_dispatch: {}
   push:
-    branches: [main]
     tags: ['sdk/v*']
 
 permissions:
@@ -46,11 +45,11 @@ jobs:
           # Check if this is a release tag (sdk/v*)
           if [[ "${GITHUB_REF}" == refs/tags/sdk/v* ]]; then
             # Extract version from release tag (remove sdk/ prefix)
-            VERSION=$(echo "${GITHUB_REF}" | sed 's|refs/tags/sdk/||')
+            VERSION=${GITHUB_REF#refs/tags/sdk/}
             echo "Using release version: ${VERSION}"
           else
             # Use git SHA for non-release builds
-            VERSION=$(echo "${GITHUB_SHA}")
+            VERSION="${GITHUB_SHA}"
             echo "Using SHA version: ${VERSION}"
           fi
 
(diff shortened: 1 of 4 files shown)


.github/workflows/publish-ros-bridge.yml
```diff
@@ -0,0 +1,77 @@
...
+        run: gcloud auth configure-docker us-central1-docker.pkg.dev --quiet
+
+      - name: Compute version for tagging
+        run: |
+          set -euo pipefail
+
+          # Check if this is a release tag (sdk/v*)
+          if [[ "${GITHUB_REF}" == refs/tags/sdk/v* ]]; then
+            # Extract version from release tag (remove sdk/ prefix)
+            VERSION=$(echo "${GITHUB_REF}" | sed 's|refs/tags/sdk/||')
```
jtbandes: nit to simplify: (I've visited [this page](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) a million times but I still have to look it up every time):
```suggestion
            VERSION=${GITHUB_REF#refs/tags/sdk/}
```

ros/src/foxglove_bridge/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+
+RUN apt-get update
+
+# Create foxglove user
+ARG USERNAME=foxglove
+ARG USER_UID=1005
+ARG USER_GID=$USER_UID
+RUN groupadd --gid $USER_GID $USERNAME \
+    && useradd --uid $USER_UID --gid $USER_GID -m $USERNAME \
+    && echo "$USERNAME ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers.d/$USERNAME
```
jtbandes, line 12: what's the purpose of all the user stuff? do things break if using the default root user?

ros/src/foxglove_bridge/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+RUN /bin/bash -c '. /opt/ros/$ROS_DISTRO/setup.bash && \
+    colcon build --packages-select foxglove_bridge'
+
+RUN echo '#!/bin/bash\n\
+set -e\n\
+\n\
+source /ros/install/setup.bash\n\
+\n\
+exec ros2 launch foxglove_bridge foxglove_bridge_launch.xml "$@"' > /ros/entrypoint.sh \
+    && chmod +x /ros/entrypoint.sh
```
jtbandes: nit: would suggest a [heredoc](https://www.docker.com/blog/introduction-to-heredocs-in-dockerfiles/) to make it a little more readable

ros/src/foxglove_bridge/README.md
```diff
@@ -19,10 +19,31 @@ Live debugging of ROS systems has traditionally relied on running ROS tooling su
...
+```
+
+where `<tag>` takes the form of:
+```
+ros-<ROS distro>-<foxglove SDK version>
+```
+
+For example, if you wanted to pull an image based on ROS Kilted and Foxglove SDK v0.11:
+```bash
+docker pull us-central1-docker.pkg.dev/foxglove-images/images/foxglove_bridge:ros-kilted-v0.11
```
jtbandes: nit: would this actually be v0.11.0? i.e. does it need to be an exact match to the package.xml version, or is there some kind of semver thing going on here too?

.github/workflows/publish-ros-bridge.yml
```diff
@@ -0,0 +1,77 @@
...
+            echo "Using release version: ${VERSION}"
+          else
+            # Use git SHA for non-release builds
+            VERSION=$(echo "${GITHUB_SHA}")
```
jtbandes:
```suggestion
            VERSION="$GITHUB_SHA"
```

.github/workflows/publish-ros-bridge.yml
ericmlujan: Thanks, I always need to find this shibboleth

ros/src/foxglove_bridge/Dockerfile
ericmlujan: Mostly pattern matching to our other Dockerfiles. I generally think it's good practice to not run as the root user in a docker container, though the foxglove user _is_ still privileged

ros/src/foxglove_bridge/README.md
ericmlujan: Good call, that was just a typo on my part

ericmlujan:
@jtbandes re: publishing every commit, I could see it bloating our storage. I'll revise to just use release tags for now, but add a workflow dispatcher so it can be manually triggered to run on a SHA if we want to ship a development version to a customer

PR #607, opened 2025-08-20
Bump schemas to 1.9.0

bryfox:
Bump minor version for schema added in #556


jeff-foxglove:
I didn't release the 1.8, as I figured the back to back release was a little pointless, so I think we can skip this and have 1.8 include both our changes

PR #609, opened 2025-08-20
Remove pinned versions from Dockerfile

bryfox:
This Dockerfile uses `rust:latest` as a base image, but pins OS dependencies. A new image was pushed last night, and CI is failing. Let's specify the major versions, at least for now to unblock CI.


bryfox:
Will include this with #608

PR #614, opened 2025-08-25
add encode method on all schemas

james-rms:
#### Added
- C: added `foxglove_..._encode` implementations for all foxglove schemas (excluding Timestamp and Duration).
- C++: added `::encode()` methods to all foxglove schemas (excluding Timestamp and Duration). 
- Python: added `self.encode()` methods to all foxglove schemas (excluding Timestamp and Duration).

These can be used to directly serialize the message as protobuf.
### Docs

None.

### Description

When attempting to use the foxglove SDK within a data loader, the user needs to encode messages and return them to the caller, rather than pass them to a Channel. The data loader framework code can't be wrapped in such a way to make the Channel abstraction make sense, because messages are "pulled" from data loaders rather than "pushed" into a channel.

Therefore, this PR adds encoding capability to C and C++ bindings (and Python for completeness). 

The C/C++ wrappers write into a preallocated buffer, and return an error if the buffer is not large enough. This allows the caller to use whatever allocation scheme suits them.

### Discussion

I'm not so sure about the language to use here in the documentation. Should I explicitly say that these are encoded as protobuf? They use whatever `Encode` implementation the underlying struct uses, which we define for ourselves for our built-in schemas. Do we ever want to implement other encodings for our own schemas and channels, or are we happy to define that as Protobuf for now?

</td><td>

</td></tr></table>


typescript/schemas/src/internal/generatePyclass.ts
```diff
@@ -289,6 +292,13 @@ function generateMessageClass(schema: FoxgloveMessageSchema): string {
     `    fn get_schema() -> PySchema {`,
     `        foxglove::schemas::${className}::get_schema().unwrap().into()`,
     `    }`,
+    `    /// Encodes the ${className}.`,
+    `    fn encode<'a>(&self, py: Python<'a>) -> Bound<'a, PyBytes> {`,
+    `        PyBytes::new_with(py, self.0.encoded_len().expect("foxglove schemas provide len"), |mut b: &mut[u8]| {`,
+    `            self.0.encode(&mut b).expect("encoding len was provided above");`,
+    `            Ok(())`,
+    `        }).expect("closure does not return error")`,
```
bryfox: This is documented as also erroring if Python raises a MemoryError. I expect the system won't really be able to recover anyway, but panicking with this message would only mask the problem.

typescript/schemas/src/internal/generateSdkRustCTypes.ts
```diff
@@ -245,6 +245,52 @@ pub extern "C" fn foxglove_channel_log_${snakeName}(channel: Option<&FoxgloveCha
...
+///
+/// # Safety
+/// The buffer starting at ptr must point to a valid memory region at least len bytes long.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn foxglove_${snakeName}_encode(
+    msg: Option<&${name}>,
+    ptr: *mut u8,
+    len: usize,
+    encoded_len: Option<&mut usize>,
+) -> FoxgloveError {
```
bryfox, line 264: Shouldn't this have a `ptr.is_null` check?

bryfox (commented):
This looks good to me in general; the APIs feel right. Should get another set of eyes on the `unsafe` parts in particular.

For documentation, I think we can document that it's protobuf today if that's helpful to callers (but subject to change in a breaking/major version).

gasmith (approved):
Looks good to me, apart from Bryan's comments.

typescript/schemas/src/internal/generateSdkRustCTypes.ts
```diff
@@ -245,6 +245,52 @@ pub extern "C" fn foxglove_channel_log_${snakeName}(channel: Option<&FoxgloveCha
...
+    let mut arena = pin!(Arena::new());
+    let arena_pin = arena.as_mut();
+    // Safety: we're borrowing from the msg, but discard the borrowed message before returning
+    match unsafe { ${name}::borrow_option_to_native(msg, arena_pin) } {
+        Ok(msg) => {
+            if len == 0 {
+                if let Some(encoded_len) = encoded_len {
+                    *encoded_len = msg.encoded_len().expect("foxglove schemas return Some(len)");
+                }
+                return FoxgloveError::BufferTooShort;
```
eloff, line 274: Nit: BufferTooSmall seems a better name to me

eloff (approved):
This looks solid. The API looks great, and I appreciate the attention to detail and performance in the implementation.

typescript/schemas/src/internal/generateSdkRustCTypes.ts
james-rms: This is an existing error name we've committed to already, so I don't feel like changing it here.

PR #616, opened 2025-08-26
C/C++ add schema() getter function for all builtin schemas

james-rms:
### Changelog
- C: add `foxglove_<schema name>_schema()` for Foxglove schemas excluding Timestamp and Duration.
- C++: add `<schema name>::schema()` for Foxglove schemas excluding Timestamp and Duration.

These allow the caller to retrieve runtime schema information that can be used with MCAP or the Data Loader API.

### Docs

None

### Description

Following up from #614 , adds a getter for the schema information to the C/C++ SDK. These mirror the similar `get_schema()` methods already available in Python and Rust. This again allows data loaders to provide schema information to the foxglove app in pull control flow rather than push.

</td><td>

</td></tr></table>

### DIFF
--- a/cpp/foxglove/src/schemas.cpp
+++ b/cpp/foxglove/src/schemas.cpp
@@ -1908,7 +1908,7 @@ Schema ArrowPrimitive::schema() {
   Schema result;
   result.name = std::string(c_schema.name.data, c_schema.name.len);
   result.encoding = std::string(c_schema.encoding.data, c_schema.encoding.len);
-  result.data = (const std::byte*)(c_schema.data);
+  result.data = reinterpret_cast<const std::byte*>(c_schema.data);
   result.data_len = c_schema.data_len;
   return result;
 }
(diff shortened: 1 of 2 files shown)


cpp/foxglove/include/foxglove/schema.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+#include <string>
+
+namespace foxglove {
+
+/// A Schema is a description of the data format of messages in a channel.
+///
+/// It allows Foxglove to validate messages and provide richer visualizations.
+/// See the [MCAP spec](https://mcap.dev/spec#schema-op0x03) for more information.
+struct FoxgloveSchema {
```
jtbandes: We already have a similar struct here: https://github.com/foxglove/foxglove-sdk/blob/312b1a3a4e0f0155deef6ea1ccebf5603aa206da/cpp/foxglove/include/foxglove/channel.hpp#L19-L36

Do we really need both?

If so, the names need to reflect the actual differences rather than differing only by a "Foxglove-" prefix

james-rms: Fixed.

PR #617, opened 2025-08-26
Export ws protocol types and WebSocketClient from sdk

eloff:
This is supporting work for this PR in agent: 

Make some supporting types public but undocumented in sdk for use in agent.

Modify the test WebSocketClient to make it more suitable for production use.


rust/examples/ws_services/src/client.rs
```diff
@@ -75,7 +75,7 @@ impl Drop for Client {
 impl Client {
     /// Connects to a websocket server and validates the subprotocol.
     async fn connect(host: &str, port: u16) -> Result<Self> {
-        let mut request = format!("ws://{host}:{port}/")
+        let mut request = format!("{host}:{port}/")
```
bryfox: Can we keep this the same? If not, the default should be updated to include the protocol so that it still works.

If this was to support wss, I think we can delay that until the feature is generally available and documented.

rust/foxglove/src/websocket_client.rs
```diff
@@ -0,0 +1,135 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum WebSocketClientError {
+    #[error("unexpected end of stream")]
+    UnexpectedEndOfStream,
+    #[error("invalid subprotocol")]
+    InvalidSubprotocol,
+    #[error(transparent)]
+    ParseError(#[from] ParseError),
+    #[error(transparent)]
+    Tungstenite(#[from] tungstenite::Error),
```
bryfox, line 29: It'd be nice to abstract this away, but I suppose it's not important today.

bryfox (approved):
Minor comment about the public-facing example, but looks good to me once that's fixed.

rust/examples/ws_services/src/client.rs
eloff: Good catch, this was a find replace that went bad, undoing the change

rust/foxglove/src/websocket_client.rs
eloff: It's internal to SDK and Agent, we can change it however we like later

PR #624, opened 2025-08-27
foxglove-c: enable building for WASM32 targets

james-rms:
### Changelog
- C: enable a minimal build for wasm32 targets, that includes Foxglove schema struct definitions and serialization functionality only.
### Description

We want to be able to use the foxglove SDK from within WASM data loaders to serialize foxglove messages. To do this we need to build the core foxglove serialization functionality as WASM.

 This PR moves all of the code that does I/O in foxglove-c out of `lib.rs` and into modules, which are conditionally included in the library based on architecture. It uses cbindgen's `[defines]` feature to put `#ifdef __wasm32__` around all of these features in the generated header file. 

NOTE: `__wasm32__` is not defined in any standard, but it is what the WASI SDK Clang implementation uses, and that's the toolchain that we specify for the data loader SDK.

</td><td>

</td></tr></table>

### DIFF
--- a/c/src/lib.rs
+++ b/c/src/lib.rs
@@ -11,30 +11,30 @@ mod arena;
 mod generated_types;
 mod util;
 pub use generated_types::*;
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(not(target_family = "wasm"))]
 mod bytes;
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(not(target_family = "wasm"))]
 mod channel;
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(not(target_family = "wasm"))]
 mod connection_graph;
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(not(target_family = "wasm"))]
 mod fetch_asset;
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(not(target_family = "wasm"))]
 mod logging;
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(not(target_family = "wasm"))]
 mod parameter;
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(not(target_family = "wasm"))]
 mod server;
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(not(target_family = "wasm"))]
 mod service;
 
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(not(target_family = "wasm"))]
 pub use server::*;
 
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(not(target_family = "wasm"))]
 pub use channel::*;
 
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(not(target_family = "wasm"))]
 pub use logging::foxglove_set_log_level;
 
 pub use foxglove::Context as FoxgloveContext;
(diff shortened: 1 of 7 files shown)


jtbandes:
I was also playing with wasm builds recently (for python) and ended up with `target_family = "wasm"` -- any preference between that and target_arch?

james-rms:
I didn't come into this with a preference, I guess because i've never heard of target_family before. I feel like `arch` is  more commonly used across build toolchains, also the description of family seems pretty wishy-washy here:

> Key-value option providing a more generic description of a target, such as the family of the operating systems or architectures that the target generally falls into.

What defines the family of operating systems? Why does this cover sets of architectures and operating systems?

c/src/lib.rs
```diff
@@ -2,32 +2,42 @@
...
+#[cfg(not(target_arch = "wasm32"))]
 mod fetch_asset;
+#[cfg(not(target_arch = "wasm32"))]
 mod logging;
+#[cfg(not(target_arch = "wasm32"))]
 mod parameter;
+#[cfg(not(target_arch = "wasm32"))]
+mod server;
+#[cfg(not(target_arch = "wasm32"))]
 mod service;
```
bryfox, line 29: A lot of these (service, server, parameter, conn graph, asset...) are specific to the WS server; we could re-organize around that and have a single module here.

typescript/schemas/src/internal/generateSdkRustCTypes.ts
```diff
@@ -125,6 +125,7 @@ pub struct ${name} {
     .join("\n\n")}
 }
 
+#[cfg(not(target_arch = "wasm32"))]
```
bryfox: We could consider generating separate schema & channel files; I don't know if that would have any impact in practice.

bryfox (commented):
> We want to be able to use the foxglove SDK from within WASM data loaders to serialize foxglove messages. To do this we need to build the core foxglove serialization functionality as WASM.

Conceptually, does it feel like we should have a separate crate for schemas & serialization that the SDK depends on? I think we want to avoid a proliferation of published crates, but I'm curious how it felt when you went through this. I like that c/src/lib is starting to be broken up more, though.

typescript/schemas/src/internal/generateSdkRustCTypes.ts
james-rms: This doesn't really do much for the resulting generated .h file, it more or less comes out the same. It would work to reduce the number of #[cfg...] declarations in this file though.

c/src/lib.rs
james-rms: I don't mind that idea, but I think it's reasonable to split these two up because one is exported with `pub use server::*` in `lib.rs` and one is not.

james-rms:
> Conceptually, does it feel like we should have a separate crate for schemas & serialization that the SDK depends on? I think we want to avoid a proliferation of published crates, but I'm curious how it felt when you went through this. I like that c/src/lib is starting to be broken up more, though.

I don't love the idea of a separate crate, because that means we need to keep a separate cbindgen.toml going, and share some of the common non-generated code that these generated files depend on - at least that's my subjective experience from the process of making this PR.

james-rms:
@eloff do you have opinions here as well? it's the base branch of #626, which you approved.

gasmith (approved):
Regarding the `target_family`/`target_arch` discussion, ChatGPT has this to say:

- Use `target_family="wasm"` to cover all wasm targets
- Use `target_arch="wasm32"` (or `"wasm64"`) to cover wasm targets of a specific word size
- Use `target_os="wasi"` to specifically target WASI

Since we're not doing anything here that cares about word size, I think it's probably better to use `target_family`.

PR #626, opened 2025-08-28
C++ data loaders: include foxglove SDK for message serialization

james-rms:
### Description

Based on https://github.com/foxglove/foxglove-sdk/pull/624

This PR changes the foxglove C++ Data loader SDK to include a limited subset of the Foxglove SDK. This allows data loader developers to serialize Foxglove message using the schemas available in the SDK.

Also:
- includes the data loader SDK as a `.zip` bundle in foxglove SDK releases.
- Changes the C++ SDK generation code to include #ifdefs around unsupported functionality for WASM builds
- Changes the Arena on-stack size to 8KB. This still allows small messages to be serialized without heap allocation, but avoids blowing up the stack on WASM (where the total stack size is 1MB)
- Updates the C++ data loader example included to demonstrate message serialization using the foxglove SDK.

</td><td>

</td></tr></table>


eloff:
> * Changes the Arena implementation used when serializing messages to allocate the arena on the heap rather than on the stack. This is because I think it's inappropriate to allocate 128KB on the stack as a library, considering that most stacks are 8MB total, and for WASM the limit is much lower (1MB).

The arena already had a fallback for heap allocation, so why don't we make the limit something small like 8kb or 4kb? It's hard to see how that would get us in trouble, while still allowing smaller messages to be serialized without any heap allocations and freeing.

cpp/foxglove/include/foxglove/arena.hpp
```diff
@@ -12,8 +12,8 @@
 
 namespace foxglove {
 
-/// A fixed-size memory arena that allocates aligned arrays of POD types on the stack.
-/// The arena contains a single inline array and allocates from it.
+/// A fixed-size memory arena that allocates aligned arrays of POD types in a contiguous array.
+/// The arena contains a single heap-allocated byte array and allocates from it.
```
eloff: See the PR level comment I made, I'm reluctant to give up on the idea entirely of a stack arena

cpp/foxglove/include/foxglove/arena.hpp
```diff
@@ -13,13 +13,15 @@
 namespace foxglove {
 
 /// A fixed-size memory arena that allocates aligned arrays of POD types on the stack.
-/// The arena contains a single inline array and allocates from it.
-/// If the arena runs out of space, it throws std::bad_alloc.
+/// The arena contains a singjle inline array and allocates from it.
```
eloff:
```suggestion
/// The arena contains a single inline array and allocates from it.
```

cpp/foxglove_data_loader/examples/data_loader.cpp
```diff
@@ -210,10 +182,31 @@ std::optional<Result<Message>> TextMessageIterator::next() {
...
+    LineIndex line = data_loader->line_indexes[index];
     // filter by channel ID
     for (const ChannelId channel_id : args.channel_ids) {
       if (channel_id == line.file) {
+        message.file = data_loader->paths[line.file];
+        message.level = foxglove::schemas::Log::LogLevel::INFO;
+        message.name = "log line";
+        message.line = index;
+        message.message = std::string(
+          (const char*)(&(data_loader->files[line.file][line.start])), line.end - line.start
```
eloff: nit: Is that a C-style cast or am I reading this wrong?

eloff (approved):
This looks good to me

cpp/foxglove/include/foxglove/arena.hpp
```diff
@@ -13,13 +13,15 @@
...
+/// If the arena runs out of space, it attempts to allocate the required space on the stack.
```
gasmith:
```suggestion
/// If the arena runs out of space, it attempts to allocate the required space on the heap.
```

cpp/foxglove/include/foxglove/schemas.hpp
```diff
@@ -13,19 +12,14 @@
 #include <string>
 #include <type_traits>
 #include <vector>
+#ifndef __wasm32__
+#include <foxglove/context.hpp>
+#endif
```
gasmith, line 17: Should this move up to be with the group of foxglove headers?

cpp/foxglove_data_loader/examples/data_loader.cpp
```diff
@@ -210,10 +182,31 @@ std::optional<Result<Message>> TextMessageIterator::next() {
...
+          message.encode(last_encoded_message.data(), last_encoded_message.size(), &encoded_len);
+        if (result == foxglove::FoxgloveError::BufferTooShort) {
+          last_encoded_message.resize(encoded_len);
+          result =
+            message.encode(last_encoded_message.data(), last_encoded_message.size(), &encoded_len);
+          if (result != foxglove::FoxgloveError::Ok) {
+            error("failed to encode message:", foxglove::strerror(result));
+            return Result<Message>{.error = "failed to encode message"};
+          }
+        }
```
gasmith: I think we want to pop that last error handler out one level.

I don't know why GH has to make this suggestion diff look so nasty:

```suggestion
            message.encode(last_encoded_message.data(), last_encoded_message.size(), &encoded_len);
        }
        if (result != foxglove::FoxgloveError::Ok) {
          error("failed to encode message:", foxglove::strerror(result));
          return Result<Message>{.error = "failed to encode message"};
        }
```

cpp/foxglove_data_loader/Makefile
```diff
@@ -1,12 +1,75 @@
...
+  target/include/foxglove/expected.hpp \
+  target/include/foxglove/error.hpp \
+  target/include/foxglove_data_loader/data_loader.hpp \
+  target/include/foxglove_data_loader/data_loader.inl \
+  target/include/foxglove_data_loader/host_internal.h \
+  target/include/foxglove_data_loader/host_internal.inl \
+  target/src/foxglove/schemas.cpp \
+  target/src/foxglove/error.cpp \
+  target/lib/libfoxglove.a \
+  target/README.md \
```
james-rms:
```suggestion
  target/README.md
```

cpp/foxglove_data_loader/README.md
gasmith, line 1: Maybe add a link out to https://docs.foxglove.dev/docs/extensions#data-loaders? Or perhaps wait until we have the C++ docs there, too.

PR #627, opened 2025-08-28
Support decoding of schemas & wkts

bryfox:
### Changelog
TBD

### Description

This adds a Decode trait with implementations for generated schemas, and implements `merge_field` for decoding of Timestamp and Duration. No support for Python/C++ yet; I'm currently using this for development of Agent features.


bryfox:
> LG. I think this would be quite useful to expose publicly, someday. Do we have a ticket tracking that?

We do now: FG-12965

PR #634, opened 2025-09-02
C++ data loaders: correct build string

james-rms:
### Docs

None.

### Description

This should help customers get their build step for C++ data loaders right the first time.


bennetthardwick (approved):
Haven't tested but looks good.

PR #635, opened 2025-09-03
SDK playground

jtbandes:
### Changelog
None

### Docs

None

### Description

- New web app that uses [Pyodide](https://pyodide.org/en/stable/) in a worker + monaco editor on the frontend.
- Build a python wheel for wasm using `maturin build --target wasm32-unknown-emscripten`.
- Sprinkle some `cfg` annotations to disable websocket stuff from the wasm build.

image

Resolves FG-8451


defunctzombie:
In my use I had an mcap named `quickstart-python.mcap` in the python but downloading it named it "output.mcap" - is there a way those two names could match?

defunctzombie:
What do I need to barter to get the source code saved in the URL so that sharing a url becomes a cheap way to share a reproduction of a visual?

jtbandes:
I’m gonna say that can be a follow up PR 😄 
Maybe we can learn how the TS playground does this? Do they really store all of it in the url?

Should we look at integrating with gist?

defunctzombie:
> Do they really store all of it in the url

From what I understand yes - maybe they gzip for larger sources? But starting somewhere is good.

👍 to follow-up PR

> Should we look at integrating with gist?

Fascinating - feels different to what I expect from playground so I don't feel strongly.

surbina (approved):
Front End code looks good!

.github/workflows/python.yml
```diff
@@ -43,6 +43,56 @@ jobs:
...
+    defaults:
+      run:
+        working-directory: python/foxglove-sdk
+    permissions:
+      deployments: write
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: 1.86.0
```
gasmith, line 59: A few comments describing why we're using these particular versions (of rustc, emsdk, pyodide, python) might be useful for future archaeologists, since it takes such effort to get a working combination. A short comment here would definitely be worthwhile (since so many of the versions are specified here), but perhaps a `python/playground/README.md` would be the better place for telling the whole story.

Might also be nice for the readme to explain how to run the playground locally.

playground/src/RunnerWorker.ts
```diff
@@ -0,0 +1,103 @@
...
+          import os, pathlib, shutil
+          shutil.rmtree("/home/pyodide/playground", ignore_errors=True)
+          pathlib.Path("/home/pyodide/playground").mkdir(parents=True)
+          os.chdir("/home/pyodide/playground")
+        `,
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+        { globals: pyodide.toPy({}) },
+      );
+    } catch (err: unknown) {
+      // ignore
```
gasmith, line 67: Why ignore?

jtbandes: I guess I could remove this and let the error propagate..I think I was just feeling that it might be worth it to try running the user's code even if this setup thing didn't work as expected.

PR #640, opened 2025-09-03
fix errors on bridge shutdown

achim-k:
### Changelog
Fix foxglove bridge node occasionally logging errors when shutting down

### Docs
None

### Description
Fixes https://github.com/foxglove/ros-foxglove-bridge/issues/370

When the foxglove bridge node is shutting down, it occasionally logs the following error(s):
```
[WS] Exception caught when closing connection: Couldn't initialize rcl timer handle: the given context is not valid, either rcl_init() was not called or rcl_shutdown() was called., at ./src/rcl/guard_condition.c:67
```

This PR should fix that by omitting to create the time when the node is shutting down.


ericmlujan (approved):
thanks for fixing this!

PR #654, opened 2025-09-10
playground: store code in url

jtbandes:
### Changelog
None

### Docs

None

### Description

- Store code in URL using zstd + base64, with a `1:` prefix to allow future evolution.
- Add cmd+S as well as a Share button to trigger saving.


playground/src/Editor.tsx
```diff
@@ -12,6 +14,10 @@ export type EditorInterface = {
 export const Editor = forwardRef<EditorInterface, EditorProps>(
   function Editor(props, ref): React.JSX.Element {
     const { initialValue } = props;
+    const latestProps = useRef(props);
+    useEffect(() => {
+      latestProps.current = props;
```
defunctzombie, line 22: react ... I die inside

jtbandes: Yeah, kinda nasty. But the idea here is to avoid redoing a bunch of monaco editor setup when all we need is for the save callback to invoke the latest version of `props.onSave`.

PR #657, opened 2025-09-15
python: Update type stubs for jedi compatibility

gasmith:
### Changelog
None

### Docs
None

### Description
This change seeks to resolve the sadness preventing jedi from loading type information from the foxglove SDK.

The two major blockers that needed to be resolved:
- Jedi does not understand `Class.__new__`; it only looks for `Class.__init__`.
- Jedi does not understand non-static defaults (`"".to_string()`, `vec![]`, etc.). Turns out sphinx doesn't either. These get converted to `Ellipsis` in the docs (cf., [SceneEntity](https://foxglove.github.io/foxglove-sdk/python/api/schemas.html#foxglove.schemas.SceneEntity)).

While I was there, I also decided to standardize on `X | Y` for `Union[X, Y]` and using standard keywords for list/dict/tuple types. The old syntax was intended for backwards compatibility for python 3.9, but it looks verbose and outdated. We can work around the compatibility issue by adding a `from __future__ import annotations` in .py files; for .pyi files, it doesn't really matter. I don't expect this to have any practical impact since type-checkers and LSPs have their own logic for parsing type hints. I've updated python to 3.13 for the docs CI workflow, to deal with these changes.


typescript/schemas/src/internal/generatePyclass.ts
```diff
@@ -97,11 +97,11 @@ export function generatePySchemaStub(schemas: FoxgloveSchema[]): string {
           [
             `class ${name}:`,
             ...doc,
-            `    def __new__(`,
-            "        cls,",
+            `    def __init__(`,
```
bryfox, line 100: Since pyO3 doesn't support `__init__` (https://pyo3.rs/v0.26.0/class.html#constructor), this is a bit of a lie. I'm not sure what problems it might cause; maybe it doesn't matter much in practice?

jtbandes (approved):
Thanks! I had noticed the `__new__` issue too, was planning to report it upstream to jedi...

jtbandes:
Works great on the playground preview deploy 🙌 

image

typescript/schemas/src/internal/generatePyclass.ts
gasmith: Yeah, I thought about that too. I think the distinction between `__init__` and `__new__` only matters if you're trying to create a subclass _and_ override the constructor, in which case the user needs to know that they have to override `__new__` instead of `__init__`.

bryfox (approved):

jtbandes:
Filed issue upstream with jedi: https://github.com/davidhalter/jedi/issues/2073

PR #660, opened 2025-09-17
Narrow channel ID to 32 bits

ericmlujan:
### Changelog
- The ID of a `Channel` is now represented as an unsigned 32-bit integer

### Docs
API docs should auto-update. I did a quick skim of docs.foxglove.dev and didn't see anything public-facing directly using channel ID.

### Description
This is in service of deprecating subscription ID and using channel ID as the main subscription identifier in the websocket protocol. Addressing subscribed messages can be accomplished by combining the 32-bit channel ID with the client ID when sending over the wire as a 64-bit integer.

Channel ID is an unsigned 64-bit integer

</td><td>

Channel ID is an unsigned 32-bit integer

</td></tr></table>

Fixes: FLE-64

### DIFF
--- a/rust/foxglove/src/channel.rs
+++ b/rust/foxglove/src/channel.rs
@@ -1,7 +1,5 @@
 use std::collections::BTreeMap;
-use std::sync::atomic::AtomicU32;
-use std::sync::atomic::Ordering::Relaxed;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 use delegate::delegate;
 use serde::{Deserialize, Serialize};
@@ -32,14 +30,19 @@ impl ChannelId {
 
     /// Allocates the next channel ID.
     pub(crate) fn next() -> Result<Self, FoxgloveError> {
-        static NEXT_ID: AtomicU32 = AtomicU32::new(1);
-        // Overflow
-        if NEXT_ID.load(Relaxed) == 0xFFFFFFFF {
+        static NEXT_ID: Mutex<u32> = Mutex::new(1);
+        let mut next_id = NEXT_ID
+            .lock()
+            .map_err(|_| FoxgloveError::ChannelError("couldn't lock channel id counter".into()))?;
+
+        // Check for overflow
+        if *next_id == u32::MAX {
             return Err(FoxgloveError::ChannelError(
                 "channel id pool is exhausted".into(),
             ));
         }
-        let id = NEXT_ID.fetch_add(1, Relaxed);
+        let id = *next_id;
+        *next_id += 1;
         Ok(Self(id))
     }
 }
--- a/rust/foxglove/src/channel.rs
+++ b/rust/foxglove/src/channel.rs
@@ -7,7 +7,9 @@ use delegate::delegate;
 use serde::{Deserialize, Serialize};
 use smallbytes::SmallBytes;
 
-use crate::{metadata::ToUnixNanos, ChannelBuilder, Encode, PartialMetadata, Schema, SinkId};
+use crate::{
+    metadata::ToUnixNanos, ChannelBuilder, Encode, FoxgloveError, PartialMetadata, Schema, SinkId,
+};
 
 mod lazy_channel;
 mod raw_channel;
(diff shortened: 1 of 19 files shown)


rust/foxglove/src/channel.rs
```diff
@@ -20,30 +22,36 @@ const STACK_BUFFER_SIZE: usize = 256 * 1024;
...
+        static NEXT_ID: AtomicU32 = AtomicU32::new(1);
+        // Overflow
+        if NEXT_ID.load(Relaxed) == 0xFFFFFFFF {
+            return Err(FoxgloveError::ChannelError(
+                "channel id pool is exhausted".into(),
+            ));
+        }
         let id = NEXT_ID.fetch_add(1, Relaxed);
-        Self(id)
+        Ok(Self(id))
```
gasmith: This isn't thread-safe; it's possible to overflow without any caller observing `0xffffffff` on L37. This isn't a performance-critical path. Let's just use a `Mutex`.

rust/foxglove/src/channel.rs
```diff
@@ -20,30 +22,36 @@ const STACK_BUFFER_SIZE: usize = 256 * 1024;
...
     /// Allocates the next channel ID.
-    pub(crate) fn next() -> Self {
-        static NEXT_ID: AtomicU64 = AtomicU64::new(1);
+    pub(crate) fn next() -> Result<Self, FoxgloveError> {
+        static NEXT_ID: AtomicU32 = AtomicU32::new(1);
+        // Overflow
+        if NEXT_ID.load(Relaxed) == 0xFFFFFFFF {
+            return Err(FoxgloveError::ChannelError(
+                "channel id pool is exhausted".into(),
+            ));
```
gasmith, line 42: Seems like we should have a new enum variant here: `FoxgloveError::ChannelIdPoolExhausted`?

gasmith (commented):
Worth noting that this is a breaking change (for C/C++/rust, anyway).

I also wonder whether this could be problematic for a long-lived process. If you ever hit the limit, the only recourse is to restart the process. I wouldn't expect any practical application to involve so many _concurrent_ channels... But imagine some sort of proxy that creates and destroys channels dynamically.

rust/foxglove/src/channel.rs
ericmlujan: I considered this but figured that there may be a handful of related errors we want to surface wrt instantiating channels. I also looked at other members of the `FoxgloveError` enum and they seemed more general than something like `ChannelIdPoolExhausted`. That was my take anyhow, but my opinion on this isn't terribly strong.

rust/foxglove/src/channel.rs
ericmlujan: Resolved in 4453dc3

rust/foxglove/src/channel.rs
gasmith: As a library user, I greatly prefer fine-grained error variants over catch-all error variants with opaque string discriminators. When handling errors, I want to know exactly what kinds of failure conditions to expect, and I want to be able treat them individually without resorting to matching/parsing strings. With `ChannelError(String)`, I have no idea what the set of possible strings might be (apart from what I observe at runtime), nor can I easily determine whether that set changes in future versions of the SDK.

How do we expect users to handle a`ChannelError`? Is it never retryable? Sometimes retryable? I'd argue that the error condition we're introducing here is _particularly_ interesting, because it is a permanent failure due to resource exhaustion, and the only correct way to handle that programmatically is to terminate (and restart) the process. That's quite different from, e.g., not providing sufficient information to instantiate the channel (cf. `MessageEncodingRequired`).

In the SDK, for better or worse, we decided to expose a single error type. The main downside is that we can't use the type system to tightly constrain the set of errors to expect, e.g. for a particular method call. Since we're already on that path, I'm also not sure I see much public-facing value in defining a sub-group of errors specific to channel instantation (e.g. `ChannelError(#[source] ChannelError)`). Building an error hierarchy might be nice for internal typing, but unpacking nested errors isn't particularly fun in Rust, and it'd be more of a challenge in C & python.

gasmith: I guess we _do_ have a shallow error hierarchy today, for wrapping external error types like `McapError` and `std::io::Error`. I actually don't know how we translate those at the FFI boundary.

ericmlujan:
@gasmith This is technically a breaking change, but:
1. The underlying type of `ChannelId` is abstracted behind a typedef in rust and C++, so folks staying within the bounds of what the SDK provides shouldn't notice a difference.
2. I'm not aware of public functions we provide where folks are manually setting channel IDs when instantiating a channel. If they're calling `Channel.id()` and storing the result using a raw datatype, that's the breaking change I can think of here.
3. Getting up to more than 4B channels, even for a long-lived proxy, seems unlikely to me. I could, of course, be wrong, but I'm not sure this is a use pattern we should be designing around vs. simplifying addressing by deprecating subscription id

rust/foxglove/src/channel.rs
ericmlujan: You make good points about retryability and recovery strategies. Given that, I can make a specific error for Channel ID pool exhaustion.

ericmlujan:
Closing, as we're no longer imminently planning on multiplexing channels from multiple SDK instances to a single agent

PR #662, opened 2025-09-19
Update Foxglove Bridge to SDK v0.14.2, change version to match foxglove_msgs

ericmlujan:
### Changelog
- Fixes linker errors when building Foxglove Bridge from source against ROS Humble

### Docs
None

### Description
Update Foxglove Bridge to point to an SDK version that builds properly n 

I'm also using this opportunity to update the ROS package's version number from `0.14.1` to `3.10.1` to match `foxglove_msgs`. This is to satisfy the constraint that ROS `bloom` will only allow publishing packages if all packages in the same repo have the exact same version number (see [the bloom documentation]()): 
> Even if you do not use catkin_prepare_release, you must have one or more valid package.xml(s) with the same version and a matching tag in your upstream repository.

This is unfortunate, as this particular version bump does not warrant jumping ahead 3 major versions. I suppose you could try to make the argument that "this is the first SDK-based Foxglove Bridge that we intend to release to rosdistro", and this does constitute a major version change from the perspective of people who apt install Foxglove Bridge, but the real reason is to work to the requirements that ROS imposes.


bryfox (approved):

bryfox:
Seems fine to me, and there was no internal pushback on the version change.

PR #663, opened 2025-09-22
notebook: set up jupyter notebook widget

surbina:
### Changelog

Add Python notebook integrations

### Docs

https://github.com/foxglove/docs/pull/1022

### Description

This PR adds the notebook integration for Python, particularly:
- Adds the notebook integration using Anywidget
- Uses @foxglove/embed to show Foxglove in a notebook
- Provides an API to load a layout from a file
- Provides an API to visualize messages from buffer associated with a Context
- Add build script and update CI jobs where necessary
- Adds API docs

The integration is shipped as an extra dependency of the foxglove-sdk package. To install, run the following command:

```bash
pip install foxglove-sdk[notebook]
```

https://drive.google.com/file/d/1BSLsBt6WSsnnj3a6G-hWsivNr8vzNZnp/view?usp=sharing

### How to test?

1. Clone this repo
1. Follow instructions [here](https://github.com/foxglove/foxglove-sdk/blob/sebastian/ert-577-set-up-widget-package/python/foxglove-sdk/CONTRIBUTING.md) to setup the repo for local development
1. Activate the virtual environment created by poetry
    ```bash
    eval $(poetry env activate)
    ```
1. Install Jupyter
    ```bash
    pip install jupyterlab
    ```
1. Build and install the sdk locally, including the notebook extra dependencies
    ```bash
    pip install -e ".[notebook]"
    ```
1. Launch Jupyter
    ```bash
    jupyter lab
    ```

Example notebooks:
1. Log-view flow: https://drive.google.com/file/d/1iBURbNQjiOtXc8Onk1vjhTwdd8q2VATq/view?usp=drive_link
2. Log-reload flow: https://drive.google.com/file/d/12Ftbu-qpXIdwSJln4_Dy7FefcoJjctO1/view?usp=drive_link


python/foxglove-sdk/tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+    "moduleResolution": "bundler",
+    "allowImportingTsExtensions": true,
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "noEmit": true,
+    "jsx": "react",
+    "strict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noFallthroughCasesInSwitch": true
```
jtbandes: a lot of these are covered by `@foxglove/tsconfig`, let's extend from that and only include the other properties if they need to be changed (like most of our TS packages)

python/foxglove-sdk/ts/widget.ts
```diff
@@ -0,0 +1,74 @@
...
+import { FoxgloveViewer, type DataSource, type OpaqueLayoutData } from "@foxglove/embed";
+
+/* Specifies attributes defined with traitlets in ../src/foxglove_notebook/__init__.py */
+interface WidgetModel {
+  width: string;
+  height: string;
+  src: string;
+  orgSlug?: string;
+  data?: DataView<ArrayBuffer>;
+  layout?: OpaqueLayoutData;
```
jtbandes: how do we decide which of these are optional and which are not? I don't think I see a difference in how they are defined in python

python/foxglove-sdk/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@foxglove/notebook",
```
jtbandes: we don't want/need to publish this, right? let's set `private: true`

python/foxglove-sdk/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "typecheck": "tsc --noEmit"
+  },
+  "dependencies": {
+    "@foxglove/embed": "0.1.2"
+  },
+  "devDependencies": {
+    "@anywidget/types": "0.2.0",
+    "esbuild": "0.25.9",
+    "typescript": "5.9.2"
+  }
```
jtbandes: probably all should be devDependencies as this is not a library package, rather it just produces an output bundle with no runtime dependencies if I understand correctly?

python/foxglove-sdk/python/foxglove/notebook/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+        self.layout = layout
+
+    def start_logging(self) -> None:
+        """
+        Prepare the widget for logging.
+        """
+        if self.writer is not None:
+            # Close the previous writer if it exists
+            self.writer.close()
+            self.writer = None
```
jtbandes: should we also delete the previous file? On the other hand, is start_logging meant to be called more than once? If not, maybe it should be prefixed with `_` and we can remove this?

indirectlylit:
hey @surbina  - I am a little out of the loop on this project but I'd be happy to do some testing.

what kind of feedback are you looking for, and how should I go about testing?

surbina:
@indirectlylit Right now I'm mostly looking feedback about the proposed API (`create_notebook_buffer` and `visualize` functions)

python/foxglove-sdk/ts/widget.ts
surbina: I've made all of these required, since these fields in the FoxgloveViewer.py class always have a default value. Thanks for calling it out.

python/foxglove-sdk/python/foxglove/__init__.py
```diff
@@ -123,13 +128,137 @@ def _level_names() -> dict[str, int]:
...
+            src=src,
+            orgSlug=orgSlug,
+            layout=layout,
+        )
+
+    except ImportError:
+        raise Exception(
+            "FoxgloveViewer is not installed."
+            "Please install it with `pip install foxglove-sdk[notebook]`"
+        )
```
indirectlylit, line 221: These two lines get concatenated together without a space:


```
FoxgloveViewer is not installed.Please install it with `pip install foxglove-sdk[notebook]`
```


Also, python best practice would be to move the `return` statement outside of the try/except and _only_ wrap the `import FoxgloveViewer` line. The way it is written now, if any import error occurs during construction of `FoxgloveViewer` we will raise an incorrect exception.

python/foxglove-sdk/python/foxglove/notebook/foxglove_viewer.py
```diff
@@ -0,0 +1,129 @@
...
+
+    Example:
+        >>> import foxglove
+        >>> buffer = foxglove.create_notebook_buffer()
+        >>> # ... log some data to the buffer ...
+        >>> viewer = foxglove.visualize(buffer, width="800px", height="600px")
+        >>> viewer  # Display the widget in the notebook
+    """
+
+    _esm = pathlib.Path(__file__).parent / "static" / "widget.js"
```
indirectlylit: when trying to run

```python
foxglove.visualize(datasource=buffer, orgSlug="foxglovehq", layout=layout)
```

I get:


```
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
Cell In[8], line 1
----> 1 foxglove.visualize(datasource=buffer, orgSlug="foxglovehq", layout=layout)

File ~/Projects/sdk/python/foxglove-sdk/python/foxglove/__init__.py:235, in visualize(datasource, width, height, src, orgSlug, layout)
    176 """
    177 Create a FoxgloveViewer widget for interactive data visualization in Jupyter notebooks.
    178 
   (...)    232     >>> viewer
    233 """
    234 try:
--> 235     from .notebook.FoxgloveViewer import FoxgloveViewer
    237     return FoxgloveViewer(
    238         datasource=datasource,
    239         width=width,
   (...)    243         layout=layout,
    244     )
    246 except ImportError:

File ~/Projects/sdk/python/foxglove-sdk/python/foxglove/notebook/FoxgloveViewer.py:16
     12 except importlib.metadata.PackageNotFoundError:
     13     __version__ = "unknown"
---> 16 class FoxgloveViewer(anywidget.AnyWidget):
     17     """
     18     A Jupyter notebook widget that embeds the Foxglove visualization app for interactive
     19     data exploration.
   (...)     38         >>> viewer  # Display the widget in the notebook
     39     """
     41     _esm = pathlib.Path(__file__).parent / "static" / "widget.js"

File ~/Projects/test-nb/.venv/lib/python3.13/site-packages/traitlets/traitlets.py:963, in MetaHasDescriptors.__new__(mcls, name, bases, classdict, **kwds)
    960         classdict[k] = v()
    961     # ----------------------------------------------------------------
--> 963 return super().__new__(mcls, name, bases, classdict, **kwds)

File ~/Projects/test-nb/.venv/lib/python3.13/site-packages/anywidget/widget.py:71, in AnyWidget.__init_subclass__(cls, **kwargs)
     67 super().__init_subclass__(**kwargs)
     68 for key in (_ESM_KEY, _CSS_KEY) & cls.__dict__.keys():
     69     # TODO(manzt): Upgrate to := when we drop Python 3.7
     70     # https://github.com/manzt/anywidget/pull/167
---> 71     file_contents = try_file_contents(getattr(cls, key))
     72     if file_contents:
     73         setattr(cls, key, file_contents)

File ~/Projects/test-nb/.venv/lib/python3.13/site-packages/anywidget/_util.py:280, in try_file_contents(x)
    278 if not path.is_file():
    279     msg = f"File not found: {path}"
--> 280     raise FileNotFoundError(msg)
    281 return FileContents(
    282     path=path,
    283     start_thread=_should_start_thread(path),
    284 )

FileNotFoundError: File not found: /Users/d/Projects/sdk/python/foxglove-sdk/python/foxglove/notebook/static/widget.js
```

drdelambre:
the use of the `buffer` variable is goofy. I dont know why we're leveraging the default context to have all of these sources and sinks automatically link together, but for this one case we have to export it into a file handler external to that concept, just to route it back in. 

it seems like the create_notebook_buffer is connected to the concept of contexts within the API (the given example being the default context). why can't we register the visualize function to a context as well? so the example becomes:

```python
foxglove.create_notebook_buffer()
# push a bunch of stuff to the buffer
foxglove.visualize(org_slug="~", layout=layout)
# push a bunch of live stuff through the context (or buffer)
```

drdelambre:
@gasmith is this a limitation of the SDK?

python/foxglove-sdk/python/foxglove/notebook/foxglove_viewer.py
surbina: I initially thought that the FE assets should be ignored since they need to be generated when we build the SDK. This is why you are getting this error.

I'll remove the ignore so that it's versioned with the rest of the code (similarly to how we include the Python code generated by Rust).

surbina: Can you try again?

surbina:
@drdelambre The way I see it:

- Context acts as an intermediary between channels (publishers) and sinks (subscribers). It doesn't store the logged messages.
- If we were implementing live visualization, we could omit the datasource and then hook into the global context with a new type of sink.
- But in this case, we are implementing visualization for messages that have already been logged, so those messages need to be stored somewhere (in this case, the notebook buffer).
- We could create a global notebook buffer by default. Still, I don't think that's right since we won't know beforehand if the logged messages will be used for an embedded visualization in Jupyter, live-streamed, or something else.

However, I do like your proposal, and I would see this API evolving in the future to be something like:

```python
# Show Foxglove and visualize the buffered
foxglove.visualize(datasource=notebook_buffer, org_slug="~", layout=layout)

# Show Foxglove and visualize live data from the notebook by hooking into the global Context
foxglove.visualize(org_slug="~", layout=layout)

# Show Foxglove and visualize live data from the notebook by hooking into my_context
foxglove.visualize(org_slug="~", layout=layout, context=my_context)
```

surbina:
@indirectlylit @drdelambre @jtbandes this is ready for another pass!

python/foxglove-sdk/python/foxglove/notebook/foxglove_viewer.py
jtbandes: Is there a way we can exclude this compiled output from the repo but automatically build it when necessary (as part of the python package build script for example)?

python/foxglove-sdk/python/foxglove/notebook/FoxgloveViewer.py
```diff
@@ -0,0 +1,155 @@
...
+        if src is not None:
+            self.src = src
+        if layout_data is not None:
+            self.layout_data = layout_data
+
+        # Use default context if no context is provided
+        ctx = context or Context.default()
+        self.set_data_from_context(ctx)
+
+    def set_data_from_context(self, context: Context) -> None:
```
jtbandes: Is the user meant to call this themselves? It kinda looks like an internal method -- in which case I think it could be prefixed with `_` and we don't need the detailed docstring unless you find it helpful for internal use.

python/foxglove-sdk/python/foxglove/notebook/NotebookBuffer.py
```diff
@@ -0,0 +1,116 @@
...
+        >>> from foxglove.schemas import SceneUpdate
+        >>>
+        >>> # Create a buffer
+        >>> buffer = foxglove.create_notebook_buffer()
+        >>>
+        >>> # Log some data (this would typically be done by your application)
+        >>> # The buffer automatically collects the logged messages
+        >>>
+        >>> # Create a viewer to visualize the data
+        >>> viewer = foxglove.visualize(buffer)
```
jtbandes: A lot of these comments seem out of date — I think it is no longer correct that you pass the buffer into visualize(), nor is it correct that you can get a reference to the buffer via the return value of `create_notebook_buffer()`?

python/foxglove-sdk/python/foxglove/notebook/notebook_buffer.py
```diff
@@ -0,0 +1,116 @@
...
+            >>> # The buffer is now reset and ready for new data
+        """
+        self._writer.close()
+
+        with open(self._file_name, "rb") as f_read:
+            # Read the entire content of the file
+            content = f_read.read()
+
+        os.remove(self._file_name)
+        self._create_writer()
```
jtbandes, line 98: wondering if this will result in temporary files being left around — what happens if you re-run a cell that creates a new context & buffer...I guess the old context might be left open & still have a reference to an old file?

python/foxglove-sdk/python/foxglove/notebook/foxglove_viewer.py
```diff
@@ -0,0 +1,155 @@
...
+    Example:
+        >>> import foxglove
+        >>> foxglove.create_notebook_buffer()
+        >>> # ... log some data to the buffer ...
+        >>> viewer = foxglove.visualize(width="800px", height="600px")
+        >>> viewer  # Display the widget in the notebook
+    """
+
+    _esm = pathlib.Path(__file__).parent / "static" / "widget.js"
+    _data = traitlets.Bytes(b"").tag(sync=True)
```
jtbandes: did we ever explore a way to send this data to the frontend without also holding it permanently in memory on the python side?

python/foxglove-sdk/python/foxglove/notebook/foxglove_viewer.py
```diff
@@ -0,0 +1,155 @@
...
+    width = traitlets.Unicode("100%").tag(sync=True)
+    height = traitlets.Unicode("500px").tag(sync=True)
+    src = traitlets.Unicode("").tag(sync=True)
+    layout_data = traitlets.Dict({}).tag(sync=True)
```
jtbandes: data has a `_` prefix while the others do not — does that mean setting the others externally is an expected workflow that actually works? Or do you always have to pass these to init? If they cannot be modified later, maybe it makes sense to mark them all with `_` (I haven't tried it)

.github/workflows/python.yml
```diff
@@ -145,8 +145,29 @@ jobs:
         with:
           packages-dir: python/foxglove-schemas-protobuf/dist
 
+  notebook-fe:
```
jtbandes: nit: I find the "fe" abbreviation a bit confusing, would prefer "frontend" throughout

python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -3,12 +3,179 @@ API Reference
...
+   .. note::
+      This function is only available when the `notebook` extra package
+      is installed. Install it with `pip install foxglove-sdk[notebook]`.
+      The widget will automatically load the data from the provided datasource
+      and display it in the embedded Foxglove viewer.
+
+Notebook Classes
+^^^^^^^^^^^^^^^^
+
+.. py:class:: NotebookBuffer
```
jtbandes, line 41: I'm not convinced we should put this in the documentation. Is it really part of the public API?

surbina: You are right; users should no longer need to interact with NotebookBuffer. I'm removing this doc.

python/foxglove-sdk/python/foxglove/notebook/foxglove_viewer.py
surbina: Yeah, manually setting these properties works:

https://github.com/user-attachments/assets/875ef3bc-e20f-43cb-9592-a5e84f5e020b

python/foxglove-sdk/python/foxglove/notebook/FoxgloveViewer.py
surbina: Yeah, users can call this method if they want to update the data in the viewer.

python/foxglove-sdk/python/foxglove/notebook/notebook_buffer.py
surbina: I tested this and the [temporary directories are deleted when the jupyter process is stopped](https://docs.python.org/3.9/library/tempfile.html#:~:text=destruction%20of%20the%20temporary%20directory%20object%20the%20newly%20created%20temporary%20directory%20and%20all%20its%20contents%20are%20removed%20from%20the%20filesystem).

python/foxglove-sdk/python/foxglove/notebook/NotebookBuffer.py
surbina: Related to a comment below, since this class is for internal use only, I slimmed down the docs a bit.

python/foxglove-sdk/python/foxglove/notebook/foxglove_viewer.py
surbina: I tried implementing this, but I ended up finding a bug when sending messages from JS to Python (which I need for synchronization). I filed a [bug](https://github.com/manzt/anywidget/issues/895), and will come back to this when it is fixed.

python/foxglove-sdk/ts/widget.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  model.on("change:width", () => {
+    parent.style.width = model.get("width");
+  });
+
+  model.on("change:height", () => {
+    parent.style.height = model.get("height");
+  });
+
+  model.on("change:data", () => {
```
gasmith:
```suggestion
  model.on("change:_data", () => {
```

python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -3,12 +3,136 @@ API Reference
...
+
+   :param context: Optional Foxglove context to use for logging. If not provided,
+       the global context will be used. This allows you to isolate the
+       buffer's data collection from other parts of your application.
+
+   .. note::
+      This function is only available when the `notebook` extra package
+      is installed.
+
+.. py:function:: visualize(context: Optional[Context] = None, width: Optional[str] = None, height: Optional[str] = None, src: Optional[str] = None, layout_data: Optional[dict] = None) -> FoxgloveViewer
```
gasmith: I feel strongly that `visualize` is too broad of a name for this function. We're returning a jupyter notebook widget. Maybe `foxgove.notebook_viewer()`? That will also help to relate it to the `[notebook]` extra.

python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -3,12 +3,136 @@ API Reference
...
+          dictionary that was exported from the Foxglove app. If not provided, uses the
+          default layout.
+      :param kwargs: Additional keyword arguments passed to the parent AnyWidget class.
+
+      .. note::
+         The widget will automatically load the data from the buffer associated with the provided
+         context and display it in the embedded Foxglove viewer. The data is loaded once
+         during initialization - to update the data, use the :meth:`set_data_from_context` method.
+
+   .. py:method:: set_data_from_context(context: Context) -> None
```
gasmith: Most people won't use multiple contexts. They'll just implicitly use the default context. When they regenerate data, they just want the viewer to be updated with the new data.

While it's possible to do so with `set_data_from_context(Context.default())`, it's a little non-obvious and very verbose. How about `reload_data(context: Context = None)`? If you don't specify a context, you get the most recently-visualized one (or the one that the viewer was initialized with).

I also feel strongly that we should separate the concerns of setting up network buffers from loading data. This method should throw an exception if the viewer doesn't have a buffer for the context.

python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -3,12 +3,136 @@ API Reference
...
    :members:
-   :exclude-members: MCAPWriter
+   :exclude-members: MCAPWriter, create_notebook_buffer, visualize
+
+Notebook Integration
+^^^^^^^^^^^^^^^^^^^^
+
+Functions and classes for integrating with Jupyter notebooks and creating interactive visualizations.
+
+.. py:function:: create_notebook_buffer(context: Context | None = None) -> None
```
gasmith: I don't like the idea of a global collection of notebook buffers. If I have multiple viewers for some reason, they can't really share notebook buffers, because the first one to load data from the buffer will drain it. I think it makes more sense for each viewer to maintain its own set of buffers.

It's also a bit non-obvious that you need to explicitly create notebook buffers in advance when you're working with multiple contexts. If you dont, the first call to `set_data_from_context()` on a new context will be an empty result.

Maybe an instance method on the viewer called `subscribe` would make it a bit more clear that this is something you need to do before attempting to log data.

```python
viewer.subscribe(ctx1)
# log to ctx1
viewer.set_data_from_context(ctx1)
```

python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -3,12 +3,136 @@ API Reference
...
+   .. note::
+      This function is only available when the `notebook` extra package
+      is installed.
+
+.. py:function:: visualize(context: Optional[Context] = None, width: Optional[str] = None, height: Optional[str] = None, src: Optional[str] = None, layout_data: Optional[dict] = None) -> FoxgloveViewer
+
+   Create a FoxgloveViewer widget for interactive data visualization in Jupyter notebooks.
+
+   This function creates an embedded Foxglove visualization widget that displays
+   the data collected in the NotebookBuffer associated with the provided context.
```
gasmith: We should remove references to `NotebookBuffer` as a symbol, if it's not part of the public API. Perhaps we say something like:

> This function creates an embedded Foxglove visualization widget.
> The widget subscribes to messages on the provided context, and stores
> them in an internal buffer.

python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -3,12 +3,136 @@ API Reference
...
+         during initialization - to update the data, use the :meth:`set_data_from_context` method.
+
+   .. py:method:: set_data_from_context(context: Context) -> None
+
+      Update the data visualized using the provided context.
+
+      This method allows you to dynamically change the data being visualized
+      without recreating the widget. The new data will be loaded from the
+      NotebookBuffer associated with the provided context and the viewer will update to
+      display it.
```
gasmith: I think we need to be explicit about a few behaviors, which I think users will find surprising:

- This method is not additive; it will remove old data from the viewer.
- This method clears the internal buffer for the specified context. If you call this method twice in a row without logging, the viewer will be empty of data.

It'd be really nice if we offered users a bit more control over those things in the future.

And maybe we should throw an exception when there's no data, before updating the embedded viewer?

python/foxglove-sdk/python/foxglove/notebook/notebook_buffer.py
```diff
@@ -0,0 +1,79 @@
...
+            # Then add each file with its separator
+            for i, file_data in enumerate(contents):
+                # Add separator with file index
+                file_separator = separator + i.to_bytes(4, byteorder="big")
+                bytes_contents += file_separator
+                # Add the file size as 8-byte integer
+                file_size = len(file_data).to_bytes(8, byteorder="big")
+                bytes_contents += file_size
+                # Add the actual file data
+                bytes_contents += file_data
```
surbina: For context, this segment was written with the help of cursor, so might be worth paying more attention here.

python/foxglove-sdk/ts/widget.ts
```diff
@@ -0,0 +1,124 @@
...
+      throw new Error(`Expected ${fileSize} bytes but got ${fileData.byteLength}`);
+    }
+
+    // Create File object
+    files.push(new File([fileData], `data-${i}.mcap`));
+    offset += fileSize;
+  }
+
+  return files;
+}
```
surbina: Similarly, this segment was also written with the help of Cursor.

python/foxglove-sdk/python/foxglove/notebook/static/widget.js
gasmith, line 1: You had asked about how to hook into the maturin build process.

I think the right way to do that is with a `build.rs` script. Something like:

```rust
use std::process::Command;

fn main() {
    for subcmd in ["typecheck", "build:prod"] {
        let status = Command::new("yarn")
            .arg(subcmd)
            .status()
            .expect("failed to run yarn");
        if !status.success() {
            panic!("yarn {subcmd} failed: {status}");
        }
    }
    println!("cargo:rerun-if-changed=ts/");
}
```

We might go further and add `widget.js` to `.gitignore`, since it can be cheaply produced at build time.

python/foxglove-sdk/python/foxglove/__init__.py
```diff
@@ -123,6 +127,80 @@ def _level_names() -> dict[str, int]:
     }
 
 
+def notebook_viewer(
+    context: Context | None = None,
+    width: Optional[str] = None,
+    height: Optional[str] = None,
+    src: Optional[str] = None,
+    layout_data: Optional[dict] = None,
```
gasmith: Nit: Prefer `T | None` over `Optional[T]`.

python/foxglove-sdk/python/foxglove/notebook/foxglove_viewer.py
surbina: I tried adding a build.rs script so that the frontend is built whenever maturing builds the sdk. However, we stumbled upon some issues with the docker images.

I ended up manually building and archiving the front-end assets so they can be downloaded whenever needed, and I repurposed the build.rs to fail if the widget.js is missing. You can find more about this here: https://foxglove.slack.com/archives/C04R1TNP2TB/p1760044379238009

surbina:
@jtbandes @gasmith This is ready for another round!

python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -3,12 +3,90 @@ API Reference
...
    :members:
-   :exclude-members: MCAPWriter
+   :exclude-members: MCAPWriter, init_notebook_sink
+
+Notebook Integration
+^^^^^^^^^^^^^^^^^^^^
+
+Functions and classes for integrating with Jupyter notebooks and creating interactive visualizations.
+
+.. py:function:: init_notebook_sink(context: Context | None = None) -> NotebookSink
```
gasmith: I'd still mildly prefer calling this a "notebook buffer", now that we've settled on its usage. I feel like "sink" is too ambiguous, and we'll likely have a different initialization method for the "live" viewer when we get there. But I won't die on this hill.

.github/workflows/ci.yml
```diff
@@ -26,6 +26,15 @@ jobs:
...
           components: rustfmt, clippy
 
+      - uses: actions/checkout@v5
+      - run: corepack enable
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: yarn
+      - run: yarn workspaces focus @foxglove/notebook
+      - run: yarn build:prod
```
jtbandes: why is this part of the `rust` job?

python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -3,12 +3,90 @@ API Reference
...
+   Foxglove visualization widget. The widget provides a fully-featured Foxglove interface
+   directly within your Jupyter notebook, allowing you to explore multi-modal robotics data
+   including 3D scenes, plots, images, and more.
+
+   .. py:method:: __init__(context: Optional[Context])
+
+      Initialize the NotebookSink with the specified context.
+
+      :param context: The Context used to log the messages. If no Context is provided, the global
+          context will be used. Logged messages will be buffered.
```
gasmith: I think we should omit documentation for `__init__`, since we want to steer users towards the `init_notebook_sink` function. For comparison, see [the documentation for `MCAPWriter`](https://foxglove.github.io/foxglove-sdk/python/api/index.html#foxglove.mcap.MCAPWriter):

> Obtain an instance by calling `foxglove.open_mcap()`.

python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -3,12 +3,90 @@ API Reference
...
+   A widget that displays a Foxglove viewer in a notebook.
+
+   .. py:method:: __init__(get_data: DataSource, width: Optional[str] = None, height: Optional[str] = None, src: Optional[str] = None, layout_data: Optional[dict] = None)
+      Initialize the FoxgloveWidget with the specified data source, width, height, source URL, and layout data.
+
+      :param get_data: A callback function that returns the data to display in the widget.
+      :param width: The width of the widget. Defaults to "100%".
+      :param height: The height of the widget. Defaults to "500px".
+      :param src: The source URL of the Foxglove viewer. Defaults to "https://embed.foxglove.dev/".
+      :param layout_data: The layout data to use for the widget.
```
gasmith: Same as above: Omit this in favor of a top-level comment about how you can get an instance from `NotebookSink.show()`.

python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -3,12 +3,90 @@ API Reference
...
+
+   .. py:method:: refresh() -> None
+      Refresh the widget by getting the data from the callback function and sending it to the widget.
```
gasmith: This should refer to the notebook sink/buffer. The user isn't providing a callback.

python/foxglove-sdk/python/foxglove/notebook/foxglove_widget.py
```diff
@@ -0,0 +1,84 @@
...
+    src = traitlets.Unicode("").tag(sync=True)
+    layout_data = traitlets.Dict({}).tag(sync=True)
+
+    def __init__(
+        self,
+        get_data: DataSource,
+        width: Optional[Union[int, Literal["full"]]] = None,
+        height: Optional[int] = None,
+        src: Optional[str] = None,
+        layout_data: Optional[dict] = None,
```
gasmith: Prefer `X | Y | None` over `Optional[Union[X, Y]]`, for consistency with type hints elsewhere in the SDK. Add a `from __future__ import annotations` at the top of the module, to enable this syntax in Python 3.9. This comment also applies to other python modules, below.

python/foxglove-sdk/build.rs
```diff
@@ -0,0 +1,6 @@
+fn main() {
+    let widget_js = "python/foxglove/notebook/static/widget.js";
+    if !std::path::Path::new(widget_js).exists() {
+        panic!("error: {widget_js} not found, run `yarn build`");
```
jtbandes: 🤔 why does the rust code care about the existence of this file?

python/foxglove-sdk/notebook-frontend/widget.ts
```diff
@@ -0,0 +1,72 @@
...
+  viewer.addEventListener("ready", () => {
+    model.send({
+      type: "ready",
+    });
+  });
+
+  model.on("msg:custom", (msg: Message, buffers: DataView<ArrayBuffer>[]) => {
+    // Only one message is supported currently, however let's keep the if clause to be explicit
+    // and avoid future pitfalls
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
gasmith, line 47: This comment can be removed now, right?

python/foxglove-sdk/notebook-frontend/package.json
jtbandes, line 1: At a high level, I found it slightly surprising/weird that this package.json lives in the top level python/foxglove-sdk. I wonder if the typescript stuff could be confined to a smaller area. Maybe there is some technical reason here but it was not obvious to me from looking at the code changes.

python/foxglove-sdk/python/foxglove/notebook/notebook_sink.py
```diff
@@ -0,0 +1,55 @@
...
+
+    :param context: The Context used to log the messages. If no Context is provided, the global
+        context will be used. Logged messages will be buffered.
+    """
+
+    def __init__(
+        self,
+        context: Optional[Context] = None,
+    ):
+        self._buffer = NotebookBuffer(context=context)
```
jtbandes: I'm not sure if there is much value in the separation of NotebookBuffer & NotebookSink, it seems like they have almost the same responsibilities except that NotebookSink also has a show() method.

A thought experiment: I know that we went through some iterations where the user interacted with the notebook buffer directly. Now that we have settled on init_notebook_sink architecture, if we had just implemented that new architecture from scratch, would you still choose to have two separate classes here?

python/foxglove-sdk/python/foxglove/notebook/foxglove_widget.py
```diff
@@ -0,0 +1,84 @@
...
+    width = traitlets.Union([traitlets.Int(), traitlets.Enum(values=["full"])]).tag(
+        sync=True
+    )
+    height = traitlets.Int(500).tag(sync=True)
+    src = traitlets.Unicode("").tag(sync=True)
+    layout_data = traitlets.Dict({}).tag(sync=True)
+
+    def __init__(
+        self,
+        get_data: DataSource,
```
jtbandes: nit: IMO you probably could just keep a reference to the NotebookSink directly, instead of storing a bound method with the intermediate DataSource type.

Or is it necessary to avoid a circular import?

python/foxglove-sdk/python/foxglove/notebook/foxglove_widget.py
jtbandes: Feels like we should update the layout stuff to match the new layout storage APIs that we just shipped.

python/foxglove-sdk/build.rs
gasmith: This was a hack I suggested to ensure that the widget.js is generated prior to the python packaging that `maturin` does. It's true that rust does not actually care about this file, but I couldn't find a better way to hook into maturin.

python/foxglove-sdk/python/foxglove/notebook/foxglove_widget.py
gasmith: The circular import shouldn't be a problem. You can always put it under a `if TYPE_CHECKING` guard.

python/foxglove-sdk/build.rs
gasmith: I agree it's rather unfortunate that this requirement now bleeds into other rust workflows (`cargo test`, `cargo clippy`, etc.).

jtbandes: At the very least, needs a comment to explain this

python/foxglove-sdk/python/docs/api/index.rst
surbina: Fair enough, I removed NotebookSink and kept only NotebookBuffer. In the same line, I renamed init_notebook_sink to init_notebook_buffer.

python/foxglove-sdk/python/docs/api/index.rst
surbina: Done!

python/foxglove-sdk/python/docs/api/index.rst
surbina: Done!

python/foxglove-sdk/python/foxglove/notebook/foxglove_widget.py
surbina: Nice, TIL about `from __future__ import annotations`!

.github/workflows/ci.yml
surbina: This is related to `python/foxglove-sdk/build.rs` we added.

python/foxglove-sdk/build.rs
surbina: I added the comment. Our first version was trying to build the js assets, but it failed on some of the Docker images used on some builds. I wonder how difficult it would be update those images.

python/foxglove-sdk/notebook-frontend/package.json
surbina: I moved the front end stuff to `notebook-frontend` as we discussed in slack.

python/foxglove-sdk/python/foxglove/notebook/notebook_sink.py
surbina: I removed NotebookSink and kept only NotebookBuffer. In the same line, I renamed init_notebook_sink to init_notebook_buffer.

python/foxglove-sdk/python/foxglove/notebook/foxglove_widget.py
surbina: Done!

python/foxglove-sdk/python/foxglove/notebook/foxglove_widget.py
surbina: I was trying to keep the `get_data` in NotebookBuffer private, but I don't feel strongly about it. I've changed following Jacob's suggestion.

python/foxglove-sdk/notebook-frontend/widget.ts
surbina: No, we still need this, because `Message` has a single member. The lint error is complaining that `type` can only have a single valid value ("update-data"), so there is no need for the check. But I'd prefer keeping it for clarity.

python/foxglove-sdk/build.rs
jtbandes: > I couldn't find a better way to hook into maturin

Just a passing thought: wonder if the switch to `uv` brings us any new options for solving this problem?

gasmith: No, uv doesn't do us any favors here. I think the answer is to write our own PEP 517 build backend to wrap our yarn+maturin build process. I'm going to give it a quick try.

jtbandes: 😱  that sounds like it might not be better than just putting this in build.rs 😄

python/foxglove-sdk/notebook-frontend/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "email": "support@foxglove.dev"
+  },
+  "homepage": "https://foxglove.dev/",
+  "private": true,
+  "scripts": {
+    "dev": "yarn run build --sourcemap=inline --watch",
+    "build": "esbuild widget.ts --format=esm --bundle --outdir='../python/foxglove/notebook/static'",
+    "build:prod": "yarn build --minify",
+    "typecheck": "tsc --noEmit",
+    "run-python-sdk-examples": "yarn workspace foxglove-sdk run-python-sdk-examples"
```
jtbandes: nit: this seems weird to have a script that runs another script from the outer workspace?

python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -3,12 +3,77 @@ API Reference
...
+   :return: A NotebookBuffer object that can be used to manage the data buffering
+       and visualization.
+   :raises Exception: If the notebook extra package is not installed. Install it
+       with `pip install foxglove-sdk[notebook]`.
+
+   .. note::
+      This function is only available when the `notebook` extra package
+      is installed. Install it with `pip install foxglove-sdk[notebook]`.
+      The widget will automatically load the data from the provided datasource
+      and display it in the embedded Foxglove viewer.
```
jtbandes: nit: I'm not sure where this sentence belongs, but I don't think it clearly follows this note about installation instructions

python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -3,12 +3,77 @@ API Reference
...
+      to display the viewer.
+
+   .. py:method:: clear() -> None
+
+      Clear the buffered data.
+
+   .. py:method:: get_data() -> list[bytes]
+
+      Retrieve all collected data.
+
```
jtbandes: why is this method exposed/documented? Are users expected to use it for something?

python/foxglove-sdk/python/docs/conf.py
```diff
@@ -32,6 +32,7 @@
     # autodoc_typehints also fails on Capability which is imported in websocket.py, but is
     # manually documented as an enum
     ("py:class", r"foxglove\.Capability"),
+    ("py:class", r"SelectLayoutParams"),
```
jtbandes: what's being ignored here? seems like this should at least have a comment explaining why it's here, like the other things do

python/foxglove-sdk/notebook-frontend/package.json
gasmith: yeah, i don't think this belongs here

python/foxglove-sdk/python/foxglove/notebook/foxglove_widget.py
```diff
@@ -0,0 +1,97 @@
...
+import anywidget
+import traitlets
+
+if TYPE_CHECKING:
+    from .notebook_buffer import NotebookBuffer
+
+try:
+    __version__ = importlib.metadata.version("foxglove")
+except importlib.metadata.PackageNotFoundError:
+    __version__ = "unknown"
```
jtbandes: why is this needed? is it because this is a package extra?

python/foxglove-sdk/python/foxglove/notebook/foxglove_widget.py
```diff
@@ -0,0 +1,97 @@
...
+    :param src: The source URL of the Foxglove viewer. Defaults to "https://embed.foxglove.dev/".
+    :param layout_data: The layout data to use for the widget.
+    """
+
+    _esm = pathlib.Path(__file__).parent / "static" / "widget.js"
+    width = traitlets.Union([traitlets.Int(), traitlets.Enum(values=["full"])]).tag(
+        sync=True
+    )
+    height = traitlets.Int(500).tag(sync=True)
+    src = traitlets.Unicode("").tag(sync=True)
```
jtbandes: is there a way to have the default be None instead of an empty string? It doesn't look like we have explicit defaults for some other values like `width`, yet we treat those ones as optional so I am wondering why we need a default value of `""` for this one?

python/foxglove-sdk/python/foxglove/notebook/foxglove_widget.py
```diff
@@ -0,0 +1,97 @@
...
+        Refresh the widget by getting the data from the callback function and sending it
+        to the widget.
+        """
+        data = self._buffer.get_data()
+        if not self._ready:
+            self._pending_data = data
+        else:
+            self.send({"type": "update-data"}, data)
+
+    def _handle_custom_msg(self, data: dict, buffers: list[bytes]) -> None:
```
jtbandes: nit: suggest renaming `data` to `msg` to avoid confusing with _pending_data etc

python/foxglove-sdk/python/foxglove/notebook/notebook_buffer.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def _create_writer(self) -> None:
+        random_id = uuid.uuid4().hex[:8]
+        file_name = f"{self._temp_directory.name}/log-{random_id}.mcap"
+        self._files.append(file_name)
+        self._writer = open_mcap(path=file_name, context=self._context)
+
+
+def is_mcap_empty(file_name: str) -> bool:
+    iter = make_reader(open(file_name, "rb")).iter_messages()
```
jtbandes: do we need to close this file? maybe use `with`?

python/foxglove-sdk/python/foxglove/notebook/foxglove_widget.py
surbina: This was added by the tool to create anywidget projects (`npm create anywidget@latest`), and it adds a `__version__` var that Hatch reads during build. I removed this since we are using a different packaging tool.

python/foxglove-sdk/python/docs/api/index.rst
surbina: This is exposed because the widget uses it to read data from the buffer during data refresh: https://github.com/foxglove/foxglove-sdk/pull/663/files#diff-23adb6182241beefcc89638c6945004a9e875e3209c97a24000776828af771b1R85

Users are not expected to use it, but since it's exposed, I added the comment. I can remove it if you think that shouldn't be the case.

python/foxglove-sdk/python/foxglove/notebook/notebook_buffer.py
surbina: Done!

python/foxglove-sdk/python/docs/conf.py
surbina: I removed this line and added some more docs.

python/foxglove-sdk/notebook-frontend/package.json
surbina: I think I added this during a previous iteration of the CI work, but it's not needed now.

.github/workflows/docs.yml
```diff
@@ -34,7 +34,15 @@ jobs:
...
+      - run: corepack enable
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: yarn
+      - run: yarn install
+      - run: yarn build:prod
       - uses: astral-sh/setup-uv@v6
+      - run: uv sync --all-extras
       - run: uv run maturin develop
```
gasmith: Whoops, missed this spot in my PR.

```suggestion
      - uses: astral-sh/setup-uv@v6
      - run: uv sync --all-extras
      - run: uv pip install --editable '.[notebook]'
```

python/foxglove-sdk/python/docs/api/index.rst
gasmith: I think we should probably omit it from this documentation; I can't imagine how users would use it.

jtbandes: I think we should probably remove it and prefix the method name with `_`, but will defer to @gasmith on python best practices

python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -3,12 +3,87 @@ API Reference
...
+.. py:class:: SelectLayoutParams
+
+   A dictionary of parameters to select a layout in the Foxglove viewer.
+
+   :param storageKey: The storage key to identify the layout in local storage. When reusing the same storage key,
+       any modifications made by the user will be restored unless `force` is true.
+   :param opaqueLayout: The layout data to load if this layout did not already exist, or if `force` is true.
+       This is an opaque JavaScript object, which should be parsed from a JSON layout file
+       that was exported from the Foxglove app.
+   :param force: If true, opaqueLayout will override the layout if it already exists. Default: false
```
gasmith: I feel like these should just be keyword arguments on `NotebookBuffer.show()`. Either that, or we need to make this easier to construct. This feels too verbose:

```python
from foxglove.notebok.foxglove_widget import SelectLayoutParams
params = SelectLayoutParams(...)
nb.show(layout=params)
```

Especially for the case where I just want to name the storage key. I'd rather do:

```python
nb.show(layout_storage_key="my-layout")
```

python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -3,12 +3,87 @@ API Reference
...
+
+   .. py:method:: refresh() -> None
+
+      Refresh the widget by reading the latest data from the buffer used to created the widget and sending it to the widget.
+
+.. py:class:: SelectLayoutParams
+
+   A dictionary of parameters to select a layout in the Foxglove viewer.
+
+   :param storageKey: The storage key to identify the layout in local storage. When reusing the same storage key,
```
gasmith: Python should use `snake_case`

python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -3,12 +3,87 @@ API Reference
...
+       any modifications made by the user will be restored unless `force` is true.
+   :param opaqueLayout: The layout data to load if this layout did not already exist, or if `force` is true.
```
gasmith: I know it's been discussed elsewhere, but I don't like the use of term "opaque" here. If we want to say that our layout json is not a stable format, that's a matter for documentation (and debate, since this is an API which takes a layout that was exported arbitrarily long ago).

python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -3,12 +3,87 @@ API Reference
...
+       This is an opaque JavaScript object, which should be parsed from a JSON layout file
```
gasmith: I'm not sure how to obtain an opaque JavaScript object in python. Is this a `str` of serialized JSON perhaps?

python/foxglove-sdk/python/docs/api/index.rst
gasmith: Are we planning to offer a way to export a layout from the widget? I can imagine someone iterating on a layout in a notebook, then trying to share it with someone. How do they accomplish that?

python/foxglove-sdk/python/docs/api/index.rst
gasmith: What is the default behavior, if I do not provide a storage key? Should we pick (and document) a sensible default, unless the user opts out? Maybe we should even make this a required parameter? My expectation would be that if I run a cell with `nb.show()`, modify the layout, and then run the cell again, I'd get the same layout.

python/foxglove-sdk/python/docs/api/index.rst
drdelambre: not actively. we have some patterns forming around that from the sdk playground work, but the idea is that they should have an easy way to declare a layout and iterate / share with that

drdelambre: that debate around layout exports is actually why this is here. currently our layout json format is very implementation specific (being tied to a specific third party library at a specific point in time). we're actively moving towards a more stable, implementation agnostic, format in the next iteration on this project (a programatic layout api). the future format can be found in this PR: https://github.com/foxglove/app/pull/10721

python/foxglove-sdk/python/docs/api/index.rst
drdelambre: @surbina since this is just a stopgap to unblock users who can't wait for the new api to drop, should it just be omitted from the official docs?

jtbandes (approved):

PR #665, opened 2025-09-22
Add note to CameraCalibration that we do not support Tx and Ty values

defunctzombie:
### Changelog
Clarify that Tx and Ty values other than 0 are not supported in `foxglove.CameraCalibration` messages.

### Docs
None

These are the docs.

### Description
Our `CameraCalibration` documentation makes it appear like we support the `Tx` and `Ty` parameters of the projection matrix when in fact we don't. This PR updates the documentation in the message to indicate we do not currently support those parameters. We do leave the parameters present so users can continue providing conventional projection matrix values.

Technically this lack of support also extends to the ROS variants so I wonder if this documentation should go elsewhere (image panel?) This is a disconnect we have today.

Reference: https://github.com/orgs/foxglove/discussions/1157#discussioncomment-14271983


typescript/schemas/src/internal/schemas.ts
```diff
@@ -825,6 +825,8 @@ For monocular cameras, Tx = Ty = 0. Normally, monocular cameras will also have R
 
 For a stereo pair, the fourth column [Tx Ty 0]' is related to the position of the optical center of the second camera in the first camera's frame. We assume Tz = 0 so both cameras are in the same stereo image plane. The first camera always has Tx = Ty = 0. For the right (second) camera of a horizontal stereo pair, Ty = 0 and Tx = -fx' * B, where B is the baseline between the cameras.
```
james-rms: I would remove this paragraph entirely. As far as I can tell we have no special treatment for stereo cameras, the comments on Tx and Ty here are moot since we only support zero for either.

typescript/schemas/src/internal/schemas.ts
```diff
@@ -825,6 +825,8 @@ For monocular cameras, Tx = Ty = 0. Normally, monocular cameras will also have R
...
 
+Note: Foxglove does not currently support Tx and Ty values other than 0.
```
james-rms: As a reader I see this and I assume this means "foxglove has undefined behavior if Tx and Ty aren't zero", which poses a problem for my existing rosbags that have nonzero Tx,Ty.

Is this true? Could we clarify and say something like:
> Foxglove currently ignores Tx and Ty values in CameraCalibration messages and assumes Tx = Ty = 0

That would at least allow me to reason about what I see in Foxglove.

jtbandes: Maybe something like:

> Foxglove currently does not support displaying stereo images, so Tx and Ty are ignored.

defunctzombie:
I've removed the paragraph explaining the unsupported parameters and updated the language to be clear that we do not support stereo camera parameters.

PR #666, opened 2025-09-23
Support referencing an enumeration declared in an external schema

angroyer:
### Changelog

Support referencing an enumeration declared in an external schema.

### Docs

None

### Description

Whenever I use an enumeration in a schema that is different from the "parentSchemaName" field, the generated .proto file fails compiling.

For example, let's suppose I have changed this code:
```diff
index 8ba41b5..69b9265 100644
--- a/typescript/schemas/src/internal/schemas.ts
+++ b/typescript/schemas/src/internal/schemas.ts
@@ -313,6 +313,11 @@ const ArrowPrimitive: FoxgloveMessageSchema = {
   name: "ArrowPrimitive",
   description: "A primitive representing an arrow",
   fields: [
+    {
+      name: "type",
+      type: { type: "enum", enum: SceneEntityDeletionType },
+      description: "Type of deletion action to perform",
+    },
     {
       name: "pose",
       type: { type: "nested", schema: Pose },
```

If I generate the files with "npm run generate", I have an error:
```diff
...
thread 'main' panicked at rust/foxglove_proto_gen/src/bin.rs:12:65:
Failed to generate protos: Failed to load protos

Caused by:
    protoc failed: foxglove/ArrowPrimitive.proto:13:3: "Type" is not defined.
...
```

Indeed, the generated schemas/proto/foxglove/ArrowPrimitive.proto file contains:
```diff
...
import "foxglove/Color.proto";
import "foxglove/Pose.proto";

package foxglove;

// A primitive representing an arrow
message ArrowPrimitive {
  // Type of deletion action to perform
  Type type = 1;
...
```

With the proposed commit, the "npm run generate" now succeeds with this generated code in schemas/proto/foxglove/ArrowPrimitive.proto:
```diff
...
import "foxglove/Color.proto";
import "foxglove/Pose.proto";
import "foxglove/SceneEntityDeletion.proto";

package foxglove;

// A primitive representing an arrow
message ArrowPrimitive {
  // Type of deletion action to perform
  foxglove.SceneEntityDeletion.Type type = 1;
...
```

### DIFF
--- a/scripts/generate.ts
+++ b/scripts/generate.ts
@@ -159,7 +159,7 @@ async function main({ clean }: { clean: boolean }) {
       );
       await fs.writeFile(
         path.join(outDir, "proto", "foxglove", `${schema.name}.proto`),
-        generateProto(schema, enums),
+        generateProto(schema, enums, foxgloveMessageSchemas),
       );
     }
   });
--- a/typescript/schemas/src/internal/generateProto.ts
+++ b/typescript/schemas/src/internal/generateProto.ts
@@ -45,6 +45,7 @@ function protoImportPath(schema: FoxgloveMessageSchema): string {
 export function generateProto(
   schema: FoxgloveMessageSchema,
   nestedEnums: FoxgloveEnumSchema[],
+  allSchemas: { [key: string]: FoxgloveMessageSchema },
 ): string {
   const enumDefinitions: string[] = [];
   for (const enumSchema of nestedEnums) {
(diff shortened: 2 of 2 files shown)


jtbandes:
Hi there,
Could you describe what you're trying to do? Since this code exists only to support code generation of existing foxglove schemas, there's not much motivation to make changes like this if they're not needed for the current set of schemas.

angroyer:
Hi Jacob,
indeed, I stumbled upon this issue whenever I tried to use this project to create my own types. Whenever an enum is shared in multiple schemas, this does not work.
This change might be necessary in case you are planning to do the same.
I let you choose what is best for you and the future of this project.

defunctzombie:
Right now we aren't expecting folks to create their own types from these types as part of the generation step. If you want to re-use the types from their packaged files that might be a more viable approach (i.e. using the pre-generated proto files).

I'll close this as out of scope but if comes up more we can revisit. Thanks for surfacing this for discussion.

PR #667, opened 2025-09-24
Fix memory leak in generic_client allocate_message

MatthewDHarrison:
### Changelog

None

### Docs

None

### Description

We have found that with a high volume of service calls the memory usage of the foxglove bridge grows unbounded. Using valgrind with memcheck, the source of some of these definitely lost blocks was found to be the `allocate_message` function in `generic_client`, where a buffer has memory allocated using the `MessageMembers` `_init_function` which is freed using `free`. This seems to mean that if a message type contains pointers, the memory pointed to will never be freed.

To remedy this, we instead call `members->_fini_function` , which the Message-specific clean-up functions (as documented [here](https://docs.ros.org/en/ros2_packages/humble/api/rosidl_typesupport_introspection_cpp/generated/structrosidl__typesupport__introspection__cpp_1_1MessageMembers__s.html) and implemented [here](https://github.com/ros2/rosidl/blob/4edcac1c0aa162d4c231389811c105214c74a439/rosidl_typesupport_introspection_cpp/resource/msg__type_support.cpp.em#L253)

</td><td>

</td></tr></table>


clalancette (approved):
This looks reasonable to me, and mimics what the upstream code in rclcpp does.  Thanks for the fix.

PR #673, opened 2025-09-25
Update the workaround to find the service typesupport handle.

clalancette:
### Changelog

None.

### Docs

None.

### Description

The GenericClient implementation in foxglove_bridge needs to get a service typesupport handle in order to initialize itself.

In late 2023, the rclcpp::get_service_typesupport_handle() API was added to ROS as a one-call way to get the typesupport handle for services.  That means that all ROS 2 releases since Iron have that API.

Humble does not have that particular API.  However, since late 2024, it does include the underlying rosidl API, so we can call that without resorting to hacks to find the symbol in the library.  In Humble prior to that 2024 release, we need that symbol lookup hack.

That leads to this patch.  If we are on a new enough rclcpp, we use rclcpp::get_service_typesupport_handle().  If we are on a newer version of Humble, we use the underlying rosidl API.  And if we are on an older version of humble, we use the symbol hack.

All tests succeed (run with `colcon test`) on Kilted, Jazzy, Humble, and older Humble (prior to the symbol being available).

### DIFF
--- a/ros/src/foxglove_bridge/src/generic_client.cpp
+++ b/ros/src/foxglove_bridge/src/generic_client.cpp
@@ -59,36 +59,57 @@ std::string getTypeIntrospectionSymbolName(const std::string& serviceType) {
          pkgName + "__" + (middleModule.empty() ? "srv" : middleModule) + "__" + typeName;
 }
 
-/**
- * The default symbol names for getting type support handles for services are missing from the
- * rosidl_typesupport_cpp shared libraries, see
- * https://github.com/ros2/rosidl_typesupport/issues/122
- *
- * We can however, as a (hacky) workaround, use other symbols defined in the shared library.
- * With `nm -C -D /opt/ros/humble/lib/libtest_msgs__rosidl_typesupport_cpp.so` we see that there is
- * `rosidl_service_type_support_t const*
- * rosidl_typesupport_cpp::get_service_type_support_handle<test_msgs::srv::BasicTypes>()` which
- * mangled becomes
- * `_ZN22rosidl_typesupport_cpp31get_service_type_support_handleIN9test_msgs3srv10BasicTypesEEEPK29rosidl_service_type_support_tv`
- * This is the same for galactic, humble and rolling (tested with gcc / clang)
- *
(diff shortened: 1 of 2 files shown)


ros/src/foxglove_bridge/src/generic_client.cpp
```diff
@@ -59,36 +59,60 @@ std::string getTypeIntrospectionSymbolName(const std::string& serviceType) {
...
+#else
+  // Humble needs to do additional work.
   const auto [pkgName, middleModule, typeName] = extract_type_identifier(serviceType);
+
+  // Humble rosidl 3.1.6 is the first version to have the appropriate
+  // __get_service_type_support_handle__ methods available.  However, rosidl does not have a version
+  // macro, so we approximate this with rclcpp versions.
+  // https://github.com/ros2/ros2/commit/9f7d334291ff77a07d68d4d8976871b99e03612b is the version
+  // where rosidl was bumped to 3.1.6, and also bumped rclcpp to version 16.0.11.  Thus, anything
+  // older than that version needs to use the old symbol hack.
```
achim-k: I feel like we shouldn't include the old symbol hack anymore. It is probably fine to assume that the humble rosidl version includes the `__get_service_type_support_handle__` methods. If people still want to build the bridge for older humble versions, they can use the non-SDK based version at https://github.com/foxglove/ros-foxglove-bridge

achim-k (approved):
LGTM. I'd probably remove the old symbol hack.

ros/src/foxglove_bridge/src/generic_client.cpp
clalancette: > I feel like we shouldn't include the old symbol hack anymore. It is probably fine to assume that the humble rosidl version includes the `__get_service_type_support_handle__` methods. If people still want to build the bridge for older humble versions, they can use the non-SDK based version at https://github.com/foxglove/ros-foxglove-bridge

That's fine with me.  I'll update this PR to remove the old hack completely.  Thanks for taking a look.

clalancette: All right, removed that compatibility bit in cc86c6e.

PR #676, opened 2025-09-29
Python: expose Context.default in types & docs

bryfox:
### Changelog
Python: expose Context.default in types & docs

### Docs
None

### Description

This exposes the `Context.default()` method in Python; a user may want to reference it explicitly.


surbina (approved):
LGTM!

PR #677, opened 2025-09-29
rustdoc: Fix for latest nightly

bryfox:
### Changelog
None

### Description

This renames the `doc_auto_cfg` feature, following https://github.com/rust-lang/rust/pull/138907. We install the latest nightly build in CI for docs generation; developers will need to update to at least 2025-09-28.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,6 +20,7 @@ Remember to publish new versions of all libraries!
 This package generates source and interface files for the Foxglove SDK, and relies on tooling from the Rust and Python ecosystems.
 
 - Rust, installed via [rustup](https://rustup.rs/)
+  - Documentation generation requires a recent `nightly` build
 - [Protobuf compiler](https://grpc.io/docs/protoc-installation/)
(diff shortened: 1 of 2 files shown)


bryfox:
> Why do we require nightly for CI?

This feature is what labels docs with the corresponding features (#344), which is pretty useful.

> seems like it would be better to pin a specific version to avoid CI breakages

Yeah, we can certainly do that. Any objections, @gasmith or @eloff ?

bryfox:
> seems like it would be better to pin a specific version to avoid CI breakages

Idea ticket for that: FG-13157. It may also not be too long before we have a stable version with this feature.

gasmith:
> > seems like it would be better to pin a specific version to avoid CI breakages
> 
> Yeah, we can certainly do that. Any objections, @gasmith or @eloff ?

I'm pretty sure that docs.rs builds our documentation against +nightly, so it behooves us to stay bleeding edge for this particular CI job. Considering this is the first such breakage we've seen, I don't think we need to work harder to avoid them.

PR #678, opened 2025-09-30
schema: Fix bayer CFA typo

gasmith:


typescript/schemas/src/internal/schemas.ts
```diff
@@ -1005,7 +1005,7 @@ For each \`encoding\` value, the \`data\` field contains image pixel data serial
 - \`32FC1\`:
   - Pixel brightness is represented as a single-channel, 32-bit little-endian IEEE 754 floating-point value, ranging from 0.0 (black) to 1.0 (white).
   - \`step\` must be greater than or equal to \`width\` * 4.
-- \`bayer_rggb8\`, \`bayer_bggr8\`, \`bayer_rggb8\`, \`bayer_gbrg8\`, or \`bayer_grgb8\`:
+- \`bayer_rggb8\`, \`bayer_bggr8\`, \`bayer_rggb8\`, \`bayer_gbrg8\`, or \`bayer_grbg8\`:
```
bryfox: Change looks right, but `bayer_rggb8` is also duplicated

gasmith: Haha! Nice catch.

PR #679, opened 2025-10-01
Upgrade jest to fix snapshot updates

bryfox:
### Changelog
None

### Description

Bump jest to v30. This allows jest to update inline snapshots in conjunction with prettier, which means we don't need to deal with updates after changing schemas.


james-rms:
It works!

jtbandes:
Amazing - it only took like a year or more for them to fix this. (at least that's what it feels like)

PR #680, opened 2025-10-01
Add unreliable metadata option when creating channels via the channel builder

eloff:
### Changelog

Add unreliable channels, a hint to sinks that log messages can be sent over unreliable transports or dropped under pressure.

### Docs

None

### Description

This add an unreliable option to ChannelBuilder, and tags channels with metadata "foxglove_reliable": "1", as well as marks them internally in the SDK with an unreliable boolean. This can be used by sinks or downstream consumers to handle the messages differently (e.g. send them over UDP). Update send_lossy to drop new unreliable when the queue is full instead of the oldest message in the queue.

Ideally we would drop older unreliable messages instead, but this is not possible with flume, we would need our own queue or to fork it and modify it.

### DIFF
--- a/rust/foxglove/src/channel_builder.rs
+++ b/rust/foxglove/src/channel_builder.rs
@@ -59,6 +59,14 @@ impl ChannelBuilder {
         self
     }
 
+    /// Prefer using a unreliable/lossy transport when a sink supports it.
+    /// This enables sending log messages over UDP or similar for sinks that support it.
+    /// It may also effect which messages are dropped when the [`crate::WebSocketServer`] queue fills.
+    /// Has no effect for MCAP sinks.
+    pub fn unreliable(self) -> Self {
+        self.add_metadata("foxglove_unreliable", "1")
+    }
+
     /// Sets the context for this channel.
     pub fn context(mut self, ctx: &Arc<Context>) -> Self {
         self.context = ctx.clone();
--- a/rust/foxglove/src/channel/raw_channel.rs
+++ b/rust/foxglove/src/channel/raw_channel.rs
@@ -42,6 +42,7 @@ pub struct RawChannel {
     metadata: BTreeMap<String, String>,
     sinks: LogSinkSet,
     closed: AtomicBool,
+    unreliable: bool,
     warn_throttler: Mutex<Throttler>,
 }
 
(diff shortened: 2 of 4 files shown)


rust/foxglove/src/channel_builder.rs
```diff
@@ -59,6 +59,14 @@ impl ChannelBuilder {
         self
     }
 
+    /// Prefer using a unreliable/lossy transport when a sink supports it.
+    /// This enables sending log messages over UDP or similar for sinks that support it.
+    /// It may also effect which messages are dropped when the [`crate::WebSocketServer`] queue fills.
+    /// Has no effect for MCAP sinks.
+    pub fn unreliable(self) -> Self {
+        self.add_metadata("foxglove_unreliable", "1")
```
gasmith, line 67: What's the rationale for stashing this in the metadata map? Up until this point, we've treated that map as opaque. If we want to change that, we should publicly document reserved keys. This metadata is also written out to mcap files, which feels questionable. Maybe it'd be better to just have it as a first class attribute on the channel (descriptor)?

gasmith, line 67: As the old adage goes, never use a boolean when you might need an enum. Should we consider making this a QoS enum or something?

eloff: This gives the agent and client a way to identify unreliable channels. There are other ways to do that, like modify ws_protocol to pass it there.

If we use metadata like this, maybe we should reserve and document the foxglove_ prefix and strip foxglove_ pairs before writing to MCAP. Another thing we're thinking of using metadata for is to mark video channels for live viz.

eloff: I meant to leave that open by using a number here (instead of "true"), I figure we could use different numeric values here in the future if we expand it that way.

defunctzombie:
I need to think deeper about this but my initial thoughts on this are that this is not something we should offer to the user to configure in the context of remote connectivity. What I mean here is that part of the operational modality of remote connectivity is to keep providing data even in the fact of poor connections. The user should not have to explicitly fumble around to make that possible because they might say that some signal needs to be "reliable" which will create head-of-line blocking.

We'll want to look at how our protocol can avoid head-of-line blocking on a topic or chanel-by-channel basis (like QUIC does) as a way around this. But that still means that someone could claim reliable on some channel when in fact it isn't desired behavior when you go to connect remotely. Said another way - I don't _know_ that this should be hard-coded by the sdk user.

defunctzombie:
> Update send_lossy to drop new unreliable when the queue is full instead of the oldest message in the queue.
> 
> Ideally we would drop older unreliable messages instead, but this is not possible with flume, we would need our own queue or to fork it and modify it.

This seems wrong - the semantics we need to work toward are to have the latest message be the one we try to deliver and re-deliver until we get another message.

eloff:
Discussion brought up a lot of questions and concerns. This does not appear to be part of an eventual QoS solution, closing it

PR #681, opened 2025-10-01
python: Make Context.default() always return the same object

gasmith:
### Changelog
- Python: Context.default() always returns the same object.

### Docs
None

### Description
There can be only one highlander.


surbina:
Thanks for implementing this so quickly!

PR #682, opened 2025-10-01
JSONSchema: mark all properties required

james-rms:
### Changelog
- *Breaking*: Foxglove JSON schemas now require all object properties to be defined. This updates the JSONSchema representation of Foxglove schemas to match the requirements of the app, as well as the requirements of ROS and Flatbuffer schemas.
  - Note: The Foxglove app does not enforce this requirement at runtime, so existing messages will not be validated differently within Foxglove.
### Docs

Self-documenting.
### Description

Today whether or not a field is required in a foxglove schema message depends on the schema encoding you use. Today what the app can assume is in a message depends on the encoding:
:white_check_mark:ROS(1|2)(msg|idl): all fields are required
❓ : Protobuf: no fields are required by the serializer. the deserializer will fill in zero values for fields with primitive types. fields with nested struct  types are set to null by the deserializer.
:x:jsonschema: no fields are required.
:white_check_mark:Flatbuffer: all fields are required.

This PR updates our JSONSchemas to require fields, which brings us closer to agreement between all encodings. It's impossible to require fields in a protobuf schema (that's left to the application to enforce).

</td><td>

</td></tr></table>

### DIFF
--- a/typescript/schemas/src/internal/types.ts
+++ b/typescript/schemas/src/internal/types.ts
@@ -20,7 +20,6 @@ export type FoxgloveMessageField = {
     | { type: "nested"; schema: FoxgloveMessageSchema }
     | { type: "enum"; enum: FoxgloveEnumSchema };
   array?: true | number;
-  required?: true;
   description: string;
   protobufFieldNumber?: number;
   defaultValue?: string | number | boolean;


jtbandes (approved):

defunctzombie:
> * Note: The Foxglove app does not enforce this requirement at runtime, so existing messages will not be validated differently within Foxglove.

Have you confirmed this part? If you write a file with a schema where you indicated a field was required in the JSON schema but you did not provide it...

This change seems fine to me from a "schema describes intent" though practically I don't see that it will do anything, will it? For folks that specify the JSON schema but fail to write a field they will still be able to do that since nothing would have forced them to write the field. So as long as it isn't breaking app functionality this LGTM.

Since we are < v1 for the sdk, _and_ I don't know of these JSON schemas being used for writer-side validation, this would be sufficient to handle with a minor version bump.

PR #684, opened 2025-10-02
Remove ROS 2 badges with inaccurate package versions

james-rms:
The [shields.io ROS index integration](https://shields.io/badges/ros-package-index) is serving up out-of-date versions for these packages, it probably takes a while for its cache to refresh. Also, since moving foxglove_msgs and foxglove_bridge into this repo, there's no longer a way to display separate badges for those versions, because the badge does not support multi-package repos.


jtbandes:
Alternative proposal: https://github.com/foxglove/foxglove-sdk/pull/686

PR #685, opened 2025-10-02
prototype including an optional C lib

eloff:
This is not for merging. It demonstrates how we can add an optional dependency on a closed-source Rust library (here https://github.com/foxglove/data-platform/pull/2208), exported as a C static lib to avoid the ABI instability that Rust currently suffers from. It uses cbindgen on the other side to generate a C header, and bindgen here to generate Rust bindings to that API automatically.


eloff:
Closing as we're not going to use this as-is or soon, but the approach has merit

PR #686, opened 2025-10-02
Update ROS badges for rolling

jtbandes:
### Changelog
None

### Docs

None

### Description

Alternative to #684 to fix the broken badge. We need to specify the name of the repository, not the package itself. These used to be the same but now the repository is named `foxglove-sdk`.


clalancette (approved):
Even better, this looks good to me.  Thanks!

jtbandes:
We'll need to do the same for the other distros once they are synced.

Issue #531, opened 2025-07-08
No cmake install target


jlack1987:
**Description**

I would like to install the SDK to the system(e.g. `/usr/local`) so I can build/install it in my docker once and not build it every time, but currently the cmake doesn't have an install target.

**Expected Behavior**

I would expect that I would be able to `make install` after running cmake and it would install headers and libraries to the designated install location.

bryfox:
You're talking about building the libraries from source using https://github.com/foxglove/foxglove-sdk/blob/main/cpp/CMakeLists.txt? The way we [recommend installing](https://docs.foxglove.dev/docs/sdk#installation) right now is to use the versioned releases (libraries and C++ source). [This example](https://docs.foxglove.dev/docs/sdk/example?lang=cpp#create-a-new-project-and-add-the-foxglove-sdk) has a CMakeLists that might be of use. Will that work for you? (For example, download/extract the sources in a separate step — cache that layer — and then link & include per the example?)

jlack1987:
Yeah that's what i'm referring to, but the fetch content method is not optimal as we'd like to build our code from scratch in an environment that doesn't have internet so we'll need to install it to the system in a docker container and depend on it from there. I'll see if I can solve it via cmake and PR it back to you all if that would be acceptable

bryfox:
I don't think this will be an install path we recommend, but it's also not something we'd necessarily be against and would be happy to review a PR. Out of curiosity, why is starting from the SDK source preferable to the versioned artifacts? Presumably you could supply either one to the build system out-of-band and then build your project from there.

jlack1987:
Yeah i'm not opposed to those artifacts, but they don't provide cmake tools for nicely finding the headers/libraries. Going to try and make some cmake changes that won't break your current recommendation, but will also make it an installable package so users can use it with `find_package`

jlack1987:
I have been unable to get this to work unfortunately. I tried to make a [vendor package](https://github.com/jlack1987/foxglove_sdk_vendor) but haven't been able to get it to work and may not be able to work on it again for awhile. I do think this is necessary though because requiring users to use `FetchContent` as the [documentation](https://docs.foxglove.dev/docs/sdk) suggests means you must have an active internet connection to build your code if it depends on this sdk and that is often not possible

valgur:
A couple of additional install() commands would be very helpful indeed. An exported CMake config would also help make the SDK easier to use by defining a consistent interface to the library.
[...]

PR #687, opened 2025-10-03
Allow the foxglove-sdk C++ library to use system libraries

clalancette:
### Changelog

* When building from source, the Foxglove SDK C++ library now attempts to find dependencies via the package manager (`apt`, `dnf`, `conan`, etc), before falling back to vendored packages.

### Docs

None.

### Description

The concept here was kicked off by attempting to package the foxglove-sdk library for C++, specifically https://github.com/conan-io/conan-center-index/pull/27946 and the maintainers' response to it.  They explicitly said that they prefer:

1.  Recipes don't fetch and use a pre-compiled library.
2.  Use the upstream CMakeLists.txt, rather that have a patched or a custom one in Conan

Point one there means that we should compile the foxglove-sdk C++ library within the confines of Conan.  Not explicitly stated, but implied and borne out by research into other Conan recipes, is that all dependencies come from Conan (not vendored via FetchContent).

This PR is one step towards solving 1.  What it does is to introduce a configuration variable called `USE_PACKAGE_MANAGER_DEPENDENCIES`, which defaults to ON.  If it is true, then we attempt to `find_package` all dependencies first.  If the dependency is found via the package manager, that version is used.  If not, it falls back to vendoring via FetchContent.  If `USE_PACKAGE_MANAGER_DEPENDENCIES` is OFF, then the `find_package` is skipped and dependencies are always vendored.  One downside to this method is that there is no way to pick and choose which dependencies is vendored; you sort of get all or nothing.  But I also didn't think adding 5 different configuration options was a good idea.

Note that this PR, on its own, is *not* enough to get foxglove-sdk C++ packaged into Conan.  Besides finding dependencies this way, we also need to:

1. Install CMake configuration glue so downstream consumers can find it (ala #531).
2. Convince upstream Conan to allow rustc/cargo as one of the compilers.
3. Package Corrosion for Conan.

For testing, I compiled the library with `USE_PACKAGE_MANAGER_DEPENDENCIES` set to ON and OFF, and also with the system packages for all of the dependencies installed or not.  The results:

|  |  `USE_PACKAGE_MANAGER_DEPENDENCIES=ON` | `USE_PACKAGE_MANAGER_DEPENDENCIES=OFF` |
| ---|-------| ------ |
| System packages installed | :heavy_check_mark: | :heavy_check_mark: |
| System packages not installed | :heavy_check_mark:   | :heavy_check_mark:        |

Related to FLE-2


cpp/CMakeLists.txt
```diff
@@ -14,61 +14,95 @@ if(CCACHE_PROGRAM)
   message(STATUS "Using ccache: ${CCACHE_PROGRAM}")
 endif()
 
+option(USE_PACKAGE_MANAGER_DEPENDENCIES
+  "Try to find dependencies provided by the local package manager, and if not found, fall back to building them from source"
+  ON)
```
clalancette, line 19: Note to reviewers: if we agree on this approach, we can certainly discuss whether we want the default here to be `ON` or `OFF`.

cpp/CMakeLists.txt
```diff
@@ -14,61 +14,95 @@ if(CCACHE_PROGRAM)
...
+endif()
+if(NOT TARGET Catch2::Catch2WithMain)
+  FetchContent_Declare(
+    Catch2
+    GIT_REPOSITORY https://github.com/catchorg/Catch2.git
+    GIT_TAG        v3.8.0
+  )
+  FetchContent_MakeAvailable(Catch2)
+  list(APPEND CMAKE_MODULE_PATH ${catch2_SOURCE_DIR}/extras)
+endif()
```
jcar87, line 48: I could be wrong - but I believe the pattern you are after is already supported by CMake - documented here https://cmake.org/cmake/help/latest/module/FetchContent.html#integrating-with-find-package

in particular, the variable  `FETCHCONTENT_TRY_FIND_PACKAGE_MODE` which has `OPT_IN`, `ALWAYS` and `NEVER` as options - it would work similar to your proposed `USE_PACKAGE_MANAGER_DEPENDENCIES`, but without having to use custom logic and conditionals.


e.g.

```
FetchContent_Declare(Catch2
  GIT_REPOSITORY xxx
  GIT_TAG yyy
  FIND_PACKAGE_ARGS
  )

FetchContent_MakeAvailable(Catch2)

find_package(Catch2)
```

When `USE_PACKAGE_MANAGER_DEPENDENCIES` is undefined, it assumes `OPT_IN` (which we have opted into by having added the `FIND_PACKAGE_ARGS` to `FetchContent_Declare`)

The `FetchContent_MakeAvailable(Catch2)` will try to call `find_package(Catch2)` first (I think with `QUIET` and all). If you want to disable this behaviour and "force" to use fetch content, you would set `USE_PACKAGE_MANAGER_DEPENDENCIES` to `NEVER`.

In case this is helpful.

clalancette: Thanks for the feedback (and sorry for the delay in responding).

I did in fact try to use `FIND_PACKAGE_ARGS` for `FetchContent_Declare` first.  But I couldn't convince myself that it was actually working; it kept making the `_deps` directory for the packages I marked that way anyway.  Maybe I was doing it wrong, or maybe that is expected.  I'm honestly not sure.  Do you know how I can tell for certain that the libraries are being used locally instead of being pulled?

gasmith:
@clalancette I suppose this needs to be updated now that #989 has landed?

clalancette:
> @clalancette I suppose this needs to be updated now that #989 has landed?

Yes, good call.  Will work on that next.

cpp/CMakeLists.txt
clalancette: It turns out that I was just doing something wrong before, and this does indeed work.  Thanks for the tip, I've switched to that now.

clalancette:
> 1. **libwebsockets URL** (line 63): The switch from `github.com/warmcat/libwebsockets.git` to `libwebsockets.org/repo/libwebsockets` is still unexplained. The GitHub mirror is more widely accessible (no corporate firewall issues, better uptime track record). What prompted this?

The one I switched to is the real upstream of libwebsockets.  It seems safer to get it straight from the horses mouth, rather than relying on a GitHub mirror.

> 2\. **libwebsockets version constraint** (line 65): `FIND_PACKAGE_ARGS QUIET` accepts any system version while FetchContent pins v4.3.3. If there are API boundaries that matter, `FIND_PACKAGE_ARGS 4.3 QUIET` would prevent silent breakage from an older system install.

Meh.  I'm not all that worried about it, but I'll change it if a human thinks I should.

cpp/CMakeLists.txt
gasmith: ON makes sense to me.

cpp/CMakeLists.txt
```diff
@@ -45,27 +60,31 @@ endif()
 FetchContent_MakeAvailable(Corrosion)
 
 FetchContent_Declare(libwebsockets
-  GIT_REPOSITORY https://github.com/warmcat/libwebsockets.git
+  GIT_REPOSITORY https://libwebsockets.org/repo/libwebsockets
   GIT_TAG v4.3.3
+  FIND_PACKAGE_ARGS QUIET
```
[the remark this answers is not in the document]
gasmith: This is just for tests, so I don't think we need to be overly exact here.

PR #688, opened 2025-10-04
Add Agent sink behind agent feature flag

eloff:
### Changelog

None

### Description

Add Agent sink for remote visualization through Foxglove Agent running on the same host (not yet released).

This adds a non-default feature agent, and behind that an Agent sink. Currently this starts the WebSocketServer, but the implementation will change in the future.

I'd like some feedback about the methods I included on the builder. I think they all make sense going forward, but I'd like a second opinion on that. For now I didn't expose anything other than stop() on the handle.

Things like services and asset handlers might be supported in the future? At any rate they're not supported now, so I left them out.

### DIFF
--- a/rust/foxglove/src/lib.rs
+++ b/rust/foxglove/src/lib.rs
@@ -347,6 +347,8 @@ pub use schema::Schema;
 pub use sink::{Sink, SinkId};
 pub(crate) use time::nanoseconds_since_epoch;
 
+#[cfg(feature = "agent")]
+mod agent;
 #[cfg(feature = "live_visualization")]
 mod runtime;
 #[cfg(feature = "live_visualization")]
@@ -355,6 +357,8 @@ pub mod websocket;
 mod websocket_client;
 #[cfg(feature = "live_visualization")]
 mod websocket_server;
+#[cfg(feature = "agent")]
+pub use agent::Agent;
 #[cfg(feature = "live_visualization")]
 pub(crate) use runtime::get_runtime_handle;
 #[cfg(feature = "live_visualization")]
--- a/rust/foxglove/Cargo.toml
+++ b/rust/foxglove/Cargo.toml
@@ -11,6 +11,7 @@ license.workspace = true
 default = ["derive", "live_visualization", "lz4", "zstd"]
 chrono = ["dep:chrono"]
 derive = ["dep:foxglove_derive"]
+agent = ["live_visualization"]
 live_visualization = [
   "dep:base64",
   "dep:flume",
--- a/rust/foxglove/src/websocket_server.rs
+++ b/rust/foxglove/src/websocket_server.rs
@@ -60,6 +60,14 @@ impl WebSocketServer {
         Self::default()
     }
 
+    /// Creates a new websocket server with the given options.
+    pub(crate) fn with_options(options: ServerOptions) -> Self {
+        Self {
+            options,
+            ..Self::default()
+        }
+    }
+
     /// Set the websocket server name to advertise to clients.
     ///
     /// By default, the server is not given a name.


defunctzombie:
How would someone use this?

eloff:
> How would someone use this?

I'll add an example

rust/examples/agent/src/main.rs
```diff
@@ -0,0 +1,126 @@
...
+    let env = env_logger::Env::default().default_filter_or("debug");
+    env_logger::init_from_env(env);
+
+    let args = Cli::parse();
+
+    // Foxglove Agent needs to be running on the same machine for this to work
+    let server = foxglove::Agent::new()
+        .start()
+        .await
+        .expect("Could not connect to agent");
```
bryfox: Would it be on the user to deal with timeouts if the agent wasn't immediately reachable? Would we provide some sort of retry-until-cancelled interface?

(This question doesn't really make sense with the current implementation, but I'm wondering if that's how we want it to work longer-term.)

defunctzombie: I agree with the comment here. This seems strange to me that I would need to "start" and also "await" here. My mental model of this is that I am simply initializing agent as a valid sink and from there connectivity will happen behind the scenes.

eloff: Yeah, that makes more sense. Let me see what I can do

eloff: I changed this to create(), and the "connection" happens in the background. I added methods to the handle to wait for the connection. They don't need to be used, but are required to make the test work, or if you want to wait for a connection before logging something to ensure it's not discarded.

bryfox: 👍 I like the general interface.

Since the method here is called "ensure_connected" I think "start" might more naturally be called "connect".

I also think connect/disconnect reflects how I think about this more, but will see what others think.

rust/foxglove/src/agent.rs
```diff
@@ -0,0 +1,269 @@
...
+        let result = self
+            .connect_rx
+            .recv()
+            .map_err(|e| FoxgloveError::Unspecified(Box::new(e)));
+        self.connecting = false;
+        self.server = Some(result??);
+        Ok(())
+    }
+
+    /// Gracefully shut down the agent, if connected. Otherwise returns None.
```
bryfox: I assume this (in a future world) might still tear down the connection, but never actually shut down the agent. As noted elsewhere, I think 'connect'/'disconnect' still feel more natural to me than 'start'/'stop'.

bryfox:
Do you wan to add Python & C++ as part of this PR?

rust/foxglove/src/agent.rs
eloff: Yeah, true. I'll rename it

eloff:
> Do you wan to add Python & C++ as part of this PR?

I'll make a new PR for those to split the work up

rust/examples/agent/src/main.rs
```diff
@@ -0,0 +1,133 @@
...
+    env_logger::init_from_env(env);
+
+    let args = Cli::parse();
+
+    // Foxglove Agent needs to be running on the same machine for this to work
+    let mut handle = foxglove::Agent::new()
+        .create()
+        .expect("Could not create agent");
+
+    // This is optional, only if you need to wait for the agent to be connected before continuing
```
defunctzombie: Connected to what?

rust/foxglove/src/lib.rs
```diff
@@ -355,6 +357,8 @@ pub mod websocket;
 mod websocket_client;
 #[cfg(feature = "live_visualization")]
 mod websocket_server;
+#[cfg(feature = "agent")]
```
defunctzombie, line 362: Why do we have a live_visualization feature and an agent feature?

eloff: live_visualization feature covers the websocketserver, we need a separate feature to control adding the new live viz (agent) functionality and linked library - we don't want to couple that to the existing websocketserver, you should be able to include either, both, or none.

rust/examples/agent/src/main.rs
eloff: Connected to agent, I should clarify that. Although currently we can't guarantee that, we can just guarantee the server is started. I think I'm going to remove that from this example for now, and make those methods private until they can do what's promised on the tin.

rust/foxglove/src/lib.rs
defunctzombie: `live_visualization feature covers the websocketserver` Does this mean that by default we don't build sdk with websocket server support? I thought sdk had websocket server built in?

eloff: It's enabled by default, but can be disabled if you don't want it.

Agent isn't enabled by default, it's opt in.

defunctzombie: 👍 I got confused with a similarly named flag in our other repo

eloff: It is confusing, NGL

rust/examples/agent/src/main.rs
eloff: I renamed stop to disconnect, renamed create to connect, and for now hid ensure_connected as it doesn't really make much sense at present. We'll make it public or add something like it before GA.

rust/foxglove/src/agent.rs
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    /// Ensures that the Foxglove Agent is connected.
+    /// If connected this returns Ok immediately.
+    /// Otherwise connects and blocks until the connection is established or fails and returns the result.
+    ///
+    /// Note: currently this starts a server for Agent to connect to, and blocks until it's started.
+    /// There isn't a way to wait until the Agent has connected to that server.
+    /// This behavior will change soon.
+    pub(crate) async fn ensure_connected(&mut self) -> Result<(), FoxgloveError> {
```
bryfox: These should be public? Should they be part of the example, or did you explicitly omit for now?

eloff: I removed it after feedback from Roman, we can can add it back when we know what interface we want and it can ensure we're connected to agent, not just listening for it (if we release it now the interface maybe stays the same, but what it means would change.)

bryfox: 👍 missed that; sounds good

rust/examples/agent/src/main.rs
eloff: I think making the connectivity happen behind the scenes was partially a mistake.

Ideally when you call connect, you can wait for it to actually connect (with a short timeout - since it's on the local host). The connect() method seems like a good place to do this.

The other problem is kind of Rust specific. If you currently call connect in  a context outside the tokio runtime, the spawn call will fail "there is no reactor running, must be called from the context of a Tokio 1.x runtime". Likewise if you were to run it inside a tokio spawn_blocking to fix that, it would fail if run inside a Tokio runtime.

The way we get around this for the WebSocketServer is have async start, for calling inside a Tokio runtime, and start_blocking for outside. We need two methods like that here as well. I think connect and connect_blocking. Being able to connect().await also gives a way to delay until it's connected to agent before logging, assuming it can connect.

The difference is I think we don't want it to be an error if you can't connect (e.g. sdk started before agent), and we do want it to reconnect automatically in the background. I think we can accomplish this by not returning a Result from the connect method, just always return a handle, and awe can have maybe a way on the handle in the future to determine if it's connected, to force a reconnect attempt maybe, etc. Currently I think just a disconnect() method and we can go from there.

I've implemented this in the last commit to what I think the interface should be.

rust/foxglove/src/agent.rs
```diff
@@ -0,0 +1,278 @@
...
+use tracing::error;
+
+use crate::{
+    websocket::{Capability, Server, ShutdownHandle},
+    Context, FoxgloveError, WebSocketServer, WebSocketServerHandle,
+};
+
+/// A handle to the Agent Remote Connection.
+///
+/// This handle can safely be dropped and the agent will run forever.
```
gasmith:
```suggestion
/// This handle can safely be dropped and the connection will run forever.
```

rust/foxglove/src/agent.rs
```diff
@@ -0,0 +1,278 @@
...
+        self
+    }
+
+    /// Set the message backlog size.
+    ///
+    /// The server buffers outgoing log entries into a queue. If the backlog size is exceeded, the
+    /// oldest entries will be dropped.
+    ///
+    /// By default, the server will buffer 1024 messages.
+    pub fn message_backlog_size(mut self, size: usize) -> Self {
```
gasmith: This configuration is extremely specific to the websocket-server-iness of the current implementation. We should probably omit it here. We'll have other knobs for controlling QoS, but let's cross that bridge when we get there.

rust/foxglove/src/agent.rs
```diff
@@ -0,0 +1,278 @@
...
+    pub fn capabilities(mut self, capabilities: impl IntoIterator<Item = Capability>) -> Self {
+        self.capabilities = capabilities.into_iter().collect();
+        self
+    }
+
+    /// Configure an event listener to receive client message events.
+    pub fn listener(mut self, listener: Arc<dyn crate::websocket::ServerListener>) -> Self {
+        self.listener = Some(listener);
+        self
+    }
```
gasmith: I suppose at some point we're due for a reorganization of modules to reflect common functionality shared between the `Agent` and `WebSocketServer` sinks. But this also begs the question: Are the set of Agent capabilities actually the same as that of the ws server? Is the ServerListener interface really the same? I feel like, even if they're the same now, they're likely to diverge in the future.

rust/examples/agent/src/main.rs
bryfox: > Ideally when you call connect, you can wait for it to actually connect (with a short timeout - since it's on the local host)

Where's the short timeout in the new code? Is that something I'd add as a user?

I think the separate non-blocking/blocking methods make sense, and seem consistent with the WS server interface. Does the automatic retrying-in-background but allowing you to drop the handle seem too magical? Are there other libraries that do anything similar?

I'm also curious if you already have a clear picture of what python would look like with this.

rust/foxglove/src/agent.rs
```diff
@@ -0,0 +1,278 @@
...
+        Self {
+            options,
+            server: None,
+            connecting: false,
+            connect_rx,
+            connect_tx,
+        }
+    }
+
+    fn start_connection(&mut self) {
```
gasmith: Do we really need all this complexity to start a websocket server? The only way it's going to fail is permanently. Why not just:

```rust
impl Agent {
    pub async fn connect(self) -> AgentHandle {
        let server = WebSocketServer::new()
            // ...
            .start()
            .await
            .expect("this goes away when we switch to a client");
        AgentHandle(server)
    }
}
```

It seems like we're trying to plan ahead for the world where we are maintaining a persistent client connection, but I think that's going to look different from this. It's probably better to wait until we're actually building the persistent client to introduce that complexity.

rust/foxglove/src/lib.rs
```diff
@@ -349,6 +349,8 @@ pub use sink::{Sink, SinkId};
 pub use sink_channel_filter::SinkChannelFilter;
 pub(crate) use time::nanoseconds_since_epoch;
 
+#[cfg(feature = "agent")]
+mod agent;
 #[cfg(feature = "live_visualization")]
```
gasmith, line 354: When we originally introduced the `live_visualization` feature, we did so under the expectation that it was going to encompass more than just the websocket server. Perhaps it's time to rethink that? What do others think of deprecating the `live_visualization` feature, in favor of `websocket_server` (or even just `websocket`)?

rust/foxglove/src/agent.rs
eloff: They're likely to diverge, I think I can make a stripped down copy of just the bits we support now, with their own types so we can diverge it safely later

rust/foxglove/src/lib.rs
eloff: Yeah, we guessed wrong with that one. We could rename it to websocket_server or websocket, but we'd still need to keep live_visualization as an alias for that for now for backward compatibility. I'm not sure that gains us much.

rust/foxglove/src/lib.rs
```diff
@@ -357,6 +359,8 @@ pub mod websocket;
 mod websocket_client;
 #[cfg(feature = "live_visualization")]
 mod websocket_server;
+#[cfg(feature = "agent")]
+pub use cloud_sink::CloudSink;
```
bryfox: This should export the listener trait as well

rust/examples/agent/src/main.rs
bryfox, line 1: I think our agent examples should illustrate a video stream and receiving messages for teleop. I'm putting an example like that together for the primary documentation; I can update this one to match (in a future PR) if you agree. And I can do the same for py/c++.

rust/examples/agent/src/main.rs
eloff: There's no point in the timeout with the code as-is but in the future we can add that (and make it configurable in the builder if we want).

It represents two things, a connection to agent and a conversion to the cloud live server. Both could have connectivity issues, but we reconnect as needed. The handle is needed only if you want to stop it or if you want to check on the connection status. Seems reasonable to me.

I have a Python PR in the works, hopefully ready for review before EOD.

rust/examples/agent/src/main.rs
eloff: Yeah, that would be perfect

rust/foxglove/src/cloud_sink.rs
```diff
@@ -0,0 +1,263 @@
...
+impl CloudSink {
+    /// Creates a new websocket server with default options.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Sets the server capabilities to advertise to the client.
+    ///
+    /// By default, the server does not advertise any capabilities.
+    pub fn capabilities(mut self, capabilities: impl IntoIterator<Item = Capability>) -> Self {
```
bryfox: This is consistent with the WS Server, but I think we can actually remove it (at least for now) and set `ClientPublish` for you if you configure a listener. It'd be nice for this sink / teleop to work without needing to introduce the concept of capabilities.

eloff: Good point, I'll make the change

PR #690, opened 2025-10-06
Fix python build issues on windows and with sdk examples

eloff:
### Changelog
None

### Docs

None

### Description

Two separate issues fixed:

1) test-sdk-examples tries to install av v16 which doesn't have wheels for all platforms (no Linux x86 or x64!)
2) windows x86 builds try to use 64bit python when they should use the 32bit version

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -233,7 +233,7 @@ jobs:
         with:
           target: ${{ matrix.platform.target }}
           working-directory: python/foxglove-sdk
-          args: --release --out dist --find-interpreter
+          args: --release --out dist -i python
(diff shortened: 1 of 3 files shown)


.github/workflows/python.yml
```diff
@@ -233,7 +233,7 @@ jobs:
         with:
           target: ${{ matrix.platform.target }}
           working-directory: python/foxglove-sdk
-          args: --release --out dist --find-interpreter
+          args: --release --out dist -i python
```
sofuture:
```suggestion
          args: --release --out dist --interpreter python
```

we should prefer the long arg in scripts/config

PR #694, opened 2025-10-08
Ensure writer is closed in examples, add a note about that to doc string

eloff:
### Changelog
None

### Docs

https://github.com/foxglove/docs/pull/1003/files

### Description

If you don't close the mcap writer, it won't finish the mcap file correctly leading to an obscure error “length out of int32 range” when you try to open it. So we add the with statement and a comment to all examples, to make it clearer for the developer how to avoid that.

### DIFF
--- a/python/foxglove-sdk-examples/quickstart/main.py
+++ b/python/foxglove-sdk-examples/quickstart/main.py
@@ -22,28 +22,28 @@ size_channel = Channel("/size", message_encoding="json")
 
 # We'll log to both an MCAP file, and to a running Foxglove app via a server.
 file_name = "quickstart-python.mcap"
-writer = foxglove.open_mcap(file_name)
-server = foxglove.start_server()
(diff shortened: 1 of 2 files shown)


python/foxglove-sdk-examples/quickstart/main.py
```diff
@@ -22,28 +22,28 @@
 
 # We'll log to both an MCAP file, and to a running Foxglove app via a server.
 file_name = "quickstart-python.mcap"
-writer = foxglove.open_mcap(file_name)
-server = foxglove.start_server()
+# Close the mcap writer with close() or the with statement
+with foxglove.open_mcap(file_name):
+    server = foxglove.start_server()
+    while True:
+        size = abs(math.sin(time.time())) + 1
```
eloff, line 29: This diff is dumb...

bryfox (approved):

PR #695, opened 2025-10-08
[FG-11597] Support mcap metadata in SDK (Rust Core + Python Bindings)

Jenn738:
### Changelog
None

### Docs
None

### Description

Problem: There's currently no support to log metadata (key, value pairs) into MCAP.

Solution:
- Added write_metadata method to mcap_sink.rs and the McapWriterHandle in mcap_writer.rs
- Supported BTreeMap<String, String> for metadata key, value pairs as described by the MCAP documentation
- Added unit tests that verify that we maintain backwards compatibility - i.e. still logging without metadata if metadata does not exist
- Added integration tests in example_mcap
- Added Python support and tests

Testing:
- Rust Unit tests:
  - write basic metadata
  - empty metadata
  - multiple metadata
  - write metadata after close

- Rust Integration tests: `run --bin example_mcap`
  - basic metadata test
  - multiple metdata records
  - write empty metadata

- Python tests: `python -m pytest python/foxglove/tests/test_mcap.py -k "write_metadata" -v`
  - basic metadata test
  - multiple metdata records
  - write empty metadata

Issue: https://linear.app/foxglove/issue/FG-11597/support-mcap-metadata-in-sdk


rust/examples/mcap/src/main.rs
```diff
@@ -73,6 +73,123 @@ fn log_until(fps: u8, stop: Arc<AtomicBool>) {
     }
 }
 
+fn verify_metadata(path: &PathBuf) {
```
ericmlujan: This should probably live in a unit test rather than in the example, but totally understand why you put it here while iterating!

rust/examples/mcap/src/main.rs
```diff
@@ -73,6 +73,123 @@ fn log_until(fps: u8, stop: Arc<AtomicBool>) {
...
+                                metadata.metadata.len()
+                            );
+
+                            // Show all key-value pairs
+                            for (key, value) in &metadata.metadata {
+                                println!("  {}: {}", key, value);
+                            }
+
+                            // Check specific tests
+                            match metadata.name.as_str() {
```
ericmlujan: Nit: I'd consider turning this into more of a check where you define a hashmap of expected keys and values, and use set boolean and/or map equality operations to see that the set of keys match and that the values are equal. This feels a bit hardcoded to me

rust/examples/mcap/src/main.rs
```diff
@@ -73,6 +73,123 @@ fn log_until(fps: u8, stop: Arc<AtomicBool>) {
...
+                                    println!("ERROR: Empty metadata should not have been written!");
+                                }
+                                _ => {
+                                    println!("  Unknown metadata: {}", metadata.name);
+                                }
+                            }
+                        }
+                    }
+
+                    println!("\n=== METADATA TEST RESULTS ===");
```
ericmlujan: Similarly this should probably be rewritten to use the [`assert!` macros](https://doc.rust-lang.org/book/ch11-01-writing-tests.html#checking-results-with-the-assert-macro) when converting to a proper test

rust/foxglove/src/mcap_writer/mcap_sink.rs
```diff
@@ -143,6 +143,40 @@ impl<W: Write + Seek> McapSink<W> {
...
+        }
+
+        let mut guard = self.inner.lock();
+        let writer = guard.as_mut().ok_or(FoxgloveError::SinkClosed)?;
+
+        writer
+            .writer
+            .write_metadata(&mcap::records::Metadata {
+                name: name.into(),
+                metadata: metadata.clone(),
```
ericmlujan: Do we want to clone here or is it okay to take ownership and consume the metadata?

rust/foxglove/src/mcap_writer/mcap_sink.rs
```diff
@@ -349,6 +383,158 @@ mod tests {
...
+            .write_metadata("test_metadata", &metadata)
+            .expect("failed to write metadata");
+
+        writer.finish().expect("failed to finish recording");
+
+        let mut metadata_count = 0;
+        let mut metadata_found = false;
+        foreach_mcap_metadata(&temp_path, |meta| {
+            metadata_count += 1;
+            if meta.name == "test_metadata" {
```
ericmlujan: Similarly I wonder if this could be simplified a bit to directly compare against the `metadata` BTreeMap defined above

rust/examples/mcap/src/main.rs
Jenn738: Would you be able to outline what the testing structure looks like for unit and integration tests including the files they live in? This wasn't so obvious to me.

rust/foxglove/src/mcap_writer/mcap_sink.rs
Jenn738: good catch, i'll change to take ownership!

rust/examples/mcap/src/main.rs
ericmlujan: For dedicated integration tests, we tend to have small files for those. For example, [logging](https://github.com/foxglove/foxglove-sdk/blob/ef448413a8a93ce36682b870afd823cf62e5a0b0/rust/foxglove/src/tests/logging.rs#L24) has a test where we open a temporary mcap and write to it.

For unit tests, the general pattern is to define them (and any related test-only utilities) in the [`mod tests`](https://github.com/foxglove/foxglove-sdk/blob/ef448413a8a93ce36682b870afd823cf62e5a0b0/rust/foxglove/src/mcap_writer/mcap_sink.rs#L145-L329) block in a file, which is guarded by a directive `#[cfg(test)]` that only compiles it in for testing.

python/foxglove-sdk/python/foxglove/tests/test_mcap.py
```diff
@@ -116,6 +116,108 @@ def test_write_to_different_contexts(make_tmp_mcap: Callable[[], Path]) -> None:
...
+        writer.write_metadata("test1", expected_metadata["test1"])
+
+        # Write multiple metadata records
+        writer.write_metadata("test2", expected_metadata["test2"])
+        writer.write_metadata("test3", expected_metadata["test3"])
+
+        # Write empty metadata (should be skipped)
+        writer.write_metadata("empty_test", {})
+
+        # Log some messages
```
bryfox, line 169: Do you think writing messages are useful to these tests? I'd assume that we could rely here on the underlying library producing correct metadata either way.

python/foxglove-sdk/python/foxglove/tests/test_mcap.py
```diff
@@ -116,6 +116,108 @@ def test_write_to_different_contexts(make_tmp_mcap: Callable[[], Path]) -> None:
...
+
+    # Verify file was created and has some content
+    assert tmp_mcap.exists()
+    assert tmp_mcap.stat().st_size > 0
+
+    # Verify no metadata was written (empty metadata should be skipped)
+    _verify_metadata_in_file(tmp_mcap, {})
+
+
+def test_write_metadata_multiple_records(tmp_mcap: Path) -> None:
```
bryfox, line 194: This case seems like it was already covered with the first test. (I suppose that's true of the empty metadata as well). Is there value in keeping both?

python/foxglove-sdk/python/foxglove/tests/test_mcap.py
bryfox: Maybe interleaving metadata & message writes would be useful.

python/foxglove-sdk/src/mcap.rs
```diff
@@ -167,6 +167,25 @@ impl PyMcapWriter {
...
     }
+
+    /// Write metadata to the MCAP file.
+    ///
+    /// Metadata consists of key-value string pairs associated with a name.
+    /// If the metadata dictionary is empty, this method does nothing.
+    fn write_metadata(
+        &self,
+        name: &str,
+        metadata: std::collections::BTreeMap<String, String>,
```
bryfox, line 178: nit: `use std::collections::BTreeMap;` to match usage elsewhere.

rust/examples/mcap/Cargo.toml
```diff
@@ -11,5 +11,6 @@ foxglove = { path = "../../foxglove", default-features = false, features = ["der
 clap = { version = "4.5", features = ["derive"] }
 ctrlc = "3.4.5"
 env_logger = "0.11.5"
+mcap = "0.23"
```
bryfox, line 14: Looks like you can remove this. That'll clarify that users shouldn't need to rely on the mcap sdk directly.

rust/examples/mcap/src/main.rs
```diff
@@ -100,6 +101,14 @@ fn main() {
         .create_new_buffered_file(&args.path)
         .expect("Failed to start mcap writer");
 
+    // Example: Write metadata to the MCAP file
+    let mut metadata = BTreeMap::new();
+    metadata.insert("os".to_string(), "linux".to_string());
+    metadata.insert("arch".to_string(), "x64".to_string());
```
bryfox, line 107: not a big deal, but you could use OS & ARCH from env: https://doc.rust-lang.org/std/env/consts/index.html

rust/foxglove/src/mcap_writer/mcap_sink.rs
```diff
@@ -349,6 +383,154 @@ mod tests {
...
+
+        // Test Case 1: Basic metadata write
+        let mut basic_metadata = BTreeMap::new();
+        basic_metadata.insert("key1".to_string(), "value1".to_string());
+        basic_metadata.insert("key2".to_string(), "value2".to_string());
+        writer
+            .write_metadata("basic_test", basic_metadata)
+            .expect("Failed to write basic metadata");
+
+        // Test Case 2: Multiple metadata records
```
bryfox, line 466: these test cases are using the same writer, so there doesn't seem to be a real need for `metadata3` to get multiple records. Likewise, if it's important to test a single record, we should make a new writer (probably in a separate test).

rust/foxglove/src/mcap_writer/mcap_sink.rs
```diff
@@ -349,6 +383,154 @@ mod tests {
...
+
+        writer.finish().expect("failed to finish recording");
+
+        // Verify Tests Above
+        // Define expected metadata using flexible HashMap approach
+        let mut expected = std::collections::HashMap::new();
+
+        // Expected basic metadata
+        let mut expected_basic = BTreeMap::new();
+        expected_basic.insert("key1".to_string(), "value1".to_string());
```
bryfox, line 495: nit: I think you could build a map from the existing metadata maps above instead of re-creating key/values; you might just need to clone them.

python/foxglove-sdk/python/foxglove/tests/test_mcap.py
Jenn738: What about 3 tests for now?
- one that interleaves metadata and messages
- one that solely tests multiple metadata -  I think the basic one is redundant
- another that tests the empty metadata edge case

There are definitely other edge case tests that can be added, but I'm saving that for future work since some of it requires additional validation logic. Some that came to mind:
- input validation
- special characters
- large inputs or metadata
- multiple threads writing metadata at the same time
- multiple metadata with the same name
- whitespace in the name or metadata
- empty metadata name

python/foxglove-sdk/python/foxglove/tests/test_mcap.py
Jenn738: yes, so true! I realized this last night :). I think some of these can be deleted.

eloff:
Superseded by https://github.com/foxglove/foxglove-sdk/pull/715

bryfox:
Thanks! These changes have been included as part of #715 and will be available in the next release.

PR #698, opened 2025-10-15
Add python agent sink wrapper

eloff:
### Changelog
None

### Docs

None

### Description

Add Python wrappers around the new CloudSink, CloudSinkHandle, and CloudSinkListener.

Also renames the agent feature to cloud and the agent examples to cloud, but they're temporary as tests for the functionality. @bryfox intends to work on better examples that will replace them.

### DIFF
--- a/python/foxglove-sdk-examples/cloud/README.md
+++ b/python/foxglove-sdk-examples/cloud/README.md
@@ -1,0 +1,15 @@
+# Log to cloud live visualization
+
+An example from the Foxglove SDK.
+
+Connect to the cloud with Foxglove agent and log messages for live visualization to be viewed in the Foxglove app.
+
+This example will soon be replaced with something better.
+
+## Usage
+
+This example uses [uv](https://docs.astral.sh/uv/).
+
+```bash
+uv run python main.py
+```
--- a/python/foxglove-sdk-examples/cloud/pyproject.toml
+++ b/python/foxglove-sdk-examples/cloud/pyproject.toml
@@ -1,0 +1,14 @@
+[project]
+name = "cloud"
+version = "0.0.0"
+description = "Example of logging to Foxglove cloud remote connection"
+authors = [{ name = "Foxglove", email = "support@foxglove.dev" }]
+requires-python = ">=3.9"
+readme = "README.md"
+dependencies = ["foxglove-sdk"]
+
+[dependency-groups]
+dev = ["mypy>=1,<2"]
+
+[tool.uv]
+package = false
--- a/python/foxglove-sdk-examples/cloud/main.py
+++ b/python/foxglove-sdk-examples/cloud/main.py
@@ -1,0 +1,80 @@
+import time
+
+import foxglove
+from foxglove.schemas import RawImage
+
+
+class MessageHandler(foxglove.CloudSinkListener):
+    def __init__(self) -> None:
+        self.topics_by_channel_id: dict[int, str] = {}
+
+    def on_client_advertise(
+        self,
+        client: foxglove.websocket.Client,
+        client_channel: foxglove.websocket.ClientChannel,
+    ) -> None:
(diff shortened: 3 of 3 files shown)


python/foxglove-sdk-examples/cloud/pyproject.toml
```diff
@@ -0,0 +1,18 @@
+[project]
+dependencies = ["foxglove-sdk @ file:///home/eloff/foxglove/sdk/python/foxglove-sdk", "mcap>=1.2"]
```
bryfox: Can this use a relative path?

If not, you can specify it as a poetry-specific dependency and then we'll update after the next release.

```
[tool.poetry.dependencies]
foxglove-sdk = { path = "../../foxglove-sdk" }
```

This is still a little cumbersome; FG-11340 tracks improving it.

python/foxglove-sdk/python/foxglove/__init__.py
```diff
@@ -86,6 +91,34 @@ def start_server(
             channel_filter=channel_filter,
         )
 
+    def start_cloud_sink(
```
bryfox, line 93: Why do we have this wrapping function rather than exporting the original and documenting the function in the stub file? (I realize the same question applies for the existing `start_server`.)

rust/foxglove/Cargo.toml
```diff
@@ -11,7 +11,7 @@ license.workspace = true
 default = ["derive", "live_visualization", "lz4", "zstd"]
 chrono = ["dep:chrono"]
 derive = ["dep:foxglove_derive"]
-agent = ["live_visualization"]
+cloud = ["live_visualization"]
```
bryfox: Is this feature intended to be temporary or long-lived? I could envision other agent functionality someday being included, and a feature specific to having an Agent installation might make sense.

python/foxglove-sdk-examples/cloud/README.md
```diff
@@ -0,0 +1,16 @@
+# Log to cloud live visualization
+
+An example from the Foxglove SDK.
+
+Connect to the cloud with Foxglove agent and log messages for live visualization to be viewed in the Foxglove app.
```
bryfox: This should mention that running an Agent is required.

python/foxglove-sdk/python/foxglove/__init__.py
```diff
@@ -86,6 +91,34 @@ def start_server(
...
+        *,
+        server_listener: ServerListener | None = None,
```
bryfox: type should be `CloudSinkListener` now. The argument name should probably change too (`listener`)?

python/foxglove-sdk/python/foxglove/__init__.py
bryfox: Oh, I wonder if the reason is the listener interface. That needs to inherit from `typing.Protocol` which maybe we couldn't do from native. I'm putting together the new example based on this branch; I'll fix things up here as needed and open a PR on top of this.

bryfox:
OK, I merged #701 into this

python/foxglove-sdk/python/foxglove/cloud.py
```diff
@@ -0,0 +1,61 @@
...
+        :param client: The client (id) that is unadvertising the channel.
+        :param client_channel_id: The client channel id that is being unadvertised.
+        """
+        return None
+
+    def on_message_data(
+        self, client: Client, client_channel_id: int, data: bytes
+    ) -> None:
+        """
+        Called by the server when a message is received from a client.
```
bryfox:
```suggestion
        Called when a message is received from a client.
```

Sorry, I missed this in #701.

rust/foxglove/Cargo.toml
eloff: Yeah, I think agent is better

PR #701, opened 2025-10-15
Python: update agent/cloud module, types and example

bryfox:
This builds on #698:

- Move the `start_cloud_sink` function to the top-level module, to match `start_server`
- Add missing type stubs & Listener definition
- Update the example to use video & client publish; the equivalent of #700

Unlike rust, the `on_message_data` callback receives only a channel ID, so the example is a little more verbose; this is a choice we made in #286.

### DIFF
--- a/python/foxglove-sdk-examples/cloud/main.py
+++ b/python/foxglove-sdk-examples/cloud/main.py
@@ -1,53 +1,79 @@
-import inspect
 import time
 
 import foxglove
-from foxglove.channels import LogChannel
-from foxglove.schemas import Log, LogLevel
+from foxglove.schemas import RawImage
 
-log_chan = LogChannel(topic="/log1")
+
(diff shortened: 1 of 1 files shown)


python/foxglove-sdk/python/foxglove/cloud.py
```diff
@@ -0,0 +1,61 @@
...
+        :param client: The client (id) that is unadvertising the channel.
+        :param client_channel_id: The client channel id that is being unadvertised.
+        """
+        return None
+
+    def on_message_data(
+        self, client: Client, client_channel_id: int, data: bytes
+    ) -> None:
+        """
+        Called by the server when a message is received from a client.
```
eloff, line 55: I'm not sure server is the right word here

eloff (approved):
LGTM

PR #703, opened 2025-10-16
Ignore foxglove_bridge in ROS1

chef-kyle:
### Changelog
Ensure that foxglove_bridge is ignored in ROS1.

### Docs

None

### Description
I'd like to add foxglove_msgs to [ros-o](https://github.com/ros-o/ros-o) via https://github.com/ubi-agni/ros-builder-action/pull/91 but catkin is picking up on the ROS2 only package `foxglove_bridge`.

For testing I'll point the ros-o build at my fork.

### DIFF
--- a/ros/src/foxglove_bridge/CATKIN_IGNORE
+++ b/ros/src/foxglove_bridge/CATKIN_IGNORE
@@ -1,0 +1,1 @@
+The version of foxglove_bridge in this repo does not support ROS 1. If you are a ROS 1 user, you can find the source code for the legacy version of foxglove_bridge in the ROS 1 foxglove_bridge repo at https://github.com/foxglove/ros-foxglove-bridge


chef-kyle:
Actually, not sure if this will fix the issue or not - but it's still useful for anyone pulling in this subfolder.

chef-kyle:
Looking at the source code for the ros-o builder this should indeed do the trick - https://github.com/ubi-agni/ros-builder-action/blob/ac34c23c75ca73c51d8417ff08d64e0805f60f35/src/build.sh#L207C1-L210C1

clalancette (approved):
This is fair enough for now, so I'll approve.

That said, there are new features to the ROS .msg format coming in the next version of ROS 2 (Lyrical).  While I'm not sure whether this repository will use those features, long-term you may have to end up forking to allow this to continue to work.

amacneil:
@chef-kyle so I guess you would be sad if we merged https://github.com/foxglove/foxglove-sdk/pull/549?

chef-kyle:
@amacneil I'd be a little sad, but also understand the reasoning.

PR #704, opened 2025-10-16
Update ROS badges for Jazzy & Kilted

jtbandes:
Same as #686 for J & K now that they have synced


ericmlujan:
Now just waiting for them to sync H and our badges will be good again :D

PR #163, opened 2025-01-31
Migrate python packages to poetry

amacneil:
### Changelog
None

### Docs
None

### Description
For consistency with our upcoming packages, migrate the existing python packages from pipenv to poetry.

PR #705, opened 2025-10-16
Switch from poetry to uv

gasmith:
Since we've been talking about the possibility of switching from poetry to uv, I figured I'd give it a quick try.

Other changes here:
- Updated some CI workflows to reuse `Container.mk` targets.
- Stripped unnecessary metadata from pyproject.toml files.
- Arranged examples in to a [uv workspace](https://docs.astral.sh/uv/concepts/projects/workspaces/), so they can share a venv.


.flake8
```diff
@@ -4,11 +4,14 @@ extend-ignore = E203
...
     __pycache__,
     .git,
+    .uv_cache,
+    .venv,
     dist,
     node_modules,
     python/foxglove-schemas-flatbuffer/foxglove_schemas_flatbuffer/*.py,
     python/foxglove-schemas-protobuf/foxglove_schemas_protobuf/*.py,
     python/foxglove-sdk/python/foxglove/_protobuf/*,
+    python/foxglove-sdk/.venv,
```
bryfox: I think this might be handled by `.venv` above, but check me on that

https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-exclude

python/foxglove-schemas-flatbuffer/pyproject.toml
```diff
@@ -5,33 +5,27 @@ description = "Precompiled flatbuffer schemas for Foxglove"
...
-    "wheel",
-  ] },
-  { path = "foxglove_schemas_flatbuffer/*.bfbs", format = [
-    "sdist",
-    "wheel",
-  ] },
-]
+[build-system]
+requires = ["uv_build>=0.9.3,<0.10.0"]
+build-backend = "uv_build"
```
bryfox, line 24: I checked the build outputs here; metadata has added classifier and `Metadata-Version: 2.4` instead of `2.3`, which is probably fine.

python/foxglove-sdk-examples/pyproject.toml
```diff
@@ -0,0 +1,14 @@
+[project]
+name = "foxglove-sdk-examples"
+version = "0.0.0"
+requires-python = ">=3.9"
+
+[tool.uv]
+package = false
+
+[tool.uv.workspace]
+members = [ "*" ]
```
bryfox: I was wondering if it's feasible to start omitting SDK version numbers in all examples (FG-11340), which would simplify the story as we add new features & examples. (The downside is that if we update an example to use a new feature, then we're still in the same position of needing to specify a min version for anyone who previously ran that example, but I don't think that situation has come up yet.)

The workspace makes that problem a little worse. It means that new examples requiring a new version will still need to specify their minimum version; specify "foxglove-sdk" will not work for anyone who ran any example before. One way to do that is to specify a local dependency, then update the example pyproject after a new release is available.

Does the workspace have any advantages other than reducing some boilerplate in the other pyproject files? If not, I wonder if we should accept the downside in the first paragraph and just use "foxglove-sdk" in all examples. I'm also open to other ideas here.

This isn't a uv-specific issue, but if we want to make this change, it's a good time to do so — before users generate lockfiles in this parent directory.

gasmith: I can think of two advantages:

- When installing the SDK by path (`--with ../../foxglove-sdk`), we have to build it for each `.venv`, which takes a few seconds. The common workspace means we only have to do that once.
- A single .venv means that you don't end up with dead `$example/.venv` dirs when we remove or rename examples, which might get picked up by the `runPythonSdkExamples` script.

Overall, I agree that the workspace isn't much of a win, and I agree that the lack of isolation is a concern. I'll get rid of it. And update the `runPythonSdkExamples` script to be a little more cautious about what it tries to run.

Regarding the version specification, I'm actually inclined towards a _tighter_ version specification. The best practice for users (who may copy-paste our examples) is to specify a range of compatible versions (`>=0.14.0,<0.15.0`). If we do that in our examples, we'll need to update the dependency versions as part of the release. The downside is that whenever we break compatibility, all of our examples are not only broken on main, but also specify the _wrong_ dependency - which is to say they're broken permanently. If we don't specify a version, our examples are still broken on main, but only until the next release.

One way out of that quagmire would be to adjust our release process to preemptively bump our version number. For example, after tagging release 0.x.y, we add a commit to bump the version to 0.x.(y+1). If, during the course of development leading up to the next release, we decide that we need to introduce a breaking change, we run another script to bump the version to 0.(x+1).0. As part of the breaking change, we'd update examples, and adjust the minimum version specification accordingly.

bryfox: >  _tighter_ version specification

Yeah, OK. I was hoping we could keep releases moving quickly enough that main-branch breakages would be minimal, but the channel filtering feature/example right now is evidence against that. (I think we'll cut a release this week.)

> `>=0.14.0,<0.15.0`

Do you think the upper bound is worthwhile as a matter of practice? I generally assume releases won't contain breaking changes affecting most examples; when they do, we'll update the examples. (We're still subject to the breakage on main until the next version is released...)

> One way out of that quagmire would be to adjust our release process to preemptively bump our version number

I'm missing something here; it sounds like you want to bump the release version and have examples reference that. Will `uv` install a version from the local package? I assumed it would only be looking at PyPI (unless we specify the local location), and error if a specified version wasn't found. Are editable installs the way around that? If there's a mechanism to load a version from PyPI but fall back to local, that sounds ideal.

gasmith: > > One way out of that quagmire would be to adjust our release process to preemptively bump our version number
> 
> I'm missing something here; it sounds like you want to bump the release version and have examples reference that. Will `uv` install a version from the local package? I assumed it would only be looking at PyPI (unless we specify the local location), and error if a specified version wasn't found. Are editable installs the way around that? If there's a mechanism to load a version from PyPI but fall back to local, that sounds ideal.

Your assumption is correct, and there doesn't seem to be a way to give uv a local fallback - I spent quite a while looking for a way to do that. The idea is that the breakage would be similar to what we have today, where the example cannot be (trivially) run on main until after we've tagged the release. The upshot is that versioning is _correct_. So once the release is tagged, the example will run, even if you're on an old commit, prior to the tag.

amacneil:
@jhurliman called it

https://github.com/foxglove/foxglove-sdk/pull/163#issuecomment-2626290869

PR #706, opened 2025-10-18
Update grid schema documentation

banisadr:
### Changelog
None

### Docs

This is a schema documentation update. 

### Description

Adding details of how the grid schema colorization works. See [motivation](https://foxglove.slack.com/archives/C079S680PC0/p1760687557078269) and [internal discussion](https://foxglove.slack.com/archives/C07RB3SSJE5/p1760739780238259).


typescript/schemas/src/internal/schemas.ts
```diff
@@ -1217,13 +1217,13 @@ const Grid: FoxgloveMessageSchema = {
...
...l four fields must be present with these exact names for RGB visualization to work. The order of fields doesn\'t matter, but the names must match exactly.\n\n:::\n\nRecommended type: `UINT8` (0-255 range) for standard 8-bit color channels.\n\nExample field definitions:\n\n**RGB color only:**\n\n```javascript\nfields: [\n  { name: "red", offset: 0, type: NumericType.UINT8 },\n  { name: "green", offset: 1, type: NumericType.UINT8 },\n  { name: "blue", offset: 2, type: NumericType.UINT8 },\n  { name: "alpha", offset: 3, type: NumericType.UINT8 },\n];\n```\n\n**RGB color with elevation (for 3D terrain visualization):**\n\n```javascript\nfields: [\n  { name: "red", offset: 0, type: NumericType.UINT8 },\n  { name: "green", offset: 1, type: NumericType.UINT8 },\n  { name: "blue", offset: 2, type: NumericType.UINT8 },\n  { name: "alpha", offset: 3, type: NumericType.UINT8 },\n  { name: "elevation", offset: 4, type: NumericType.FLOAT32 },\n];\n```\n\nWhen these fields are present, the 3D panel will offer additional "Color Mode" options including "RGBA (separate fields)" to visualize the RGB data directly. For elevation visualization, set the "Elevation field" to your elevation layer name.',
```
bryfox: Can we avoid markup which is specific to the docs site? This has a URL which will only work from docs pages; maybe that could be an absolute url, but could also just omit the link? The `::: note` admonition won't render correctly elsewhere. Maybe that's not a big deal, but it looks a little strange to me as plain text. We can adjust the script to strip or modify these from other schemas, but the TS schemas would still have them.

typescript/schemas/src/internal/schemas.ts
```diff
@@ -1283,7 +1283,7 @@ const VoxelGrid: FoxgloveMessageSchema = {
       type: { type: "nested", schema: PackedElementField },
       array: true,
       description:
-        "Fields in `data`. `red`, `green`, `blue`, and `alpha` are optional for customizing the grid's color.",
+        "Fields in `data`. See [Color fields](#color-fields) below for RGB visualization.",
```
bryfox: Same here; this doesn't make any sense outside the context of the documentation pages. Maybe we should strip all relative or anchor links in the script, but "below" is confusing.

bryfox (commented):
Ideally these schemas remain self-documenting in their comments, and we don't need to include content which is specific to the docs site. (These are also used to generate language-specific SDK reference documentation.)

We could update the generation script to improve this a little, but curious what you think. Another alternative is to produce output specific to the docs site here, and not leverage TS schemas for that.

typescript/schemas/src/internal/schemas.ts
banisadr: Good call out, fixed this

typescript/schemas/src/internal/schemas.ts
banisadr: Also fixed

banisadr:
@bryfox do you know why the cpp and python docs might be failing? Does it matter?

bryfox:
> @bryfox do you know why the cpp and python docs might be failing? Does it matter?

It does matter because this is what produces the [cpp](https://foxglove.github.io/foxglove-sdk/cpp/) and [python](https://foxglove.github.io/foxglove-sdk/python/) reference docs.

C++:
```
/home/runner/work/foxglove-sdk/foxglove-sdk/cpp/foxglove/include/foxglove/schemas.hpp:740: error: unable to resolve reference to 'color-fields' for \ref command
```

That anchor reference `(#color-fields)` is still in the doc comments.

Python:
```
/home/runner/work/foxglove-sdk/foxglove-sdk/python/foxglove-sdk/python/foxglove/schemas/__init__.py:docstring of foxglove.schemas.Grid:24:Literal block expected; none found.
```

The [literal block](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks) is `::`, so I think this is the `:::note` admonition tripping up sphinx.

It looks like you commented above that both of these were removed/resolved; maybe something went wrong with the merge?

I also see `poetry` lockfiles here — we just switched to `uv` in #705 (sorry for the timing!). Easiest thing to do might be to start fresh with only your internal schema changes and re-generate the rest.

PR #707, opened 2025-10-18
C++ support for CloudSink

eloff:
### Changelog
None

### Docs

None

### Description

This implements C++ wrappers for cloud sink and callbacks, very similar to what we have for WebSockerServer. It adds a rudimentary example based on the ws_server example.

I didn't add channel filter support. I can look into adding that on Monday.

### DIFF
--- a/cpp/foxglove/src/cloud_sink.cpp
+++ b/cpp/foxglove/src/cloud_sink.cpp
@@ -154,7 +154,10 @@ FoxgloveResult<CloudSink> CloudSink::create(
   return CloudSink(sink, std::move(callbacks), std::move(sink_channel_filter));
 }
 
-CloudSink::CloudSink(foxglove_cloud_sink* sink, std::unique_ptr<CloudSinkCallbacks> callbacks, std::unique_ptr<SinkChannelFilterFn> sink_channel_filter)
+CloudSink::CloudSink(
+  foxglove_cloud_sink* sink, std::unique_ptr<CloudSinkCallbacks> callbacks,
+  std::unique_ptr<SinkChannelFilterFn> sink_channel_filter
+)
     : callbacks_(std::move(callbacks))
     , sink_channel_filter_(std::move(sink_channel_filter))
     , impl_(sink, foxglove_cloud_sink_stop) {}


.github/workflows/ci.yml
```diff
@@ -164,7 +164,7 @@ jobs:
           git add --all
           changes=$(git status --porcelain)
           if [ -n "$changes" ]; then
-            echo "::error::The following generated files need to be updated. Run 'make generate' to update them."
+            echo "::error::The following generated files need to be updated. Run 'yarn generate' to update them."
```
bryfox: `make generate` is the recommended command to avoid versioning issues with rust, python, clang (which can produce differences in formats)

c/cbindgen.prelude.toml
```diff
@@ -39,6 +39,7 @@ prefix_with_name = false
 "target_family = wasm" = "__wasm__"
 
 [export.rename]
+FoxgloveServerStatusLevel = "foxglove_server_status_level"
```
bryfox: was this moved intentionally?

cpp/foxglove/include/foxglove/cloud_sink.hpp
```diff
@@ -0,0 +1,76 @@
...
+namespace foxglove {
+
+/// @brief The callback interface for a Cloud Sink.
+///
+/// These methods are invoked from the client's main poll loop and must be as low-latency as
+/// possible.
+///
+/// @note These callbacks may be invoked concurrently from multiple threads.
+/// You must synchronize access to your mutable internal state or shared resources.
+struct CloudSinkCallbacks {
```
bryfox, line 19: Any thoughts on the "Listener"/"Callback" naming now that you've revisited in this context? I recall this being a conscious decision with the WS server, but the inconsistency is slightly unfortunate when documenting concepts.

c/src/cloud_sink.rs
```diff
@@ -0,0 +1,255 @@
...
+///
+/// Resources must later be freed by calling `foxglove_server_stop`.
+///
+/// `port` may be 0, in which case an available port will be automatically selected.
+///
+/// Returns 0 on success, or returns a FoxgloveError code on error.
+///
+/// # Safety
+/// If `name` is supplied in options, it must contain valid UTF8.
+/// If `host` is supplied in options, it must contain valid UTF8.
```
bryfox: name & host aren't fields of these options

c/src/cloud_sink.rs
```diff
@@ -0,0 +1,255 @@
...
+    unsafe {
+        let result = do_foxglove_cloud_sink_start(options);
+        result_to_c(result, server)
+    }
+}
+
+unsafe fn do_foxglove_cloud_sink_start(
+    options: &FoxgloveCloudSinkOptions,
+) -> Result<*mut FoxgloveCloudSink, foxglove::FoxgloveError> {
+    let mut server = foxglove::CloudSink::new();
```
bryfox: nit: do we want to call this `server` (vs `handle`)?

c/src/cloud_sink.rs
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    let server = server.start_blocking()?;
+    Ok(Box::into_raw(Box::new(FoxgloveCloudSink(Some(server)))))
+}
+
+/// Stop and shut down `server` and free the resources associated with it.
+#[unsafe(no_mangle)]
+pub extern "C" fn foxglove_cloud_sink_stop(
+    server: Option<&mut FoxgloveCloudSink>,
```
bryfox: nit: `sink`? `handle`?

c/src/cloud_sink.rs
```diff
@@ -0,0 +1,255 @@
...
+) -> FoxgloveError {
+    let Some(server) = server else {
+        tracing::error!("foxglove_server_stop called with null server");
```
bryfox: "foxglove_cloud_sink_stop" (and update "server" if renamed above)

c/src/cloud_sink.rs
```diff
@@ -0,0 +1,255 @@
...
+        return FoxgloveError::ValueError;
+    };
+
+    // Safety: undo the Box::into_raw in foxglove_server_start, safe if this was created by that method
```
bryfox: foxglove_cloud_sink_start

bryfox:
> channel filter support

That doesn't exist for other languages yet either, right? I don't think that needs to happen here.

.github/workflows/ci.yml
eloff: That's make generate in Container.mk? The dev container?

The only other make generate I see is in the Python Makefile.

c/cbindgen.prelude.toml
eloff: No, definitely not. Not sure how I managed that

.github/workflows/ci.yml
bryfox: Yeah, the Container.mk. And then the primary Makefile defines a target for each of those that runs in a container:

https://github.com/foxglove/foxglove-sdk/blob/803ee03dbabc51908f808e0cb2b38fd381cdd9fb/Makefile#L21-L31

cpp/foxglove/include/foxglove/cloud_sink.hpp
eloff: Yeah, that's a little unfortunate. Being consistent with the naming matters more to me than listener vs callbacks as a name. Not something I want to address here though.

PR #709, opened 2025-10-20
Update build badges for Humble

ericmlujan:
### Changelog
None

### Docs
None

### Description
Like #704, update badges for ROS Humble now that the package sync is completed. Also remove badges from foxglove_bridge readme, as they're redundant with those at the top level.


jtbandes (approved):
🎉  we made it!

PR #711, opened 2025-10-21
Fix urdf workflow for SO100, add SO101 files

msadowski:
This PR updates the SO100 examples to work with updated URDF joint names. It also adds the files for SO101 visualization.

Fixes DR-21

When reviewing, you can skip SO100 and SO101 directories - these are directly copied for the SO100 repo and contain updated URDF structures.


python/foxglove-sdk-examples/so100-visualization/SO100/so100.urdf
bryfox, line 1: Do we need to include the license for these this content in the repository here? Maybe the example readme should also explain the source.

python/foxglove-sdk-examples/so101-visualization/main.py
```diff
@@ -155,16 +159,16 @@ def main():
...
             print(f"Failed to setup environment camera: {e}")
 
-    # Connect to SO-100 arm
-    config = SO100FollowerConfig(
+    # Connect to SO-101 arm
+    config = SO101FollowerConfig(
         port=args.robot_port, id=args.robot_id, use_degrees=True
     )
-    follower = SO100Follower(config)
+    follower = SO101Follower(config)
```
bryfox, line 162: There are commented-out lines above for users who want to use SO-100, but it looks like other code and messaging needs to change as well. Should we try to put this all behind a single boolean option (assuming it's worth continuing to support SO-100)?

bryfox: Also this example is still named so100; should it be updated?

python/foxglove-sdk-examples/so101-visualization/foxglove/lerobot_layout.json
```diff
@@ -254,7 +254,7 @@
           "label": "URDF",
           "layerId": "foxglove.Urdf",
           "sourceType": "url",
-          "url": "https://storage.googleapis.com/assets.foxglove.dev/demo-datasets/SO100/so100.urdf",
+          "url": "https://raw.githubusercontent.com/TheRobotStudio/SO-ARM100/refs/heads/main/Simulation/SO101/so101_new_calib.urdf",
```
bryfox: Do you want to link this to the copy added in this PR, or use a stable ref instead of main — in case this changes in the future?

Or you could serve the asset from the SDK :) ...but that might distract from the example.

python/foxglove-sdk-examples/so100-visualization/SO100/so100.urdf
msadowski: I added the licence to SO-101 files and mentioned where we took the files from in the README.md

python/foxglove-sdk-examples/so101-visualization/main.py
msadowski: Having slept on this, I removed any references to SO101, if anyone decides to use SO100, they should be quite easily able to modify the example

python/foxglove-sdk-examples/so101-visualization/foxglove/lerobot_layout.json
msadowski: Good call, updated it!

defunctzombie:
@msadowski Bryan is no longer at Foxglove so we need another reviewer here

python/foxglove-sdk-examples/so101-visualization/SO101/LICENSE
gasmith, line 1: Is this intentionally left blank?

msadowski: It's not blank, it was just moved from one directory to another, and github is apparently not showing the diff

PR #713, opened 2025-10-21
Python example: logging config and filtering

bryfox:
This adds a Python example demonstrating logging configuration and filtering which might be closer to a real-world application. 

For convenience in scripts, `foxglove.set_log_level()` will configure basic logging if it hasn't already been; for brevity, this is what our other examples use. The reference docs note this, and recommend configuring logging yourself, but we don't show you how.

This also adds rudimentary support for configuration of logging via `FOXGLOVE_LOG_LEVEL` (env var). It supports a subset of the env_logger format, excluding regex filters. (Note there's an `env_filter` crate, but it doesn't expose an API to let us work with the pyo3 logger here.) This moves python towards consistency with other languages, but one issue is that module names don't match the user-facing python libraries. We could document the level-only arguments ("debug", "info", etc) and omit mention of more targeted config — this is what we do with C++. But I've omitted it from the example for this reason.


python/foxglove-sdk/Cargo.toml
```diff
@@ -10,10 +10,12 @@ crate-type = ["cdylib"]
 
 [lints]
 workspace = true
+
 [dependencies]
 base64 = "0.22.1"
 bytes.workspace = true
 log.workspace = true
+maplit = "1.0.2"
```
gasmith: Could move to `[dev-dependencies]`

gasmith (approved):

PR #715, opened 2025-10-22
Support write metadata in c++ and add it to examples

eloff:
### Changelog

Add support to MCAP sink for write MCAP metadata.

### Docs

None

### Description

This builds on Jennifer's work to add support for write metadata to c++. I also add writing metadata to the mcap examples for all 3 languages, it's simple enough that I think it's more helpful than harmful, and doesn't warrant a full standalone example.

I have a chicken and egg problem with the Python example. It needs the new released version of the library from this PR so it can build. I will split the change to the example into its own PR so this can be merged and released first.

### DIFF
--- a/cpp/foxglove/include/foxglove/mcap.hpp
+++ b/cpp/foxglove/include/foxglove/mcap.hpp
@@ -17,8 +17,8 @@ enum foxglove_error : uint8_t;
 struct foxglove_mcap_writer;
 
 foxglove_error foxglove_mcap_write_metadata(
-  foxglove_mcap_writer* writer, const foxglove_string* name,
-  const foxglove_key_value* metadata, size_t metadata_len
+  foxglove_mcap_writer* writer, const foxglove_string* name, const foxglove_key_value* metadata,
+  size_t metadata_len
 );
 /// @endcond
 


bryfox:
I guess that c++ branch was built on #695, so all python changes are here, too. Would you plan to merge that as part of this PR, all together? There was some out-of-band feedback on that one; I could summarize it here or there.

bryfox:
> I have a chicken and egg problem with the Python example. It needs the new released version of the library from this PR so it can build. I will split the change to the example into its own PR so this can be merged and released first.

Note the python examples use `uv` as of 56 minutes ago 😄 ; the pyproject config will take a few tweaks following https://github.com/foxglove/foxglove-sdk/pull/705. @gasmith also wrote some thoughts on how to address this chicken & egg problem: https://github.com/foxglove/foxglove-sdk/pull/705#discussion_r2448927433

eloff:
> I guess that c++ branch was built on #695, so all python changes are here, too. Would you plan to merge that as part of this PR, all together? There was some out-of-band feedback on that one; I could summarize it here or there.

Yeah, I think we'll merge this all together here. I'll check out the feedback on that PR and make any changes I think are required.

bryfox:
> Yeah, I think we'll merge this all together here. I'll check out the feedback on that PR and make any changes I think are required.

👍 Out-of-band feedback on that PR: we should check the behavior of the MCAP libraries when writing multiple metadata records with the same name. SDK should probably match that behavior. If the MCAP libraries aren't consistent with each other, I think the SDK should be internally consistent; we should document the behavior in any case.

eloff:
> > Yeah, I think we'll merge this all together here. I'll check out the feedback on that PR and make any changes I think are required.
> 
> 👍 Out-of-band feedback on that PR: we should check the behavior of the MCAP libraries when writing multiple metadata records with the same name. SDK should probably match that behavior. If the MCAP libraries aren't consistent with each other, I think the SDK should be internally consistent; we should document the behavior in any case.

We're just handing off to the mcap library, so the behavior will match. I can look into what that is and document it.

python/foxglove-sdk/python/foxglove/tests/test_mcap.py
```diff
@@ -116,6 +116,55 @@ def test_write_to_different_contexts(make_tmp_mcap: Callable[[], Path]) -> None:
     assert len(contents1) < len(contents2)
 
 
+def test_write_metadata(tmp_mcap: Path) -> None:
```
bryfox: The branch from #695 must have some changes that weren't in the the base branch here. Notably, the python tests include an mcap reader to verify the metadata written ([ref](https://github.com/foxglove/foxglove-sdk/pull/695/files#diff-ead84cf03cd31c8dc4278b9b392452ab43d1da1d456f228e69f11a1e6af23567R119)). Maybe you can cherry-pick the later commits that seem useful, or just pull that code in?

python/foxglove-sdk/python/foxglove/tests/test_mcap.py
```diff
@@ -116,6 +116,55 @@ def test_write_to_different_contexts(make_tmp_mcap: Callable[[], Path]) -> None:
...
+
+        # Log some messages
+        chan.log({"test": "data"})
+
+    # Verify file was created and has some content
+    assert tmp_mcap.exists()
+    assert tmp_mcap.stat().st_size > 0
+
+
+def test_write_metadata_multiple_records(tmp_mcap: Path) -> None:
```
bryfox: I'd delete this one; `test_write_metadata` already tests this. Some other similar comments in this discussion: https://github.com/foxglove/foxglove-sdk/pull/695#pullrequestreview-3323397399

cpp/examples/mcap/src/main.cpp
```diff
@@ -15,6 +15,14 @@ int main() {
   }
   auto writer = std::move(writer_result.value());
 
+  // If you want to add some MCAP metadata: https://mcap.dev/spec#metadata-op0x0c
+  std::map<std::string, std::string> metadata = {{"os", "linux"}, {"arch", "x64"}};
```
bryfox, line 19: I like your change to the rust example that uses the actual os/arch

c/src/channel.rs
```diff
@@ -182,6 +182,62 @@ pub unsafe extern "C" fn foxglove_mcap_close(
...
+    metadata_len: usize,
+) -> Result<(), foxglove::FoxgloveError> {
+    let name = unsafe { name.as_utf8_str() }.map_err(|e| {
+        foxglove::FoxgloveError::Utf8Error(format!("metadata name is invalid: {e}"))
+    })?;
+
+    // Convert C key-value array to BTreeMap
+    let mut metadata_map = std::collections::BTreeMap::new();
+    for i in 0..metadata_len {
+        let kv = unsafe { &*metadata.add(i) };
```
bryfox: Can we add a null check for metadata?

bryfox (commented):
There are some good changes we can pull in for the python tests (see comment there) but this looks good to me overall. Would like @ericmlujan to review at least the c++ changes.

python/foxglove-sdk/python/foxglove/tests/test_mcap.py
eloff: Merged

c/src/channel.rs
eloff: good catch

bryfox (approved):
I like that the rust example uses os & arch from env rather than hardcoding values, but that's a tiny nit; it doesn't distract from the example.

cpp/examples/mcap/src/main.cpp
ericmlujan:
```suggestion
  const std::map<std::string, std::string> metadata = {{"os", "linux"}, {"arch", "x64"}};
```

cpp/examples/mcap/src/main.cpp
```diff
@@ -15,6 +15,14 @@ int main() {
...
+  foxglove::FoxgloveError err = writer.writeMetadata("platform", metadata.begin(), metadata.end());
+  if (err != foxglove::FoxgloveError::Ok) {
+    std::cerr << "Failed to write metadata: " << foxglove::strerror(err) << '\n';
```
ericmlujan:
```suggestion
    std::cerr << "Failed to write metadata: " << foxglove::strerror(err) << std::endl;
```

cpp/foxglove/include/foxglove/mcap.hpp
```diff
@@ -6,12 +6,22 @@
...
 
 #include "channel.hpp"
 
+/// @cond foxglove_internal
 enum foxglove_error : uint8_t;
 struct foxglove_mcap_writer;
 
+foxglove_error foxglove_mcap_write_metadata(
+  struct foxglove_mcap_writer* writer, const struct foxglove_string* FOXGLOVE_NONNULL name,
+  const struct foxglove_key_value* metadata, size_t metadata_len
```
ericmlujan: A lot of these C-like constructs in your forward declaration I believe are unnecessary in C++ (like prefixing the typename with `struct`). I made this change locally and it seemed to link OK.

```suggestion
  foxglove_mcap_writer* writer, const foxglove_string* name,
  const foxglove_key_value* metadata, size_t metadata_len
```

PR #716, opened 2025-10-22
Include ROS distro information in server info message

bryfox:
### Changelog
Fix: ROS Foxglove bridge supports click-to-publish in 3d panel

### Description

In order to support click-to-publish in the 3d panel on a WebSocket connection, the ServerInfo message needs to include the value of the ROS_DISTRO environment variable. (Publishing is only supported for ROS sources right now.) Prior versions of the Bridge included this but the new releases for ROS2 do not, and the SDK doesn't expose a mechanism for it.

This PR adds an undocumented `server_info` option to the Rust and C++ server options. This metadata will be included in the ServerInfo message sent to the app. The only use case today is to convey ROS distro, but we may extend in the future. I named this "server_info" instead of "metadata" since it's specifically for information sent to clients, but am open to better names.

This does not add python support, and we have no workarounds for people wanting to spoof support for 3d publishing there. ERT-689 covers this (bringing 3d publish support to non-ROS systems).

#### Alternatives considered

1. Set ROS_DISTRO in the SDK directly. This PR originally did this, which opens the door for easy workarounds for people, but has the downsides noted in comments.

2. Expose a ROS-specific option or some general metadata option, with no clear reason (or use) for the interface beyond the ROS distro. Since the server here is intended to serve the Foxglove app, including other arbitrary metadata isn't useful.


gasmith:
> One concern I have with this approach is that it is "leaking" ROS details into the low-level websocket server implementation.

I'm inclined to agree. I also worry about some non-bridge, user-implemented application picking up this metadata by mistake, simply because it's running in a ROS environment.

I'd prefer to see the bridge plumb this metadata explicitly, so that the SDK remains agnostic to ROS things. Regarding @bryfox's concern about this not being generally useful for SDK users, I think it'd be sufficient to hide it using `#[doc(hidden)]` in rust and `@cond foxglove_internal` in C++. And simply avoid exposing it in Python.

ericmlujan (commented):
+1 to @clalancette's opinion, I feel this lives more naturally in Foxglove Bridge. For some of our other work, I've also filed FLE-107 to expose `metadata` within ServerInfo from the SDK; Bridge could then use this.

bryfox:
It's unanimous, then. I updated the changes per suggestions above. TIL that we can use `@cond` on struct fields; I removed tls_identity from docs too since it was documented as internal — but we could leave that alone if desired.

I called the new option `server_info`; I think 'metadata' is too generic here, and want to convey somehow that this information will be sent to clients. Open to suggestions.

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -114,6 +114,11 @@ FoxgloveBridge::FoxgloveBridge(const rclcpp::NodeOptions& options)
   sdkServerOptions.supported_encodings = {"cdr", "json"};
   sdkServerOptions.capabilities = _capabilities;
   sdkServerOptions.context = _serverContext;
+
+  if (rosDistro && strlen(rosDistro) > 0) {
```
bryfox, line 118: I don't think the null check is technically needed here since we don't have one above; ROS_DISTRO must be guaranteed to be set? I added it anyway.

@ericmlujan since you have a bridge set up locally, would you mind running a quick manual verification of click-to-publish from a 3d panel?

clalancette: I think that a check for NULL is a good idea; in practice it won't really ever be a problem, but we may as well be robust.

rust/foxglove/src/websocket/server.rs
```diff
@@ -507,17 +512,17 @@ impl Server {
 
     /// Builds a server info message.
     fn server_info(&self) -> ServerInfo {
+        let mut metadata = self.server_info.clone();
+        metadata.insert("fg-library".into(), get_library_version());
```
clalancette, line 520: We should probably add a check and panic here to make sure that `self.server_info` doesn't already contain `fg-library`.  It is highly unlikely that a user of the API would set this, but if they did, it could be confusing why their value was overridden.

rust/foxglove/src/websocket/server.rs
```diff
@@ -218,6 +222,7 @@ impl Server {
             services: parking_lot::RwLock::new(ServiceMap::from_iter(opts.services.into_values())),
             fetch_asset_handler: opts.fetch_asset_handler,
             tasks: parking_lot::Mutex::default(),
+            server_info: opts.server_info.unwrap_or_default(),
             stream_config,
```
clalancette: Nit: just to be consistent with the struct fields above, I'd suggest putting this after stream_config.




```suggestion
            stream_config,
            server_info: opts.server_info.unwrap_or_default(),
```

rust/foxglove/src/websocket/server.rs
bryfox: Good point. Is a panic what you'd want? We do have precedent for 'reserving' prefixes in metadata via documentation. Since this is internal and not for client use (at least yet), I think either is reasonable.

clalancette: I don't have a strong preference.  A panic seems the easiest to debug, but if we have precedent for doing it via documentation, that's OK with me.  If we go the documentation route, I'll suggest outputting a warning message to the console.

bryfox: Since this value is only used internally by the app (a user wouldn't even know if the value was overridden or not), I'm inclined to not panic. But happy to change if you think that's the correct approach to signal that this shouldn't be used.

bryfox: Sorry, I replied without seeing your follow-up above, but I went with the warning message.

clalancette: Looks good to me, thanks.

c/src/server.rs
```diff
@@ -91,6 +94,14 @@ pub struct FoxgloveServerOptions<'a> {
     pub supported_encodings: *const FoxgloveString,
     pub supported_encodings_count: usize,
 
+    /// Optional information about the server, which is shared with clients.
+    ///
+    /// # Safety
+    /// - If provided, the `server_info` must be a valid pointer to an array of valid
+    ///   `FoxgloveKeyValue`s with `server_info_count` elements.
+    pub server_info: *const FoxgloveKeyValue,
+    pub server_info_count: usize,
```
ericmlujan, line 103: Do we want to wrap these in a struct?

ericmlujan:
Doing a quick code review here, but was able to E2E test click-to-publish on a local build of this PR replaying a nuScenes mcap and that seemed to work swimmingly (seeing output in the `/clicked_point` topic in the terminal)
image

cpp/foxglove/include/foxglove/server.hpp
```diff
@@ -216,12 +216,20 @@ struct WebSocketServerOptions {
   std::vector<std::string> supported_encodings;
   /// @brief A fetch asset handler callback.
   FetchAssetHandler fetch_asset;
+  /// @brief A sink channel filter callback.
+  SinkChannelFilterFn sink_channel_filter;
+  /// @cond foxglove_internal
```
ericmlujan: Did we want to make `tls_identity` foxglove internal? I thought this was shipped. It's in the documentation, at least?

cpp/foxglove/include/foxglove/server.hpp
```diff
@@ -216,12 +216,20 @@ struct WebSocketServerOptions {
...
   /// This option is under active development and may change.
   std::optional<TlsIdentity> tls_identity = std::nullopt;
-  /// @brief A sink channel filter callback.
-  SinkChannelFilterFn sink_channel_filter;
+  /// @endcond
+  /// @cond foxglove_internal
+  /// @brief (internal) Information about the server, which is shared with clients.
+  ///
+  /// This option is for internal use only and may change.
+  std::optional<std::map<std::string, std::string>> server_info = std::nullopt;
```
ericmlujan, line 229: Nit: std::map is ordered by default, do we want ordering here?

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
ericmlujan: Alternatively, we could load ROS_DISTRO into a `std::string` after reading it from the env. Probably is nicer than it hanging around in a C-style string

rust/foxglove/src/websocket/server.rs
```diff
@@ -507,17 +513,20 @@ impl Server {
 
     /// Builds a server info message.
     fn server_info(&self) -> ServerInfo {
+        let mut metadata = self.server_info.clone();
+        if metadata.contains_key("fg-library") {
```
ericmlujan, line 517: Maybe not for now, but if we're going to be exposing `server_info` more broadly, it may make sense to have a static set of reserved keys defined that we check against and either warn or flat-out don't allow overriding. Thoughts?

c/src/server.rs
bryfox: I was following the example of `supported_encodings`/`supported_encodings_count` above (also in cloud sink). I think if we want to change these together in the future, we can do so freely.

cpp/foxglove/include/foxglove/server.hpp
bryfox: I'll revert this change.

It's in the documentation as "(internal)" — only because I didn't realize we could use `@cond` on an individual struct field. But I suppose there's no good reason to remove it now that it's there.

cpp/foxglove/include/foxglove/server.hpp
bryfox: Since this is internal & undocumented right now and we expect at most one entry, I'm inclined to leave it. I think we could freely update this with FLE-107 if we go that route. Please let me know if you disagree, though.

rust/foxglove/src/websocket/server.rs
bryfox: Yeah, see also https://github.com/foxglove/foxglove-sdk/pull/716#discussion_r2455358525. We could also consider reserving all "fg-" prefixed keys. I think we should revisit if/once we make this a publicly documented interface.

cpp/foxglove/include/foxglove/server.hpp
clalancette: My hot take on `std::map` vs `std::unordered_map` is that `std::map` is more balanced in terms of performance.  That is, a red-black tree is `O(log n)` for all operations in both average and worst case, while a hash table is `O(1)` in the average case but `O(n)` in the worst case.  Thus my general preference is to stick with `std::map` in all circumstances unless there is a demonstrated need for the best/average case performance of a hash table.

All of this context is probably moot for this use-case, as I don't expect that there will be a ton of keys in this map.  So I'd be fine with either (though `std::map` is shorter to type :rofl: ).

PR #718, opened 2025-10-22
notebook: Add custom build backend for codegen

gasmith:
This is built on top of #663, and attempts to address the awkward interaction between the python build process and the frontend codegen that we need to do as part of packaging.

The original approach, of putting a guard in `build.rs` has numerous disadvantages. Chief among them is that it completely breaks building from sdist, because maturin rearranges our sources files in such a way that `widget.js` is not where `build.rs` expects it to be. It's also a great disappointment that `cargo test` and friends now have an artificial dependency on a frontend codegen step that they don't really require. And none of this gets us to the point where `uv build` does what you'd expect - it just yells at the developer to go run `yarn`.

Ideally, we want this all to be baked into the python build system, so that when you run `uv build`, you get a distribution with the latest up-to-date frontend assets.

Maturin doesn't offer any hooks for running a codegen step, so this change wraps maturin in our own custom PEP 517 build backend. Thankfully, there's a way to do this entirely within tree, so we don't need a separate package on pypi. The custom build backend only attempts to do codegen when running in-tree. When building from sdist, it's just a passthrough to maturin.

I've also updated the CI python workflow. We build the sdist, and then use that to build wheels. This ensures that sdist builds are working, and ensures that we're always generating the frontend assets in the same way. It's a little hacky that we're using PyO3/maturin-action (which bypasses our build system) for building wheels, but it's safe since our build system would effectively just call maturin anyway. I tried using cibuildwheel instead of maturin-action, and it was a bit of a disaster. And in any case, we need to invoke `maturin build` directly for cross-compiling the wasm binary, because maturin's PEP 517 backend doesn't seem support cross compilation.

https://linear.app/foxglove/issue/FG-13367/integrate-front-end-assets-build-into-maturins-workflow

### DIFF
--- a/python/foxglove-sdk/custom_build.py
+++ b/python/foxglove-sdk/custom_build.py
@@ -46,7 +46,7 @@ def _frontend_codegen(editable: bool = False) -> None:
     # For editable installs, don't minify.
     build_target = "build" if editable else "build:prod"
     cmds = [
-        ["yarn"],
+        ["yarn", "install"],
         ["yarn", build_target],
     ]
     for cmd in cmds:


surbina:
If we have this custom backend to take care of building the front-end assets, do we need the [notebook-frontend](https://github.com/foxglove/foxglove-sdk/pull/663/files#diff-092659a9bd0068791326fb1d08f005532e5aeb983b999a7bd32c51deee0a71d5R23) job?

> We could perhaps merge this into @surbina's branch and hack on it more there.

Feel free to do this, and we can both keep working on that branch.

gasmith:
> If we have this custom backend to take care of building the front-end assets, do we need the [notebook-frontend](https://github.com/foxglove/foxglove-sdk/pull/663/files#diff-092659a9bd0068791326fb1d08f005532e5aeb983b999a7bd32c51deee0a71d5R23) job?

Good call. I've reworked the CI workflow, so that we build an sdist (which generates the frontend assets), and then pass that to the various wheel-building jobs. I think this is better, since CI no longer needs to care about where the frontend assets live. It also gives us an opportunity to validate that builds from sdist work properly, which we weren't doing before.

python/foxglove-sdk/CONTRIBUTING.md
```diff
@@ -6,50 +6,30 @@
...
 To test the notebook integration:
 
 ```sh
-# Install Jupyter
-uv pip install jupyterlab
-
-# Build SDK with notebook extra
-uv run maturin develop -E notebook
+# Install Jupyter and SDK with notebook extra
+uv pip install jupyterlab -e '.[notebook]'
```
surbina: I got the following error while running these instructions:

Image

Should we add this to this readme as well?

gasmith: I decided to put a section closer to the top, reminding the user to do a `uv sync`, which is a little more ergonomic than using `uv venv` directly.

python/foxglove-sdk/custom_build.py
```diff
@@ -0,0 +1,61 @@
...
+    # sources.
+    notebook_frontend = Path.cwd() / "notebook-frontend"
+    if not notebook_frontend.exists():
+        print("[custom-build] no frontend sources; skipping frontend codegen.")
+        return
+
+    # For editable installs, don't minify.
+    build_target = "build" if editable else "build:prod"
+    cmds = [
+        ["yarn"],
```
jtbandes: nit: maybe `yarn install` for clarity

python/foxglove-sdk/custom_build.py
```diff
@@ -0,0 +1,61 @@
...
+    # pyproject.toml.
+    os.environ["MATURIN_NO_MISSING_BUILD_BACKEND_WARNING"] = "1"
+
+    # If we're building from an sdist, there's nothing to do here. We packaged
+    # the compiled frontend assets when packaging the sdist, and we omit the
+    # sources.
+    notebook_frontend = Path.cwd() / "notebook-frontend"
+    if not notebook_frontend.exists():
+        print("[custom-build] no frontend sources; skipping frontend codegen.")
+        return
```
jtbandes, line 44: wonder if we should also check that the build products we expect to exist actually do exist?

gasmith: Yes. Done!

PR #726, opened 2025-10-28
python: add api to create layouts

surbina:
TODO:
- [ ] Document panels/props
- [x] Add classes to create config for panels:
    - [x] 3D
    - [x] Audio
    - [x] Diagnostics Details ROS
    - [x] Diagnostics Summary ROS
    - [x] Image
    - [x] Indicator
    - [x] Gauge
    - [x] Teleop
    - [x] Map
    - [x] Markdown
    - [x] Parameters
    - [x] Plot
    - [x] Publish
    - [x] Service Call
    - [x] Raw Messages
    - [x] Log
    - [x] State Transitions
    - [x] Table
    - [x] Topic Graph
    - [x] Transform Tree
    - [x] Data Source Info
    - [x] Variable Slider
    - [x] User Scripts

### Changelog

Python SDK: adds classes to programmatically create layouts that can be passed to the Jupyter embedded viewer

### Docs

https://linear.app/foxglove/issue/ERT-1034/document-python-api-to-create-layouts-programmatically

### Description

This PR adds classes to aid in creating layouts that can be passed to the Jupyter embedded viewer. The classes will provide some defaults when appropriate and run some minor validations.

Here is a simple example of how the API would be used:

```python
from foxglove.layouts import Layout, Variables, ItemList, Item
from foxglove.layouts.panels import MarkdownPanel, RawMessagesPanel

# Simple example:
# - A horizontal row with two items:
#     - a markdown panel
#     - a raw messages panel
simple_layout = Layout(
  Variables(test="test-value"),
  ItemList(
    Item(
      ratio=1,
      content=MarkdownPanel(
        markdown="My panel markdown",
        font_size=14
      )
    ),
    Item(
      ratio=1,
      content=RawMessagesPanel(
        expansion="all",
        topic_path="/clicked_point"
      )
    )
  )
)

print(simple_layout.to_json())
```

The following notebooks includes a more complex example and can be used to test the code:

[test-api-layout.ipynb](https://github.com/user-attachments/files/23197802/test-api-layout.ipynb)

### DIFF
--- a/python/foxglove-sdk/python/foxglove/layouts/__init__.py
+++ b/python/foxglove-sdk/python/foxglove/layouts/__init__.py
@@ -175,7 +175,7 @@ class Layout:
             "version": self.version,
             "variables": self.variables,
             "items": self.items.to_dict(),
-            "user_nodes": {k: v.to_dict() for k, v in self.user_nodes.items()},
+            "userNodes": {k: v.to_dict() for k, v in self.user_nodes.items()},
         }
 
 


python/foxglove-sdk/python/foxglove/layouts/__init__.py
```diff
@@ -0,0 +1,201 @@
...
+
+from __future__ import annotations
+
+import json
+import random
+import string
+from typing import Any, Literal
+
+
+class Variables:
```
james-rms: This class seems redundant IMO.

james-rms:
My thoughts, from most important to least:
* Please establish what high-level goals we are trying to achieve here. Do we want them to be able to specify panels  intuitively, given the panel docs page? If i want an indicator, I should be able to look at https://docs.foxglove.dev/docs/visualization/panels/indicator and know what all these arguments mean? do the users really want a JSON file as their output, or is there a corresponding embed API that they can feed this into?
* I think you could reduce a lot of boilerplate using Python's dataclasses or record types etc.
* Careful about using classes to specify API arguments, this implies that the user can create their own subclasses and they'll do something useful. Consider using a union of all panel types instead.

python/foxglove-sdk/python/foxglove/layouts/__init__.py
```diff
@@ -0,0 +1,201 @@
...
+
+    The ratio determines the proportion of space this item takes up, similar to CSS flex ratio.
+    For example, if a list has 3 items with ratios 1, 2, and 3, the first item takes 1/6 of the
+    space, the second item takes 2/6, and the third item takes 3/6.
+
+    :param ratio: Proportion of space this item occupies
+    :param content: Content of the item (panel, tabs, or item list)
+    """
+
+    def __init__(self, ratio: float, content: Panel | Tabs | ItemList) -> None:
```
james-rms:
```suggestion
    def __init__(self, ratio: float, content: UnionOfAllSupportedPanels | Tabs | ItemList) -> None:
```
I'd opt to define a type as the union of all supported panels, and accept that type here. The panel superclass implies that users can create their own panel subclasses, which will not go well for them.

python/foxglove-sdk/python/foxglove/layouts/__init__.py
```diff
@@ -0,0 +1,201 @@
...
+class Panel:
+    """
+    A panel in a layout.
+
+    :param id: Unique identifier for the panel
+    :param type: Type of the panel (e.g., "Plot", "3D", "Raw Messages")
+    :param config: Configuration dictionary for the panel
+    """
+
+    def __init__(self, type: str, id: str | None = None, **panel_config: Any) -> None:
```
james-rms: This kinda implies that you can create a panel from any set of args, which isn't true - there's only a few supoprted panels. I'd create a panel [protocol](https://typing.python.org/en/latest/spec/protocol.html) and have the supported foxglove panels implement it.

python/foxglove-sdk/python/foxglove/layouts/panels.py
```diff
@@ -0,0 +1,40 @@
...
+
+from . import Panel
+
+
+class MarkdownPanel(Panel):
+    def __init__(self, *, markdown: str | None = None, font_size: int | None = None):
+        super().__init__("Markdown", markdown=markdown, fontSize=font_size)
+
+
+class RawMessagesPanel(Panel):
```
james-rms, line 120: how would one support [schema converters](https://docs.foxglove.dev/docs/visualization/panels/raw-messages#message-conversion) in this scheme?

surbina:
@james-rms 

> Please establish what high-level goals we are trying to achieve here.

Please refer to the [project overview page](https://linear.app/foxglove/project/programmatically-create-layouts-in-jupyter-notebooks-528fd4fe3cc0/overview) and this [thread](https://foxglove.slack.com/archives/C0469RNBPEW/p1758815770729249) with prior discussion.

> Do we want them to be able to specify panels intuitively, given the panel docs page?

I think the panel docs page documents our UI, and the API should have [its own documentation](https://linear.app/foxglove/issue/ERT-1034/document-python-api-to-create-layouts-programmatically). This is because the UI can change frequently (reorganizing components, changing copy, etc.). We don't want to burden the API with keeping up with that pace.

> do the users really want a JSON file as their output, or is there a corresponding embed API that they can feed this into?

The embedded viewer will provide an API to feed the layouts created by this API. This allows us to abstract the implementation detail (e.g., mosaic's binary tree). Users have asked for a way to manipulate ("merge") layouts, which will be possible either via Python code or via the JSON output ([ref](https://linear.app/foxglove/project/programmatically-create-layouts-in-jupyter-notebooks-528fd4fe3cc0/overview#background-5635fcc9)).

> Careful about using classes to specify API arguments, this implies that the user can create their own subclasses and they'll do something useful. Consider using a union of all panel types instead.

The API is intended to allow users to define their own panel configurations. This is to support panels from third-party extensions. ([ref](https://linear.app/foxglove/project/programmatically-create-layouts-in-jupyter-notebooks-528fd4fe3cc0/overview#requirements-aa96b9ec))

python/foxglove-sdk/python/foxglove/layouts/__init__.py
surbina: We want to support panels defined in third-party extensions that we are not aware of. At the same time, we want to allow users to create their own helpers for these panel configurations. The union type here would prevent us from achieving this.

python/foxglove-sdk/python/foxglove/layouts/__init__.py
surbina: > This kinda implies that you can create a panel from any set of args, which isn't true - there's only a few supoprted panels.

Actually, you can create a panel config from any set of args. This is how we support panels from third-party extensions.

python/foxglove-sdk/python/foxglove/layouts/panels.py
surbina: The selected converter is kept in memory rather than in the layout.

python/foxglove-sdk/python/foxglove/layouts/__init__.py
surbina: I replaced this with a dict.

python/foxglove-sdk/python/foxglove/layouts/__init__.py
james-rms: ok, then this API doesn't give you any certainty over whether your config is valid or not? how are users expected to sync this up with what the app expects?

PR #727, opened 2025-10-28
playground: styling updates

jtbandes:
### Changelog
None

### Docs

None

### Description

Playground visual updates:
- Use MUI components with Foxglove theme
- Top bar only covers editor, not embed
- Resizable split view
- Removed output pane at the bottom
- Toast styling tweaks


julielwong (commented):
I played around with the prototype and it matches the design well. Looks good to me!

achim-k:
Usability feedback: As a user, I would expect the `Run` button to also start playback

PR #728, opened 2025-10-28
Introduce PlaybackControlRequest types

ericmlujan:
### Changelog
None

### Docs
None

### Description
Introduce new PlaybackControlRequest protocol message. Currently guarding behind the `unstable` ff and excluding from documentation.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -26,9 +26,6 @@ jobs:
 
   build-python-docs:
     runs-on: ubuntu-latest
-    defaults:
-      run:
-        working-directory: python/foxglove-sdk
     steps:
       - uses: actions/checkout@v5
       - uses: actions/setup-python@v6
@@ -40,10 +37,7 @@ jobs:
           node-version: 22
           cache: yarn
       - uses: astral-sh/setup-uv@v7
-      - run: uv sync --all-extras
-      - run: uv pip install --editable '.[notebook]'
-      - name: Sphinx build
-        run: uv run sphinx-build --fail-on-warning ./python/docs ./python/docs/_build
+      - run: make -f Container.mk docs-python
       - name: Upload artifacts
         uses: actions/upload-artifact@v4
         with:
--- a/.github/workflows/publish-ros-bridge.yml
+++ b/.github/workflows/publish-ros-bridge.yml
@@ -33,7 +33,7 @@ jobs:
           credentials_json: ${{ secrets.PUBLISHER_KEY }}
 
       - name: Setup gcloud
-        uses: google-github-actions/setup-gcloud@v2
+        uses: google-github-actions/setup-gcloud@v3
 
       - name: Configure gcloud docker
         run: gcloud auth configure-docker us-central1-docker.pkg.dev --quiet
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -357,7 +357,7 @@ jobs:
           path: python/foxglove-sdk
       - run: find python/foxglove-sdk/wheels-*
       - name: Generate artifact attestation
-        uses: actions/attest-build-provenance@v2
+        uses: actions/attest-build-provenance@v3
         with:
           subject-path: "python/foxglove-sdk/wheels-*/*"
       - name: Publish to PyPI
--- a/Container.mk
+++ b/Container.mk
@@ -31,6 +31,17 @@ benchmark-python:
 	uv --directory python/foxglove-sdk pip install --editable '.[notebook]'
 	uv --directory python/foxglove-sdk run pytest --with-benchmarks
 
+.PHONY: docs-python
+docs-python:
+	uv --directory python/foxglove-sdk lock --check
+	uv --directory python/foxglove-sdk sync --all-extras
+	uv --directory python/foxglove-sdk pip install --editable '.[notebook]'
+	uv --directory python/foxglove-sdk run sphinx-build --fail-on-warning ./python/docs ./python/docs/_build
+
+.PHONY: clean-docs-python
+clean-docs-python:
+	rm -rf python/foxglove-sdk/python/docs/_build
+
 .PHONY: lint-rust
 lint-rust:
 	cargo fmt --all --check
--- a/cpp/Doxyfile
+++ b/cpp/Doxyfile
@@ -1066,7 +1066,7 @@ EXCLUDE_SYMLINKS       = NO
 # Note that the wildcards are matched against the file with absolute path, so to
 # exclude all test directories for example use the pattern */test/*
 
-EXCLUDE_PATTERNS       = */tests/* */expected.hpp
+EXCLUDE_PATTERNS       = */tests/* */expected.hpp */cloud_sink.*
 
 # The EXCLUDE_SYMBOLS tag can be used to specify one or more symbol names
 # (namespaces, classes, functions, etc.) that should be excluded from the
--- a/playground/src/Editor.tsx
+++ b/playground/src/Editor.tsx
@@ -1,3 +1,4 @@
+import { useTheme } from "@mui/material";
 import * as monaco from "monaco-editor";
 import { forwardRef, useEffect, useImperativeHandle, useRef } from "react";
 
@@ -80,6 +81,11 @@ export const Editor = forwardRef<EditorInterface, EditorProps>(
         editorRef.current = null;
       };
     }, [initialValue, runner]);
+
+    const isDark = useTheme().palette.mode === "dark";
+    useEffect(() => {
+      monaco.editor.setTheme(isDark ? "vs-dark" : "vs");
+    }, [isDark]);
 
     useImperativeHandle(
       ref,
--- a/playground/src/Playground.tsx
+++ b/playground/src/Playground.tsx
@@ -46,6 +46,9 @@ const useStyles = tss.create(({ theme }) => ({
   },
   toast: {
     fontSize: theme.typography.body1.fontSize,
+    backgroundColor: theme.palette.background.paper,
+    color: theme.palette.text.primary,
+    boxShadow: theme.shadows[4],
   },
   toastMonospace: {
     maxWidth: "none",
@@ -273,7 +276,6 @@ export function Playground(): React.JSX.Element {
         <FoxgloveViewer
           ref={viewerRef}
           style={{ width: "100%", height: "100%", overflow: "hidden" }}
-          colorScheme="light"
           data={dataSource}
           layout={selectedLayout}
         />
@@ -405,12 +407,12 @@ const DEFAULT_LAYOUT_DATA = {
   layout: {
     first: "3D!1ehnpb2",
     second: {
-      direction: "column",
+      direction: "row",
       second: "Plot!30ea437",
       first: "RawMessages!2zn7j4u",
-      splitPercentage: 67.0375521557719,
+      splitPercentage: 33,
     },
-    direction: "row",
+    direction: "column",
     splitPercentage: 60.57971014492753,
   },
 };
--- a/playground/src/RunnerWorker.ts
+++ b/playground/src/RunnerWorker.ts
@@ -120,7 +120,7 @@ export class RunnerWorker {
       },
     });
     this.#abortController.signal.throwIfAborted();
-    await pyodide.loadPackage(["jedi", "micropip", "protobuf"]);
+    await pyodide.loadPackage(["jedi", "micropip"]);
     this.#abortController.signal.throwIfAborted();
     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
     const micropip = pyodide.pyimport("micropip");
@@ -132,6 +132,7 @@ export class RunnerWorker {
 
   async run(code: string): Promise<string | undefined> {
     const pyodide = await this.#pyodide;
+    await pyodide.loadPackagesFromImports(code);
     pyodide.runPython(
       `
         import os, pathlib, shutil
--- a/playground/webpack.config.ts
+++ b/playground/webpack.config.ts
@@ -103,11 +103,19 @@ export default (_env: unknown, argv: WebpackArgv): Configuration => {
         // Pyodide is distributed with a list of packages that it knows about. These filenames match
         // the ones it will try to download at runtime when calling pyodide.loadPackage(). See the
         // list at: https://pyodide.org/en/stable/usage/packages-in-pyodide.html
-        "protobuf-5.29.2-cp312-cp312-pyodide_2024_0_wasm32.whl",
         "jedi-0.19.1-py2.py3-none-any.whl",
         "micropip-0.9.0-py3-none-any.whl",
+        "numpy-2.0.2-cp312-cp312-pyodide_2024_0_wasm32.whl",
+        "openblas-0.3.26.zip",
+        "opencv_python-4.10.0.84-cp312-cp312-pyodide_2024_0_wasm32.whl",
         "packaging-24.2-py3-none-any.whl",
+        "pandas-2.2.3-cp312-cp312-pyodide_2024_0_wasm32.whl",
         "parso-0.8.4-py2.py3-none-any.whl",
+        "protobuf-5.29.2-cp312-cp312-pyodide_2024_0_wasm32.whl",
+        "python_dateutil-2.9.0.post0-py2.py3-none-any.whl",
+        "pytz-2024.1-py2.py3-none-any.whl",
+        "scipy-1.14.1-cp312-cp312-pyodide_2024_0_wasm32.whl",
+        "six-1.16.0-py2.py3-none-any.whl",
       ]),
     ],
   };
--- a/python/foxglove-sdk/python/docs/api/index.rst
+++ b/python/foxglove-sdk/python/docs/api/index.rst
@@ -12,7 +12,7 @@ foxglove
 
 .. automodule:: foxglove
    :members:
-   :exclude-members: MCAPWriter, init_notebook_buffer
+   :exclude-members: CloudSink, CloudSinkListener, MCAPWriter, init_notebook_buffer, start_cloud_sink
 
 Notebook Integration
 ^^^^^^^^^^^^^^^^^^^^


rust/foxglove/src/websocket/ws_protocol/client/player_state.rs
```diff
@@ -0,0 +1,144 @@
+use serde::{Deserialize, Serialize};
+
+use crate::websocket::ws_protocol::JsonMessage;
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum Playing {
```
achim-k: I'd rename that enum. `Playing.Playing` doesn't look right.

```suggestion
pub enum PlaybackState {
```

rust/foxglove/src/websocket/ws_protocol/client/player_state.rs
```diff
@@ -0,0 +1,144 @@
...
+    Playing,
+    Paused,
```
achim-k: maybe also `Buffering` and `Ended` ? The server could emit buffering state after a client has made a seek request

rust/foxglove/src/websocket/ws_protocol/client/player_state.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// Playing state
+    pub playing: Playing,
+    /// Playback speed
+    pub playback_speed: f32,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    /// Seek playback time
+    pub seek_playback_time: Option<PlayerTime>,
+    /// Previous playback time (will be defined if a seek has occurred)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub seek_previous_playback_time: Option<PlayerTime>,
```
achim-k: Wouldn't it be enough to have a single `last_seek_time` field to let clients know that a seek has occurred?

rust/foxglove/src/websocket/ws_protocol/client/snapshots/foxglove__websocket__ws_protocol__client__player_state__tests__encode.snap
```diff
@@ -0,0 +1,14 @@
+---
+source: rust/foxglove/src/websocket/ws_protocol/client/player_state.rs
+assertion_line: 96
+expression: message()
+---
+{
+  "op": "playerState",
+  "playing": "playing",
+  "playbackSpeed": 1.0,
+  "currentPlaybackTime": {
```
achim-k: this field seems to be missing in the PlayerState struct definition above

ericmlujan: This snapshot is out-of-date, was renamed to `seekPlaybackTime`

rust/foxglove/src/websocket/ws_protocol/client/player_state.rs
ericmlujan: Agreed, I waffled on this one a lot and was still not happy. PlaybackState is better, though maybe the wrapping message needs to be renamed to avoid confusion between `PlaybackState` and `PlayerState`.... Naming is hard™

rust/foxglove/src/websocket/ws_protocol/client/player_state.rs
ericmlujan: I thought about this, and my bias was towards including the previous playback time and the seek time so (for example), the user could tell if a seek went backwards in time, or forwards in time, and could do any necessary computations if messages at the seek time required computations between the previous playback time and the seek time, or if state clearing were necessary. That said, if the user here is responsible for owning the clock, they could refer to the current state of their clock and the seek time and make that determination. I presume that's what you're saying here?

rust/foxglove/src/websocket/ws_protocol/client/player_state.rs
jtbandes: yeah, I was pondering suggesting changing PlayerState to PlaybackState 🤔

rust/foxglove/src/websocket/ws_protocol/client/player_state.rs
```diff
@@ -0,0 +1,144 @@
...
+    Playing,
+    Paused,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PlayerTime {
+    pub sec: u64,
+    pub nsec: u64,
+}
```
jtbandes: wonder if some consolidation is possible here – we already have:
```rs
pub struct Time {
    /// Timestamp in nanoseconds.
    pub timestamp: u64,
}
```
and
```rs
pub struct Timestamp {
    /// Seconds since epoch.
    sec: u32,
    /// Additional nanoseconds since epoch.
    nsec: u32,
}
```

rust/foxglove/src/websocket/ws_protocol/client/player_state.rs
```diff
@@ -0,0 +1,144 @@
...
+    pub playback_speed: f32,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    /// Seek playback time
+    pub seek_playback_time: Option<PlayerTime>,
+    /// Previous playback time (will be defined if a seek has occurred)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub seek_previous_playback_time: Option<PlayerTime>,
+}
+
+impl JsonMessage for PlayerState {}
```
jtbandes: I suspect we should lean towards binary for new messages for performance -- see also https://linear.app/foxglove/issue/FG-2314/potential-motivations-for-a-new-ws-protocol-subprotocol-version

rust/foxglove/src/websocket/ws_protocol/client/player_state.rs
ericmlujan: Re this: I think we should remove this field and only provide the requested seek time, providing the other field feels unnecessary

rust/foxglove/src/websocket/ws_protocol/client/player_state.rs
ericmlujan: Heard, I'm down with that. Thanks for linking that issue, it was helpful

rust/foxglove/src/websocket/ws_protocol/client/player_state.rs
ericmlujan: I'm leaning nanoseconds as u64 for consistency with timestamps in other protocol messages, as @achim-k pointed out

ericmlujan:
@achim-k @jtbandes 
Thanks for the reviews, I went ahead and made some changes:
- Changed the underlying representation to being binary
- Introduced enum states for `Buffering` and `Ended` in the `PlaybackState` enum (no longer `Playing` 😅)
- Changed the seek time to be a primitive u64 and eliminated the redundant previous seek time field (we can add it in if needed). Still making it optional and signaling with a flag as to whether a seek occurred (in which case that field should be interpreted). Notably, I've erred towards serializing the seek time even if a seek hasn't occurred, to keep the message size constant. I figured 8 bytes was pretty minimal overhead to simplify parsing logic and allow for preallocating a fixed-size memory block.

rust/foxglove/src/websocket/ws_protocol/client/player_state.rs
```diff
@@ -0,0 +1,201 @@
...
+            return Err(ParseError::BufferTooShort);
+        }
+
+        let state_byte = data.get_u8();
+        let playback_state = PlaybackState::try_from(state_byte)
+            .map_err(|_| ParseError::InvalidPlaybackState(state_byte))?;
+
+        let playback_speed = data.get_f32_le();
+        let had_seek = data.get_u8() != 0;
+        let seek_time = if had_seek && data.len() >= 8 {
```
gasmith: Since you validated the length on L53, I think you can do:

```suggestion
        let seek_time = if had_seek {
```

rust/foxglove/src/websocket/ws_protocol/client/playback_control_request.rs
```diff
@@ -0,0 +1,201 @@
...
+#[doc(hidden)]
+/// Player state message.
+#[derive(Debug, Clone, PartialEq)]
+pub struct PlayerState {
+    /// Playback state
+    pub playback_state: PlaybackState,
+    /// Playback speed
+    pub playback_speed: f32,
+    /// Seek playback time in nanoseconds (only set if a seek has been performed)
+    pub seek_time: Option<u64>,
```
achim-k, line 35: I'm wondering if we should not add start & end time to this message / struct. It would be more explicit than adding it to the server info.

achim-k: Do we want to add a field here indicating what is being played back? E.g. the recording name? Or would the server respond communicate that through server info message?

achim-k: Should we also add the current time here? It kind of says "we are now playing from time X" or "at time X we are buffering" etc.

ericmlujan: I don't think that would work, in that this is a message being sent _from_ Foxglove to the server. The server needs to communicate the start and end time of the data to Foxglove so it can hold onto it and use it to set up the play bar.

ericmlujan: > ... add a field here indicating what is being played back?

I didn't think much about the name of the recording, I could see that being useful to send through ServerInfo. We'd then need UI support to display that name at the top of the viewport

ericmlujan: >Should we also add the current time here? It kind of says "we are now playing from time X" or "at time X we are buffering" etc.

That's sort of what I was thinking about when I had introduced a `previous_seek_time` field (perhaps it was better named `current_time`). But now that we've established that it makes more sense for the server to own/drive the clock, and the player's UI controls simply requesting playback actions like pausing or seeking, I can't think of an obvious need for communicating the current time from the client back to the server. Is there an application you had in mind?

ericmlujan: Thinking more about this, I could imagine a better workflow where the callback that takes `PlayerState` also returns a `PlayerState`, which then is sent back to the app and used to synchronize time/the UI. I had originally intended `PlayerState` to go from client->server but there's a good argument for it to be bidirectional.

For example, if a user clicks the pause button, we send a request to the server to pause playback, but the app shouldn't change the UI element to a play button until the server indicates that it's paused. Similarly, if a seek request occurs, the server could decide to handle it by jumping to the nearest message, which may not be the exact time clicked on the play bar. In that case, you'd want the play head to reflect where the server actually seeked and is playing from, not where the user clicked.

rust/foxglove/src/websocket/ws_protocol/client/playback_control_request.rs
```diff
@@ -0,0 +1,201 @@
...
+/// Player state message.
+#[derive(Debug, Clone, PartialEq)]
+pub struct PlaybackControlRequest {
+    /// Playback state
+    pub playback_state: PlaybackState,
+    /// Playback speed
+    pub playback_speed: f32,
+    /// Seek playback time in nanoseconds (only set if a seek has been performed)
+    pub seek_time: Option<u64>,
+}
```
gasmith, line 39: Do we need a `loop: bool`? Or would this be implemented by a seek(0)? If the latter, how does the client know when we've reached the end of the recording?

rust/foxglove/src/websocket/ws_protocol/client/playback_control_request.rs
```diff
@@ -0,0 +1,201 @@
...
+
+#[cfg(feature = "unstable")]
+#[doc(hidden)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u8)]
+pub enum PlaybackState {
+    Playing = 0,
+    Paused = 1,
+    Buffering = 2,
+    Ended = 3,
```
gasmith: How would these be used? This seems more like the sort of status a server would send a client.

rust/foxglove/src/websocket/ws_protocol/client/playback_control_request.rs
```diff
@@ -0,0 +1,201 @@
...
+#[cfg(feature = "unstable")]
+use super::BinaryOpcode;
+
+#[cfg(feature = "unstable")]
+#[doc(hidden)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u8)]
+pub enum PlaybackState {
+    Playing = 0,
+    Paused = 1,
```
gasmith: Should this be phrased as a command? `Pause`, `Play`, etc.

rust/foxglove/src/websocket/ws_protocol/client/playback_control_request.rs
gasmith: I agree that we need a response from the server, to indicate that it has (or hasn't?) handled the request. In addition to the reason you describe, there's also the matter of this being a full-duplex channel: the client will be receiving messages concurrently while it waits for the server to handle the playback control request, and the client (presumably) needs to identify the cut-off between old messages (before the server handled the playback control request) and new messages.

Especially since the request handling is being written by a third party, we also need to guard against a faulty server implementation that fails to reply, or fails to do so in a timely manner. We should consider debuggability here, because the implementer is going to have to troubleshoot why something isn't working as expected. We put some safeguards in place for service calls (like the [`Responder` drop handler](https://github.com/foxglove/foxglove-sdk/blob/72a311a9abd475b3251796a5c6fc3afe319f8cef/rust/foxglove/src/websocket/service/response.rs#L81-L86)), and we might want to considering doing similar here.

Which leads us to a question: Instead of making a new primitive ws_protocol message type, could this ride on top of the ServiceCallRequest/Response/Failure messages? Perhaps we introduce a new well-known service name (`/foxglove/playback_control`?) and schema?

rust/foxglove/src/websocket/ws_protocol/client/playback_control_request.rs
```diff
@@ -0,0 +1,201 @@
...
+
+#[cfg(feature = "unstable")]
+#[doc(hidden)]
+/// Player state message.
+#[derive(Debug, Clone, PartialEq)]
+pub struct PlaybackControlRequest {
+    /// Playback state
+    pub playback_state: PlaybackState,
+    /// Playback speed
+    pub playback_speed: f32,
```
gasmith, line 33: There's certainly an upper bound on playback speed. There may also be situations where the simulation is incapable at running at non-realtime speeds (fast or slow). I wonder whether a server needs to be able to communicate these limitations to the client somehow, so that we can adjust the range of UI speed controls accordingly.

rust/foxglove/src/websocket/ws_protocol/client/playback_control_request.rs
achim-k: Yeah it doesn't make sense here. I was incorrectly assuming that these messages are sent from the server to the client :man_facepalming:

rust/foxglove/src/websocket/ws_protocol/client/playback_control_request.rs
ericmlujan: I think there will need to be a (separate, to be implemented) playback message from the server to the client that broadcasts that it's ended, or saying what state playback is in in regards to a PlaybackControlRequest, for example. I think the client should handle requesting a loop as a seek, once the server indicates that playback is over.

rust/foxglove/src/websocket/ws_protocol/client/playback_control_request.rs
ericmlujan: Yeah, I like that better

rust/foxglove/src/websocket/ws_protocol/client/playback_control_request.rs
ericmlujan: That's an interesting point, and you're right. I'd also argue that this message represents a _request_ for a playback speed, and the playback speeds that humans will be comfortable interacting with are also probably bounded. For something like a local simulation, the highest achievable playback speed may well be variable throughout a recording, and knowing that a priori can difficult. 

My preference is for Foxglove to present a reasonable default range of _requestable_ playback speeds (ranging from maybe 0.1x to 100x with reasonable increments between), and adjust the UI based on user feedback. It's possible (even likely!) that you can request 5x speed, and the server can only provide playback at 0.3x, and time advances more slowly than you're requesting. I think most people using this feature would understand that the limitation is on the server/user end

gasmith: > My preference is for Foxglove to present a reasonable default range of _requestable_ playback speeds (ranging from maybe 0.1x to 100x with reasonable increments between), and adjust the UI based on user feedback.

Fair. Surfacing some notion of inferred playback speed in the UI would help the user to understand when the server can't keep up with the speed they're requesting. This might even make sense across the board (e.g., when playing back regular ol' recordings from the data platform).

> I think most people using this feature would understand that the limitation is on the server/user end

I agree, with one exception. There may be server implementations that only implement realtime playback, and in that situation, it'd be preferable to completely disable speed controls. Otherwise, a user would likely be confused if they request a slower-than-realtime playback speed, and the server serves up realtime anyway.

ericmlujan: >Surfacing some notion of inferred playback speed in the UI would help the user to understand when the server can't keep up with the speed they're requesting. This might even make sense across the board (e.g., when playing back regular ol' recordings from the data platform).

I think this is a good idea regardless. Perhaps we could take the logic we use for displaying the topic Hz and repurpose it across all messages (I noticed when publishing a 1Hz topic in my local spike this also tracked the playback speed nicely with no explicit feedback).

ericmlujan:
@gasmith Thanks again for the feedback and earlier in-person discussion on this. I added a new `PlaybackState` message and did some renaming some fields in the `PlaybackControlRequest` message, in addition to adding a request_id string field. Let me know what you think

rust/foxglove/src/websocket/ws_protocol/client/playback_control_request.rs
```diff
@@ -0,0 +1,201 @@
...
+        };
+        insta::assert_snapshot!(format!("{:#04x?}", message.to_bytes()));
+    }
+
+    #[test]
+    fn test_encode_pause() {
+        let message = PlaybackControlRequest {
+            playback_command: PlaybackCommand::Pause,
+            playback_speed: 1.0,
+            seek_time: None,
```
gasmith, line 131: Nit: Maybe provide Some seek time here, so we can snapshot that?

rust/foxglove/src/websocket/ws_protocol/client/playback_control_request.rs
```diff
@@ -0,0 +1,201 @@
...
+    // Message layout:
+    //   opcode (1 byte)
+    // + playback_command (1 byte)
+    // + playback_speed (4 bytes)
+    // + had_seek (1 byte)
+    // + seek_time (8 bytes)
+    // + request_id_len (4 bytes)
+    // + request_id
+    fn parse_binary(mut data: &'a [u8]) -> Result<Self, ParseError> {
+        if data.len() < 1 + 4 + 1 + 8 {
```
gasmith: Tack on 4 more for `request_id_len`

```suggestion
        if data.len() < 1 + 4 + 1 + 8 + 4 {
```

rust/foxglove/src/websocket/ws_protocol/client/playback_control_request.rs
```diff
@@ -0,0 +1,201 @@
...
+/// A request to control playback from the client
+#[derive(Debug, Clone, PartialEq)]
+pub struct PlaybackControlRequest {
+    /// Playback state
+    pub playback_command: PlaybackCommand,
+    /// Playback speed
+    pub playback_speed: f32,
+    /// Seek playback time in nanoseconds (only set if a seek has been performed)
+    pub seek_time: Option<u64>,
+    /// Unique string identifier, used to indicate that a PlaybackState is in response to a particular request from the client
```
gasmith: We should probably indicate that this should not be an empty string, since that's indistinguishable from None in the `PlaybackState` message.

gasmith (approved):
LG, with some nits.

rust/foxglove/src/websocket/ws_protocol/client/playback_control_request.rs
ericmlujan: Resolved in [92baafa](https://github.com/foxglove/foxglove-sdk/pull/728/commits/92baafa5c8c81ef81309188cea2895bf368e5228)

rust/foxglove/src/websocket/ws_protocol/client/playback_control_request.rs
ericmlujan: Resolved in 88ef06d

rust/foxglove/src/websocket/ws_protocol/client/playback_control_request.rs
ericmlujan: Resolved in 88ef06d

PR #730, opened 2025-10-29
python: remove stray lockfiles

sofuture:
### Changelog
None

### Docs
None

### Description

Looks like some `poetry.lock` files were mistakenly added in #706 , remove them.


jtbandes:
I've seen some of these in my working tree too. Should we gitignore them? Or another question – why do we not want to check in these lockfiles? (since we check in most lockfiles)

sofuture:
@jtbandes Two of the three I deleted were for examples that don't otherwise exist. For the third I deleted and the other examples (without checked in lockfiles) -- those are great questions that I don't know the answer to.

gasmith:
These probably got picked up because of #705; I removed `poetry.lock` from .gitignore, and replaced it with `uv.lock`.

> why do we not want to check in these lockfiles? (since we check in most lockfiles)

I guess because it seemed like bloat? We don't really benefit from the overhead of maintaining pinned dependencies for examples. I'm fine running against the latest available version, and in the unlikely event that something breaks, we'll go fix it.

PR #734, opened 2025-10-30
Fix clippy error for implicit string clone

ericmlujan:
### Changelog
None

### Docs
None

### Description
Looks like this is breaking in CI due to a toolchain upgrade, flagged by @jtbandes. Quick fix to resolve the clippy error.


ericmlujan:
Race condition 😁, closing in favor of #733

PR #735, opened 2025-10-30
Create WebSocket server handler for PlaybackControlRequest

ericmlujan:
### Changelog
None

### Docs
None

### Description
Add in Rust SDK functionality for the messages introduced in #728; namely:
- Introduce the `RangedPlayback` capability, which indicates that a server connected over WebSocket is providing data from a fixed playback time range.
- Introduce the `playback_time_range` option in `ServerOptions`, where the server is expected to declare the time range of the data it is playing back. This option will implies the `RangedPlayback` capability and will enable it if it is not already enabled
- Add the `on_playback_control_request` function to the `ServerListener` trait, which allows a server to receive playback control requests from the client. It requires the server to send a `PlaybackState` message in response.
- Add the `broadcast_playback_state` function, which allows a server to broadcast a change in its playback state in response to anything other than a `PlaybackControlRequest` message (for example, reaching the end of playback or an external process pausing playback).

These changes are currently guarded behind the `unstable` FF and excluded from documentation.

### DIFF
--- a/rust/foxglove/src/websocket/ws_protocol/server/server_info.rs
+++ b/rust/foxglove/src/websocket/ws_protocol/server/server_info.rs
@@ -132,6 +132,7 @@ pub enum Capability {
     Assets,
     /// Indicates that the server is sending data within a fixed time range. This requires the
     /// server to specify the `data_start_time` and `data_end_time` fields in its `ServerInfo` message.
+    /// Playback control requests are only accepted when this capability is enabled.
     RangedPlayback,
 }
 


rust/foxglove/src/websocket/server.rs
```diff
@@ -164,6 +167,9 @@ pub(crate) struct Server {
     /// Information about the server, which is shared with clients.
     /// Keys prefixed with "fg-" are reserved for internal use.
     server_info: HashMap<String, String>,
+    /// Time range of data being played back, in absolute nanoseconds.
+    /// Implies the [`RangedPlayback`](crate::websocket::Capability::Time) capability if set.
```
gasmith:
```suggestion
    /// Implies the [`RangedPlayback`](crate::websocket::Capability::RangedPlayback) capability if set.
```

rust/foxglove/src/websocket_server.rs
```diff
@@ -124,6 +130,18 @@ impl WebSocketServer {
...
 
+    /// Declare the time range for playback, in absolute nanoseconds. This applies if the server is playing back a fixed time range of data.
+    /// This will add the RangedPlayback capability to the server.
+    pub fn playback_time_range(mut self, start_time: u64, end_time: u64) -> Self {
+        self.options.playback_time_range = Some((start_time, end_time));
+        if let Some(capabilities) = self.options.capabilities.as_mut() {
+            capabilities.insert(Capability::RangedPlayback);
+        } else {
+            self.options.capabilities = Some(HashSet::from([Capability::RangedPlayback]));
+        }
```
gasmith: Let's not put this logic in three places. My gut feels that it belongs in `Server::new()`, which is where we handle the enforced advertisement of the `Services` and `Assets` capabilities.

rust/foxglove/src/websocket/ws_protocol/server/server_info.rs
```diff
@@ -69,6 +77,19 @@ impl ServerInfo {
...
+
+    #[must_use]
+    pub fn with_playback_time_range(mut self, time_range: Option<(u64, u64)>) -> Self {
+        if let Some((start_time, end_time)) = time_range {
+            self.data_start_time = Some(start_time);
+            self.data_end_time = Some(end_time);
+
+            if !self.capabilities.contains(&Capability::RangedPlayback) {
+                self.capabilities.push(Capability::RangedPlayback);
+            }
```
gasmith: I don't think this business logic belongs here. Either that, or we need to update `.with_capabilities()` to account for the fact that the methods could be called in either order. See also my comment in `websocket_server.rs`: I'd prefer we do this in a single place.

rust/foxglove/src/websocket/server.rs
```diff
@@ -196,6 +202,10 @@ impl Server {
             );
         }
 
+        if opts.playback_time_range.is_some() {
+            capabilities.insert(Capability::RangedPlayback);
+        }
```
gasmith, line 211: I think we also need to validate that the user doesn't advertise support for `RangedPlayback` without providing the time range.

Maybe we shouldn't even expose the `RangedPlayback` variant in `websocket::Capability` (while keeping it as a low-level protocol detail in `ws_protocol::server::server_info::Capability`, of course). That way there's no possibility of misconfiguration. If we go this route, maybe also rename `WebSocketServer::playback_time_range` to be something a little more generic, in case we decide to offer other parameters in the future... Maybe just `WebSocketServer::playback`?

The downside of going this direction is that it might hamper discoverability, somewhat. As a developer, I'd probably look at the `Capability` enum to get a sense for what kinds of optional functionality a server implementation might support.

Anyway, food for thought.

rust/foxglove/src/websocket/server_listener.rs
```diff
@@ -61,4 +65,11 @@ pub trait ServerListener: Send + Sync {
     /// Callback invoked when the last client unsubscribes from the connection graph. Requires
     /// [`Capability::ConnectionGraph`][super::Capability::ConnectionGraph].
     fn on_connection_graph_unsubscribe(&self) {}
+    #[cfg(feature = "unstable")]
+    #[doc(hidden)]
+    /// Callback invoked when a client sends a player state message.
+    /// Requires [`Capability::RangedPlayback`][super::Capability::RangedPlayback].
+    fn on_playback_control_request(&self, _request: PlaybackControlRequest) -> PlaybackState {
+        Default::default()
```
gasmith: I think it might be better to return `Option<PlaybackState>`, so that we can identify a buggy server implementation that advertises `RangedPlayback`, but forgets to implement this callback. Sending the "default" value is almost certainly not the right thing to do.

rust/foxglove/src/websocket/ws_protocol/server/playback_state.rs
```diff
@@ -115,6 +115,17 @@ impl<'a> BinaryMessage<'a> for PlaybackState {
     }
 }
 
+impl Default for PlaybackState {
```
gasmith: I'm not sure this should have a default implementation, tbh.

rust/foxglove/src/websocket/server_listener.rs
```diff
@@ -61,4 +65,11 @@ pub trait ServerListener: Send + Sync {
     /// Callback invoked when the last client unsubscribes from the connection graph. Requires
     /// [`Capability::ConnectionGraph`][super::Capability::ConnectionGraph].
     fn on_connection_graph_unsubscribe(&self) {}
+    #[cfg(feature = "unstable")]
+    #[doc(hidden)]
+    /// Callback invoked when a client sends a player state message.
+    /// Requires [`Capability::RangedPlayback`][super::Capability::RangedPlayback].
+    fn on_playback_control_request(&self, _request: PlaybackControlRequest) -> PlaybackState {
```
gasmith: Do we want this to synchronously return the new PlaybackState? I suppose if it isn't possible to handle the request synchronously, the implementation can always return `Status::Buffering`, but I wonder whether we're painting developers into a bit of a corner.

An alternative to consider, would be something like [the `Responder` object we pass for service calls](https://github.com/foxglove/foxglove-sdk/blob/c2fba752d16ce9ca6520ccb6fde31d890b3f66e4/rust/foxglove/src/websocket/service/handler.rs#L9-L19), so that the implementation can choose how they want to handle the request. But maybe that complexity isn't justified here.

rust/foxglove/src/websocket/connected_client.rs
```diff
@@ -673,6 +677,19 @@ impl ConnectedClient {
...
+    #[cfg(feature = "unstable")]
+    fn on_playback_control_request(&self, server: Arc<Server>, msg: PlaybackControlRequest) {
+        if !server.has_capability(Capability::RangedPlayback) {
+            self.send_error("Server does not support ranged playback capability".to_string());
+            return;
+        }
+
+        if let Some(handler) = server.listener() {
+            let playback_state = handler.on_playback_control_request(msg);
+            self.send_control_msg(&playback_state);
```
gasmith: If we're taking care of sending the response, should we also take care of populating `request_id`? If that's the developer's responsibility, we should call that out.

rust/foxglove/src/websocket/ws_protocol/server/playback_state.rs
ericmlujan: Yeah, I went back and forth on this one. I didn't love having the `on_playback_control_request` return an Option, since it didn't feel like something we wanted to make optional. But having a default playback state also doesn't make a lot of sense, since it's so context dependent and there's not a universal trivial definition of it

rust/foxglove/src/websocket/ws_protocol/server/server_info.rs
ericmlujan: Agreed, I played whac-a-mole with this in a few places when writing tests, and now that everything's green, it's worth consolidating

rust/foxglove/src/websocket/connected_client.rs
ericmlujan: That's an interesting point. I'm personally fine making that the developer's responsibility, in principle, for the sake of being explicit and not introducing a code path that populates it in a nonobvious way

rust/foxglove/src/websocket/server.rs
ericmlujan: I think enforcement is a good idea, especially since this can be checked at construction time. I do think exposing it is nice for discoverability's sake, as you mentioned. And of course, this all should be validated as well on the app side, in case there is some implementation out there that doesn't conform.

rust/foxglove/src/websocket/ws_protocol/server/server_info.rs
```diff
@@ -26,6 +26,12 @@ pub struct ServerInfo {
     /// Optional string.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub session_id: Option<String>,
+    /// Optional timestamp, in absolute nanoseconds, indicating the start of the data range.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub data_start_time: Option<u64>,
```
achim-k: u64 is problematic for parsing the JSON in javascript, because javascripts number type can't hold integers higher than 2^53 - 1 without loosing precision.

`{ sec: u32, nsec: u32 }` would work, but I wonder if we should not put the start and end time in the playback state struct?

ericmlujan: We could do that too, but other binary messages in the protocol (namely, `Time` and `MessageData` use u64 nanosecond fields for their timestamps). Wouldn't that also be a problem there?

achim-k: It's not a problem for binary messages as those don't have to be JSON deserialized

rust/foxglove/src/websocket/connected_client.rs
achim-k: The updated playback state should be sent to all clients
```suggestion
            server.broadcast_playback_state(playback_state);
```

rust/foxglove/src/websocket_server.rs
ericmlujan: Agreed, added a validation check and panic to `Server::new()`

rust/foxglove/src/websocket/ws_protocol/server/server_info.rs
ericmlujan: Consolidated to `Server::new()`

rust/foxglove/src/websocket/server_listener.rs
ericmlujan: Ended up going this way, and emitting an error if your callback returns None. It's a little nonobvious from the developer's perspective ("why is it an option, but I can't return None?!", but it was the least bad interface)

rust/foxglove/src/websocket/ws_protocol/server/server_info.rs
ericmlujan: Right, this is for ServerInfo. I added in a SerializedTimestamp struct to `server_info.rs` which now does that split

rust/foxglove/src/websocket/connected_client.rs
ericmlujan: Resolved

rust/foxglove/src/websocket/server_listener.rs
ericmlujan: My thought is that this can be accomplished with `Buffering`, and since we provide a `broadcast_playback_state()` function, they're free to update the state basically at any time (for example, at the end of an expensive seek or buffering operation)

rust/foxglove/src/websocket/ws_protocol/server/server_info.rs
```diff
@@ -6,6 +6,25 @@ use serde::{Deserialize, Serialize};
...
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct SerializedTimestamp {
+    pub sec: u32,
+    pub nsec: u32,
+}
+
+impl SerializedTimestamp {
+    fn from_nsecs(timestamp: u64) -> Self {
+        SerializedTimestamp {
+            sec: (timestamp / 1e9 as u64) as u32,
```
gasmith: There's a chance of overflow here, because `u64::MAX/1e9` requires 34 bits. It'd be better to saturate instead truncating:


```suggestion
            sec: u32::try_from(timestamp / 1e9 as u64).unwrap_or(u32::MAX),
```

rust/foxglove/src/websocket/ws_protocol/server/server_info.rs
```diff
@@ -69,6 +96,15 @@ impl ServerInfo {
         self.session_id = Some(session_id.into());
         self
     }
+
+    #[must_use]
+    pub fn with_playback_time_range(mut self, time_range: Option<(u64, u64)>) -> Self {
+        if let Some((start_time, end_time)) = time_range {
+            self.data_start_time = Some(SerializedTimestamp::from_nsecs(start_time));
+            self.data_end_time = Some(SerializedTimestamp::from_nsecs(end_time));
+        }
```
gasmith, line 108: Nit: Handle the else case:
```suggestion
        if let Some((start_time, end_time)) = time_range {
            self.data_start_time = Some(SerializedTimestamp::from_nsecs(start_time));
            self.data_end_time = Some(SerializedTimestamp::from_nsecs(end_time));
        } else {
            self.data_start_time = None;
            self.data_end_time = None;
        }
```

rust/foxglove/src/websocket/connected_client.rs
```diff
@@ -673,6 +677,41 @@ impl ConnectedClient {
...
     }
 
+    #[cfg(feature = "unstable")]
+    fn on_playback_control_request(&self, server: Arc<Server>, msg: PlaybackControlRequest) {
+        if !server.has_capability(Capability::RangedPlayback) {
+            self.send_error("Server does not support ranged playback capability".to_string());
+            return;
+        }
+
+        if let Some(handler) = server.listener() {
```
gasmith, line 687: Should we also send back an error when the server doesn't have a listener configured?

rust/foxglove/src/websocket_server.rs
```diff
@@ -113,7 +115,11 @@ impl WebSocketServer {
     ///
     /// By default, the server does not advertise any capabilities.
     pub fn capabilities(mut self, capabilities: impl IntoIterator<Item = Capability>) -> Self {
-        self.options.capabilities = Some(capabilities.into_iter().collect());
+        if let Some(capabilities_inner) = self.options.capabilities.as_mut() {
+            capabilities_inner.extend(capabilities);
+        } else {
+            self.options.capabilities = Some(capabilities.into_iter().collect());
+        }
```
gasmith: I think we can revert this hunk, now that we're only injecting the `RangedPlayback` capability during `Server::new`.

gasmith (approved):
LG, with some nits.

rust/foxglove/src/websocket/connected_client.rs
```diff
@@ -673,6 +677,41 @@ impl ConnectedClient {
...
+                tracing::error!(
+                        "Request id in playback state returned in response to playback control request id {} didn't match",
+                        request_id,
+                    );
+                self.send_error(
+                    "Server sent non-matching request id in response to playback control request"
+                        .to_string(),
+                );
+                return;
+            }
```
gasmith: If we're going to be this pedantic, we might as well just forcibly set it.

ericmlujan: Fair enough, given that there's no real option

rust/foxglove/src/websocket/ws_protocol/server/server_info.rs
```diff
@@ -91,6 +130,9 @@ pub enum Capability {
     ConnectionGraph,
     /// Allow clients to fetch assets.
     Assets,
+    /// Indicates that the server is sending data within a fixed time range. This requires the
+    /// server to specify the `data_start_time` and `data_end_time` fields in its `ServerInfo` message.
+    RangedPlayback,
```
achim-k, line 136: Probably should also mention here that playback control requests are only accepted when the server advertises this capability

achim-k: Should the server be required to sent time messages when this capability is advertised?

achim-k (approved):

PR #736, opened 2025-10-31
playground: include current layout when constructing share url

jtbandes:
### Changelog
None

### Docs

None

### Description

Now clicking "share" will create a url that contains the latest layout, including any user modifications 🎉


defunctzombie:
what a time to be alive

Issue #729, opened 2025-10-29
Windows wheels only distributed for Python 3.9


SimenZhor:
**Description**
I was about to write a bug report that there are no wheels distributed for Python 3.14, but I had a [closer look](https://pypi.org/project/foxglove-sdk/#files) and it seems the only Python version that has distributed wheels is Python 3.9 which has recently reached [end-of-life](https://devguide.python.org/versions/).

- Version: 0.15.2 (probably more)
- Platform: Windows x64

Quick test with 4 python versions:
<img width="1956" height="1023" alt="Image" src="https://github.com/user-attachments/assets/b2068fb2-1384-438e-be5b-a93150ec7b5f" />

Screenshot from Pypi:
<img width="808" height="423" alt="Image" src="https://github.com/user-attachments/assets/2c62f308-5edb-4f57-a614-74b008430131" />

**Steps To Reproduce**
1. Make sure no Rust toolchain is available on PATH (I used a Windows sandbox in the screenshot above)
2. Install Python 3.14
3. Install foxglove-sdk

```
pip install foxglove-sdk
Collecting foxglove-sdk
  Downloading foxglove_sdk-0.15.2.tar.gz (343 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [31 lines of output]
      Python reports SOABI: cp314-win_amd64
      Computed rustc target triple: x86_64-pc-windows-msvc
      Installation directory: C:\Users\WDAGUtilityAccount\AppData\Local\puccinialin\puccinialin\Cache
      Downloading rustup-init from https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-msvc/rustup-init.exe
      Checking for Rust toolchain....
      Rust not found, installing into a temporary directory

      Downloading rustup-init:   0%|          | 0.00/13.6M [00:00<?, ?B/s]
      Downloading rustup-init:  27%|##6       | 3.60M/13.6M [00:00<00:00, 35.3MB/s]
      Downloading rustup-init:  58%|#####7    | 7.86M/13.6M [00:00<00:00, 39.5MB/s]
      Downloading rustup-init:  87%|########7 | 11.8M/13.6M [00:00<00:00, 39.2MB/s]
[...]

PR #738, opened 2025-11-03
python: Fix windows wheel builds

gasmith:
In #690, we worked around a CI issue, but we went about it the wrong way. By turning off `maturin-action`'s `--find-interpreter`, we ended up building only a single wheel against python 3.9 for windows x64.

The root of the original problem that #690 was trying to tackle, is that github's `windows-latest` runner doesn't have _any_ 32-bit pythons installed. This change uses `setup-python` to install the versions we care about, and restores the use of `maturin-action`'s `--find-interpreter` so that we build against them.

```
Archive:  wheels-windows-x86_64.zip
    testing: foxglove_sdk-0.15.2-cp310-cp310-win_amd64.whl   OK
    testing: foxglove_sdk-0.15.2-cp312-cp312-win_amd64.whl   OK
    testing: foxglove_sdk-0.15.2-cp311-cp311-win_amd64.whl   OK
    testing: foxglove_sdk-0.15.2-cp313-cp313-win_amd64.whl   OK
    testing: foxglove_sdk-0.15.2-cp314-cp314-win_amd64.whl   OK
    testing: foxglove_sdk-0.15.2-cp39-cp39-win_amd64.whl   OK

Archive:  wheels-windows-x86.zip
    testing: foxglove_sdk-0.15.2-cp310-cp310-win32.whl   OK
    testing: foxglove_sdk-0.15.2-cp311-cp311-win32.whl   OK
    testing: foxglove_sdk-0.15.2-cp312-cp312-win32.whl   OK
    testing: foxglove_sdk-0.15.2-cp313-cp313-win32.whl   OK
    testing: foxglove_sdk-0.15.2-cp314-cp314-win32.whl   OK
    testing: foxglove_sdk-0.15.2-cp39-cp39-win32.whl   OK
```

Fixes: FG-13419
Fixes: #729

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -245,16 +245,16 @@ jobs:
           - runner: windows-latest
             target: x86_64
           - runner: windows-latest
-            target: i686
+            target: x86
     steps:
       - uses: actions/setup-python@v6
         if: ${{ matrix.platform.target == 'x86_64' }}
         with:
           python-version: "3.9"
 
-      # The 64-bit windows-latest image doesn't have 32-bit pythons.
+      # The windows-latest image doesn't have 32-bit pythons.
       - uses: actions/setup-python@v6
-        if: ${{ matrix.platform.target == 'i686' }}
+        if: ${{ matrix.platform.target == 'x86' }}
         with:
           architecture: x86
           python-version: |


jtbandes (approved):
So are we just getting lucky that the other runners have all the pythons we want to build for? Wondering if we should be explicit about which pythons to install for all the platforms?

gasmith:
> So are we just getting lucky that the other runners have all the pythons we want to build for? Wondering if we should be explicit about which pythons to install for all the platforms?

I don't think it's luck, per se. I think it's reasonable to assume that runners have pythons available for cross-compilation, dating back to 3.9 or earlier. I'm guessing they removed 32-bit pythons from the windows image because windows itself is no longer officially supported on 32-bit platforms. Come to think of it, we might consider dropping support for win32 targets too, for the same reason.

PR #740, opened 2025-11-03
playground: adjust height of toolbar to match app bar

defunctzombie:
### Changelog
None

### Docs
None

### Description

Update the toolbar height to match the embed app bar height. Creates a visually consistent bar across the top between the two halves.

Screenshot 2025-11-03 at 10 53 40 AM

</td><td>

Screenshot 2025-11-03 at 10 57 42 AM

</td></tr></table>

### DIFF
--- a/playground/src/Playground.tsx
+++ b/playground/src/Playground.tsx
@@ -22,7 +22,9 @@ const useStyles = tss.create(({ theme }) => ({
   topBar: {
     flex: "0 0 auto",
     display: "flex",
-    padding: "8px 8px 8px 16px",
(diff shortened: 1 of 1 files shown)


defunctzombie:
> My only sadness now is the padding on the top/bottom doesn't match the padding on the right

It didn't before did it? At least in css it was 8px/16px. I would not say I had the same sadness. Might even say these buttons could be full-bar height.

jtbandes:
> It didn't before did it?

I think it did, since it was specified as `8px 8px 8px 16px` with no fixed height

PR #741, opened 2025-11-03
ci: Don't deploy playground for dependabot PRs

gasmith:
Dependabot can't deploy the playground, because it doesn't have access to secrets. So don't deploy the playground for dependabot PRs.


jtbandes (approved):
Thanks, I was just noticing we need to do this as well. We do the same in other repos/workflows.

PR #743, opened 2025-11-04
ci: git rid of deprecated macos runner

sofuture:
### Changelog
None
### Docs
None

### Description
Remove deprecated `macos-13` runner. Replace with `macos-15-intel`.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -293,10 +293,10 @@ jobs:
     strategy:
       matrix:
         platform:
-          - runner: macos-13
-            target: x86_64
           - runner: macos-14
             target: aarch64
+          - runner: macos-15-intel
+            target: x86_64
     steps:
       - uses: actions/setup-python@v6
         with:


jtbandes (approved):
I wonder if cross-compiling would work fine for macos and we wouldn't need a matrix here? anyway - I suppose this is a fine incremental change

.github/workflows/python.yml
```diff
@@ -293,10 +293,10 @@ jobs:
     strategy:
       matrix:
         platform:
-          - runner: macos-13
-            target: x86_64
           - runner: macos-14
```
gasmith, line 296: Might as well update this as well?

.github/workflows/python.yml
```diff
@@ -293,10 +293,10 @@ jobs:
...
             target: aarch64
+          - runner: macos-15-intel
```
gasmith, line 298: Looks like cross-compile also works fine, fwiw: https://github.com/foxglove/foxglove-sdk/actions/runs/19081439997/job/54510933922?pr=744

sofuture:
Ope, the timing there didn't work out.

PR #744, opened 2025-11-04
python: Use macos-15 runners

gasmith:


jtbandes:
#743

jtbandes:
I guess this is an alternative where we try cross-compiling?

gasmith:
lost the race

PR #746, opened 2025-11-05
C/C++ bindings for PlaybackControlRequest listener

ericmlujan:
### Changelog
None

### Docs
Documentation comments are added, but hidden from public-facing docs until we've had more time to iron out issues from integration testing + add equivalent python bindings

### Description
Add C FFI bindings for the `PlaybackControlRequest` server listener and  `broadcast_playback_state` functions, the `PlaybackControlRequest` and `PlaybackState` structs, and wrappers/equivalent classes around these bindings to expose the functionality in the  C++ API.

Unit tests added to the C++ side.

### DIFF
--- a/cpp/foxglove/tests/test_server.cpp
+++ b/cpp/foxglove/tests/test_server.cpp
@@ -1519,9 +1519,7 @@ std::vector<std::byte> playbackControlRequestToBinary(
   writeIntLE(
     msg, playback_control_request.seek_time.has_value() ? *playback_control_request.seek_time : 0x0
   );
-  auto request_id_size = std::min<uint32_t>(
-    playback_control_request.request_id.size(), (std::numeric_limits<uint32_t>::max)()
-  );
+  auto request_id_size = static_cast<uint32_t>(playback_control_request.request_id.size());
   writeIntLE(msg, request_id_size);
   for (char c : playback_control_request.request_id) {
     msg.emplace_back(std::byte{static_cast<std::byte>(c)});


c/src/lib.rs
```diff
@@ -193,9 +196,10 @@ impl Clone for FoxgloveStringBuf {
 impl Drop for FoxgloveStringBuf {
     fn drop(&mut self) {
         let FoxgloveString { data, len } = self.0;
-        assert!(!data.is_null());
-        // SAFETY: This was constructed from valid `String`.
-        drop(unsafe { String::from_raw_parts(data as *mut u8, len, len) })
+        if !data.is_null() {
```
gasmith: How would such a `FoxgloveStringBuf` be created?

c/src/playback_state.rs
```diff
@@ -0,0 +1,43 @@
...
+    pub request_id: FoxgloveString,
+}
+
+impl FoxglovePlaybackState {
+    /// Create a native playback state from a FoxglovePlaybackState.
+    ///
+    /// # Safety
+    /// - `self.request_id` must wrap a valid UTF-8 string that lives for the duration of the call.
+    ///   This function copies request_id into a separately allocated Rust string in the output native PlaybackState.
+    pub(crate) fn as_native(
```
gasmith: If these safety guarantees are the responsibility of the caller, then this should be an `unsafe fn`. If one or more safety guarantees are met by construction (which does not seem to be the case here), then you put a `/// SAFETY` comment to explain why an `unsafe { ... }` block is actually safe.

https://std-dev-guide.rust-lang.org/policy/safety-comments.html

c/src/playback_state.rs
```diff
@@ -0,0 +1,43 @@
...
+        &self,
+    ) -> Result<foxglove::websocket::PlaybackState, foxglove::FoxgloveError> {
+        let status = foxglove::websocket::PlaybackStatus::try_from(self.status).map_err(|e| {
+            foxglove::FoxgloveError::ValueError(format!("invalid playback status {e}"))
+        })?;
+
+        let request_id = if self.request_id.is_empty() {
```
gasmith, line 30: NYC, but I'm noticing that `FoxgloveString::is_empty()` and `FoxgloveString::as_utf8_str()` are not _quite_ consistent with each other. Seems like we might want to treat a `FoxgloveString` as empty whenever either `data` is null or `len` is 0.

c/src/server.rs
```diff
@@ -38,6 +40,9 @@ pub const FOXGLOVE_SERVER_CAPABILITY_SERVICES: u8 = 1 << 4;
 /// Allow clients to request assets. If you supply an asset handler to the server, this capability
 /// will be advertised automatically.
 pub const FOXGLOVE_SERVER_CAPABILITY_ASSETS: u8 = 1 << 5;
+/// Indicates that the server is sending data within a fixed time range. This requires the
+/// server to specify the `data_start_time` and `data_end_time` fields in its `ServerInfo` message.
```
gasmith:
```suggestion
/// server to specify the `data_start_time` and `data_end_time` fields in `foxglove_server_options`.
```

c/src/server.rs
```diff
@@ -159,6 +168,13 @@ pub struct FoxgloveServerOptions<'a> {
             channel: *const FoxgloveChannelDescriptor,
         ) -> bool,
     >,
+
+    /// If the server is sending data from a fixed time range, and has the RangedPlayback capability,
+    /// the start time of the data range.
+    pub data_start_time: Option<&'a u64>,
+    /// If the server is sending data from a fixed time range, and has the RangedPlayback capability,
+    /// the end time of the data range.
+    pub data_end_time: Option<&'a u64>,
```
gasmith: Should we name these to more closely reflect the `playback_time_range` in the rust API? `playback_start_time` & `playback_end_time`, perhaps?

c/src/server.rs
```diff
@@ -293,6 +309,13 @@ pub struct FoxgloveServerCallbacks {
     >,
     pub on_connection_graph_subscribe: Option<unsafe extern "C" fn(context: *const c_void)>,
     pub on_connection_graph_unsubscribe: Option<unsafe extern "C" fn(context: *const c_void)>,
+    pub on_playback_control_request: Option<
```
gasmith, line 322: Deserves a docstring.

c/src/server.rs
```diff
@@ -464,6 +494,37 @@ pub extern "C" fn foxglove_server_broadcast_time(
...
+/// Publishes the current playback state to all clients.
+///
+/// Requires the `FOXGLOVE_CAPABILITY_RANGED_PLAYBACK` capability.
+///
+/// # Safety
+/// - `playback_state` must be a valid pointer to a playback state that lives for the duration of the call.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn foxglove_server_broadcast_playback_state(
+    server: Option<&FoxgloveWebSocketServer>,
+    playback_state: *const FoxglovePlaybackState,
```
gasmith: My cbindgen memory is a bit dusty, but I think this could be `Option<&FoxglovePlaybackState>`, and do the same `let Some(playback_state) = ...` that we do for the `server` pointer.

c/src/playback_state.rs
gasmith: Nit: Might be slightly more accurate to name this `to_native()`, since we're taking copies instead of borrowing.

https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

gasmith (commented):
Got through the C part, will look through C++ later.

c/src/lib.rs
eloff: It came up when setting it from the C side

c/src/playback_state.rs
ericmlujan: Aha, yes @eloff and I discussed this last night and came to the same conclusion. It was initially unclear to us whether the data ptr or the length had more semantic meaning in determining null-ness, but when we talked it out, it made sense to treat either being zero-valued as an empty string

eloff: If data is null with non zero len that would be an invalid state. Not against handling it defensively though.

c/src/lib.rs
ericmlujan: In an earlier version of this code, I was using `FoxgloveStringBuf` as the `request_id` member and it was being both allocated and freed on the C++ side (resulting in a double-free when rust attempted to drop it). We were able to luckily simplify the code to use `FoxgloveString` instead, which I believe makes this change no longer relevant. I will revert this hunk.

c/src/playback_state.rs
ericmlujan: Added a small change to defend against this

c/src/playback_state.rs
ericmlujan: Thanks for pointing out the convention, that was clarifying!

c/src/server.rs
ericmlujan: Neat! That made the code a little nicer

ericmlujan:
@gasmith thanks for the initial feedback, pushed [9c56e77](https://github.com/foxglove/foxglove-sdk/pull/746/commits/9c56e7707454a1c3fe4603eae75df9e9483f0117) to incorporate it

c/src/lib.rs
```diff
@@ -94,7 +97,7 @@ impl FoxgloveString {
     /// The [`data`] field must be valid UTF-8, correctly aligned, and have a length equal to
     /// [`FoxgloveString.len`].
     unsafe fn as_utf8_str(&self) -> Result<&str, std::str::Utf8Error> {
-        if self.data.is_null() {
+        if self.data.is_null() || self.len == 0 {
```
gasmith, line 100: Can we also update `FoxgloveString::is_empty()` to return true when `self.data.is_null()`?

c/src/server.rs
```diff
@@ -464,6 +504,37 @@ pub extern "C" fn foxglove_server_broadcast_time(
...
+///
+/// # Safety
+/// - `playback_state` must be a valid pointer to a playback state that lives for the duration of the call.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn foxglove_server_broadcast_playback_state(
+    server: Option<&FoxgloveWebSocketServer>,
+    playback_state: Option<&FoxglovePlaybackState>,
+) -> FoxgloveError {
+    let Some(server) = server else {
+        return FoxgloveError::ValueError;
```
eloff, line 519: Nit: Could be written more succinctly using map_err and ?

c/src/server.rs
```diff
@@ -916,4 +987,38 @@ impl foxglove::websocket::ServerListener for FoxgloveServerCallbacks {
...
+    fn on_playback_control_request(
+        &self,
+        playback_control_request: foxglove::websocket::PlaybackControlRequest,
+    ) -> Option<foxglove::websocket::PlaybackState> {
+        let on_playback_control_request = self.on_playback_control_request?;
+
+        let c_playback_control_request = FoxglovePlaybackControlRequest {
+            playback_command: playback_control_request.playback_command as u8,
+            playback_speed: playback_control_request.playback_speed,
+            seek_time: playback_control_request.seek_time.as_ref(),
```
eloff, line 1000: Nit: could assign from request_id in the C struct and delete this var

c/src/playback_state.rs
```diff
@@ -0,0 +1,43 @@
+use crate::FoxgloveString;
+
+#[repr(C)]
+pub struct FoxglovePlaybackState {
+    /// The status of server data playback
+    pub status: u8,
+    /// The current time of playback, in absolute nanoseconds
+    pub current_time: u64,
+    /// The speed of playback, as a factor of realtime
+    pub playback_speed: f32,
```
eloff, line 10: Nit: move this field up one adjacent to the status and this struct is a word smaller

cpp/foxglove/src/server.cpp
```diff
@@ -223,6 +223,36 @@ FoxgloveResult<WebSocketServer> WebSocketServer::create(
...
+          const foxglove_playback_control_request* c_playback_control_request,
+          foxglove_playback_state* c_playback_state
+        ) {
+          try {
+            if (c_playback_control_request == nullptr) {
+              return;
+            }
+
+            std::optional<PlaybackState> maybe_playback_state =
+              (static_cast<const WebSocketServerCallbacks*>(context))
```
eloff: Nit: I think this would be easier on the eyes if the cast was on a dedicated line with an auto var.

cpp/foxglove/src/server.cpp
```diff
@@ -223,6 +223,36 @@ FoxgloveResult<WebSocketServer> WebSocketServer::create(
...
+
+            if (!maybe_playback_state.has_value()) {
+              return;
+            }
+
+            const PlaybackState& playback_state = maybe_playback_state.value();
+            c_playback_state->status = static_cast<uint8_t>(playback_state.status);
+            c_playback_state->current_time = playback_state.current_time;
+            c_playback_state->playback_speed = playback_state.playback_speed;
+          } catch (const std::exception& exc) {
```
eloff, line 243: Could we narrow the try catch to just around the callback (is that the only thing that might throw here?)

cpp/foxglove/tests/test_server.cpp
```diff
@@ -1488,3 +1499,181 @@ TEST_CASE("Server info") {
...
+  )});
+
+  writeFloatLE(msg, playback_control_request.playback_speed);
+  msg.emplace_back(
+    playback_control_request.seek_time.has_value() ? std::byte{0x1} : std::byte{0x0}
+  );
+  writeIntLE(
+    msg, playback_control_request.seek_time.has_value() ? *playback_control_request.seek_time : 0x0
+  );
+  uint32_t request_id_size =
```
eloff: This is never this large in the tests, could simplify this

eloff (commented):
My feedback was all minor, this is looking good now. Greg had some good comments.

c/src/server.rs
ericmlujan: I don't think we can use `?`, since this C-bound function returns an Error, not a Result.

c/src/server.rs
ericmlujan: I'm able to inline it, but I don't think I can have `c_playback_state.request_id` populated from `c_playback_control_request.request_id`. I tried it and failed compilation because the instance is partially moved.

cpp/foxglove/include/foxglove/server.hpp
```diff
@@ -228,6 +252,12 @@ struct WebSocketServerOptions {
   /// This option is for internal use only and may change.
   std::optional<std::map<std::string, std::string>> server_info = std::nullopt;
   /// @endcond
+
+  /// @cond foxglove_internal
+  /// @brief The time range for playback. This applies if the server is playing back a fixed time
+  /// range of data.
```
gasmith, line 257: Mention that this is required for RangedPlayback, and also implies RangedPlayback.

cpp/foxglove/include/foxglove/server.hpp
```diff
@@ -69,6 +71,12 @@ enum class WebSocketServerCapabilities : uint8_t {
   /// Allow clients to request assets. If you supply an asset handler to the
   /// server, this capability will be advertised automatically.
   Assets = 1 << 5,
+  /// @cond foxglove_internal
+  /// Indicates that the server is sending data within a fixed time range. This requires the
+  /// server to specify the `data_start_time` and `data_end_time` fields in its `ServerInfo`
+  /// message.
```
gasmith:
```suggestion
  /// server to specify the `playback_time_range` field in `WebSocketServerOptions`.
```

cpp/foxglove/tests/test_server.cpp
```diff
@@ -1488,3 +1499,181 @@ TEST_CASE("Server info") {
...
+    REQUIRE(received_playback_control_request->seek_time.has_value());
+    REQUIRE(received_playback_control_request->seek_time.value() == 42);
+    REQUIRE(received_playback_control_request->request_id == "a_request_id");
+  }
+
+  auto received_binary_playback_state = client.filterRecv([](const std::string& payload) {
+    return payload[0] == '\x05';  // PlaybackState binary opcode
+  });
+
+  REQUIRE(received_binary_playback_state.has_value());
```
gasmith: Nit: Should we also validate that the request ID is correct?

cpp/foxglove/tests/test_server.cpp
ericmlujan: The problem is that the windows build complains any time you narrow from size_t (uint64_t) to uint32_t, I figured doing something more principled here would be better than trying to throw casts at it until the compiler accepted it

ericmlujan: I was able to make the code a bit more terse though by using `std::min`

c/src/server.rs
eloff: Right, because FoxgloveString is not a copy type, (even although it is in C, we're not in C). This is fine then.

cpp/foxglove/tests/test_server.cpp
ericmlujan: Heh, and that caused MSVC to complain again. Looks like `static_cast` is the way to go here

PR #747, opened 2025-11-05
Python bindings for PlaybackControlRequest listener

ericmlujan:
### Changelog
None

### Docs
Python docstrings updated, but hidden from public documentation

### Description
Add Python bindings for the RangedPlayback capability and the supporting ServerListener callbacks and classes introduced in #728.

---

> [!NOTE]
> Introduce RangedPlayback with playback control request/state types, listener callback, server broadcast API, and a validated playback_time_range option.
> 
> - **Python API**:
>   - Add `Capability.RangedPlayback` and new types: `PlaybackCommand`, `PlaybackControlRequest`, `PlaybackStatus`, `PlaybackState` in `foxglove.websocket`.
>   - Extend `ServerListener` with `on_playback_control_request()` returning optional `PlaybackState`.
>   - Expose `WebSocketServer.broadcast_playback_state()`.
>   - `start_server()` gains `playback_time_range: tuple[int, int] | None` (validated) and forwards to core.
> - **Type stubs / bindings**:
>   - Update `.pyi` stubs to include new enums/classes and server method.
>   - Rust-Py bindings: define/playback enums and structs, conversions, and wire `on_playback_control_request` callback; map `Capability.RangedPlayback`; add tuple parsing for `playback_time_range`.
> - **Docs**:
>   - Exclude new playback-related members from auto-docs in `docs/api/index.rst`.
> - **Tests**:
>   - Update server interface test to call `broadcast_playback_state`.
>   - Add tests validating invalid `playback_time_range` inputs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 98ffe39ceb0722d9ab44680b11b61708f96cffef. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/python/foxglove-sdk/python/foxglove/tests/test_server.py
+++ b/python/foxglove-sdk/python/foxglove/tests/test_server.py
@@ -128,11 +128,14 @@ def test_server_with_invalid_playback_time_range() -> None:
         # Tuple of a single element
         start_server(port=0, playback_time_range=(123,))
 
+    with pytest.raises(TypeError):
         # Tuple with invalid types
         start_server(port=0, playback_time_range=("not-a-time", None))
 
+    with pytest.raises(TypeError):
         # Not a tuple
         start_server(port=0, playback_time_range=23443)
 
+    with pytest.raises(TypeError):
         # Tuple with too many elements
         start_server(port=0, playback_time_range=(123, 456, 789))
--- a/python/foxglove-sdk/src/websocket.rs
+++ b/python/foxglove-sdk/src/websocket.rs
@@ -154,6 +154,7 @@ pub struct PyPlaybackControlRequest {
     playback_command: PyPlaybackCommand,
     playback_speed: f32,
     seek_time: Option<u64>,
+    request_id: String,
 }
 impl From<PlaybackControlRequest> for PyPlaybackControlRequest {
     fn from(value: PlaybackControlRequest) -> PyPlaybackControlRequest {
@@ -161,6 +162,7 @@ impl From<PlaybackControlRequest> for PyPlaybackControlRequest {
             playback_command: value.playback_command.into(),
             playback_speed: value.playback_speed,
             seek_time: value.seek_time,
+            request_id: value.request_id,
         }
     }
 }
@@ -409,14 +411,14 @@ impl ServerListener for PyServerListener {
         playback_control_request: PlaybackControlRequest,
     ) -> Option<PlaybackState> {
         let py_playback_control_request: PyPlaybackControlRequest = playback_control_request.into();
-        let result: PyResult<PyPlaybackState> = Python::with_gil(|py| {
+        let result: PyResult<Option<PyPlaybackState>> = Python::with_gil(|py| {
             let result = self.listener.bind(py).call_method(
                 "on_playback_control_request",
                 (py_playback_control_request,),
                 None,
             )?;
 
-            result.extract::<PyPlaybackState>()
+            result.extract::<Option<PyPlaybackState>>()
         });
 
         match result {
@@ -424,7 +426,8 @@ impl ServerListener for PyServerListener {
                 tracing::error!("Callback failed: {}", err.to_string());
                 None
             }
-            Ok(playback_state) => Some(playback_state.into()),
+            Ok(None) => None,
+            Ok(Some(playback_state)) => Some(playback_state.into()),
         }
     }
 }


python/foxglove-sdk/python/foxglove/_foxglove_py/websocket.pyi
```diff
@@ -187,6 +190,66 @@ class ParameterValue:
 
         def __init__(self, value: dict[str, AnyParameterValue]) -> None: ...
 
+class PlaybackCommand:
```
gasmith:
```suggestion
class PlaybackCommand(Enum):
```

python/foxglove-sdk/python/foxglove/tests/test_server.py
```diff
@@ -47,6 +47,14 @@ def test_server_interface() -> None:
...
     server.publish_status("test message", StatusLevel.Info, "some-id")
     server.broadcast_time(time.time_ns())
+    server.broadcast_playback_state(
+        PlaybackState(
+            status=PlaybackStatus.Paused,
+            playback_speed=1.0,
+            current_time=time.time_ns(),
+            request_id=None,
+        )
+    )
```
gasmith, line 59: I was going to point out that this is a no-op, for lack of the `RangedPlayback` capability, but then I realized we're also not setting the `Time` capability for the line above, so...

I guess it's not like we're validating anything. It's sort of a "if it doesn't raise Exceptions it's probably OK" kind of test. We should file a ticket to add proper test coverage, particularly to ensure that the `ServerListener` callbacks are working properly.

python/foxglove-sdk/src/websocket.rs
```diff
@@ -375,6 +500,7 @@ pub fn start_server(
     context: Option<PyRef<PyContext>>,
     session_id: Option<String>,
     channel_filter: Option<Py<PyAny>>,
+    playback_time_range: Option<Py<PyTuple>>,
```
gasmith, line 511: I think this also needs to be plumbed through `python/foxglove/__init__.py`, where we have a wrapper around `start_server`; and we also need to update the interface typing in `python/foxglove/_foxglove_py/__init__.pyi`.

I wonder whether those wrappers in `__init__.py` could simply be replaced by something like:

```python
start_server = _foxglove.start_server
start_cloud_sink = _foxglove.start_cloud_sink
```

I suppose we'd move these more elaborate docstrings to the interface definitions in `python/foxglove/_foxglove_py/__init__.pyi`. But it'd be one less copy of these args to maintain.

python/foxglove-sdk/src/websocket.rs
```diff
@@ -417,6 +543,18 @@ pub fn start_server(
...
 
+    if let Some(playback_time_range) = playback_time_range {
+        let bound_time_range = playback_time_range.bind(py);
+        if bound_time_range.len() != 2 {
+            return Err(PyTypeError::new_err(
+                "playback_time_range must be a tuple of (start_time, end_time)",
+            ));
+        }
+        let start_time = bound_time_range.get_item(0)?.extract::<u64>()?;
+        let end_time = bound_time_range.get_item(1)?.extract::<u64>()?;
```
gasmith, line 562: Might be nice to add some test coverage for the error handling here.

python/foxglove-sdk/Cargo.toml
```diff
@@ -29,6 +29,7 @@ tokio-tungstenite.workspace = true
 tokio-util.workspace = true
 tokio.workspace = true
 foxglove = { path = "../../rust/foxglove", default-features = false, features = [
+  "unstable",
```
gasmith: I thought we opted for `doc(hidden)` over `cfg(feature = "unstable")` in the rust interface?

python/foxglove-sdk/python/foxglove/tests/test_server.py
ericmlujan: Yeah, the existing tests are primarily focused on making sure that things don't throw a runtime error when the codepaths are exercised, and that mypy analysis passes, and within the scope of this PR I just wanted to match that level of scrutiny. I'd like to see more effective testing here, and can file a ticket for that.

python/foxglove-sdk/Cargo.toml
ericmlujan: Good call, this was leftover from a previous version that predated that shift

python/foxglove-sdk/src/websocket.rs
ericmlujan: I'd be curious to know if we can swing that too. My (mis)adventures on the Python side of things similarly make me apprehensive about all the parallel places things exist/are documented/are stubbed (though I know pyo3 stub generation support is only experimental atm)

gasmith: Yeah, the hope is that there is a glorious future in which pyo3 stub generation saves us from our toil. Maybe we should take another look at https://crates.io/crates/pyo3-stub-gen, now that we've bumped the minimum python version to 3.10. (See also: FG-9900)

python/foxglove-sdk/src/websocket.rs
ericmlujan: Resolved in [b1f72ff](https://github.com/foxglove/foxglove-sdk/pull/747/commits/b1f72ff5628aefca97e8ab6ee0a4f0d5da5aa30e)

ericmlujan:
@gasmith thanks for the thorough review, this is ready for another pass whenever you get a chance!

gasmith (approved):

PR #749, opened 2025-11-06
playground: add more packages

jtbandes:
### Changelog
None

### Docs

None

### Description

Added numpy, scipy, opencv, & pandas. Switch to `loadPackagesFromImports` to load the packages on demand.

### DIFF
--- a/playground/src/RunnerWorker.ts
+++ b/playground/src/RunnerWorker.ts
@@ -120,15 +120,7 @@ export class RunnerWorker {
       },
     });
     this.#abortController.signal.throwIfAborted();
-    await pyodide.loadPackage([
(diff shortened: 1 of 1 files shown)


gasmith:
> I wonder if we might be able to do something more clever like on-demand installation — open to any suggestions/ideas on how to do that?

Could we perhaps use https://pyodide.org/en/stable/usage/api/js-api.html#pyodide.loadPackagesFromImports?

jtbandes:
😮

jtbandes:
nice find, seems to work.

Throws this error if you import something we don't bundle:
image

msadowski:
This is amazing! Thanks!

PR #755, opened 2025-11-11
embed: add api to create layouts schemas and code generation

surbina:
### Docs

TODO

### Description

- Use JSONSchemas to describe the layout format
    - Includes two panel examples (Markdow and StateTransitions)
- Use [datamodel-code-generator](https://koxudaxi.github.io/datamodel-code-generator/) to generate Python code based on the JSON Schemas
- Bumps minimum supported Python version to 3.10

Use the following command to generate the API code:

```shell
cd python/foxglove-sdk/ &&
uv run datamodel-codegen \
	--input ../../layout-schemas/ \
	--input-file-type jsonschema \
	--output ./python/foxglove/layouts/models/ \
	--output-model-type dataclasses.dataclass \
	--snake-case-field \
	--target-python-version 3.10 \
	--use-double-quotes \
	--base-class ..dataclass_model.DataclassModel \
	--use-union-operator \
	--use-schema-description \
	--use-field-description \
	--keyword-only \
	--enum-field-as-literal all \
	--disable-timestamp
```

_Plesa note that the autogenerated code won't be commited and should instead be generated as part of our CI process. It's included in this PR so that people can review the output of the generation process_

This is an alternative approach to the one in #726

#### Why bumping Python?

The autogenerated code uses dataclasses. Dataclaases are a standard for representing object structs. The schema definitions includes a `base_panel` schema from which all panels inherit. This `base_panel` add a single required field: `type`. The each class needs to overwrite it with a proper value. This creates a problem during type checking as described [here](https://stackoverflow.com/questions/51575931/class-inheritance-in-python-3-7-dataclasses).

The simplest solution is using the `kw_only` parameter introduced in Python 3.10, as described [here](https://stackoverflow.com/a/69822584).

</td><td>

</td></tr></table>


james-rms:
> Plesa note that the autogenerated code won't be commited and should instead be generated as part of our CI process. It's included in this PR so that people can review the output of the generation process

why not? the usual pattern for me at least is to commit generated code and have a CI step that verifies that the genereated code is up-to-date.

james-rms:
Also you may want to tweak the linguist gitattribute in this PR to make the .json files visible in PR reviews.

schemas/layouts/layouts.json
```diff
@@ -0,0 +1,106 @@
...
+        }
+      },
+      "required": ["tabs"]
+    },
+    "Item": {
+      "type": "object",
+      "properties": {
+        "ratio": {
+          "type": "number",
+          "description": "The ratio of the item"
```
james-rms: is this an aspect ratio?

schemas/layouts/panel.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "Panel",
+  "description": "A generic panel in Foxglove",
+  "allOf": [{ "$ref": "./base_panel.json" }],
+  "properties": {
+    "config": {
+      "type": "object",
+      "description": "The config of the panel. This can be an arbitrary JSON serializable object.",
+      "additionalProperties": true
```
james-rms:
```suggestion
```
`additionalProperties` is implied by default

schemas/layouts/base_panel.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "Base Panel",
+  "description": "A base panel in Foxglove",
```
defunctzombie: What is a base panel?

layout-schemas/layout.json
```diff
@@ -0,0 +1,106 @@
+{
```
defunctzombie, line 1: meta: why's this file `layouts.json` rather than `layout.json`? Your other files are singular rather than plural.

layout-schemas/layout.json
```diff
@@ -0,0 +1,106 @@
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "Layout",
+  "description": "A layout in Foxglove",
+  "$defs": {
```
defunctzombie, line 20: should these defs be after the properties so a user can learn about the layout object structure first and then go to the specific things they want?

schemas/layouts/layouts.json
```diff
@@ -0,0 +1,106 @@
...
+          "description": "The source code of the user script"
+        }
+      },
+      "required": ["name", "sourceCode"]
+    }
+  },
+  "type": "object",
+  "properties": {
+    "version": {
+      "type": "string",
```
defunctzombie: Is this a string our should it be an enum of values? My knowledge of how you can use jsonschema for this is to have a union type with explicit version values that are the discriminator.

schemas/layouts/layouts.json
```diff
@@ -0,0 +1,106 @@
...
+      "description": "The version of the layout format"
+    },
+    "variables": {
+      "$ref": "#/$defs/Variables"
+    },
+    "items": {
+      "$ref": "#/$defs/ItemList"
+    },
+    "userNodes": {
```
defunctzombie: We don't call these "nodes" for public facing. These are "scripts".

I don't feel strongly that we need scripts for this v1 or for this effort yet. The user already produces the data in a "script" context so they can turn it into the format needed for visualization which tends to negate the needs of scripts.

schemas/layouts/markdown_panel.json
```diff
@@ -0,0 +1,36 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "MarkdownPanel",
+  "description": "A Markdown panel in Foxglove",
```
defunctzombie: Should we be linking to the docs pages for each panel?

Makes me also think about us including this config in each panel's docs?

schemas/layouts/base_panel.json
surbina: I added this base class so that all panels extend `BasePanel`. This is to ensure:
1. All panels have a `type`
2. The content of an item is either a `BasePanel` (or something that inherited from it), a `Tabs` or an `ItemsList`.

There is also a `Panel` which is just an escape hatch to create configs for panels without python classes (e.g., panels from custom extensions).

schemas/layouts/markdown_panel.json
surbina: Yeah, this makes sense.

schemas/layouts/layouts.json
surbina: Sounds good, I can remove this so we focus on a minimal set of features.

schemas/layouts/layouts.json
surbina: Makes total sense!

schemas/layouts/layouts.json
surbina: This a ratio similar to the `flex` css property. It indicates the proportion of space taken by the item within the item list. I improved this description to make it clearer.

layout-schemas/layout.json
surbina: No reason, I renamed this to layout.json

layout-schemas/layout.json
surbina: Done!

PR #763, opened 2025-11-15
rust: Enable rust-2024-compatibility lints

gasmith:
### Changelog
None

### Docs
None

### Description
Enables lint rules to ease migration to the 2024 edition.

This change disables several lints that don't benefit us until we're actually ready to cut over to the 2024 edition, at which point they'll be useful for flagging areas for a one-time manual inspection.

This is the spiritual successor to #248.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,35 @@ tokio-util = { version = "0.7", features = ["rt"] }
 tracing = { version = "0.1", features = ["log"] }
 tracing-test = "0.2.5"
 
+[workspace.lints.rust]
+# Group to warn about changes coming in Rust 2024 edition. Having this on will
+# make a future transition smoother. Lowering the priority allows overriding
+# several lints in the group that shouldn't be automatically enabled.
+rust-2024-compatibility = { level = "warn", priority = 0 }
+
+# This lint flags all tail expressions that drop tempraries with a custom Drop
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -36,6 +36,35 @@ tokio-util = { version = "0.7", features = ["rt"] }
...
+# make a future transition smoother. Lowering the priority allows overriding
+# several lints in the group that shouldn't be automatically enabled.
+rust-2024-compatibility = { level = "warn", priority = 0 }
+
+# This lint flags all tail expressions that drop tempraries with a custom Drop
+# implementation. We'll manually review these we upgrade to 2024, but generally
+# don't expect any of our code to depend on drop-ordering in this way.
+#
+# See also: https://doc.rust-lang.org/edition-guide/rust-2024/temporary-tail-expr-scope.html
+tail-expr-drop-order = { level = "allow", priority = 1 }
```
jtbandes, line 50: > We'll manually review these we upgrade to 2024

Is the upgrade gonna force us to review them manually, or is having this disabled going to mean we just have to remember to come back and look at it? 😅

jtbandes (approved):

Cargo.toml
gasmith: No, the upgrade won't force us to review them. Hopefully a wall of text in Cargo.toml will be enough to help us remember. :prayer_beads:

PR #764, opened 2025-11-17
C/C++: allow custom write implementations for MCAP writer

james-rms:
### Changelog
- C++: added CustomWriter to allow C++ clients to write MCAP to targets other than the filesystem.
### Docs

Detailed in docstrings. Will add an example before merge.

### Description

Today Rust clients can create MCAP writers that wrap an arbitrary std::io::Write implementation of their own. This does not work in C++ or in python. This PR adds that ability to the C++ SDK. This is useful for anyone who want to log MCAP but not to a traditional filesystem - they may want to log to a socket, or they may want to intercept writes before they go to the filesystem. 

## TODO
- [x] Unit tests
- [x] Example code

</td><td>

</td></tr></table>

### DIFF
--- a/cpp/foxglove/tests/test_mcap.cpp
+++ b/cpp/foxglove/tests/test_mcap.cpp
@@ -9,8 +9,6 @@
 #include <catch2/matchers/catch_matchers_string.hpp>
 
 #include <array>
-#include <atomic>
-#include <cstring>
 #include <filesystem>
 #include <fstream>
 #include <optional>


c/src/channel.rs
```diff
@@ -14,13 +15,121 @@ pub enum FoxgloveMcapCompression {
...
 
+/// Custom writer function pointers for MCAP writing.
+/// write_fn and flush_fn must be non-null. Seek_fn may be null iff `disable_seeking` is set to true.
+/// These function pointers may be called from multiple threads.
+/// They will not be called concurrently with themselves or each-other.
+#[repr(C)]
+#[derive(Clone, Copy)]
+pub struct FoxgloveCustomWriter {
+    /// User-provided context pointer, passed to all callback functions
+    pub user_data: *mut std::ffi::c_void,
```
gasmith: FWIW, we call this opaque callback data "context" in [other](https://github.com/foxglove/foxglove-sdk/blob/cf6d7a339f587b6df527149bc45a982acc7d9354/c/src/service.rs#L182) [parts](https://github.com/foxglove/foxglove-sdk/blob/cf6d7a339f587b6df527149bc45a982acc7d9354/c/src/server.rs#L128) of the SDK.

The naming is unfortunate; I think we settled on it before we added the notion of "logging context" as a first-class thing in the API. Maybe worth a linear ticket for a rename, though it's also not a huge deal since this is only visible in the C API.

gasmith: Why `*mut` instead of `*const`? We don't need to mutate this on the rust side, and the annotation is lost on the C side.

c/src/channel.rs
```diff
@@ -14,13 +15,121 @@ pub enum FoxgloveMcapCompression {
...
+            user_data: *mut std::ffi::c_void,
+            data: *const u8,
+            len: usize,
+            error: *mut i32,
+        ) -> usize,
+    >,
+    /// Flush function: ensure all buffered data is written
+    pub flush_fn: Option<unsafe extern "C" fn(user_data: *mut std::ffi::c_void) -> i32>,
+    /// Seek function: change the current position in the stream
+    /// whence: 0=SEEK_SET, 1=SEEK_CUR, 2=SEEK_END
```
gasmith, line 42: Should we add a first-class enum for that?

c/src/channel.rs
```diff
@@ -119,42 +255,76 @@ pub unsafe extern "C" fn foxglove_mcap_open(
...
+                "write_fn and flush_fn must be provided".to_string(),
+            ));
+        }
+        let custom_writer = unsafe { CustomWriter::new(custom_writer_callbacks) };
+
+        let writer = builder.create(custom_writer)?;
+
+        McapWriterVariant::Custom(writer)
+    } else {
+        // Use file path (existing behavior)
```
gasmith, line 287:
```suggestion
        // Use file path
```

c/src/channel.rs
```diff
@@ -14,13 +15,121 @@ pub enum FoxgloveMcapCompression {
...
+        Ok(())
+    }
+}
+
+impl Seek for CustomWriter {
+    fn seek(&mut self, pos: SeekFrom) -> std::io::Result<u64> {
+        let seek_fn = self
+            .callbacks
+            .seek_fn
+            .expect("seek_fn checked in do_foxglove_mcap_open");
```
gasmith, line 101: Nit: Since `seek_fn` is optional, perhaps we return `ESPIPE` here instead of panicking. A bit pedantic, since I realize that we have safeguards to ensure that `disable_seeking` is set.

c/src/channel.rs
```diff
@@ -88,22 +197,49 @@ impl FoxgloveMcapOptions {
...
+        }
+    }
+}
+
+/// Create or open an MCAP writer for writing to a file or custom destination.
 /// Resources must later be freed with `foxglove_mcap_close`.
 ///
+/// If `custom_writer` is provided, the MCAP data will be written using the provided
+/// function pointers instead of to a file.
+///
```
gasmith, line 231: We should also mention that the callbacks will be invoked synchronously as part of this call. This is an important detail for any implementation that expects to use synchronization in callbacks.

c/src/channel.rs
```diff
@@ -119,42 +255,76 @@ pub unsafe extern "C" fn foxglove_mcap_open(
...
+            .map_err(foxglove::FoxgloveError::IoError)?;
+        if let Some(sink_channel_filter) = options.sink_channel_filter {
+            builder = builder.channel_filter(Arc::new(ChannelFilter::new(
+                options.sink_channel_filter_context,
+                sink_channel_filter,
+            )));
+        }
+        let writer = builder
+            .create(BufWriter::new(file))
+            .expect("Failed to create writer");
```
gasmith: Looks like this was the original behavior, but this shouldn't be a panic.

cpp/foxglove/include/foxglove/mcap.hpp
```diff
@@ -27,6 +29,37 @@ namespace foxglove {
...
 class Context;
 
+/// @brief Custom writer for writing MCAP data to arbitrary destinations.
+///
+/// This provides a simple function pointer interface that matches the C API.
+/// Users are responsible for managing the lifetime of user_data and ensuring
+/// thread safety if needed.
+struct CustomWriter {
+  /// @brief User-provided context pointer, passed to all callback functions
+  void* user_data = nullptr;
```
gasmith: This is a bit of C-ism. Other callbacks in the C++ SDK do not take an explicit `void *` like this. The expectation is that users would typically reach for a closure or `std::bind` to bind additional context.

cpp/foxglove/include/foxglove/mcap.hpp
```diff
@@ -27,6 +29,37 @@ namespace foxglove {
...
+
+  /// @brief Write function: write data to the custom destination
+  /// @param user_data User-provided context pointer
+  /// @param data Pointer to data to write
+  /// @param len Number of bytes to write
+  /// @param error Pointer to error code (set to an error number defined in errno.h if write fails)
+  /// @return Number of bytes actually written
+  std::function<size_t(void* user_data, const uint8_t* data, size_t len, int32_t* error)> write_fn;
```
gasmith: We use `camelCase` for callbacks in C++, even though we use `snake_case` for struct fields.

cpp/foxglove/include/foxglove/mcap.hpp
jtbandes: +1, especially since you are already ignoring this struct field and doing `static_cast<CustomWriter*>(user_data)` to invoke the callback, there is no need to also pass the user_data in to the callback itself.

cpp/foxglove/src/mcap.cpp
```diff
@@ -6,13 +6,47 @@
...
   c_options.context = options.context.getInner();
   c_options.path = {options.path.data(), options.path.length()};
   c_options.profile = {options.profile.data(), options.profile.length()};
+
+  // Handle custom writer if provided
+  FoxgloveCustomWriter c_custom_writer = {};
+  if (options.custom_writer.has_value()) {
+    const auto& custom_writer = options.custom_writer.value();
+    c_custom_writer.user_data =
+      const_cast<CustomWriter*>(&custom_writer);  // Safe: we control the lifetime
```
jtbandes: _how_ are we controlling this lifetime? maybe I skimmed too fast, but I don't see where the `options` / `options.custom_writer` would be preserved beyond the end of this constructor

Compare with `sink_channel_filter` below, which is stored in a unique_ptr in a class member to extend its lifetime.

c/src/channel.rs
eloff: Naming things...

james-rms: It feels a little clearer to me where we call it that the callback is going to mutate the contents - i can see your argument though.

c/src/channel.rs
james-rms: We could but that feels like gilding the lily to me. I'm trying to make any reader recall the behavior of `fseek`, where the canonical type of `whence` is `int` with a few well-defined values.

c/src/channel.rs
james-rms: I may have hit a merge conflict here, will double check.

cpp/foxglove/src/mcap.cpp
james-rms: I've since moved to the same approach as sink_channel_filter.

c/src/channel.rs
james-rms: I think we should assert because we can only get here if there's a bug in other code.

c/src/channel.rs
james-rms: I explain this more in the docstring for CustomWriter.

james-rms: I explain this more in the docstring for CustomWriter.

c/src/channel.rs
```diff
@@ -18,12 +18,13 @@ pub enum FoxgloveMcapCompression {
 /// Custom writer function pointers for MCAP writing.
 /// write_fn and flush_fn must be non-null. Seek_fn may be null iff `disable_seeking` is set to true.
 /// These function pointers may be called from multiple threads.
-/// They will not be called concurrently with themselves or each-other.
+/// These functions are called synchronously with respect to each other within the SDK, but these
+/// calls are not synchronized with other SDK function calls.
 #[repr(C)]
 #[derive(Clone, Copy)]
 pub struct FoxgloveCustomWriter {
     /// User-provided context pointer, passed to all callback functions
```
gasmith, line 26:
```suggestion
    /// User-provided context pointer, passed to the write_fn callback.
```

And similarly, below.

c/src/channel.rs
gasmith: Should probably name this `write_context` to match. Same below.

cpp/foxglove/src/mcap.cpp
```diff
@@ -34,15 +30,19 @@ FoxgloveResult<McapWriter> McapWriter::create(const McapWriterOptions& options)
...
+    custom_writer = std::make_unique<CustomWriter>(options.custom_writer.value());
+    c_custom_writer = std::make_unique<FoxgloveCustomWriter>();
+
+    c_custom_writer->write_context = &(custom_writer.get()->write);
+    c_custom_writer->write_fn = &custom_write;
+    c_custom_writer->flush_context = &(custom_writer.get()->flush);
+    c_custom_writer->flush_fn = &custom_flush;
+    c_custom_writer->seek_context = &(custom_writer.get()->seek);
+    c_custom_writer->seek_fn = &custom_seek;
+    c_options.custom_writer = c_custom_writer.get();
```
gasmith: Nit: I feel like having a single common context pointer for all callbacks (to the `FoxgloveCustomWriter`) would've been sufficient, and more aligned with what we do for `WebSocketServerCallbacks`, for example.

cpp/foxglove/src/mcap.cpp
```diff
@@ -92,13 +92,18 @@ FoxgloveResult<McapWriter> McapWriter::create(const McapWriterOptions& options)
...
 }
 
 McapWriter::McapWriter(
-  foxglove_mcap_writer* writer, std::unique_ptr<SinkChannelFilterFn> sink_channel_filter
+  foxglove_mcap_writer* writer, std::unique_ptr<SinkChannelFilterFn> sink_channel_filter,
+  std::unique_ptr<CustomWriter> custom_writer, std::unique_ptr<FoxgloveCustomWriter> c_custom_writer
 )
     : sink_channel_filter_(std::move(sink_channel_filter))
+    , custom_writer_(std::move(custom_writer))
+    , c_custom_writer_(std::move(c_custom_writer))
```
gasmith: Is it necessary to store `c_custom_writer_` as well? I would've expected that the `foxglove_mcap_options.custom_writer` pointer only needs to live through the duration of the `foxglove_mcap_open()` call.

c/include/foxglove-c/foxglove-c.h
```diff
@@ -1704,6 +1695,44 @@ typedef struct foxglove_raw_image {
...
 
+#if !defined(__wasm__)
+/**
+ * Custom writer function pointers for MCAP writing.
+ * write_fn and flush_fn must be non-null. Seek_fn may be null iff `disable_seeking` is set to true.
+ * These function pointers may be called from multiple threads.
+ * These functions are called synchronously with respect to each other within the SDK, but these
+ * calls are not synchronized with other SDK function calls.
+ */
+typedef struct FoxgloveCustomWriter {
```
gasmith: I think we need an entry in `cbindgen.toml` to rename this as `foxglove_custom_writer`.

c/src/channel.rs
```diff
@@ -88,22 +204,49 @@ impl FoxgloveMcapOptions {
...
 /// Returns 0 on success, or returns a FoxgloveError code on error.
 ///
 /// # Safety
 /// `path` and `profile` must contain valid UTF8. If `context` is non-null,
 /// it must have been created by `foxglove_context_new`.
+/// If `custom_writer` is non-null, its function pointers must be valid and
+/// all `context` pointers must remain valid for the lifetime of the writer.
 #[unsafe(no_mangle)]
 #[must_use]
 pub unsafe extern "C" fn foxglove_mcap_open(
```
gasmith, line 241: Nit: Not your change, but... `foxglove_mcap_open` feels a bit out of place in a file called `channel.rs`. Perhaps we should split the mcap stuff into a new `mcap.rs` file.

eloff:
The changes look good to me, I'd like @gasmith to give it another look though

PR #768, opened 2025-11-18
playground: allow changing embed url

jtbandes:
### Description

Taken over by @snosenzo 

Add the ability to specify an embed url in the playground viewer. This makes it easier to test embed preview deploy features.

I also fixed an incorrect directory structure in the setup readme.

Related foxglove preview deploy PR: https://github.com/foxglove/app/pull/11113

</td><td>

</td></tr></table>

https://linear.app/foxglove/issue/FOU-28/make-signin-work-for-embed-preview-deploys

### DIFF
--- a/playground/src/urlState.ts
+++ b/playground/src/urlState.ts
@@ -78,13 +78,9 @@ function deserializeState(serialized: string): UrlState | undefined {
   const layout = layoutJson ? (JSON.parse(layoutJson) as unknown) : undefined;
 
   let embedURL: URL | undefined;
-  try {
-    const embedStr = params.get("embed");
-    if (embedStr) {
-      embedURL = new URL(embedStr);
-    }
-  } catch {
-    // ignore
+  const embedStr = params.get("embed");
+  if (embedStr) {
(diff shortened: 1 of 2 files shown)


playground/src/urlState.ts
```diff
@@ -67,11 +71,23 @@ function deserializeState(serialized: string): UrlState | undefined {
...
-  return { code, layout };
+
+  let embedURL: URL | undefined;
+  try {
+    const embedStr = params.get("embed");
+    if (embedStr) {
+      embedURL = new URL(embedStr);
+    }
+  } catch {
+    // ignore
```
achim-k: why not throw here?

jtbandes: IMO if one of the url params is invalid, I'm not sure we should completely ignore all the other ones

achim-k: shouldn't we at least show an error toast or something if the embed URL is an invalid URL?

achim-k:
Is there a way to detect if the loaded website is actually a foxglove embedded instance?

image

PR #769, opened 2025-11-19
foxglove_bridge: fix message format error when using cache

chencaidy:
### Changelog

Fix message format error when using cache, foxglove shows "Failed to parse channel schema".

### Docs

None

### Description

foxglove_bridge contains a hard-coded issue when retrieving cached message files, causing IDL files to be correctly recognized only on the first attempt. Subsequent cached requests return an incorrect format.

Here is the first discussions:
https://github.com/orgs/foxglove/discussions/1195

This fix was tested on my computer, both MSG and IDL worked well.

### DIFF
--- a/ros/src/foxglove_bridge/src/message_definition_cache.cpp
+++ b/ros/src/foxglove_bridge/src/message_definition_cache.cpp
@@ -341,8 +341,8 @@ const MessageSpec& MessageDefinitionCache::load_message_spec(
 
 std::pair<MessageDefinitionFormat, const std::string&> MessageDefinitionCache::get_full_text(
   const std::string& root_package_resource_name) {
-  if (full_text_cache_.find(root_package_resource_name) != full_text_cache_.end()) {
-    return {MessageDefinitionFormat::MSG, full_text_cache_[root_package_resource_name]};
+  if (auto it = full_text_cache_.find(root_package_resource_name); it != full_text_cache_.end()) {
+    return it->second;
   }
 
   std::unordered_set<DefinitionIdentifier, DefinitionIdentifierHash> seen_deps;
@@ -376,8 +376,9 @@ std::pair<MessageDefinitionFormat, const std::string&> MessageDefinitionCache::g
     result = delimiter(root_definition_identifier) + append_recursive(root_definition_identifier);
   }
 
-  auto [it, _] = full_text_cache_.emplace(root_package_resource_name, result);
-  return {format, it->second};
+  auto [it, _] =
+    full_text_cache_.emplace(root_package_resource_name, std::make_pair(format, result));
+  return it->second;
 }
 
 }  // namespace foxglove_bridge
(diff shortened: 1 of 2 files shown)


ros/src/foxglove_bridge/src/message_definition_cache.cpp
```diff
@@ -341,8 +341,8 @@ const MessageSpec& MessageDefinitionCache::load_message_spec(
 
 std::pair<MessageDefinitionFormat, const std::string&> MessageDefinitionCache::get_full_text(
   const std::string& root_package_resource_name) {
-  if (full_text_cache_.find(root_package_resource_name) != full_text_cache_.end()) {
-    return {MessageDefinitionFormat::MSG, full_text_cache_[root_package_resource_name]};
+  if (auto it = full_text_cache_.find(root_package_resource_name); it != full_text_cache_.end()) {
+    return {it->second.first, it->second.second};
```
clalancette: I'm not a huge fan of using semicolons inside of if statements.  Further, I think we can just return the second part of the iterator:


```suggestion
  auto found_it = full_text_cache_.find(root_package_resource_name);
  if (found_it != full_text_cache_.end()) {
    return found_it->second;
```

ros/src/foxglove_bridge/src/message_definition_cache.cpp
```diff
@@ -376,8 +376,9 @@ std::pair<MessageDefinitionFormat, const std::string&> MessageDefinitionCache::g
     result = delimiter(root_definition_identifier) + append_recursive(root_definition_identifier);
   }
 
-  auto [it, _] = full_text_cache_.emplace(root_package_resource_name, result);
-  return {format, it->second};
+  auto [it, _] =
+    full_text_cache_.emplace(root_package_resource_name, std::make_pair(format, result));
+  return {it->second.first, it->second.second};
```
clalancette: Same here, we can just return the second part of the iterator:


```suggestion
  return it->second;
```

ros/src/foxglove_bridge/src/message_definition_cache.cpp
chencaidy: Using semicolons internally is to maintain consistency in coding style. besauce `msg_specs_by_definition_identifier_` used it. Did you think this modification is necessary?

chencaidy:
> The code changes look good to me, thanks for the fix and for iterating.
> 
> I'm thinking we should add a test case for this, as it isn't a common scenario and thus may regress in the future. I think we could use the ROS 2 `test_msgs` package as a test dependency to get access to the `.idl` files in there, and then write a test that uses those. @chencaidy would you mind giving that a shot and seeing if you can write a test? Thanks.

@clalancette I noticed that humble and jazzy have no idl files in test_msgs repo, and I think add a idl to foxglove_msgs also a quirk, This unit test will break the testing framework. hmm... I have no better idea now.

clalancette:
> @clalancette I noticed that humble and jazzy have no idl files in test_msgs repo, and I think add a idl to foxglove_msgs also a quirk, This unit test will break the testing framework. hmm... I have no better idea now.

Oh, you are right.  I misremembered how that was all put together.

OK, in that case, I'm good with this as-is.  Thanks for taking a look.

PR #770, opened 2025-11-19
C++ example for playback over WebSocket

ericmlujan:
### Changelog
None

### Docs
In-line documentation provided

### Description
Annotated C++ example using the functionality introduced in #746. 

Fixes: FLE-142

### DIFF
--- a/cpp/examples/playback-ws-server/src/main.cpp
+++ b/cpp/examples/playback-ws-server/src/main.cpp
@@ -131,7 +131,7 @@ public:
         // current time and playback index to the start of the data buffer, and enter a Paused
         // state.
         server_->broadcastPlaybackState(
-          {foxglove::PlaybackStatus::Ended, current_time_, playback_speed_, std::nullopt}
+          {foxglove::PlaybackStatus::Ended, current_time_, playback_speed_, false, std::nullopt}
         );
         return;
       }
@@ -167,7 +167,12 @@ public:
       seekInternal(*request.seek_time);
     }
 
-    return currentPlaybackStateInternal();
+    auto playback_state = currentPlaybackStateInternal();
+    if (request.seek_time.has_value()) {
+      playback_state.did_seek = true;
+    }
+
+    return playback_state;
   }
 
   void seek(uint64_t seek_time) {
@@ -218,6 +223,7 @@ private:
       playing_ ? foxglove::PlaybackStatus::Playing : foxglove::PlaybackStatus::Paused,
       current_time_,
       playback_speed_,
+      false,
       std::nullopt,
     };
   }


cpp/examples/playback-ws-server/src/main.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+using namespace std::chrono_literals;
+
+// Example class for playing back a fixed interval of robot stack data.
+//
+// More practical implementations would load data from a file on disk, but for the sake of
+// illustration, this example generates a fixed buffer of example data in-memory.
+class DataPlayer {
+public:
+  explicit DataPlayer(size_t num_timesteps, foxglove::RawChannel&& channel)
```
eloff: Don't think you need explicit here

cpp/examples/playback-ws-server/src/main.cpp
```diff
@@ -0,0 +1,257 @@
...
+    server_ = std::make_unique<foxglove::WebSocketServer>(std::move(server_result.value()));
+  }
+
+  static std::string toMessage(const std::pair<uint64_t, double>& data) {
+    return "{\"val\": " + std::to_string(data.second) + "}";
+  }
+
+  void process() {
+    if (!server_) {
+      throw std::runtime_error("Tried to spin with uninitialized server");
```
eloff: Run or process might be better words than spin here

cpp/examples/playback-ws-server/src/main.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+private:
+  std::vector<std::pair<uint64_t, double>> data_;
+
+  foxglove::RawChannel channel_;
+  std::unique_ptr<foxglove::WebSocketServer> server_ = nullptr;
+
+  // Internal variables for orchestrating playback. In addition to accessing data_, these are are
+  // used to generate a foxglove::PlaybackState to send to the Foxglove player.
+  std::atomic<size_t> current_playback_index_ = 0;
```
eloff: Using multiple atomics like this leads to all kinds of strange race conditions, seeing partial or out of order updates.

As an example I'd be more confortable if it was a mutex.

Maybe nothing bad happens here, but we should set a conservative example.

cpp/examples/playback-ws-server/src/main.cpp
ericmlujan: Good catch, this was left over from when the channel lived outside the class

cpp/examples/playback-ws-server/src/main.cpp
ericmlujan: I decided to use `tick` to show that this is being invoked by an outer execution loop

eloff (approved):
LGTM

cpp/examples/playback-ws-server/src/main.cpp
```diff
@@ -0,0 +1,295 @@
...
+using namespace std::chrono_literals;
+
+// Example class for playing back a fixed interval of robot stack data.
+//
+// More practical implementations would load data from a file on disk, but for the sake of
+// illustration, this example generates a fixed buffer of example data in-memory.
+class DataPlayer {
+public:
+  DataPlayer(size_t num_timesteps, foxglove::RawChannel&& channel)
+      : channel_(std::move(channel)) {
```
clalancette, line 26: I'm going to suggest that we instantiate the `server_` in the constructor here, and set it to `nullptr` in the destructor.  That will both make it RAII, and also get rid of the checks for `if (!server_)` in a few other places in the code.

cpp/examples/playback-ws-server/src/main.cpp
```diff
@@ -0,0 +1,295 @@
...
+  }
+
+  void tick() {
+    if (!server_) {
+      throw std::runtime_error("Tried to tick with uninitialized server");
+    }
+
+    if (data_.empty()) {
+      throw std::runtime_error("Tried to tick with empty data");
+    }
```
clalancette, line 101: It might make sense to check for `num_timesteps == 0` in the constructor instead, then we could remove this check.

cpp/examples/playback-ws-server/src/main.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+    if (currentPlaybackState().status == foxglove::PlaybackStatus::Paused) {
+      std::this_thread::sleep_for(50ms);
```
clalancette, line 104: It is not immediately clear to me why 50ms was chosen here.  I'll suggest either adding a comment, or changing this to use `std::this_thread::sleep_for(1000ms / std::max<double>(playback_speed, 0.1));` so it is consistent.

cpp/examples/playback-ws-server/src/main.cpp
```diff
@@ -0,0 +1,295 @@
...
+    } else if (it->first > seek_time && it != data_.begin()) {
+      --it;
+    }
+
+    current_playback_index_ = static_cast<size_t>(std::distance(data_.begin(), it));
+    current_time_ = it->first;
+  }
+
+  // Gets the current playback state; assumes that the playback_mutex_ is locked.
+  [[nodiscard]] foxglove::PlaybackState currentPlaybackStateInternal() const {
```
clalancette, line 221: Just out of curiousity, why `[[nodiscard]]` here, and not on e.g. `seekInternal`?

clalancette (changes requested):
I've left some minor things to think about.

PR #771, opened 2025-11-21
Add agent C staticlib as optional dependency 

eloff:
### Changelog
None

### Docs

None

### Description

Requires: https://github.com/foxglove/data-platform/pull/2208

It uses cbindgen to generate Rust types for use the exported C functions and types from the C staticlib.

Currently to use this you must first build the header file and C staticlib using the agent branch above, and then copy them into the foxglove_agent crate, and rename the staticlib to end in the target triple for the platform you're using. The next task is to modify the CI build to automate the building of the library for each platform we want to support and publishing those binaries to a dedicated public repo for that purpose. Then we'll update the build script to download the files for the right platform automatically from that repo.

As an example, this uses the example foxglove_add function from the staticlib to add two integers in a test.

To run it you must run/build the tests with the --all-features flag.

I'm seeking feedback on the general approach, and the naming of things.

This is behind an agent_unstable flag, so we could merge it, but I would rather wait until it's more refined as it's not holding anything up either. We can base the next set of changes off this branch.


rust/foxglove_agent/include/foxglove_enterprise.h
```diff
@@ -0,0 +1,11 @@
+#include <cstdarg>
```
ericmlujan, line 1: I imagine this file is a leftover from an earlier version of this work?

rust/foxglove_agent/agent_sdk.h
```diff
@@ -0,0 +1,32 @@
+/*
```
ericmlujan, line 1: It feels a little odd for the C deps to live loose in the directory structure alongside Rust files; what do you think about adding them to a dedicated subdirectory?

rust/foxglove_agent/build.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    // The bindgen::Builder is the main entry point
+    // to bindgen, and lets you build up options for
+    // the resulting bindings.
+    let bindings = bindgen::Builder::default()
+        // The input header we would like to generate
+        // bindings for.
+        .header("wrapper.h")
+        .allowlist_function("foxglove_.*")
```
ericmlujan, line 64: Nit this (as well as some of the file paths) feel like good candidates for obvious constants at the top of the file

rust/foxglove_agent/build.rs
```diff
@@ -0,0 +1,77 @@
...
+            PathBuf::from(&out_dir).join(format!("libfoxglove_agent_sdk_{target}_localized.a"));
+
+        // Copy the library first
+        std::fs::copy(&original_lib, &localized_lib).expect("Failed to copy static library");
+
+        // Localize Rust std library symbols explicitly
+        // This approach only touches symbols we know conflict, leaving foxglove_* intact
+        if !Command::new("objcopy")
+            .arg("--wildcard")
+            .arg("--localize-symbol=!foxglove_*")
```
ericmlujan, line 28: I suppose this puts a limitation on us to name our functions/symbols with that particular prefix. That's managable (though should be documented/perhaps enforced on the other repo's end somehow). 

If that is indeed the case, we should consider implementing a [ParseCallback](https://docs.rs/bindgen/latest/bindgen/callbacks/trait.ParseCallbacks.html#method.generated_name_override) that strips the `foxglove_` prefix in the generated Rust bindings. It feels a bit redundant to have that leak from the C FFI to the native Rust crate where we typically don't do that kind of prefixing

defunctzombie:
Closing. We went a different direction.

PR #772, opened 2025-11-21
Update mcap stream example to support RangedPlayback capability

ericmlujan:
### Description

</td><td>

</td></tr></table>

### DIFF
--- a/rust/examples/ws_stream_mcap/src/main.rs
+++ b/rust/examples/ws_stream_mcap/src/main.rs
@@ -28,6 +28,7 @@ struct Inner {
     current_time: u64,
     pending_seek_time: Option<u64>,
     playback_speed: f32,
+    time_tracker: Option<TimeTracker>,
 }
 
 struct StreamMcapListener {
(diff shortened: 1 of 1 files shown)


rust/examples/ws_stream_mcap/src/main.rs
```diff
@@ -1,26 +1,134 @@
...
 };
-use mcap::records::{MessageHeader, Record, SchemaHeader};
-use mcap::sans_io::linear_reader::{LinearReadEvent, LinearReader, LinearReaderOptions};
+use mcap::records::MessageHeader;
+use mcap::sans_io::indexed_reader::{IndexedReadEvent, IndexedReader, IndexedReaderOptions};
+use mcap::sans_io::summary_reader::{SummaryReadEvent, SummaryReader, SummaryReaderOptions};
+use mcap::Summary as McapSummary;
 use tracing::info;
 
+struct StreamMcapListener {
```
eloff, line 34: When I see a lot of values atomically updated in the same struct, often it's better to put them in a Inner struct behind the same mutex. Lock it once, update all the values, unlock it.

Then it's a lot easier to reason about what state things could be in and keep values consistent relative to each other.

Might that apply here?

ericmlujan: For sure! I think I made that change in the C++ example, but didn't do it over in the corresponding Rust example. Will make a change now

PR #778, opened 2025-11-28
rust: add McapStream writer

bennetthardwick:
### Changelog

rust: add an MCAP writer that will write to a `futures::Stream`

### Docs

None

### Description

This change adds an MCAP writer that can write to an async stream. This makes it easier to return a streaming MCAP HTTP response from something like axum, using the Foxglove SDK.

Fixes FIRE-224.

---

> [!NOTE]
> Introduce `stream` feature with `create_mcap_stream`, `McapStreamHandle`, and `McapStream` to produce a futures::Stream of MCAP bytes for HTTP streaming.
> 
> - **New MCAP streaming API** (feature-gated via `stream`):
>   - Add `rust/foxglove/src/stream.rs` with `create_mcap_stream`, `McapStreamHandle`, and `McapStream` implementing a `futures::Stream` of MCAP bytes using a shared buffer and tokio mpsc; supports `flush`, `close`, and buffer sizing; includes unit test.
> - **Public exports**:
>   - Gate `pub mod stream` behind `cfg(feature = "stream")` in `rust/foxglove/src/lib.rs`.
> - **Features/Deps**:
>   - Add `stream` feature in `Cargo.toml` and optional `futures` dependency.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9babb8941d6096d4f3392eb26b3ccb6fa6ac30a4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/python/foxglove-sdk-examples/write-mcap-file/main.py
+++ b/python/foxglove-sdk-examples/write-mcap-file/main.py
@@ -1,5 +1,6 @@
 import argparse
 import inspect
+import time
 
 import foxglove
 from foxglove.channels import LogChannel
@@ -18,6 +19,17 @@ def main() -> None:
     with foxglove.open_mcap(args.path) as writer:
         # If you want to add some MCAP metadata: https://mcap.dev/spec#metadata-op0x0c
         writer.write_metadata("platform", {"os": "linux", "arch": "x64"})
+
+        # If you want to attach arbitrary files to the recording
+        # Common uses include configuration files, calibration data, or other reference material
+        now_ns = time.time_ns()
+        writer.attach(
+            log_time=now_ns,
+            create_time=now_ns,
+            name="config.json",
+            media_type="application/json",
+            data=b'{"robot_name": "example_robot", "version": "1.0"}',
+        )
 
         for i in range(10):
             frame = inspect.currentframe()
--- a/python/foxglove-sdk/python/foxglove/_foxglove_py/mcap.pyi
+++ b/python/foxglove-sdk/python/foxglove/_foxglove_py/mcap.pyi
@@ -94,3 +94,27 @@ class MCAPWriter:
         :param metadata: Dictionary of key-value pairs to store
         """
         ...
+
+    def attach(
+        self,
+        *,
+        log_time: int,
+        create_time: int,
+        name: str,
+        media_type: str,
+        data: bytes,
+    ) -> None:
+        """
+        Write an attachment to the MCAP file.
+
+        Attachments are arbitrary binary data that can be stored alongside messages.
+        Common uses include storing configuration files, calibration data, or other
+        reference material related to the recording.
+
+        :param log_time: Time at which the attachment was logged, in nanoseconds since epoch.
+        :param create_time: Time at which the attachment data was created, in nanoseconds since epoch.
+        :param name: Name of the attachment (e.g., "config.json").
+        :param media_type: MIME type of the attachment (e.g., "application/json").
+        :param data: Binary content of the attachment.
+        """
+        ...
--- a/python/foxglove-sdk/python/foxglove/tests/test_mcap.py
+++ b/python/foxglove-sdk/python/foxglove/tests/test_mcap.py
@@ -146,6 +146,51 @@ def _verify_metadata_in_file(file_path: Path, expected_metadata: dict) -> None:
             ), f"Metadata '{name}' has wrong key-value pairs"
 
 
+def _verify_attachments_in_file(
+    file_path: Path, expected_attachments: list[dict]
+) -> None:
+    """Helper function to verify attachments in MCAP file match expected."""
+    import mcap.reader
+
+    with open(file_path, "rb") as f:
+        reader = mcap.reader.make_reader(f)
+
+        found_attachments = []
+        for attachment in reader.iter_attachments():
+            found_attachments.append(
+                {
+                    "log_time": attachment.log_time,
+                    "create_time": attachment.create_time,
+                    "name": attachment.name,
+                    "media_type": attachment.media_type,
+                    "data": attachment.data,
+                }
+            )
+
+        # Verify count
+        assert len(found_attachments) == len(
+            expected_attachments
+        ), f"Expected {len(expected_attachments)} attachments, found {len(found_attachments)}"
+
+        # Verify each attachment matches expected
+        for expected in expected_attachments:
+            matching = [a for a in found_attachments if a["name"] == expected["name"]]
+            assert len(matching) == 1, f"Attachment '{expected['name']}' not found"
+            actual = matching[0]
+            assert (
+                actual["log_time"] == expected["log_time"]
+            ), f"Attachment '{expected['name']}' has wrong log_time"
+            assert (
+                actual["create_time"] == expected["create_time"]
+            ), f"Attachment '{expected['name']}' has wrong create_time"
+            assert (
+                actual["media_type"] == expected["media_type"]
+            ), f"Attachment '{expected['name']}' has wrong media_type"
+            assert (
+                actual["data"] == expected["data"]
+            ), f"Attachment '{expected['name']}' has wrong data"
+
+
 def test_write_metadata(tmp_mcap: Path) -> None:
     """Test writing metadata to MCAP file."""
     # Define expected metadata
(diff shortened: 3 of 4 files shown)


rust/foxglove/src/mcap_writer/mcap_stream.rs
```diff
@@ -0,0 +1,298 @@
...
+use bytes::{Bytes, BytesMut};
+use futures::{ready, Stream};
+use parking_lot::Mutex;
+
+use crate::{Context, FoxgloveError, McapWriteOptions, McapWriter, McapWriterHandle};
+
+#[derive(Default)]
+struct Inner {
+    buffer: BytesMut,
+    position: u64,
```
eloff, line 25: Since this is a memory buffer you could also use usize here. Instead of converting on store, you'd convert on load where needed.

Either way is fine, no need to change this unless you can see a good reason to do so.

rust/foxglove/src/mcap_writer/mcap_stream.rs
```diff
@@ -0,0 +1,298 @@
...
+    }
+}
+
+impl Seek for SharedBuffer {
+    fn seek(&mut self, pos: SeekFrom) -> std::io::Result<u64> {
+        let inner = self.0.lock();
+        match pos {
+            SeekFrom::Start(n) if inner.position == n => Ok(n),
+            SeekFrom::Current(0) => Ok(inner.position),
+            _ => Err(std::io::Error::other("seek on unseekable file")),
```
eloff, line 56: We could support End(i64) here, for negative values. Maybe an error or noop for seeking beyond end of the buffer.

rust/foxglove/src/mcap_writer/mcap_stream.rs
```diff
@@ -0,0 +1,298 @@
+use std::{
+    fmt::Debug,
+    io::{self, Seek, SeekFrom, Write},
+    pin::Pin,
+    sync::Arc,
+    task::Poll,
+};
+
+use tokio::sync::mpsc::{error::SendError, Receiver as TokioReceiver, Sender as TokioSender};
```
eloff: flume crate is also an option, we use that in the sdk already. The reason that might be interesting is flume queues can be used from async and sync code, as it has both blocking and async methods.

Not sure if that's helpful here, but mentioning it.

eloff (commented):
There is another way I could imagine to do this:

The mcap writer itself is also buffering, and it's only blocking when it dumps its buffer by calling finish_chunk() internally.

This solution works, but it's effectively double-buffering. One could modify the mcap writer in the mcap crate, either so it buffers forever like this until an explicit flush, or it swaps the internal buffer wit a fresh one and does the flush on a background thread when it reaches the threshold instead of blocking.

What are your thoughts about that?

bennetthardwick:
> This solution works, but it's effectively double-buffering. One could modify the mcap writer in the mcap crate, either so it buffers forever like this until an explicit flush, or it swaps the internal buffer wit a fresh one and does the flush on a background thread when it reaches the threshold instead of blocking.

@eloff I don't think it's possible to get rid of double buffering entirely because the writer needs to seek back to the beginning of a chunk to write the length before being flushed. I don't think this is a big deal because the space is reused internally and shouldn't really allocate that frequently. When these chunks are flushed there will be a single realloc instead of many to grow the output buffer.

One of the big issues I was dealing with is how do we manage back pressure between the output `McapStream` reading bytes and the `McapWriter` providing them. I wanted to make it possible for the user to be able to wait for the `McapStream` to have capacity before future messages are written as the buffer would grow indefinitely. It also needs to be able to wait with `.await` because synchronously blocking on writes to a channel would likely break user code as they'd likely be using tokio.

eloff:
> > This solution works, but it's effectively double-buffering. One could modify the mcap writer in the mcap crate, either so it buffers forever like this until an explicit flush, or it swaps the internal buffer wit a fresh one and does the flush on a background thread when it reaches the threshold instead of blocking.
> 
> @eloff I don't think it's possible to get rid of double buffering entirely because the writer needs to seek back to the beginning of a chunk to write the length before being flushed. I don't think this is a big deal because the space is reused internally and shouldn't really allocate that frequently. When these chunks are flushed there will be a single realloc instead of many to grow the output buffer.

Maybe I'm missing something here. Doesn't the mcap writer currently buffer that way before writing a chunk? If you thought of that as playing the role of your SharedBuffer here letting you log to memory, and then finish_chunk as flush here, you really just need a way to expose finish_chunk() as a an pub async flush method and you have a similar setup to this, but with only one level of buffering?

> 
> One of the big issues I was dealing with is how do we manage back pressure between the output `McapStream` reading bytes and the `McapWriter` providing them. I wanted to make it possible for the user to be able to wait for the `McapStream` to have capacity before future messages are written as the buffer would grow indefinitely. It also needs to be able to wait with `.await` because synchronously blocking on writes to a channel would likely break user code as they'd likely be using tokio.

You've hit on the really thorny problem. We felt this acutely with the WebSocketServer because the TCP stream can be a lot lower throughput than the filesystem.

Ideally to provide back pressure, you want an async log method, which we don't provide in the SDK. It's a bit wierd, no logging library in Rust gives you an async log method. It's also not great if you suspend the client process in an await to log something because a buffer is full. On the other hand, the other options are buffer forever, and get OOM killed if you keep over-running the system, or you drop messages. We opted for the latter in the WebSocketServer, with a fixed length queue and dropping the oldest message. It's not great for bursty workloads, and it's really not great for something like an mcap file where you don't want an mcap file with arbitrary logs missing.

It's really a hard tradeoff, three bad options, you must pick one.

rust/foxglove/src/mcap_writer/mcap_stream.rs
```diff
@@ -0,0 +1,298 @@
...
+///
+/// Logged messages are buffered in memory until a call to [`McapStreamHandle::flush`] is called.
+/// If this method is not called messages will buffer indefinitely.
+#[must_use]
+#[derive(Debug)]
+pub struct McapStreamBuilder {
+    writer: McapWriter,
+}
+
+impl McapStreamBuilder {
```
gasmith: Could expose this functionality as part of the existing `McapWriter` builder, rather than wrapping it, so that we don't have to duplicate that interface? Consider:

```rust
let (stream_handle, stream) = McapWriter::with_options(options)
    .context(context)
    .channel_filter(filter)
    .create_stream()?;
```

I think it would be reasonable to forcibly set `disable_seeking` as part of `create_stream()`.

rust/foxglove/src/mcap_writer/mcap_stream.rs
```diff
@@ -0,0 +1,298 @@
...
+            inner.buffer.split()
+        };
+
+        if bytes.is_empty() {
+            return Ok(());
+        }
+
+        if let Err(SendError(bytes)) = sender.send(bytes).await {
+            let mut inner = buffer.0.lock();
+            inner.buffer.unsplit(bytes);
```
gasmith: Is this safe? What if `inner.buffer` was written to while the lock was dropped? If it was, we need to insert `bytes` at the beginning of the buffer.

On the other hand, I suppose the data is unrecoverable once the receiver is gone. Maybe just drop `bytes` instead of trying to push it back into the buffer?

bennetthardwick:
> Do we ever plan to expose this functionality in C++ or Python?

@gasmith I don't think so as from @james-rms it sounds like most other ecosystems use a writer interface for stream responses - so these would be supported by default.

eloff:
I've been thinking about this and https://github.com/foxglove/foxglove-sdk/pull/782 which are two different solutions for a similar problem - at least on the surface.

I think the Stream of Bytes as the consumer is very specific to wanting to stream the results out from an Axum handler, and not much more broadly applicable - most other SDK users would then have to write code to forward that into a file.

It also requires the user to manually flush in an async context periodically. I'm not sure how many of our customers are using async Rust, we would need to support both.

Perhaps we're trying too hard to fit this use case into the log() and Sink API. We don't need to also log these messages to other sinks like a WebSocketServer at the same time. You could get better performance and lose the mutex acquire/release on each log call by implementing a dedicated wrapper around the mcap library writer. You could also have an async log method that handles the buffering and flushing transparently if you did that.

Something like:

```rust
// in an AsyncMcapWriter or something

pub async fn log<T: Encode>(&mut self, channel: &Channel<T>, msg: &T) -> Result<(), FoxgloveError> {
    todo!("call log_with_time with current time")
}

pub async fn log_with_time<T: Encode>(&mut self, channel: &Channel<T>, msg: &T, timestamp: impl ToUnixNanos) -> Result<(), FoxgloveError> {
    let mut buf: SmallBytes<STACK_BUFFER_SIZE> = SmallBytes::new();
    if let Some(estimated_size) = msg.encoded_len() {
        buf.reserve(estimated_size);
    }

    msg.encode(&mut buf).unwrap();
    
    todo("log buf to mcap writer");
}
```

You only really benefit from the Encode implementation for the Foxglove schemas in this case, but that's the important part that you were most interested in for this use case. You'd also have to add log_raw methods for RawChannels. This code could still live in the SDK. It gives you a lot more control over what the implementation does.

I think we also need to solve this problem for log() with the Sink interface, but the requirements are somewhat different, and we don't need the Stream of Bytes there. I'm not sure if it's better to move the writing and blocking to a background thread like Stephen's work trial approach, or buffer the mcap writer output like this in chunks and write that in a background thread more like this approach. I think we'd have to try both and measure them and take our time to get that right. But that doesn't have to block your work here.

What do you think?

rust/foxglove/src/mcap_writer/mcap_stream.rs
bennetthardwick: Good call, removed!

rust/foxglove/src/mcap_writer/mcap_stream.rs
bennetthardwick: Modified to separate this out into it's own separate. You now call the free function `create_mcap_stream` instead of going via any other known interface.

bennetthardwick:
As discussed I've modified this to not be a member on `Context` or `McapWriter`. It's now in a module called `stream` and gated by a feature flag. You have to call `create_mcap_stream()` which will return the handle and stream pair.

I felt like the `channel.log` and `flush` combo was the one with the fewest footguns (instead of having a log message that passed channels). You now create channels with `handle.channel_builder(..)` which handles the context for you.

rust/foxglove/src/stream.rs
```diff
@@ -0,0 +1,266 @@
...
+                channel.log(&Message { data: i as f64 });
+                handle.flush().await.unwrap();
+            }
+
+            handle.close().await.unwrap();
+        });
+
+        let mut out = vec![];
+
+        while let Some(bytes) = stream.next().await {
```
james-rms, line 260: I think this test serves as useful example code and needs comments to explain to readers exactly what's happening and where you'd expect to use this stream. Maybe explain that the stream can be used as an axum response body.

james-rms (approved):

rust/foxglove/src/mcap_writer/mcap_stream.rs
bennetthardwick: The position is a "global" position of the stream and not an index into the memory buffer so this could technically be above a usize in some cases.

rust/foxglove/src/mcap_writer/mcap_stream.rs
bennetthardwick: This was inspired by `binrw::NoSeek` which doesn't include `SeekFrom::End(0)`. I guess it's not super clear what it should be returning in that case?

PR #791, opened 2025-12-10
MCAP attachments

sanderpick:
### Changelog

Add `McapWriterHandle::attach()` method and `McapAttachment` type to enable writing attachments to MCAP files via the Rust SDK.

### Docs

Is there a docs repo?

### Description

This PR exposes the mcap crate's attachment functionality through the foxglove SDK, allowing users to store arbitrary binary data (configuration files, calibration data, images, etc.) alongside messages in MCAP recordings.

#### Changes
- Added `McapAttachment` type re-export from the mcap crate
- Added `McapSink::attach()` internal implementation
- Added `McapWriterHandle::attach()` public API with documentation and example
- Added comprehensive tests for attachment functionality

#### Usage example
```rust
use std::borrow::Cow;
use foxglove::{McapWriter, McapAttachment};

let mcap = McapWriter::new()
    .create_new_buffered_file("test.mcap")
    .expect("create failed");

mcap.attach(&McapAttachment {
    log_time: 0,
    create_time: 0,
    name: "config.json".to_string(),
    media_type: "application/json".to_string(),
    data: Cow::Borrowed(br#"{"setting": true}"#),
}).expect("attach failed");

mcap.close().expect("close failed");
```

#### Testing
Added unit tests for basic attachment, multiple attachments, and error handling when attaching after close. All 212 tests pass with `--features unstable`. Manual testing works:
```
$ mcap info test.mcap 
library:   foxglove-sdk-rust/v0.0.0                                    
profile:                                                               
messages:  3204                                                        
duration:  1m22.913993797s                                             
start:     2025-12-10T13:34:08.054447689-07:00 (1765398848.054447689)  
end:       2025-12-10T13:35:30.968441486-07:00 (1765398930.968441486)  
compression:
	lz4: [732/732 chunks] [1.36 GiB/968.95 MiB (30.44%)] [11.69 MiB/sec] 
chunks:
	max uncompressed size: 1.91 MiB
	max compressed size: 1.34 MiB
	overlaps: no
channels:
	[redacted]
channels: 5
attachments: 1
metadata: 1
```

---

> [!NOTE]
> Adds support for writing MCAP attachments via `McapWriterHandle::attach` and re-exports `McapAttachment`, with sink implementation and tests.
> 
> - **Public API**
>   - Re-export `McapAttachment` from `mcap` in `rust/foxglove/src/lib.rs` and `rust/foxglove/src/mcap_writer.rs`.
[...]

PR #795, opened 2025-12-16
add MCAP attachment support for C++

eloff:
### Changelog

Add C++ support for MCAP attachment writing.

### Docs

None

### Description

Builds on the Rust MCAP attachment support added in https://github.com/foxglove/foxglove-sdk/pull/791

Wraps it into C++, following the lead of how we did it for metadata, and adds tests for it.

---

> [!NOTE]
> Adds MCAP attachment writing via new C API and Rust impl, with C++ wrapper method and tests.
> 
> - **C API (`c/include/foxglove-c/foxglove-c.h`)**:
>   - Add `struct foxglove_mcap_attachment` and `foxglove_mcap_attach` to write attachments (non-WASM).
>   - Update cbindgen mappings (`c/cbindgen*.toml`).
> - **Rust FFI (`c/src/channel.rs`)**:
>   - Implement `FoxgloveMcapAttachment` and `foxglove_mcap_attach`, including UTF-8 validation and data handling.
>   - Extend `McapWriterVariant` with `attach` passthrough.
> - **C++ SDK (`cpp/foxglove/include/foxglove/mcap.hpp`, `cpp/foxglove/src/mcap.cpp`)**:
>   - Introduce `foxglove::Attachment` and `McapWriter::attach(const Attachment&)` calling C API.
>   - Declare C bindings for `foxglove_mcap_attach`.
> - **Tests (`cpp/foxglove/tests/test_mcap.cpp`)**:
>   - Add tests for single/multiple/empty attachments and basic verification in written MCAP.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 07be51d04f8e4d061d0fa59cc56ed85bb437cfdd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/cpp/foxglove/include/foxglove/mcap.hpp
+++ b/cpp/foxglove/include/foxglove/mcap.hpp
@@ -85,7 +85,7 @@ struct Attachment {
   /// @brief Media type of the attachment, e.g. "application/json".
   std::string_view media_type;
   /// @brief Pointer to the attachment data.
-  const uint8_t* data = nullptr;
+  const std::byte* data = nullptr;
   /// @brief Length of the attachment data in bytes.
   size_t data_len = 0;
 };
--- a/cpp/foxglove/tests/test_mcap.cpp
+++ b/cpp/foxglove/tests/test_mcap.cpp
@@ -650,7 +650,7 @@ TEST_CASE("Write single attachment to MCAP") {
   attachment.create_time = 900000000;
   attachment.name = "config.json";
   attachment.media_type = "application/json";
-  attachment.data = reinterpret_cast<const uint8_t*>(attachment_data.data());
+  attachment.data = reinterpret_cast<const std::byte*>(attachment_data.data());
   attachment.data_len = attachment_data.size();
 
   auto error = writer->attach(attachment);
@@ -684,7 +684,7 @@ TEST_CASE("Write multiple attachments to MCAP") {
   config_attachment.create_time = 900000000;
   config_attachment.name = "config.yaml";
   config_attachment.media_type = "application/yaml";
-  config_attachment.data = reinterpret_cast<const uint8_t*>(config_data.data());
+  config_attachment.data = reinterpret_cast<const std::byte*>(config_data.data());
   config_attachment.data_len = config_data.size();
 
   auto error1 = writer->attach(config_attachment);
@@ -697,7 +697,7 @@ TEST_CASE("Write multiple attachments to MCAP") {
   calibration_attachment.create_time = 1800000000;
   calibration_attachment.name = "calibration.bin";
   calibration_attachment.media_type = "application/octet-stream";
-  calibration_attachment.data = reinterpret_cast<const uint8_t*>(calibration_data.data());
+  calibration_attachment.data = reinterpret_cast<const std::byte*>(calibration_data.data());
   calibration_attachment.data_len = calibration_data.size();
 
   auto error2 = writer->attach(calibration_attachment);


cpp/foxglove/include/foxglove/mcap.hpp
```diff
@@ -64,6 +69,27 @@ enum class McapCompression : uint8_t {
...
+  uint64_t log_time = 0;
+  /// @brief Timestamp at which the attachment was created, in nanoseconds since epoch.
+  /// If not available, set to 0.
+  uint64_t create_time = 0;
+  /// @brief Name of the attachment, e.g. "config.json".
+  std::string_view name;
+  /// @brief Media type of the attachment, e.g. "application/json".
+  std::string_view media_type;
+  /// @brief Pointer to the attachment data.
+  const std::byte* data = nullptr;
```
clalancette: This is `const uint8_t *` basically everywhere else.  If we made it a `const uint8_t *` here, we could get rid of the `reinterpret_cast` below.  Any particular reason to use `const std::byte *` here?

c/src/channel.rs
```diff
@@ -410,6 +439,69 @@ unsafe fn do_foxglove_mcap_write_metadata(
...
+
+    let Some(writer_handle) = writer.0.as_ref() else {
+        return Err(foxglove::FoxgloveError::SinkClosed);
+    };
+
+    let data = if attachment.data.is_null() || attachment.data_len == 0 {
+        &[]
+    } else {
+        unsafe { std::slice::from_raw_parts(attachment.data, attachment.data_len) }
+    };
```
clalancette, line 501: This is minor, but what if `attachment.data != NULL` and `attachment.data_len == 0`, or `attachment.data == NULL` and `attachment.data_len != 0` ?  In both cases, no data will be copied (so we'll avoid UB), but they are both nonsensical operations.  Should we explicitly return an error for these cases?

cpp/foxglove/include/foxglove/mcap.hpp
eloff: Good point

c/src/channel.rs
eloff: I think elsewhere we follow this behavior instead of returning an error, but it's an interesting point.

We could log an error, or do that and return an error code, or keep the current behavior of ignoring it. My preference is treat it as an error, but I don't feel strongly about it.

clalancette: I see what you mean about following this behavior elsewhere.

I don't feel strongly about it either way.  I also prefer to treat this as an error (much easier to debug).  If you do want to make this an error, I think a way to avoid the Cursor error is to do something like this:

```
let data = if attachment.data.is_null() {
    if attachment.data_len != 0 {
        return Err(foxglove::FoxgloveError::ValueError(
            "attachment data is null but data_len is not 0".to_string(),
        ));
    }
    &[]
} else {
    if attachment.data_len == 0 {
        return Err(foxglove::FoxgloveError::ValueError(
            "attachment data is not null but data_len is 0".to_string(),
        ));
    }
    unsafe { std::slice::from_raw_parts(attachment.data, attachment.data_len) }
};
```

cpp/foxglove/include/foxglove/mcap.hpp
eloff: Actually it seems there's a lot (more) std::byte usage too. 

std::byte is a little more opaque (requires casts to perform arithmetic) but that doesn't much matter one way or another. uint8_t is problematic sometimes because it's UB to cast uint8_t to an arbitrary type. std::byte, unsigned char are allowed to be cast like that.

I'm tempted to change this back to std::byte fit the other code better.

Slack thread for context: https://foxglove.slack.com/archives/C04R1TNP2TB/p1756332887723899

clalancette: For what it is worth, I don't see anything in that thread that suggests it is UB to cast uint8_t to arbitrary types.  In point of fact, the linked [Chromium thread](https://groups.google.com/a/chromium.org/g/cxx/c/bBY0gZa1Otk) suggests the opposite.  Also, the [Stackoverflow thread](https://stackoverflow.com/questions/77097673/a-byte-type-stdbyte-vs-stduint8-t-vs-unsigned-char-vs-char-vs-stdbitset8) states that typeof(uint8_t) === typeof(std::byte) === unsigned char for all common platforms.  The only downside to uint8_t is that it does not necessarily exist on all platforms, but I don't think that applies to anywhere this would run.

All of that said, I really don't feel strongly about it.  My preference is always to reduce casts, and the best way in my experience to reduce casts is to make sure the types are consistent all the way through the stack.  But if we have a preference for using `std::byte`, then we can add that back with casts.

clalancette (approved):
I'm going to approve this, as I don't really care which way we go with respect to `uint8_t` vs. `std::byte`.

cpp/foxglove/include/foxglove/mcap.hpp
eloff: I read differently from the C++ thread, see: https://groups.google.com/a/chromium.org/g/cxx/c/bBY0gZa1Otk/m/IE7i5xtqAAAJ and https://groups.google.com/a/chromium.org/g/cxx/c/bBY0gZa1Otk/m/NXgxRpJsAAAJ.

Since we support C++ 17 as the minimum, we can choose std::byte here. I share your preference for keeping it simpler and avoiding casts, and I think in the real world uint8_t will just work. But it does seem to be more problematic with respect to UB and the language spec. Even though it's certainly all unsigned char on any platform we care about.

Given we already seem to prefer std::byte (and changing all that to uint8_t is a breaking change) I think we should just double down on it rather than mix types or try to convert it all to uint8_t.

PR #796, opened 2025-12-20
playground: add set_layout method

jtbandes:
### Changelog
The [playground](https://playground.foxglove.dev) now offers a `playground.set_layout()` method.

### Docs

Included

### Description

This `playground` module does not exist in the SDK, it's just injected at runtime.

Example:

```py
from foxglove.layouts import *
playground.set_layout(Layout(content=RawMessagesPanel()))
```

### DIFF
--- a/playground/src/RunnerWorker.ts
+++ b/playground/src/RunnerWorker.ts
@@ -53,39 +53,20 @@ export class RunnerWorker {
   };
   constructor() {
     this.#pyodide = this.#setup();
+    // Inject this at the beginning of scripts to allow jedi completions to understand the
+    // playground module already available as a global. The jedi.Interpreter class might be able to help
+    // with this, but it breaks completion for * imports:
+    // https://github.com/davidhalter/jedi/issues/2087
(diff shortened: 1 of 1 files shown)


jtbandes:
cc @gasmith @surbina  -- this is not ready for prime time (since the new APIs have not been released), but let me know if you think this would be a good future direction

defunctzombie:
Lots of feedback in the slack thread: https://foxglove.slack.com/archives/C0469RNBPEW/p1769728084840319

The only thing I felt was a _blocker_ is that there is this rouge import of "playground" that is meaningless to the sdk user. If there's a way to have this be a special "global" available in the playground environment that seems better to me to contain this away from "sdk".

`from foxglove import playground`

Everything else from there is about how we iterate on the sdk notebook layout apis.

jtbandes:
Seems like the mechanism I used to provide completions for the `playground` module has broken completions for `import *`... https://github.com/davidhalter/jedi/issues/2087

PR #799, opened 2026-01-07
Move protocol types to `protocol` directory

ericmlujan:
### Changelog
None

### Docs
None

### Description
This PR exists to set up the structure for iterating on future versions of the Foxglove protocol independently of the v1 protocol currently used in the WebSocket implementation, without making breaking API changes while in development. All changes should either be moving existing logic or renaming messages specific to V1;  no message definition changes or new messages are introduced in this PR.

It contains the following changes:

1. Move Foxglove protocol message definitions from `websocket::ws_protocol` to a new top-level `protocol` module. Facilities and message definitions shared across protocol versions will live in the `protocol::common` module.
2. Move `Subscribe`, `Unsubscribe`, and `MessageData` to their respective `client` and `server` modules  in a new `protocol::v1` module, and rename them to add a `V1` suffix. The `ServerMessage` and `ClientMessage` enums are similarly renamed to `ServerMessageV1` and `ClientMessageV1` to reflect that they are refer to the set of message definitions specific to protocol v1.
3. Update `websocket::ws_protocol` to provide re-exports and renaming of types that now live in `protocol` for backwards-compatibility. **All business logic in the `websocket` module still uses these re-exports for now to keep the diff size sane; we will mechanically update them to directly reference the `protocol` module later.**
4. Update top-level module and public exports in `lib.rs`.
5. Re-generate snapshots in their new subdirectories.

Fixes: FLE-164

---

> [!NOTE]
> Establishes a new `protocol` module for Foxglove protocol types while preserving backward compatibility.
> 
> - Introduces `protocol::{common, v1}` and migrates client/server messages (JSON/Binary, schema, parameters, tungstenite) from `websocket::ws_protocol` into these modules
> - Updates `websocket::ws_protocol` to re-export from `protocol::v1` for non-breaking API usage; adjusts imports in code (e.g., `cloud_sink` test) and enables `pub mod protocol` in `lib.rs`
> - Regenerates and relocates snapshots to new paths; no functional changes to message formats or behavior
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b1ae3decc6012c0492c0359af61413a813f71b4c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/rust/foxglove/src/lib.rs
+++ b/rust/foxglove/src/lib.rs
@@ -358,7 +358,7 @@ pub(crate) use time::nanoseconds_since_epoch;
 #[cfg(feature = "agent")]
 mod cloud_sink;
 #[cfg(feature = "live_visualization")]
-pub mod protocol;
+mod protocol;
 #[cfg(feature = "live_visualization")]
 mod runtime;
 #[cfg(feature = "live_visualization")]


rust/foxglove/src/protocol/common/client/unadvertise.rs
```diff
@@ -25,7 +25,7 @@ impl JsonMessage for Unadvertise {}
 
 #[cfg(test)]
 mod tests {
-    use crate::websocket::ws_protocol::client::ClientMessage;
+    use crate::protocol::v1::client::ClientMessageV1;
```
ericmlujan: Note to reviewer: I'm aware that it'd be ideal to decouple the tests in `common` from a dependency on `ClientMessageV1`. I opted not do to that in this PR to keep the diff as mechanical as possible and easy-ish to review, but if you'd prefer I tackle that now, lmk.

gasmith: It was the first thing that went through my head as I started reading. I'm fine leaving it as it is for now.

rust/foxglove/src/protocol/v1/client.rs
```diff
@@ -0,0 +1,140 @@
...
+pub use unsubscribe::UnsubscribeV1;
+
+/// A representation of a client message useful for deserializing.
+#[derive(Debug, Clone, PartialEq)]
+#[allow(missing_docs)]
+pub enum ClientMessageV1<'a> {
+    Subscribe(SubscribeV1),
+    Unsubscribe(UnsubscribeV1),
+    Advertise(Advertise<'a>),
+    Unadvertise(Unadvertise),
```
gasmith: As I read through, I'm starting to dislike the V1 suffix. I feel like there should be a `protocol::v1::client::Advertise`, regardless of whether it is a re-export of `protocol::common::client::Advertise` or a distinct struct from `protocol::v2::client::Advertise`. If we decide in the future we need to change that message type for v2, it'd be nice for all the v1-specific code to remain untouched. I guess we get that property by virtue of the aliased re-exports in ws_protocol.rs, but I'd be curious to hear your take on this.

rust/foxglove/src/lib.rs
```diff
@@ -367,6 +369,9 @@ mod websocket_client;
 mod websocket_server;
 #[cfg(feature = "agent")]
 pub use cloud_sink::{CloudSink, CloudSinkHandle, CloudSinkListener};
+#[doc(hidden)]
+#[cfg(feature = "live_visualization")]
+pub use protocol::common;
```
gasmith: This export (`foxglove::common`) definitely doesn't belong here.

rust/foxglove/src/protocol.rs
```diff
@@ -0,0 +1,9 @@
+//! Implementation of the Foxglove protocol
+
+#[doc(hidden)]
+pub mod common;
```
gasmith: I don't think we should expose the common module publicly. Code outside of this crate should always be written against either v1 or v2 explicitly. A message that's common today might be version-specific in the future.

gasmith (commented):
LG in general, though I'm still mulling over naming and public interface just a bit.

rust/foxglove/src/protocol.rs
ericmlujan: Agreed, and that's an excellent point. Users should explicitly target a version. I resolved that in fcd6cbc858de63d55fec2e0988f9b580940c089d.

rust/foxglove/src/lib.rs
ericmlujan: Addressed in a816d03d353ccddd1076a19be170fc96ced491ad.

rust/foxglove/src/protocol/v1/client.rs
ericmlujan: Addressed in 3ce586f0d890c1933ae3c341c06bac6151200317.

ericmlujan:
@gasmith thanks for the review; this should be good for another pass by you when you get a chance 😄

rust/foxglove/src/lib.rs
```diff
@@ -358,6 +358,8 @@ pub(crate) use time::nanoseconds_since_epoch;
 #[cfg(feature = "agent")]
 mod cloud_sink;
 #[cfg(feature = "live_visualization")]
+pub mod protocol;
```
gasmith: Does this need to be exported? Feels like the answer is no, since we re-export the parts we need from the websocket module.

ericmlujan: Technically not, and we can keep internal for now

PR #798, opened 2026-01-06
[python] Support opening MCAP writers on file-like objects

Muon:
### Changelog
* Python: extend `open_mcap()` to also accept file-like objects. 

### Docs
None

### Description

This PR extends `open_mcap()` to accept file-like objects in addition to file paths, enabling MCAP recording to any Python object implementing `write()`, `seek()`, and `flush()`. This allows using in-memory buffers, network streams, or similar as MCAP writing sinks. 

Closes: FIRE-248

---

> [!NOTE]
> Adds file-like object support for MCAP writing in Python.
> 
> - Extends `open_mcap(path, ...)` to accept `str | Path | BinaryIO`; updates stubs and docstrings in `python/foxglove/_foxglove_py/__init__.pyi`
> - Implements `PyFileLikeWriter` and `WriterInner` in Rust (`src/mcap.rs`) to adapt Python objects with `write/seek/flush` to `Write + Seek`; updates `PyMcapWriter` to use `BufWriter<WriterInner>`
> - Updates `open_mcap` in `src/lib.rs` to handle either filesystem paths (with `allow_overwrite`) or file-like objects (ignoring overwrite)
> - Adds tests in `python/foxglove/tests/test_mcap.py` for writing to `BytesIO`, validating MCAP magic bytes, and reading back via `mcap.reader`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1e7f13b31cef6989b20283fbf3e2a4499d9898cc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #800, opened 2026-01-07
[python] Add support for `disable_seeking` MCAP writer option

Muon:
### Changelog
* python: add support for the `disable_seeking` MCAP writer option, allowing writing to non-seekable streams (e.g. network sockets).
* python: add `McapWritable` protocol describing minimal IO needed for writing MCAP. 

### Docs
None

### Description

#798 added support for opening an MCAP writer on file-like objects. However, an important use case is writing to network streams (e.g. HTTP or WebSocket connections). This PR allows for this in Python by exposing the  `disable_seeking` option from Rust. Similarly to Rust, the file-like object must still have a `seek` method, except it only needs to support `seek(0, SEEK_CUR)` and `seek(current_position, SEEK_SET)`.

This PR also loosens the signature of `open_mcap` by allowing it to take an `McapWritable`, a newly added Python protocol which only requires `write`, `seek` and `flush`. This is strictly more general than `BinaryIO`, but `BinaryIO` is retained in the signature for documentation clarity.

Fixes: FIRE-247

### DIFF
--- a/python/foxglove-sdk/python/foxglove/_foxglove_py/__init__.pyi
+++ b/python/foxglove-sdk/python/foxglove/_foxglove_py/__init__.pyi
@@ -1,7 +1,25 @@
 from pathlib import Path
-from typing import Any, BinaryIO, Callable
+from typing import Any, BinaryIO, Callable, Protocol
 
 from foxglove.websocket import AssetHandler
+
+class McapWritable(Protocol):
+    """A writable and seekable file-like object.
+
+    This protocol defines the minimal interface required for writing MCAP data.
+    """
+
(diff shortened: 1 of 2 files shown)


james-rms (approved):

Muon:
@james-rms @gasmith I've made a signature change to `open_mcap` to deal with the linter/type errors. `BinaryIO` has a bunch of extra methods that are annoying to implement, so I've added an `McapWritable` protocol, which is strictly more general than `BinaryIO`. I think we should keep `BinaryIO` in the signature of `open_mcap` though, so it's still clear we support arbitrary file-like objects.

PR #803, opened 2026-01-07
Update foxglove_bridge to use new ament_index API

ericmlujan:
### Changelog

None.

### Docs

None.

### Description

Upstream ROS 2 has changed the ament_index_cpp API in two ways:
* `ament_index_cpp::get_package_share_directory` -> `ament_index_cpp::get_package_share_path`
* `ament_index_cpp::get_resource(type, package, contents)` -> `ament_index_cpp::get_resource(type, package)` (the contents are returned)

Because we want this package to work on older ROS 2 versions, we use conditional compilation on the AMENT_INDEX_CPP version to fix things so it works for all currently supported ROS 2 (Humble, Jazzy, Kilted, and Rolling).

Fixes: FLE-165

### DIFF
--- a/ros/src/foxglove_bridge/src/message_definition_cache.cpp
+++ b/ros/src/foxglove_bridge/src/message_definition_cache.cpp
@@ -231,6 +231,9 @@ const MessageSpec& MessageDefinitionCache::load_message_spec(
 #if AMENT_INDEX_CPP_VERSION_GTE(1, 13, 0)
   ament_index_cpp::PathWithResource path_with_resource =
     ament_index_cpp::get_resource("rosidl_interfaces", package);
+  if (path_with_resource.resourcePath == std::nullopt) {
+    throw DefinitionNotFoundError(definition_identifier.package_resource_name);
+  }
   index_contents = path_with_resource.contents;
 #else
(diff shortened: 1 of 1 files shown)


ros/src/foxglove_bridge/src/message_definition_cache.cpp
```diff
@@ -215,7 +215,12 @@ const MessageSpec& MessageDefinitionCache::load_message_spec(
   }
 
   // Get the package share directory, or throw a PackageNotFoundError
-  const std::string share_dir = ament_index_cpp::get_package_share_directory(package);
+  std::filesystem::path share_dir;
+  #if AMENT_INDEX_CPP_VERSION_GTE(1, 13, 0)
```
ericmlujan: This macro should become available after ament/ament_index#105 merges

clalancette:
Just as an FYI, the upstream PR merged and is available in Jazzy, Kilted, and Rolling.  We are still waiting for a Humble patch release; once that happens we should be able to move forward with this PR.

clalancette:
I'm going to put this one into review now, since all of the upstream changes we need are now in place.

PR #809, opened 2026-01-12
Rust utilities for data providers for remote data loaders

Muon:
### Changelog
* Add `data_provider` Rust module with utilities for writing data providers for remote data loaders.

### Docs
FIRE-291

### Description

Adds a new module to the `foxglove` crate with utilities for writing data providers for remote data loaders. Currently, this module contains the following utilities for manifest construction:

1. Structs matching the JSON schema for manifests. These implement `Deserialize`, so they can be used with `serde_json`.
2. `ChannelSet`, which is a utility for declaring channels while deduplicating schemas.

No utilities for building data endpoints are included. (Adding support for automatic stream flushing was attempted, but is currently infeasible because stream flushing is asynchronous.)

This PR includes an example in Rust using `axum` for serving HTTP and `foxglove` for streaming MCAP.

#### Design rationale

Attempts at a more opinionated lightweight SDK did not provide a meaningfully better experience than writing data providers manually. The main benefit that could have been obtained was eliminating type mismatches between manifest and data endpoints in the case where the same server provides both endpoints. However, this gave a restrictive and unfamiliar API which still exposed details of the data provider HTTP API and the streaming transport.

Since the data provider HTTP API is just two endpoints, it would be best to provide some more universal building blocks instead and encourage Rust users to use `axum` directly for now (as shown in the example in this PR).

#### Testing
This PR includes a snapshot test for manifest building using `ChannelSet`. E2E testing against `remote-data-loader` planned in FIRE-283.

Closes: FIRE-292

### DIFF
--- a/rust/foxglove/src/data_provider/manifest.rs
+++ b/rust/foxglove/src/data_provider/manifest.rs
@@ -67,6 +67,7 @@ pub struct Topic {
     /// Message encoding (e.g. `"protobuf"`).
     pub message_encoding: String,
     /// Schema ID, if this topic has an associated schema.
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub schema_id: Option<NonZeroU16>,
 }
 


rust/foxglove_remote_data_loader_upstream/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+//! Every request first calls [`Authenticator::authenticate`] with the bearer token (if present).
+//! The returned identity is passed to the [`ManifestProvider::get_manifest`] and
+//! [`DataProvider::authorize_data`] methods.
+
+pub mod manifest;
+
+use std::{error::Error as StdError, future::Future, net::SocketAddr, sync::Arc};
+
+use axum::{
+    body::{Body, Bytes},
```
james-rms: need to ensure that Bytes and other required third-party types are re-exported from this library.

rust/foxglove_remote_data_loader_upstream/examples/demo.rs
```diff
@@ -0,0 +1,121 @@
...
+        // 3. Stream data (only runs for data requests)
+        let Some(handle) = source.into_stream_handle() else {
+            // Manifest request - we're done
+            return Ok(());
+        };
+
+        // Log some demo data
+        tracing::info!(flight_id = %params.flight_id, "streaming data");
+        for i in 0..10 {
+            channel.log(&DemoMessage {
```
james-rms: i would have expected to await at this point. Without yielding, is there a way for an upstream source to properly wait for backpressure to subside before continuing to write?

james-rms: Following our previous discussion, we should add the logic to the example that monitors the output buffer size and flushes when it gets large.

rust/foxglove_remote_data_loader_upstream/src/blocking.rs
james-rms: fwiw, can lose the mut

Muon: Not without changing the `foxglove` crate. That method takes `&mut self` for some reason. I'm not sure if that's a breaking change, but if it is, I shouldn't do it in this PR, and if it isn't, I can demut this later.

rust/foxglove_remote_data_loader_upstream/examples/demo.rs
```diff
@@ -0,0 +1,145 @@
...
+        _params: &FlightParams,
+    ) -> Result<(), AuthError> {
+        // No authentication in this demo.
+        Ok(())
+    }
+
+    async fn initialize(
+        &self,
+        params: FlightParams,
+        reg: &mut impl ChannelRegistry,
```
bennetthardwick: It's odd to me for `ChannelRegistry` to be a trait here. I feel like a struct that has a trait object inside it might make more sense if we need alternate implementations? I think that way there would be less risk of breaking API changes and maybe ffi would be slightly easier?

Muon: `ChannelRegistry` has two implementations currently. How is risk of breaking API changes reduced by using a struct with a `dyn ChannelRegistry`? FFI is the same, I'll still need to write a concrete Rust impl using opaque pointers and bind to C.

rust/examples/remote_data_loader/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+use chrono::{DateTime, DurationRound, Utc};
+use foxglove::data_provider::{ChannelSet, Manifest, StreamedSource, UpstreamSource};
+use foxglove::stream::create_mcap_stream;
+use futures::StreamExt;
+use serde::{Deserialize, Serialize};
+
+/// A simple message type for this example. In a real implementation, you should probably use one of
+/// the built-in message types from the [`foxglove`] crate for out-of-the-box visualization support.
+#[derive(foxglove::Encode)]
+struct DemoMessage {
```
james-rms: IMO the value of the example would be more immediately obvious if it used one of the foxglove schemas directly - in the docs i've been using Vector3. This more clearly connects this example to "useful things happening on the screen"

rust/foxglove/src/data_provider/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+        }
+    }
+
+    /// Insert a channel for message type `T`.
+    ///
+    /// Extracts the schema from `T: Encode`, assigns a schema ID, and deduplicates schemas
+    /// automatically. If multiple channels share the same schema, only one schema entry will be
+    /// created.
+    pub fn insert<T: Encode>(&mut self, topic: impl Into<String>) {
+        let schema_id = T::get_schema().map(|s| self.add_schema(s));
```
james-rms, line 74: nit: i'm generally against using a map on one value  (s) to mutate another value (self). Flow control would be easier to see with an `if let`

```rust
let schema_id = if let Some(s) = T::get_schema() {
  Some(self.add_schema(s))
} else {
  None
};

File: rust/examples/data_provider/src/main.rs [hunk 6]
```diff
@@ -0,0 +1,195 @@
...
+    });
+
+    Body::from_stream(mcap_stream.map(Ok::<_, Infallible>)).into_response()
+}
+
+#[tokio::main]
+async fn main() -> std::io::Result<()> {
+    tracing_subscriber::fmt::init();
+
+    let app = Router::new()
```
james-rms, line 191: Worth a comment explaining these paths aren't determined by the data provider API, we choose them here for ourselves. Maybe defining them as constants at the top (and reusing the data endpoint in the manifest generation) helps make this clear?

james-rms (approved):

PR #810, opened 2026-01-13
Move message definitions, introduce version-specific binary message framing

ericmlujan:
### Changelog
None

### Docs
None

### Description
Preparation work for introducing v2 binary message definitions. This PR includes the following changes intended to make it easier to introduce new binary message definitions in the upcoming v2 protocol:

1. Create a v1-specific specialization for `{server, client}::BinaryOpcode`. This enables us to migrate some messages like `Subscribe` and `Unsubscribe` in v2 to binary messages without breaking backwards compatibility.
2. Rename the `BinaryMessage` trait in `common` to `BinaryPayload`, which contains logic for serializing the message's contents without hardcoding version-specific opcodes or framing. This trait can be used by binary messages shared across protocol versions. Each protocol version implements its own `BinaryMessage` trait, which encompasses logic for wrapping inner messages with version-appropriate framing.
3. Move tungstenite support to the v1 mod, to reflect its usage of v1-specific `BinaryMessage` traits.
4. Move `v1::client::MessageData` to `common::client`, as it doesn't use subscription IDs and the message definition will end up being identical in v1 and v2.

Fixes: FLE-169

### DIFF
--- a/rust/foxglove/src/protocol/common/client/playback_control_request.rs
+++ b/rust/foxglove/src/protocol/common/client/playback_control_request.rs
@@ -78,16 +78,17 @@ impl<'a> BinaryPayload<'a> for PlaybackControlRequest {
         })
     }
 
-    fn to_payload(&self) -> Vec<u8> {
-        let mut buf = Vec::with_capacity(1 + 4 + 1 + 8 + 4 + self.request_id.len());
+    fn payload_size(&self) -> usize {
+        1 + 4 + 1 + 8 + 4 + self.request_id.len()
+    }
 
+    fn write_payload(&self, buf: &mut impl BufMut) {
         buf.put_u8(self.playback_command as u8);
         buf.put_f32_le(self.playback_speed);
         buf.put_u8(if self.seek_time.is_some() { 1 } else { 0 });
         buf.put_u64_le(self.seek_time.unwrap_or(0));
         buf.put_u32_le(self.request_id.len() as u32);
         buf.put_slice(self.request_id.as_bytes());
-        buf
     }
 }
 
--- a/rust/foxglove/src/protocol/common/client/message_data.rs
+++ b/rust/foxglove/src/protocol/common/client/message_data.rs
@@ -45,12 +45,13 @@ impl<'a> BinaryPayload<'a> for MessageData<'a> {
         })
     }
 
-    fn to_payload(&self) -> Vec<u8> {
-        let size = 4 + self.data.len();
-        let mut buf = Vec::with_capacity(size);
+    fn payload_size(&self) -> usize {
+        4 + self.data.len()
+    }
+
+    fn write_payload(&self, buf: &mut impl BufMut) {
         buf.put_u32_le(self.channel_id);
         buf.put_slice(&self.data);
-        buf
     }
 }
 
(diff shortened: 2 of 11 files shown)


eloff (changes requested):
The problem with this BinaryPayload trait is to_payload allocates a Vec internally to copy the data to, but to add the frame size and opcode we then need to allocate a second Vec and perform a second copy.

I think that extra copy will be an issue. One copy is hard to avoid when framing and serializing the message, but I think we can eliminate the second copy.

I think there's a lot of ways to achieve that, I can help brainstorm options.

ericmlujan:
@eloff thanks for the review, addressed in c71ac506a0133fe78e59134edc9b038c6581b140

ericmlujan:
Note that there's an unrelated CI flake here for the ROS Bridge @ humble, filed FLE-180 to investigate in an upcoming cycle

rust/foxglove/src/protocol/common/client/playback_control_request.rs
```diff
@@ -81,23 +78,23 @@ impl<'a> BinaryMessage<'a> for PlaybackControlRequest {
         })
     }
 
-    fn to_bytes(&self) -> Vec<u8> {
-        let mut buf = Vec::with_capacity(1 + 1 + 4 + 1 + 8 + 4 + self.request_id.len());
+    fn payload_size(&self) -> usize {
+        1 + 4 + 1 + 8 + 4 + self.request_id.len()
```
clalancette, line 82: This isn't at all blocking, but I see at least 2 uses of this "1 + 4 + 1 + 8 + 4" size.  Would it make sense to introduce a structure representing the payload, and then use `std::mem::size_of::Payload`?  We could also possibly use that struct to serialize/deserialize.

(same goes for all of the other impl of `BinaryPayload` below)

ericmlujan: Would that work? The request ID is a variable length item and _is_ a part of the payload. I suppose we could create a struct for all the fixed-size and contiguous elements of the payload, but I'm a little hesitant to add another layer of nesting here (since now we already have BinaryMessage vs. BinaryPayload).

I do want to take another pass at all this code (and the impl BinaryMessage for all of these) and see if, for example, a few helper macros could help reduce some duplication

clalancette: > Would that work? The request ID is a variable length item and is a part of the payload. I suppose we could create a struct for all the fixed-size and contiguous elements of the payload, but I'm a little hesitant to add another layer of nesting here (since now we already have BinaryMessage vs. BinaryPayload).

In C, I would use flexible array members and do something like this:
```
struct playback_control_request
{
  uint8_t playback_command;
  float playback_speed;
  uint8_t had_seek;
  uint64_t seek_time;
  uint32_t request_id_len;
  uint8_t request_id[];
} __attribute__((packed));
```

Then `sizeof(struct playback_control_request)` == 1+4+1+8+4.  Also, you could then do a cast from the byte buffer to the struct, and access the fields instead of doing `data.get_<type>()`/`data.put_<type>()`.  I don't know if that kind of structure is possible/desirable in Rust.

> I do want to take another pass at all this code (and the impl BinaryMessage for all of these) and see if, for example, a few helper macros could help reduce some duplication

That's fair, and maybe a simpler way to reduce this duplication.

PR #806, opened 2026-01-09
add python api to select a layout using the public layout

surbina:
TODO

### Description

<table><tr><th>Before</th><th>After</th></tr><tr><td>

</td><td>

</td></tr></table>

PR #814, opened 2026-01-14
notebook: remove layout_storage_key, add layout stub

jtbandes:
### Changelog
* notebook: Removed `layout_storage_key` parameter and made all parameters kw-only.
* notebook: Manual changes to the layout are not persisted.

### Docs

None

### Description

Adapted from https://github.com/foxglove/foxglove-sdk/pull/806

Based on internal feedback, changing the notebook APIs so the layout is always set and user changes are not restored from the cache. (so the `layout_storage_key` and `force` parameters are removed – the latter was not documented)

Currently this is not publicly documented since the layout APIs are not yet available. This PR adds a stub `Layout` class to allow us to define the method signature.

I also made some tweaks to docs:
- Use `autoclass` instead of manually documenting layout classes.
- Add `autodoc_class_signature = "separated"` option which allows us to hide the `__init__` methods which are treated as private. This also led me to move some init params' docs into the `__init__` docstring instead of the class docstring.

### DIFF
--- a/python/foxglove-sdk/notebook-frontend/widget.ts
+++ b/python/foxglove-sdk/notebook-frontend/widget.ts
@@ -16,7 +16,7 @@ type Message = {
   type: "update-data";
 };
 
-function createSelectLayoutParams(layoutJson: string | undefined): SelectLayoutParams | undefined {
+function createSelectLayoutParams(layoutJson: string | undefined): SelectLayoutParams {
   // Even if no layout is provided, we want to always provide our storageKey and force=true so that
   // the embed doesn't fall back to its default caching behavior.
   return {
@@ -72,9 +72,7 @@ function render({ model, el }: RenderProps<WidgetModel>): void {
   model.on("change:_layout", () => {
     const layoutJson = model.get("_layout");
     const selectParams = createSelectLayoutParams(layoutJson);
-    if (selectParams) {
-      viewer.selectLayout(selectParams);
-    }
+    viewer.selectLayout(selectParams);
   });
 
   el.appendChild(parent);


python/foxglove-sdk/python/docs/api/index.rst
```diff
@@ -19,55 +19,18 @@ Notebook Integration
...
-       context will be used. Logged messages will be buffered.
-   :return: A NotebookBuffer object that can be used to manage the data buffering
-       and visualization.
-   :raises Exception: If the notebook extra package is not installed. Install it
-       with `pip install "foxglove-sdk[notebook]""`.
-
-   .. note::
-      This function is only available when the `notebook` extra package
-      is installed. Install it with `pip install "foxglove-sdk[notebook]""`.
+.. autofunction:: init_notebook_buffer
```
surbina, line 22: Neat!

python/foxglove-sdk/python/foxglove/notebook/foxglove_widget.py
```diff
@@ -27,26 +22,25 @@ class FoxgloveWidget(anywidget.AnyWidget):
...
     ):
-        super().__init__(**kwargs)
+        """
+        :param buffer: The NotebookBuffer object that contains the data to display in the widget.
+        :param width: The width of the widget. Defaults to "full".
+        :param height: The height of the widget in pixels. Defaults to 500.
+        :param src: The source URL of the Foxglove viewer. Defaults to
+            "https://embed.foxglove.dev/".
+        """
+        super().__init__()
```
surbina, line 43: Is there a way to pass the kwargs here?

python/foxglove-sdk/python/foxglove/notebook/foxglove_widget.py
```diff
@@ -67,25 +62,12 @@ def __init__(
...
-            "force": force_layout,
-        }
-
     def refresh(self) -> None:
         """
         Refresh the widget by getting the data from the callback function and sending it
         to the widget.
         """
-        data = self._buffer.get_data()
+        data = self._buffer._get_data()
```
surbina, line 70: Nit: it looks a bit odd that the widget is accessing a private method from the buffer.

python/foxglove-sdk/notebook-frontend/widget.ts
```diff
@@ -1,39 +1,42 @@
...
+  _layout?: string;
 }
 
+const DEFAULT_NOTEBOOK_LAYOUT_STORAGE_KEY = "foxglove-notebook-default-layout";
+
 type Message = {
   type: "update-data";
 };
 
+function createSelectLayoutParams(layoutJson: string | undefined): SelectLayoutParams | undefined {
```
surbina: This doesn't return undefined anymore

python/foxglove-sdk/python/foxglove/notebook/foxglove_widget.py
jtbandes: What kwargs would the user want to pass here?

python/foxglove-sdk/python/foxglove/notebook/foxglove_widget.py
jtbandes: I guess so, on the other hand, we don't want users to call this method, right? (and want to hide it from the docs)

I think I could also do `:meta private:` in the docstring, would you prefer that?

python/foxglove-sdk/python/foxglove/notebook/foxglove_widget.py
surbina: Yeah, you are right, let's leave it as is.

python/foxglove-sdk/python/foxglove/notebook/foxglove_widget.py
surbina: Nah, don't worry, it's more of a doubt I had.

PR #817, opened 2026-01-15
rust: Add serde for schema message types

gasmith:
### Changelog
- rust: Added serde implementations for schema messages

### Docs
None

### Description
This change adds `Serialize` and `Deserialize` implementations for foxglove schemas in rust, under the `serde` feature.

For human-readable formats like JSON:
 - Enums are serialized as their string names (e.g., `"LINE_STRIP"`)
 - Binary data are serialized as base64-encoded strings

For binary formats:
 - Enums are serialized as i32 values
 - Binary data are serialized as raw bytes

I'm still not entierly sure I want to claim this is the basis for a stable wire protocol, but I feel a bit better about it now that we have the format-aware serde implementations for enums & binary data.

Resolves: https://github.com/orgs/foxglove/discussions/1207

### DIFF
--- a/rust/foxglove/src/schemas.rs
+++ b/rust/foxglove/src/schemas.rs
@@ -13,8 +13,8 @@
 //! For human-readable formats (e.g., JSON), enums are serialized as string names, and binary data
 //! are serialized as base64. For binary formats, enums are serialized as i32 values.
 //!
-//! Note that CDR is not compatible with these schemas, because it does not support optional
-//! fields.
+//! Note that [CDR](https://docs.rs/cdr) is not compatible with these schemas, because it does
+//! not support optional fields.
 
 pub(crate) mod descriptors;
 #[allow(missing_docs)]


rust/foxglove/Cargo.toml
```diff
@@ -72,8 +73,10 @@ rcgen = { version = "0.14.3", features = ["crypto", "pem", "x509-parser"], optio
 
 [dev-dependencies]
 assert_matches = "1.5.0"
+bincode = "1.3"
```
eloff: Bincode is not maintained anymore, but for a dev dep that's fine

rust/foxglove/src/schemas.rs
```diff
@@ -5,6 +5,23 @@
...
+//! logging, and integration with tools that consume JSON or other serde-compatible formats.
+//!
+//! For human-readable formats like JSON:
+//! - Enums are serialized as their string names (e.g., `"LINE_STRIP"`)
+//! - Binary data are serialized as base64-encoded strings
+//!
+//! For binary formats:
+//! - Enums are serialized as i32 values
+//! - Binary data are serialized as raw bytes
+//!
```
eloff, line 15: Maybe consider adding a note that we don't yet guarantee serde serialization format compatibility between major versions of the SDK - if you're still uncomfortable committing to that

rust/foxglove/src/schemas_wkt.rs
```diff
@@ -108,6 +108,7 @@ impl NormalizeResult {
 /// assert_eq!(duration, Duration::MAX);
 /// ```
 #[derive(Debug, Default, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[cfg_attr(feature = "serde", derive(::serde::Serialize, ::serde::Deserialize))]
```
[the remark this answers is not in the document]
gasmith: Resolved by adding normalization methods for well-known types.

rust/foxglove/src/schemas.rs
```diff
@@ -15,3 +35,199 @@ mod impls;
...
+    use base64::Engine;
+    use bytes::Bytes;
+    use serde::de::{Error as _, Visitor};
+    use serde::{Deserialize, Deserializer, Serializer};
+
+    pub fn serialize<S>(bytes: &Bytes, s: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        if s.is_human_readable() {
```
eloff, line 43: I didn't know this was a thing. Very nice!

rust/foxglove/Cargo.toml
gasmith: Yeah, we can probably drop it tbh. The cbor coverage is enough.

rust/foxglove/src/schemas.rs
gasmith: I think I'm comfortable with it, actually. It matches what we have under schemas/jsonschema.

PR #821, opened 2026-01-16
rust: Add img2yuv

gasmith:
### Changelog
None

### Docs
None

### Description
This change adds the `img2yuv` module, which will be used in the future for transcoding image data. The implementation is basically lifted from the data-platform repository, with some [minor adjustments](https://github.com/foxglove/foxglove-sdk/pull/821/commits/eaa49973fb17babb9c19c26bb326612222e2dc94) based on Claude's review of the code, and the lack of `is_multiple_of` in our MSRV.

### DIFF
--- a/rust/foxglove/src/img2yuv/compressed.rs
+++ b/rust/foxglove/src/img2yuv/compressed.rs
@@ -52,15 +52,11 @@ impl FromStr for Compression {
 impl Compression {
     /// Parses a format string from a ROS `CompressedImage` message.
     ///
-    /// The format string is described at length in the ROS1 [sensor_msgs/CompressedImage][ros1]
+    /// The format string is described at length in the ROS 2 [sensor_msgs/CompressedImage][ros2]
     /// definition.
     ///
-    /// While the ROS2 [sensor_msgs/msg/CompressedImage][ros2] documentation suggests that the
-    /// format string can only be "jpeg" or "png", we've observed ROS2 compressed images in the
-    /// wild that use ROS1-style format specifications.
-    ///
     /// [ros1]: https://docs.ros.org/en/noetic/api/sensor_msgs/html/msg/CompressedImage.html
-    /// [ros2]: https://docs.ros2.org/foxy/api/sensor_msgs/msg/CompressedImage.html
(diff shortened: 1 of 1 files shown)


rust/foxglove/src/img2yuv/compressed.rs
```diff
@@ -0,0 +1,162 @@
...
+    ///
+    /// The format string is described at length in the ROS1 [sensor_msgs/CompressedImage][ros1]
+    /// definition.
+    ///
+    /// While the ROS2 [sensor_msgs/msg/CompressedImage][ros2] documentation suggests that the
+    /// format string can only be "jpeg" or "png", we've observed ROS2 compressed images in the
+    /// wild that use ROS1-style format specifications.
+    ///
+    /// [ros1]: https://docs.ros.org/en/noetic/api/sensor_msgs/html/msg/CompressedImage.html
+    /// [ros2]: https://docs.ros2.org/foxy/api/sensor_msgs/msg/CompressedImage.html
```
clalancette: This whole comment is somewhat out-of-date now.  The comments in the ROS 2 CompressedImage msg file now match ROS 1, and there is a new URL for ROS 2.  Thus I think:


```suggestion
    /// The format string is described at length in the ROS 2 [sensor_msgs/CompressedImage][ros2]
    /// definition.
    ///
    /// [ros1]: https://docs.ros.org/en/noetic/api/sensor_msgs/html/msg/CompressedImage.html
    /// [ros2]: https://docs.ros.org/en/rolling/p/sensor_msgs/msg/CompressedImage.html
```

gasmith:
> Besides what I have inline, it would also be nice to see a PR to data-platform which removes it there and makes the agent depend on the version here. Then we could verify it end-to-end that it still works. Let me know if you'd like me to do that, I'm happy to give it a shot.

While I agree with the spirit, I don't really want to expose img2yuv as a public interface of the SDK. We could keep it semi-private by marking it `#[doc(hidden)]`, but I'm also not sure there's a great deal of value there. Rather than spending time to port the agent to use this very-nearly-exact-replica of the library it's already using, I'd prefer we spend our time porting the remaining agent code into the SDK so that we can test it here. I don't think there's really much risk of regression in this migration.

eloff (approved):
LGTM

PR #827, opened 2026-01-22
docs: update notebook docs

jtbandes:
### Changelog

Added documentation for layout-related types in the notebook module docs.

### Docs

N/A

### Description

Resolves ERT-1347

~Blocked by #840~
~Blocked by #841~

* Moved notebook-related info to a new page.
* Added a custom sphinx plugin to group docs by panel and create a page for each group.


surbina:
I like how the docs are structured. One of the things I noticed is that the generated docs is missing important information, take the `ThreeDeeConfig` as an example:

Screenshot 2026-01-22 at 11 55 27

Most of the config options are not documented because they are inherited from `BaseRendererConfig` or lack docstrings. Is there a way to add at least the type hint to the signature?

PR #829, opened 2026-01-22
Increase the timeouts in the smoke tests.

clalancette:
### Changelog

None.

### Docs

None.

### Description

We've noticed some flakiness in the smoke_test tests, particularly when running in the ROS infrastructure. When they fail, they are failing on some of the timeouts waiting for things to connect.

We've also noticed that this is most likely to happen when the ROS buildfarm is busy.  That leads to the supposition that the 1 second timeout we have on some of these checks is too short when the host is busy.

The fix implemented here is to change the one second timeouts to 10 seconds everywhere.  We don't expect we'll hit 10 seconds often (if ever), but it should be enough buffer for the tests to pass even on loaded systems.

In combination with this, we also remove the repeated tests. Running the tests 50x on the ROS infrastructure seems excessive and also leads to more chances to fail.  As something of a compromise, we instead set our GitHub workflows to run 50 times instead of the default being 50x.

Fixes: [FLE-180](https://linear.app/foxglove/issue/FLE-180/investigate-flake-in-foxglove-bridge-smoketesttestsubscription)

@jtbandes FYI, particularly about reducing the number of times we run these tests by default.

### DIFF
--- a/ros/src/foxglove_bridge/CMakeLists.txt
+++ b/ros/src/foxglove_bridge/CMakeLists.txt
@@ -208,7 +208,6 @@ if(BUILD_TESTING)
 
   ament_lint_auto_find_test_dependencies()
 
-  # Repeat tests several times to catch nondeterministic issues
   ament_add_gtest(smoke_test tests/smoke_test.cpp TIMEOUT 600)
   target_link_libraries(smoke_test
     foxglove_bridge_component


.github/workflows/ros.yml
```diff
@@ -33,6 +33,6 @@ jobs:
         working-directory: ros
 
       - name: Run make docker-test
-        run: make docker-test-${{ matrix.ros_distribution }}
+        run: make GTEST_REPEAT=50 docker-test-${{ matrix.ros_distribution }}
```
gasmith, line 36: Actually, looking at CI, I'm not sure this was plumbed through properly. Might need to update the makefile to pass this env var into the docker container.

clalancette: I did plumb it through; see the changes in https://github.com/foxglove/foxglove-sdk/pull/829/changes#diff-26a140816629b5745638f3e53cee7e2e76c6f5562fb606951c4677adeb534807.

Locally, running the `foxglove_bridge` tests once takes 4.39 seconds, so naively we expect 50 runs to take 3 minutes 40 seconds (4.39*50 == 219.5 seconds == 3 minutes 40 seconds).  Locally, running it 50 times takes 3 minutes 16 seconds, which is in the ballpark (I expect the difference is due to setup time, which is reduced when you do repeated tests).

The run in CI took 3 minutes 18 seconds, which is pretty close to my local tests.  So I'm fairly confident this is running 50 times.

ericmlujan (approved):

ros/src/foxglove_bridge/CMakeLists.txt
```diff
@@ -209,7 +209,7 @@ if(BUILD_TESTING)
   ament_lint_auto_find_test_dependencies()
 
   # Repeat tests several times to catch nondeterministic issues
```
jtbandes: this comment can be removed now

ros/src/foxglove_bridge/tests/smoke_test.cpp
```diff
@@ -560,7 +559,7 @@ TEST_F(ParameterTest, testParameterSubscription) {
   _wsClient->setParameters(setParams2);
 
   future = _wsClient->waitForParameters();
-  ASSERT_EQ(std::future_status::timeout, future.wait_for(ONE_SECOND));
+  ASSERT_EQ(std::future_status::timeout, future.wait_for(std::chrono::seconds(1)));
```
jtbandes, line 562: why only 1 here?

ros/src/foxglove_bridge/CMakeLists.txt
clalancette: Good call.  Fixed in 8394919adf5b7b07d0f463e12263b5109f71bf01

ros/src/foxglove_bridge/tests/smoke_test.cpp
clalancette: This one tripped me up as well at first.  Unlike all of the other tests in this file, this is a *negative* test; it is testing for something to timeout.  While increasing this to 10 seconds is more "fair" (in that we give `waitForParameters` more time to potentially bug out and return data), I think that particular kind of failure will be noticed pretty quickly anyway.  So I kept this at 1 second just to keep running the tests short (this second is a full 1/4 of the total time it takes to run tests as it is).

clalancette:
> Maybe 50 is excessive? feel free to change if desired -- I think it was just because some race conditions were so uncommon that they ended up not being caught for a while without a bunch of repeats.

Yeah, understood.  Given the fact that running this 50 times only takes ~4 minutes on the GitHub runners, I think we can leave it; it is not the long pole here anyway.

.github/workflows/ros.yml
gasmith: My mistake; I jumped to my conclusion based on the brevity of the test output. LG!

clalancette: All good!  The comment was good in that it caused me to go back and revisit my assumptions.

PR #692, opened 2025-10-07
Add Client Count Topic to Foxglove_bridge

ndmmdn:
### Changelog

- Foxglove_bridge now publishes the number of foxglove clients connected to it under `/foxglove_bridge/client_count`
### Docs

- I added a mention of the topic to the foxglove_bridge [readme](https://github.com/ndmmdn/foxglove-sdk/blob/feat-client-count/ros/src/foxglove_bridge/README.md) and that it requires a parameter to be set.
### Description

- There are several examples where it is useful to no if a client is connected to the foxglove bridge, for example to pause other bandwidth intensive processes. Additionally, similar to the `who` command in bash, it can also be useful to know if there are multiple people connected to the same robot. Both of these cases can now be detected via the new client count topic.

there exists the [test_callbacks](https://github.com/ndmmdn/foxglove-sdk/blob/feat-client-count/cpp/test_callbacks.cpp) test to allow testing on the c++ server level. 

Additionally, the new functionality was tested by build and running the foxglove_bridge as described in the readme. In a separate terminal, the `/foxglove_bridge/client_count` topic was subscribed to. Then using several foxglove instances on the same laptop, connections and disconnections were made to verify the client count is updated correctly.  

<table><tr><th>Before</th><th>After</th></tr><tr><td>

</td><td>

</td></tr></table>

This is a refactor/adaptation of [this foxglove_bridge PR](https://github.com/foxglove/ros-foxglove-bridge/pull/376)

PR #831, opened 2026-01-22
Bump ros packages to 3.2.4, sdk dep to v0.16.5

ericmlujan:
### Changelog
None

### Docs
None

### Description
Update ROS packages to depend on SDK v0.16.5, which introduces the client count functionality from #692. ROS changelogs are updated.

Tested build succeeded locally on jazzy, humble, and kilted with `make USE_LOCAL_SDK=OFF docker-build` to verify that pulling SDK binaries is successful.

### DIFF
--- a/Dockerfile.ros
+++ b/Dockerfile.ros
@@ -1,7 +1,26 @@
 ARG ROS_DISTRIBUTION=rolling
 FROM ros:$ROS_DISTRIBUTION-ros-base
 
-RUN apt-get update
+# Prevent errors from apt-get.
+# See: http://askubuntu.com/questions/506158/unable-to-initialize-frontend-dialog-when-using-ssh
+ENV DEBIAN_FRONTEND=noninteractive
+
+# The ROS docker images are only updated when the underlying
+# Ubuntu layer changes; this is a limitation of being an
+# "official" docker image.  Additionally, ROS packages do
+# not have minimum dependency requirements.  That is,
+# ROS package "A" may have a dependency on "B", but it cannot
+# easily specify what version of "B" that it needs.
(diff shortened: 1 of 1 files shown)


clalancette (approved):

ros/src/foxglove_bridge/CHANGELOG.rst
```diff
@@ -2,6 +2,12 @@
 Changelog for package foxglove_bridge
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+3.2.4 (2026-01-22)
+------------------
+* Add client count topic (`#692 <https://github.com/foxglove/foxglove-sdk/pull/692>`_)
+* Fix build warning on newer versions of CMake (`#828 <https://github.com/foxglove/foxglove-sdk/pull/828>`_)
+* Contributors: Andrew Madden, Chris Lalancette
```
gasmith, line 10: Are there any other core SDK changes that might be relevant? How do we currently answer that question?

Maybe we should also call out that we've upgraded the foxglove SDK to v0.16.5 here, so that someone reviewing the changelog has a quick reference for which underlying SDK version was used for a particular bridge version?

ericmlujan: There's not a end-all be-all answer to the first, other than skimming the changelogs for the SDK between version bumps and assuming that anything that majorly directly a ROS package's capabilities will probably have a code change on the ROS side as well. But things like internal refactors or performance enhancements wouldn't be caught by that.

I agree we should reflect SDK version bumps that in the changelog as well.

PR #832, opened 2026-01-23
Add livekit as optional dependency and fix dockerfiles and CI builds

eloff:
### Changelog

None

### Docs

None

### Description

Adding libwebrtc and livekit as optional dependencies turns out to require a lot of Dockerfile and CI changes. I'm splitting this work out here to make it easier to review.


eloff:
This is not needed anymore, Greg fixed it

PR #836, opened 2026-01-23
Add live visualization credentials provider

ericmlujan:
### Changelog
- None

### Docs
- None

### Description
Introduce a mechanism for fetching and refreshing live visualization credentials from the Foxglove API using a device token.

This PR introduces an internal `api_client` mod that allows for communication with the Foxglove API in service of:

1. Retrieving device metadata from a token using the `/internal/platform/v1/device-info` endpoint
2. Retrieving device-specific RTC credentials from the `/internal/platform/v1/devices/:device_id/remote-sessions` endpoint.

This mod is gated by the `agent` flag, hidden from documentation, and comments have been added to warn users that this is not intended to be used directly.

Not all Foxglove API endpoints are implemented in this mod, just the bare essentials required to authorize a device and retrieve credentials. The only supported authentication method is a device token, as this client is currently only intended for use by the live visualization feature.

In addition to tokio tests that mock the above endpoints, I also made a local test executable that was able to successfully retrieve credentials from the production Foxglove API.

Note that this is only implemented in Rust until an initial Rust implementation is complete and the interfaces settle down. Will then port data types and interfaces to Python and C++ if needed.

### DIFF
--- a/rust/foxglove/src/api_client/types.rs
+++ b/rust/foxglove/src/api_client/types.rs
@@ -5,7 +5,9 @@ use serde::{Deserialize, Serialize};
 #[derive(Deserialize, Serialize, Clone)]
 #[serde(rename_all = "camelCase")]
 pub(crate) struct RtcCredentials {
+    /// Expiring access token (JWT)
     pub token: String,
+    /// URL of the RTC server where these credentials are valid.
     pub url: String,
 }
 


rust/foxglove/src/lib.rs
```diff
@@ -361,6 +361,9 @@ pub use sink_channel_filter::SinkChannelFilter;
 pub use std::collections::BTreeMap;
 pub(crate) use time::nanoseconds_since_epoch;
 
+#[cfg(feature = "agent")]
+#[doc(hidden)]
```
gasmith: Probably not necessary since it isn't public.
```suggestion
```

rust/foxglove/Cargo.toml
```diff
@@ -85,15 +85,19 @@ image = { version = "0.25.9", default-features = false, optional = true }
...
 [dev-dependencies]
 assert_matches = "1.5.0"
+axum = "0.8.8"
 hexdump = "0.1.2"
 insta = { version = "1.42.2", features = ["json"] }
 maplit = "1.0.2"
 serde_cbor = "0.11.2"
 serde_json = "1.0"
 tempfile = "3.15.0"
+tokio = { workspace = true, features = ["rt-multi-thread", "macros", "net"] }
```
gasmith: Probably reasonable to just add the `net` feature to the workspace Cargo.toml.

rust/foxglove/src/api_client/client.rs
```diff
@@ -0,0 +1,458 @@
...
+            }
+        }
+
+        Ok(response)
+    }
+}
+
+pub(super) fn default_user_agent() -> String {
+    format!("foxglove-sdk/{}", env!("CARGO_PKG_VERSION"))
+}
```
gasmith: We should use `get_library_version()` or similar here, so that we also have the language binding as part of the user-agent.

We might also want to schematize this a little differently, since our user agent strings for data platform services is rendered as `<service> (<version>)` (cf., the [device agent](https://github.com/foxglove/data-platform/blob/c4c11e27b70137d7a92187bbef11209b791b31a2/rust/agent/src/client.rs#L45)). Looks like our python client library doesn't specify a user-agent, surprisingly - otherwise, I also would've considered that for precedent.

rust/foxglove/src/api_client/client.rs
```diff
@@ -0,0 +1,458 @@
...
+
+/// Internal API client for communicating with the Foxglove platform.
+///
+/// This client is intended for internal use only to support the live visualization feature
+/// and is subject to breaking changes at any time. Do not depend on the stability of this type.
+#[derive(Clone)]
+pub(super) struct FoxgloveApiClient {
+    http: reqwest::Client,
+    device_token: Option<DeviceToken>,
```
gasmith: I mildly prefer the way this is done in the data-platform repo, where the client is generic over auth scheme, and certain methods are only implemented for certain auth schemes. This makes the lack of an API token for a method that requires one a compile-time error rather than a runtime error (`FoxgloveApiClientError::NoToken`).

rust/foxglove/src/api_client/mod.rs
```diff
@@ -0,0 +1,16 @@
+//! Internal API client for the live visualization feature.
+//!
+//! This module is intended for internal use only and is subject to breaking changes at any time.
+//! Do not depend on the stability of any types or functions in this module.
+
+#![allow(dead_code, unused_imports)]
```
gasmith: Could do simply
```suggestion
#![allow(unused)]
```

rust/foxglove/src/api_client/client.rs
```diff
@@ -0,0 +1,458 @@
+#![allow(dead_code)]
```
gasmith: Nit: This is redundant to the `client` module's lint configuration.

rust/foxglove/src/api_client/credentials_provider.rs
gasmith, line 1: TBH, this feels like it belongs outside the api client module. The foxglove API client is the primary implementation, but we'll probably want to stub it out with an (internal) trait for testing. We can move it after Dan's change lands.

rust/foxglove/src/api_client/client.rs
```diff
@@ -0,0 +1,458 @@
...
+            device_token,
+            base_url: base_url.into(),
+            user_agent: user_agent.into(),
+        })
+    }
+
+    pub fn set_device_token(&mut self, token: DeviceToken) -> &mut Self {
+        self.device_token = Some(token);
+        self
+    }
```
gasmith: Feels like more of a builder-time thing. Do we need the ability to change the token?

rust/foxglove/src/api_client/client.rs
```diff
@@ -0,0 +1,458 @@
...
+
+    pub fn get(&self, endpoint: &str) -> RequestBuilder {
+        self.request(Method::GET, endpoint)
+    }
+
+    pub fn post(&self, endpoint: &str) -> RequestBuilder {
+        self.request(Method::POST, endpoint)
+    }
+
+    pub fn device_token(&self) -> Option<&DeviceToken> {
```
gasmith: Seems unnecessary?

rust/foxglove/src/api_client/credentials_provider.rs
```diff
@@ -0,0 +1,62 @@
...
+
+        let _refresh_guard = self.refresh_lock.lock().await;
+        if let Some(credentials) = self.current_credentials() {
+            return Ok(credentials);
+        }
+
+        self.refresh().await
+    }
+
+    pub async fn refresh(&self) -> Result<Arc<RtcCredentials>, CredentialsError> {
```
gasmith: This probably doesn't need to be public, if we intend to call it internally on demand.

rust/foxglove/src/api_client/client.rs
```diff
@@ -0,0 +1,458 @@
...
+}
+
+impl FoxgloveApiClient {
+    pub fn new(
+        base_url: impl Into<String>,
+        device_token: Option<DeviceToken>,
+        user_agent: impl Into<String>,
+    ) -> Result<Self, FoxgloveApiClientError> {
+        Ok(Self {
+            http: reqwest::ClientBuilder::new().build()?,
```
gasmith: We definitely should not use the default reqwest timeout, which is unlimited. We should probably set a sane default and expose an override as part of the FoxgloveApiClientBuilder.

rust/foxglove/src/api_client/client.rs
```diff
@@ -0,0 +1,458 @@
...
+#[derive(Clone)]
+pub(super) struct FoxgloveApiClient {
+    http: reqwest::Client,
+    device_token: Option<DeviceToken>,
+    base_url: String,
+    user_agent: String,
+}
+
+impl FoxgloveApiClient {
+    pub fn new(
```
gasmith: Shouldn't be public. This should only be constructed via the builder.

rust/foxglove/src/api_client/client.rs
```diff
@@ -0,0 +1,458 @@
...
+        endpoint: &str,
+        handler: impl Handler<T, ()>,
+    ) -> String {
+        let app = Router::new().route(endpoint, axum::routing::any(handler));
+
+        let listener = TcpListener::bind("0.0.0.0:0").await.unwrap();
+        let addr = listener.local_addr().unwrap();
+
+        tokio::spawn(async move {
+            axum::serve(listener, app).await.unwrap();
```
gasmith: These server instances are effectively leaked. I'd prefer that they were scoped to each test case.

rust/foxglove/src/api_client/device.rs
```diff
@@ -0,0 +1,46 @@
...
+        &self.info.id
+    }
+
+    pub fn name(&self) -> &str {
+        &self.info.name
+    }
+
+    pub fn project_id(&self) -> &str {
+        &self.info.project_id
+    }
```
gasmith: If we intend to use project ID, we should plan ahead for the future in which devices can be moved between projects, and not treat it as static. But I don't imagine we're actually making use of this.

rust/foxglove/src/api_client/device.rs
```diff
@@ -0,0 +1,46 @@
...
+use super::types::{DeviceResponse, RtcCredentials};
+
+#[derive(Clone)]
+pub(crate) struct Device {
+    info: DeviceResponse,
+    client: FoxgloveApiClient,
+}
+
+impl Device {
+    pub async fn new(token: DeviceToken) -> Result<Self, FoxgloveApiClientError> {
```
gasmith: I don't think this struct's implementation offers enough value to justify its existence. Seems to be merely a wrapper around two client methods and a handful of fields in a device-info response.

cpp/CMakeLists.txt
```diff
@@ -111,7 +111,7 @@ endif()
 if(WIN32)
   target_link_libraries(foxglove-static INTERFACE Bcrypt SChannel Crypt32 Ncrypt)
 elseif(APPLE)
-  target_link_libraries(foxglove-static INTERFACE "-framework Security" "-framework CoreFoundation")
```
gasmith, line 114: Why is this necessary?

rust/foxglove/src/api_client/client.rs
ericmlujan: Cool, I'll update the format. I hadn't seen any prior art in this repo, didn't think to check agent

cpp/CMakeLists.txt
gasmith: As suggested elsewhere, let's just drop the agent feature from c/Cargo.toml and python/foxglove-sdk/Cargo.toml for now. We can work on enabling those language bindings as separate and orthogonal work. For now, let's just get the basic pieces into rust.

ericmlujan: As far as I can gather, adding `reqwest` introduced a transitive dep on `hyper`, which on macOS tries to link against the SystemConfiguration framework. I was seeing the following errors on macOS builds:
```
Undefined symbols for architecture arm64:
  "_SCDynamicStoreCopyProxies", referenced from:
      system_configuration::dynamic_store::SCDynamicStore::get_proxies::h57d52d9911dd50a1 in libfoxglove.a[821](system_configuration-364efb3db73e4944.system_configuration.97bde9b32ae9a12c-cgu.0.rcgu.o)
  [... snip ...]
  "_kSCPropNetProxiesHTTPSProxy", referenced from:
      hyper_util::client::proxy::matcher::mac::with_system::h9aa81bbe67f50e06 in libfoxglove.a[814](hyper_util-a748f471dd29bcf6.hyper_util.b5b8457c749cddd-cgu.09.rcgu.o)
ld: symbol(s) not found for architecture arm64
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```
Looked up the symbols and traced them to the [missing framework](https://developer.apple.com/documentation/systemconfiguration).

rust/foxglove/src/api_client/mod.rs
ericmlujan: Hah, TIL

rust/foxglove/src/api_client/client.rs
ericmlujan: I was considering this but pared it back since we only really are hitting device-authenticated frameworks today. I take the point though that, since the token is settable at runtime here, we're introducing surface area for invalid configuration at runtime, which also sucks. I'll make the device token required and enabled through a generic.

rust/foxglove/src/api_client/credentials_provider.rs
ericmlujan: Yeah, that was my plan. I just didn't have a better place to park it in the interim and figured we could move it once the intended callsite was in.

rust/foxglove/src/api_client/client.rs
ericmlujan: Probably not. This is best left as a builder/construction-time thing. I think this was left over from before our in-person conversation about the prevalence of the Builder pattern in Rust; now that that's been clarified, totally agree.

rust/foxglove/src/api_client/credentials_provider.rs
ericmlujan: I remember earlier discussions about us wanting to also rotate on-device credentials out on a cadence, but I suppose the same could be accomplished by calling `clear()` on that cadence and having the refresh actually occur when `load_credentials()` gets called next.

rust/foxglove/src/api_client/client.rs
ericmlujan: Cool, will set a 30s timeout in the Default impl for now (seems sane for something like fetching dev info and creds) and make it settable

rust/foxglove/src/api_client/types.rs
```diff
@@ -0,0 +1,27 @@
+#![allow(dead_code)]
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Deserialize, Serialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct RtcCredentials {
+    pub token: String,
+    pub url: String,
```
clalancette, line 11: Minor nit, but should we add comments here for what each of the fields represents?  Copying from the agent code, something like:


```suggestion
    /// Expiring access token (JWT)
    pub token: String,
    /// URL of the RTC server where these credentials are valid.
    pub url: String,
```

rust/foxglove/src/api_client/credentials_provider.rs
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        self.refresh().await
+    }
+
+    async fn refresh(&self) -> Result<Arc<RtcCredentials>, CredentialsError> {
+        let credentials = Arc::new(self.client.authorize_remote_viz(&self.device.id).await?);
+        self.credentials.store(Some(credentials.clone()));
+        Ok(credentials)
+    }
```
clalancette: This is very subjective, but since this is a private function, only called in one place, and only 3 lines, I would inline it into `load_credentials`.

rust/foxglove/src/api_client/credentials_provider.rs
```diff
@@ -0,0 +1,133 @@
...
+        if let Some(credentials) = self.current_credentials() {
+            return Ok(credentials);
+        }
+
+        let _refresh_guard = self.refresh_lock.lock().await;
+        if let Some(credentials) = self.current_credentials() {
+            return Ok(credentials);
+        }
+
+        self.refresh().await
```
clalancette: This sequence seems odd.  The logic as I see it:
1.  Atomically try to load the credentials.  If it succeeds, return a valid copy, if not, continue.  This is the "fast-path", though it still involves manipulation of an atomic and a copy.
2.  Take the refresh_lock to guard against races with `clear` and other callers of `load_credentials`, atomically try to load the credentials.  If it succeeds, return a valid copy, it not, continue.
3.  Try to refresh (still under the refresh_lock), which will store to self.credentials.

I would almost get rid of the first step, and unconditionally take the `refresh_lock`.  Compared to the rest of the things we are doing, it probably isn't very expensive.  But that is a light suggestion, don't block on it.

rust/foxglove/src/api_client/client.rs
```diff
@@ -0,0 +1,335 @@
...
+                Ok(error) => {
+                    return Err(RequestError::ErrorResponse {
+                        status,
+                        error,
+                        headers,
+                    });
+                }
+                Err(_) => {
+                    let body = String::from_utf8_lossy(&body).to_string();
+                    return Err(RequestError::MalformedErrorResponse {
```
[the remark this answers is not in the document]
clalancette: This seems like a valid complaint from Cursor, though this code is a copy from data-platform.  My suggestion is to open a ticket to follow-up and make this nicer in a future PR.

rust/foxglove/src/api_client/client.rs
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    pub async fn send(self) -> Result<reqwest::Response, RequestError> {
+        let response = self.0.send().await.map_err(RequestError::SendRequest)?;
+
+        let status = response.status();
+        if status.is_client_error() || status.is_server_error() {
+            let headers = Box::new(response.headers().clone());
+            let body = response.bytes().await.unwrap_or_default();
+            match serde_json::from_slice::<ErrorResponse>(&body) {
```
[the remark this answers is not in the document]
clalancette: I kind of see where this is coming from, though I also think that since this is a generic function, any limit we apply here is probably eventually going to be wrong.  But maybe we should consider a few checks:
1.  Allow passing an optional max_size into the struct (either for `send` or for `new`); individual API calls will have a better idea what maximum size they should expect.
2.  Make sure the header size doesn't exceed max_size.
3.  Make sure we never read more than the header_size bytes.

All of that said, I think this is fine as a follow-up.  So I'll suggest opening a ticket to follow-up on this.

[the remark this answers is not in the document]
ericmlujan: Opened [FLE-228](https://linear.app/foxglove/issue/FLE-228) to capture this

rust/foxglove/src/api_client/client.rs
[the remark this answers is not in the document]
ericmlujan: This was resolved in d5b48567a5001aceccc34454af871db10fe1e3ab

PR #837, opened 2026-01-27
Rust example for steaming CSV data over WebSocket with RangedPlayback capability

ericmlujan:
### Description

</td><td>

</td></tr></table>

### DIFF
--- a/rust/examples/ws_stream_csv/src/main.rs
+++ b/rust/examples/ws_stream_csv/src/main.rs
@@ -63,38 +63,43 @@ fn load_csv(path: &PathBuf) -> Result<Vec<DataPoint>, Box<dyn std::error::Error>
     Ok(data)
 }
 
-/// Shared playback state that can be accessed by both the listener and the playback loop.
+/// Simple controller for managing the playback of data. This is an example stand-in
+/// for your own custom data playback stack.
 struct PlaybackController {
-    /// Current playback status
+    inner: Inner,
+}
+
+/// Internal variables owned by PlaybackController required for tracking the current
+/// playback time, status, speed, etc. These will be copied over to an emitted PlaybackState message.
+#[derive(Clone, Copy)]
+struct Inner {
     status: PlaybackStatus,
-    /// Current playback position (nanoseconds)
-    current_time: u64,
-    /// Playback speed multiplier (1.0 = realtime)
     speed: f32,
-    /// Set to true when a seek operation occurs
+    current_time: u64,
     did_seek: bool,
-    /// Time range of the data (start, end) in nanoseconds
     time_range: (u64, u64),
 }
 
 impl PlaybackController {
     fn new(time_range: (u64, u64)) -> Self {
         Self {
-            status: PlaybackStatus::Paused,
-            current_time: time_range.0,
-            speed: 1.0,
-            did_seek: false,
-            time_range,
+            inner: Inner {
+                status: PlaybackStatus::Paused,
+                current_time: time_range.0,
+                speed: 1.0,
+                did_seek: false,
+                time_range,
+            },
         }
     }
 
     /// Create a PlaybackState message from the current state.
-    fn to_state(&self, request_id: Option<String>) -> PlaybackState {
+    fn to_playback_state(&self, request_id: Option<String>) -> PlaybackState {
         PlaybackState {
-            status: self.status,
-            current_time: self.current_time,
-            playback_speed: self.speed,
-            did_seek: self.did_seek,
+            status: self.inner.status,
+            current_time: self.inner.current_time,
+            playback_speed: self.inner.speed,
+            did_seek: self.inner.did_seek,
             request_id,
         }
     }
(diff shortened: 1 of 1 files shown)


rust/examples/ws_stream_csv/src/main.rs
```diff
@@ -0,0 +1,349 @@
...
+
+    // Wait for user to seek or restart
+    drop(ctrl);
+    loop {
+        thread::sleep(Duration::from_millis(100));
+        let ctrl = controller.lock().unwrap();
+        if ctrl.inner.did_seek || ctrl.inner.status == PlaybackStatus::Playing {
+            break;
+        }
+    }
```
gasmith, line 226: Nice use for a condition variable, though maybe overkill for an example.

rust/examples/ws_stream_csv/src/main.rs
```diff
@@ -0,0 +1,349 @@
...
+    // Create the channel for publishing CSV data
+    // Using JSON encoding with a simple schema
+    let channel = ChannelBuilder::new("/csv_data")
+        .message_encoding("json")
+        .schema(Schema::new(
+            "CsvDataPoint",
+            "jsonschema",
+            r#"{"type":"object","properties":{"timestamp":{"type":"integer"},"value":{"type":"number"}}}"#.as_bytes(),
+        ))
+        .build_raw()
```
gasmith, line 329: I realize we're trying to "show one thing" in this example, but I feel like we could eliminate a bit of boilerplate if we derived `foxglove::Encode` for `DataPoint` and declared a typed channel here. Also, typed channels are preferred in a program where you have control over the message types.

rust/examples/ws_stream_csv/src/main.rs
```diff
@@ -0,0 +1,349 @@
...
+        );
+
+        // Log the data to the channel with the data's timestamp
+        channel.log_with_meta(
+            json_payload.as_bytes(),
+            PartialMetadata::with_log_time(point.timestamp),
+        );
+
+        // IMPORTANT: Broadcast time after each message so Foxglove knows the current time
+        server.broadcast_time(point.timestamp);
```
gasmith, line 190: I think we actually want to broadcast time before sending the message. Otherwise the app receives the message "in the past" before its clock is updated. Not sure if that has any practical impact, but maybe worth looking into. When you run this, do the timestamps in the plot panel line up the way you expect?

rust/examples/ws_stream_csv/src/main.rs
```diff
@@ -0,0 +1,349 @@
...
+        if inner.status != PlaybackStatus::Playing {
+            last_wall_time = Instant::now();
+            continue;
+        }
+
+        // Calculate how much playback time has elapsed based on wall time and speed
+        let wall_elapsed = last_wall_time.elapsed();
+        let playback_elapsed_ns = (wall_elapsed.as_nanos() as f64 * inner.speed as f64) as u64;
+        let new_playback_time =
+            last_playback_time.unwrap_or(inner.current_time) + playback_elapsed_ns;
```
gasmith, line 271: No love for the TimeTracker from the ws_stream_mcap example, huh? ;)

rust/examples/ws_stream_csv/src/main.rs
```diff
@@ -0,0 +1,349 @@
...
+        );
+
+        // IMPORTANT: Broadcast time after each message so Foxglove knows the current time
+        server.broadcast_time(point.timestamp);
+
+        *current_index += 1;
+    }
+}
+
+fn update_current_time(
```
gasmith, line 196: Nit: Maybe better as a method on the controller?

rust/examples/ws_stream_csv/src/main.rs
```diff
@@ -0,0 +1,349 @@
...
+                seek_time.clamp(controller.inner.time_range.0, controller.inner.time_range.1);
+            controller.inner.did_seek = true;
+        }
+
+        // Return the new playback state with the request_id for correlation
+        Some(controller.to_playback_state(Some(request.request_id)))
+    }
+}
+
+/// Main playback loop that publishes data and time updates.
```
gasmith, line 151: This comment seems misplaced.

rust/examples/ws_stream_csv/src/main.rs
```diff
@@ -0,0 +1,349 @@
...
+fn handle_seek(
+    data: &[DataPoint],
+    current_time: u64,
+    current_index: &mut usize,
+    last_wall_time: &mut Instant,
+    last_playback_time: &mut Option<u64>,
+) {
+    *current_index = data
+        .iter()
+        .position(|d| d.timestamp >= current_time)
```
gasmith, line 161: Since we're sorted: https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search_by

gasmith (commented):
I like how the CSV-specific part is actually pretty small and self-contained. The remainder of the example just operates on a `Vec<DataPoint>`. That's nice.

I sort of wish that these examples built on each other, though. In the abstract, there's nothing different about reading records from an MCAP file, vs reading them from a (csv-populated) Vec, vs reading them from a parquet file. I think it'd be useful to write this as a common harness for playing back file data, and then have a couple backends for the underlying file formats. The real value of this example is in showing people how to write that harness.

Edit: maybe that means, for MCAP, only supporting a particular channel with a particular schema and including the sample data in the repo, as you've done here for the CSV file.

ericmlujan:
Closing as obsoleted

PR #842, opened 2026-01-29
Revert CI, ros changes

gasmith:
### Changelog
None

### Docs
None

### Description
Revert unnecessary changes to build scripts. We'll revisit these when we get around to plumbing these features to the C++ bindings.

### DIFF
--- a/Container.mk
+++ b/Container.mk
@@ -42,22 +42,27 @@ docs-python:
 clean-docs-python:
 	rm -rf python/foxglove-sdk/python/docs/_build
 
+# Use clang-19 for Rust builds that include livekit (anything with --all-targets
+# or --all-features). This is required because webrtc-sys uses compiler flags
+# like -Wno-changes-meaning that aren't available in gcc-12.
+CLANG_19 := CC=clang-19 CXX=clang++-19
+
 .PHONY: lint-rust
 lint-rust:
 	cargo fmt --all --check
-	cargo clippy --no-deps --all-targets --tests -- -D warnings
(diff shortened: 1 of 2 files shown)


.github/workflows/ros.yml
```diff
@@ -21,16 +21,18 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      - name: Build Foxglove C++ SDK
+        run: make build-cpp
```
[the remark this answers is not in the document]
gasmith: It's containerized.

.github/workflows/ros.yml
```diff
@@ -21,16 +21,18 @@ jobs:
...
+
       - name: Run make lint
         run: make lint
         working-directory: ros
 
-      # The C++ SDK is built inside the Docker container to ensure glibc compatibility.
-      # Each ROS distribution uses a different Ubuntu version with different glibc.
       - name: Run make docker-build
         run: make USE_LOCAL_SDK=ON docker-build-${{ matrix.ros_distribution }}
```
eloff, line 32: I think we can remove USE_LOCAL_SDK here, it's not used anymore if I'm correct

ros/Makefile
eloff, line 2: Can remove this too

eloff (approved):
LGTM

.github/workflows/ros.yml
gasmith: It is being used. In CI, we do `make build-cpp`, which drops staticlibs under cpp/build, which get volume-mounted into the ros build container along with the rest of the SDK repo. We want the ros build to use these, rather than downloading them from the latest SDK release from github.

PR #851, opened 2026-01-30
Added schemars-chrono feature

joe-saronic:
### Changelog

A new `schemars-chrono` Cargo feature enables the `chrono04` feature on the underlying `schemars` dependency so that we can have automatic schema generation for types including `chrono` timestamps.

### Docs

None

### Description

Building with `schemars` feature is nice, but sometimes you need `chrono` support too. Enabling `chrono` feature does not enable `chorno04` on the underlying `schemars` dependency, so you miss out on some possibilities. The new feature ensures that you can generate schemas with `chrono` date-time types.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2409,6 +2409,7 @@ version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "82d20c4491bc164fa2f6c5d44565947a52ad80b9505d8e36f8d54c27c739fcd0"
 dependencies = [
+ "chrono",
  "dyn-clone",
  "ref-cast",
  "schemars_derive",


rust/foxglove/Cargo.toml
```diff
@@ -32,6 +32,7 @@ tls = [
 ]
 stream = ["dep:futures"]
 schemars = ["dep:schemars"]
+schemars-chorono = ["schemars", "chrono", "dep:schamars/chrono04"]
```
sofuture: Is this supposed to be the following?
```suggestion
schemars-chrono = ["schemars", "chrono", "dep:schemars/chrono04"]
```

joe-saronic: Yes. I've noticed that I consistently spell `chrono` weird at small hours. This isn't the first time :)

rust/foxglove/Cargo.toml
```diff
@@ -32,6 +32,7 @@ tls = [
 ]
 stream = ["dep:futures"]
 schemars = ["dep:schemars"]
+schemars-chrono = ["schemars", "chrono", "dep:schemars/chrono04"]
```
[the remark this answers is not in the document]
gasmith: Oops, that's a nice catch. Let's fix that.

joe-saronic:
```suggestion
schemars-chrono = ["schemars", "chrono", "schemars/chrono04"]
```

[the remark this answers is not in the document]
joe-saronic: Done

joe-saronic:
Looks like I'll need to update cargo.lock somewhere

PR #852, opened 2026-01-30
derive: Implement ProtobufField for other well-known types

gasmith:
### Changelog
- rust: Added derive(Encode) support for more well-known types

### Docs
None

### Description
When we implemented the derive macro for `Encode`, we didn't implement it for the `Timestamp` and `Duration` well-known types, their chrono equivalents, nor `Option<T> where T: Encode`. This change implements `ProtobufField` for these types, so that the can be used as fields in a struct that derives `Encode`.

### DIFF
--- a/rust/foxglove_derive/src/lib.rs
+++ b/rust/foxglove_derive/src/lib.rs
@@ -8,31 +8,51 @@ use syn::{
     GenericArgument, GenericParam, Generics, PathArguments, Type,
 };
 
+/// Extract the inner type from a wrapper type like `Vec<T>` or `Option<T>`.
+/// Returns the wrapper name and inner type if it matches the pattern.
+fn unwrap_generic_type<'a>(ty: &'a Type, wrapper: &str) -> Option<&'a Type> {
+    let Type::Path(type_path) = ty else {
+        return None;
+    };
+    let segment = type_path.path.segments.last()?;
+    if segment.ident != wrapper {
+        return None;
+    }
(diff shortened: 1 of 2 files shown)


rust/foxglove/src/schemas/data/Duration.bin
joe-saronic, line 1: Just to make sure, are these checked in intentionally instead of coming from a dependency?

rust/foxglove_derive/Cargo.toml
```diff
@@ -18,7 +18,8 @@ prost-types.workspace = true
 
 [dev-dependencies]
 bytes = "1.4"
-foxglove = { path = "../foxglove" }
+chrono = "0.4"
```
joe-saronic, line 21: Why not put chrono behind a feature here too?

rust/foxglove/src/schemas/data/Duration.bin
gasmith: I don't think any of our existing protobuf crates expose the descriptors for the well-known types. `prost-reflect` might, but I don't want to add a new dependency for this. It's straightforward to just write out the descriptor as we're generating the descriptors for the other foxglove protos.

rust/foxglove_derive/Cargo.toml
gasmith: These are dev dependencies. For running tests, I'd like to make sure we cover chrono by default.

eloff (approved):
LGTM

PR #854, opened 2026-01-30
derive: Implement ProtobufField for foxglove schemas

gasmith:
### Changelog
- rust: Added derive(Encode) support for structs with nested foxglove schemas

### Docs
None

### Description
We currently don't support deriving `Encode` for structs that contain nested foxglove schema types, which is a bit of a shame. This change enables that support by implementing `ProtobufField` for foxglove schemas.

### DIFF
--- a/rust/foxglove_derive/tests/foxglove_schema_test.rs
+++ b/rust/foxglove_derive/tests/foxglove_schema_test.rs
@@ -1,0 +1,188 @@
+use bytes::BytesMut;
+use foxglove::schemas::{Log, Point3, Pose, Quaternion, Timestamp, Vector3};
+use foxglove::Encode;
+use prost::Message;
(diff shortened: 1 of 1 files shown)


rust/foxglove_derive/tests/foxglove_schema_test.rs
```diff
@@ -0,0 +1,188 @@
...
+struct MessageWithPose {
+    pose: Pose,
+    name: String,
+}
+
+/// Test Vec of foxglove schema types.
+#[derive(Encode)]
+struct MessageWithPoints {
+    points: Vec<Point3>,
+}
```
eloff, line 25: This is cool that we support custom types containing our WKTs

eloff (approved):
LGTM

PR #859, opened 2026-02-04
New example for mcap using PlaybackControl

ericmlujan:
### Changelog
- Add Rust example for using the `RangedPlayback` capability

### Docs
- This is an example.

### Description
Add a Rust example using the RangedPlayback websocket capability to load and play back an mcap file. This example introduces a documented `PlaybackSource` trait, which others can pattern match to when implementing their own loaders for custom data formats.

This PR supersedes and simplifies #772.

A Python version of this example exists at #862.
A C++ version of this example exists at #861.

Fixes: FLE-131

### DIFF
--- a/rust/examples/ws_stream_mcap/src/main.rs
+++ b/rust/examples/ws_stream_mcap/src/main.rs
@@ -48,7 +48,10 @@ impl ServerListener for Listener {
 
         // Handle seek first, before play/pause. This is important for looping behavior,
         // where Foxglove sends a seek to the beginning followed by a Play command.
-        // Setting this flag to true clears panels in the Foxglove player. For simplicity, we set this every time a seek is requested from Foxglove. In your application, consider implementing logic that determines whether a seek represents a significant jump in time for the data you're playing back.
+        // Setting this flag to true clears panels in the Foxglove player. For simplicity, we set
+        // this every time a seek is requested from Foxglove. In your application, consider
+        // implementing logic that determines whether a seek represents a significant jump in time
+        // for the data you're playing back.
         let mut did_seek = request.seek_time.is_some();
 
         if let Some(seek_time) = request.seek_time {


rust/examples/ws_stream_mcap/src/main.rs
```diff
@@ -57,256 +91,56 @@ fn main() -> Result<()> {
...
-        }
-
-        // Seek to summary section.
-        let summary_start = buf.as_slice().get_u64_le();
-        if summary_start == 0 {
-            return Err(anyhow!("missing summary section"));
-        }
-        file.seek(SeekFrom::Start(summary_start))
-            .context("seek summary")?;
+        let status = { mcap_player.lock().unwrap().status() };
```
ericmlujan: @cursoragent combine acquiring this lock on player with the acquisition below when we broadcast the ended state

rust/examples/ws_stream_mcap/src/main.rs
```diff
@@ -57,256 +91,60 @@ fn main() -> Result<()> {
...
+
+    let mcap_player = McapPlayer::new(&args.file)?;
+    let (start_time, end_time) = mcap_player.time_range();
+
+    let mcap_player = Arc::new(Mutex::new(mcap_player));
+    let listener = Arc::new(Listener::new(mcap_player.clone()));
+
     let server = WebSocketServer::new()
         .name(file_name)
         .capabilities([Capability::Time])
```
gasmith: I think we should explicitly add the RangedPlayback capability here, to be more transparent about the capabilities we're advertising.

```suggestion
        .capabilities([Capability::RangedPlayback, Capability::Time])
```

rust/examples/ws_stream_mcap/src/main.rs
```diff
@@ -1,26 +1,63 @@
...
+}
+
+impl Listener {
+    fn new(player: Arc<Mutex<dyn Send + PlaybackSource>>) -> Self {
+        Self { player }
+    }
+}
+
+impl ServerListener for Listener {
+    fn on_playback_control_request(
```
gasmith, line 43: Since this example is about illustrating this interface in particular, we should include a brief commentary about this method, and how we've chosen to implement it here.

rust/examples/ws_stream_mcap/src/main.rs
```diff
@@ -1,26 +1,63 @@
...
+        match request.playback_command {
+            PlaybackCommand::Play => player.play(),
+            PlaybackCommand::Pause => player.pause(),
+        };
+
+        Some(PlaybackState {
+            current_time: player.current_time(),
+            playback_speed: player.playback_speed(),
+            status: player.status(),
+            did_seek: request.seek_time.is_some(),
```
gasmith: Should this be `true` even when the specified seek is a no-op (e.g., due to clamping)?

rust/examples/ws_stream_mcap/src/main.rs
```diff
@@ -1,26 +1,63 @@
...
+
+impl ServerListener for Listener {
+    fn on_playback_control_request(
+        &self,
+        request: PlaybackControlRequest,
+    ) -> Option<PlaybackState> {
+        let mut player = self.player.lock().unwrap();
+
+        if let Some(seek_time) = request.seek_time {
+            player.seek(seek_time).ok()?
```
gasmith: We should log the error here.

```suggestion
            if let Err(err) = player.seek(seek_time) {
                tracing::warn!("failed to seek: {err:?}");
                return None;
            }
```

rust/examples/ws_stream_mcap/src/main.rs
```diff
@@ -57,256 +91,60 @@ fn main() -> Result<()> {
...
+        let status = {
+            let player = mcap_player.lock().unwrap();
+            let status = player.status();
+
+            // Broadcast state change when playback ends
+            if status == PlaybackStatus::Ended && last_status != PlaybackStatus::Ended {
+                server.broadcast_playback_state(PlaybackState {
+                    current_time: player.current_time(),
+                    playback_speed: player.playback_speed(),
+                    status: player.status(),
```
gasmith:
```suggestion
                    status,
```

rust/examples/ws_stream_mcap/src/main.rs
```diff
@@ -57,256 +91,60 @@ fn main() -> Result<()> {
...
-    /// Sleeps until the specified offset.
-    fn sleep_until(&mut self, offset_ns: u64) {
-        let abs = Duration::from_nanos(offset_ns.saturating_sub(self.offset_ns));
-        let delta = abs.saturating_sub(self.start.elapsed());
-        if delta >= Duration::from_micros(1) {
-            std::thread::sleep(delta);
+        // Log next message, sleeping outside the lock if needed
+        let sleep_duration = mcap_player.lock().unwrap().log_next_message(&server)?;
+        if let Some(duration) = sleep_duration {
+            std::thread::sleep(duration);
```
gasmith: Super-minor-nit, since the inter-message time on an mcap file is generally very short, but the player will be unresponsive during this sleep. Maybe clamp it to a reasonable max.

rust/examples/ws_stream_mcap/src/mcap_player.rs
```diff
@@ -0,0 +1,342 @@
...
+        }
+        self.status = PlaybackStatus::Paused;
+    }
+
+    fn seek(&mut self, log_time: u64) -> Result<()> {
+        let log_time = log_time.clamp(self.time_range.0, self.time_range.1);
+        self.current_time = log_time;
+        self.reset_reader(log_time)?;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
gasmith: Maybe just move the `current_time` assignment into `reset_reader`.

gasmith (approved):
Looks great. The trait does a lot to improve the readability. I have a few comments, but they're mostly nit-picky.

rust/examples/ws_stream_mcap/src/main.rs
ericmlujan: I left a comment about this in the code in 0a3cee44b44603a2f011f200d9639b424ca97704. For simplicity in the context of an example, I'm assuming every time the user seeks in Foxglove, they want panels to clear. A more sophisticated application would determine whether a given seek didn't cause any messages to be skipped and set to false.

rust/examples/ws_stream_mcap/src/main.rs
ericmlujan: Resolved in 0a3cee44b44603a2f011f200d9639b424ca97704

rust/examples/ws_stream_mcap/src/main.rs
ericmlujan: Resolved in 0a3cee44b44603a2f011f200d9639b424ca97704

rust/examples/ws_stream_mcap/src/main.rs
ericmlujan: Resolved in 0a3cee44b44603a2f011f200d9639b424ca97704

rust/examples/ws_stream_mcap/src/mcap_player.rs
```diff
@@ -0,0 +1,355 @@
...
+            tt.set_speed(speed);
+        }
+        self.playback_speed = speed;
+    }
+
+    fn play(&mut self) {
+        if let Some(tt) = &mut self.time_tracker {
+            tt.resume();
+        }
+        self.status = PlaybackStatus::Playing;
```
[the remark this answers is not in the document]
ericmlujan: @cursoragent please fix this bug by adding a check for `Ended` in `play()`. Make sure the ordering in the listener is correct, since Foxglove implements looping by sending a seek to the beginning of the time range combined with a Play command

rust/examples/ws_stream_mcap/src/mcap_player.rs
```diff
@@ -0,0 +1,355 @@
...
+        self.current_time = start_time;
+        self.reader = IndexedReader::new_with_options(
+            &self.summary,
+            IndexedReaderOptions::new().log_time_on_or_after(start_time),
+        )
+        .map_err(|e| anyhow!("failed to create indexed reader: {e}"))?;
+        self.time_tracker = None;
+        self.pending_message = None;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
ericmlujan: @cursoragent resolve this by reordering the lines

rust/examples/ws_stream_mcap/src/mcap_player.rs
```diff
@@ -0,0 +1,364 @@
...
+        }
+        self.status = PlaybackStatus::Playing;
+    }
+
+    fn pause(&mut self) {
+        if let Some(tt) = &mut self.time_tracker {
+            tt.pause();
+        }
+        self.status = PlaybackStatus::Paused;
+    }
```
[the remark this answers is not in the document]
ericmlujan: @cursoragent add the guard here too

rust/examples/ws_stream_mcap/src/mcap_player.rs
```diff
@@ -0,0 +1,369 @@
...
+impl PlaybackSource for McapPlayer {
+    fn time_range(&self) -> (u64, u64) {
+        self.time_range
+    }
+
+    fn status(&self) -> PlaybackStatus {
+        self.status
+    }
+
+    fn current_time(&self) -> u64 {
```
dave-foxglove: For all references to a u64 as the player time, can we define and use a named type instead of a raw u64.

Maybe a "newtype" like `McapTimestamp` and `McapDuration` (similar to rust standard timestamps and durations).

P.S. I'm also up for a "type alias" like `type McapNanos = u64` but it doesn't offer the whoops-protection a newtype does.

dave-foxglove: For all references to a u64 as the player time, can we define and use a named type instead of a raw u64.

Maybe a "newtype" like `McapTimestamp` and `McapDuration` (similar to rust standard timestamps and durations).

P.S. I'm also up for a "type alias" like `type McapNanos = u64` but it doesn't offer the whoops-protection a newtype does.

rust/examples/ws_stream_mcap/src/playback_source.rs
```diff
@@ -0,0 +1,61 @@
...
+    ///
+    /// Used to send a PlaybackState to Foxglove
+    fn current_time(&self) -> u64;
+
+    /// Returns the current playback speed multiplier.
+    ///
+    /// Used to send a PlaybackState to Foxglove
+    fn playback_speed(&self) -> f32;
+
+    /// Logs the next message for playback if it's ready, or returns a duration to wait.
```
dave-foxglove: Sorry if I'm missing something, but could/should this mention something about playback speed? E.e. if the playback speed is 10x should I sleep for 1/10 the returned Duration (or is it already accounting for that?)

rust/examples/ws_stream_mcap/src/playback_source.rs
```diff
@@ -0,0 +1,61 @@
+use std::time::Duration;
+
+use anyhow::Result;
+use foxglove::{websocket::PlaybackStatus, WebSocketServerHandle};
+
+/// A data source that supports ranged playback with play/pause, seek, and variable speed.
+///
```
dave-foxglove, line 10: It looks like `log_next_message` is expected to be called in a loop(?). If true, let's mention that in the overall doc comment.

rust/examples/ws_stream_mcap/src/playback_source.rs
```diff
@@ -0,0 +1,61 @@
...
+    fn playback_speed(&self) -> f32;
+
+    /// Logs the next message for playback if it's ready, or returns a duration to wait.
+    ///
+    /// Returns `Ok(Some(duration))` if the caller should sleep before calling again.
+    /// Returns `Ok(None)` if a message was logged or playback is not active.
+    ///
+    /// The caller should sleep outside of any lock to allow control requests to be processed.
+    /// This method also broadcasts time updates via `server.broadcast_time()`.
+    fn log_next_message(&mut self, server: &WebSocketServerHandle) -> Result<Option<Duration>>;
```
dave-foxglove, line 65: I'm likely missing some context in the wider code base, but in just this interface/api - I'm not clear on why this functions has a "log_*" prefix on it. If it's just pushing a message to a log file, then it's perfectly named, but I have an impression that it's doing something more or different(?)

rust/examples/ws_stream_mcap/src/mcap_player.rs
```diff
@@ -0,0 +1,369 @@
...
+    time_tracker: Option<TimeTracker>,
+    time_range: (u64, u64),
+    status: PlaybackStatus,
+    current_time: u64,
+    playback_speed: f32,
+    /// Buffered message that was read but not yet ready to emit.
+    pending_message: Option<(mcap::records::MessageHeader, Vec<u8>)>,
+}
+
+impl McapPlayer {
```
dave-foxglove, line 37: Can we describe what an McapPlayer is and what it is used for (asking for doc comments), either here, or on the struct (line 19) or at the top of the file (i.e. `//! comments`).

ericmlujan: Addressed in 22218a6d37271136cadeca001f92c67f3293f71b

rust/examples/ws_stream_mcap/src/playback_source.rs
ericmlujan: Addressed in 22218a6d37271136cadeca001f92c67f3293f71b

rust/examples/ws_stream_mcap/src/playback_source.rs
ericmlujan: Addressed in 22218a6d37271136cadeca001f92c67f3293f71b. It is already accounting for that

rust/examples/ws_stream_mcap/src/mcap_player.rs
ericmlujan: Addressed in 25abf130e5d912d2014158fe67023031f2efeeb3. I decided against introducing a full tuple struct for nanoseconds and duration; I found it added a bit too much boilerplate in the context of a code example.

rust/examples/ws_stream_mcap/src/playback_source.rs
ericmlujan: Good question. ["Logging"](https://docs.foxglove.dev/docs/sdk/logging-messages) is the general term for sending your robotics data over to Foxglove. We have the notion of a "channel", which you can think of as a topic that data messages can be logged to/sent over, and a channel can be backed by one or more sinks, which handle actually doing something with that data (i.e. saving to an MCAP file or sending over a WebSocket). In this case, this function calls a `Channel::log()` on a bunch of channels that are set up to publish the data over a WebSocket; hence the name.

PR #861, opened 2026-02-07
C++ example for mcap using PlaybackControl

ericmlujan:
### Changelog
- Add C++ example for using the `PlaybackControl` capability

### Docs
- This is an example.

### Description
Add a C++ example using the PlaybackControl websocket capability to load and play back an mcap file. This example introduces an annotated `PlaybackSource` base class, which others can pattern match to when implementing their own loaders for custom data formats.

This PR ports #859.

### DIFF
--- a/cpp/foxglove/include/foxglove/server.hpp
+++ b/cpp/foxglove/include/foxglove/server.hpp
@@ -96,9 +96,8 @@ inline WebSocketServerCapabilities operator|(
   // We have to disable the EnumCastOutOfRange check here since the current version of
   // clang-analyzer doesn't yet support exempting enums marked as bitflags. See:
   // https://github.com/llvm/llvm-project/issues/76208
-  return WebSocketServerCapabilities(
-    uint8_t(a) | uint8_t(b)
-  );  // NOLINT(clang-analyzer-optin.core.EnumCastOutOfRange)
+  // NOLINTNEXTLINE(clang-analyzer-optin.core.EnumCastOutOfRange)
+  return WebSocketServerCapabilities(uint8_t(a) | uint8_t(b));
 }
 
 /// @brief Check if a capability is set.


cpp/examples/ws-stream-mcap/src/mcap_player.cpp
```diff
@@ -0,0 +1,198 @@
...
+  }
+  status_ = foxglove::PlaybackStatus::Playing;
+}
+
+void McapPlayer::pause() {
+  if (time_tracker_.has_value()) {
+    time_tracker_->pause();
+  }
+  status_ = foxglove::PlaybackStatus::Paused;
+}
```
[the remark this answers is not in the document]
ericmlujan: @cursoragent fix this issue in pause()

cpp/examples/ws-stream-mcap/src/main.cpp
```diff
@@ -0,0 +1,185 @@
...
+  auto player_ptr = std::shared_ptr<McapPlayer>(std::move(player));
+
+  foxglove::WebSocketServerOptions options = {};
+  options.name = server_name;
+  options.host = host;
+  options.port = port;
+  // Explicitly advertise capabilities used by this example.
+  // Advertise Time updates and playback time range (inclusive start/end, in nanoseconds
+  // since epoch). Setting playback_time_range implies PlaybackControl capability.
+  options.capabilities = foxglove::WebSocketServerCapabilities::Time;
```
gasmith: Nit: "Explicitly advertise capabilities" seems at odds with "Setting playback_time_range implies PlaybackControl". Let's just explicitly advertise both capabilities, and move the commentary about the playback time range to live with that assignment.
```suggestion
  // PlaybackControl: allows clients to send play/pause/seek/speed requests.
  // Time: allows the server to broadcast the current playback timestamp to clients.
  options.capabilities = foxglove::WebSocketServerCapabilities::PlaybackControl |
                         foxglove::WebSocketServerCapabilities::Time;
  // The playback time range tells clients the start/end bounds of the data (nanoseconds).

```

cpp/examples/ws-stream-mcap/src/mcap_player.cpp
```diff
@@ -0,0 +1,201 @@
...
+  return true;
+}
+
+void McapPlayer::resetMessageView(uint64_t start_time) {
+  iterator_.reset();
+  message_view_.reset();
+
+  mcap::ReadMessageOptions opts;
+  opts.startTime = start_time;
+  opts.endTime = time_range_.second + 1;
```
gasmith, line 89: Maybe a comment about why we need the +1.

ericmlujan:
Note: had to disable the `EnumCastOutOfRange` check in 4dafec1c8d331753335623883ab7c4e152fa84ae to pass clang-analyzer. This is because the checker sees us doing `Capability::PlaybackControl | Capability::Time` (which computes to 72), which is not explicitly represented in the enum, because we're using it as a bitflag. 

This is a [known limitation](https://github.com/llvm/llvm-project/issues/76208), and the appropriate fix (which is marking the enum as `flag_enum` and the analyzer respecting that use pattern) hasn't landed in our version of clang.

PR #862, opened 2026-02-07
Python example for mcap using PlaybackControl

ericmlujan:
### Changelog
- Add Python example for using the `PlaybackControl` capability

### Docs
- This is an example.

### Description
Add a Python example using the PlaybackControl websocket capability to load and play back an mcap file. This example introduces an annotated `PlaybackSource` ABC, which others can pattern match to when implementing their own loaders for custom data formats.

This PR ports #859.

### DIFF
--- a/python/foxglove-sdk-examples/ws-playback-control-mcap/mcap_player.py
+++ b/python/foxglove-sdk-examples/ws-playback-control-mcap/mcap_player.py
@@ -157,6 +157,8 @@ class McapPlayer(PlaybackSource):
         self._status = PlaybackStatus.Playing
 
     def pause(self) -> None:
+        if self._status == PlaybackStatus.Ended:
+            return
         if self._time_tracker is not None:
             self._time_tracker.pause()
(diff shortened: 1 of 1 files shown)


python/foxglove-sdk-examples/ws-playback-control-mcap/mcap_player.py
```diff
@@ -0,0 +1,281 @@
...
+        # To restart playback, the caller must seek first.
+        if self._status == PlaybackStatus.Ended:
+            return
+        if self._time_tracker is not None:
+            self._time_tracker.resume()
+        self._status = PlaybackStatus.Playing
+
+    def pause(self) -> None:
+        if self._time_tracker is not None:
+            self._time_tracker.pause()
```
gasmith, line 163: Nit: The rust version is a no-op when playback has ended.
```suggestion
    def pause(self) -> None:
        if self._status == PlaybackStatus.Ended:
            return
        if self._time_tracker is not None:
            self._time_tracker.pause()
```

PR #865, opened 2026-02-10
introduce optional fields and mark LocationFix 'color' as optional

indirectlylit:
### Changelog

Make `color` field optional in `LocationFix` messages

### Docs

updated in PR

### Description

In https://github.com/foxglove/foxglove-sdk/pull/556 we added a new `color` field to `LocationFix` messages. However, introducing this meant that users could not produce messages which rely on Foxglove's automatic color assignment logic.

This PR adds a new `optional` attribute to the `FoxgloveMessageField` type and applies this to the `color` field.

For ROS, we continue requiring users to provide those values. This is documented in the schemas README


mjsobrep:
Optional fields will also be needed for:
- lineage
- Map heading w/ magnitude
- map shape

Seems like we definitely need optional fields...

typescript/schemas/src/internal/generateFlatbufferSchema.test.ts
```diff
@@ -93,22 +93,25 @@ describe("generateFlatbuffers", () => {
         /// length 3
         field_string_fixed_array:[string] (id: 20);
 
+        /// An optional string field
+        field_optional_string:optional string (id: 21);
+
         /// An enum field
-        field_enum:ExampleEnum (id: 21);
+        field_enum:ExampleEnum (id: 22);
```
mjsobrep: is it preferable to put the new field at the end of the list so we aren't changing IDs?

schemas/README.md
```diff
@@ -4,6 +4,12 @@ See [Foxglove Schemas documentation](https://docs.foxglove.dev/docs/visualizatio
 
 All schemas are generated from [schemas.ts](/internal/schemas.ts).
 
+## Optional fields
+
+Some message fields are **optional**: they may be omitted and the consumer (e.g. a Foxglove panel) will use a default. Optional fields are marked in the table below.
```
defunctzombie:
```suggestion
**optional** message fields may be omitted. When omitted, the consumer (e.g. a Foxglove panel) determines the behavior.
```

schemas/README.md
```diff
@@ -4,6 +4,12 @@ See [Foxglove Schemas documentation](https://docs.foxglove.dev/docs/visualizatio
...
+
+**IDLs without optional** (e.g. ROS 1 .msg): all fields must still be present in the message. To get the same effect as omitting an optional field, use an empty string or leave the nested message absent if the format allows.
```
defunctzombie:
```suggestion
If the IDL does not support optional fields (e.g. ROS 1 .msg) you must specify a value for the field.
```

defunctzombie (approved):
LGTM - few notes on some readme language. I'd get another stamp given this is a schemas level change that has broad impact.

I didn't see any rust changes in this change - is that something we get for free with how proto is handled? Like I can log a LocationFix and omit a color now via the sdk?

indirectlylit:
> I didn't see any rust changes in this change - is that something we get for free with how proto is handled? Like I can log a LocationFix and omit a color now via the sdk?

I believe this was already possible, because Prost already mapped fields to `Option<T>`

https://github.com/foxglove/foxglove-sdk/blob/37a902680978e48c79bb2007147b369b4cbe67df/rust/foxglove/src/schemas/foxglove.rs#L554-L556

indirectlylit:
related cc @james-rms - https://github.com/foxglove/foxglove-sdk/pull/699

should we start marking some of these as optional?

typescript/schemas/src/internal/generateJsonSchema.ts
```diff
@@ -92,6 +92,6 @@ export function generateJsonSchema(schema: FoxgloveMessageSchema): Record<string
     $comment: "Generated by https://github.com/foxglove/foxglove-sdk",
     type: "object",
     properties,
-    required: schema.fields.map((field) => field.name),
+    required: schema.fields.filter((f) => !f.optional).map((f) => f.name),
```
jesse-foxglove: I vaguely remember that we prefer no single letter variables but not sure if this is an exception

jesse-foxglove (approved):
Had a minor comment but it looks good

typescript/schemas/src/internal/generateProto.ts
```diff
@@ -93,6 +93,8 @@ export function generateProto(
     const qualifiers: string[] = [];
     if (field.array != undefined) {
       qualifiers.push("repeated");
+    } else if (field.optional) {
+      qualifiers.push("optional");
     }
```
jtbandes, line 102: are these meant to be mutually exclusive?

edit: I think it may be correct that you can't have `repeated` and `optional` on the same field, but that should probably cause an error and not silently pick one of the two

typescript/schemas/src/internal/generateOmgIdl.test.ts
```diff
@@ -120,6 +120,15 @@ describe("generateOmgIdl", () => {
...
         sequence<NestedMessage> field_nested_array;
+
+        // An optional string field
+        string field_optional_string;
+
+        // An optional bool field
+        boolean field_optional_bool;
+
+        // An optional float64 field
+        double field_optional_float64;
```
jtbandes, line 134: is there a reason not to use `@optional` here? cc @snosenzo  https://github.com/foxglove/omgidl/blob/c53f902f741139c0040961aa14c85939412a1e2c/packages/omgidl-serialization/src/MessageReader.test.ts#L1200-L1201

typescript/schemas/src/internal/generateProto.test.ts
```diff
@@ -106,6 +106,15 @@ describe("generateProto", () => {
         // long
         // description
         repeated foxglove.NestedMessage field_nested_array = 4;
+
+        // An optional string field
+        optional string field_optional_string = 26;
```
jtbandes, line 111: 🤔   Interestingly, the protobuf docs suggest that `optional` should always be used (rather than the implicit behavior): https://protobuf.dev/programming-guides/proto3/#field-labels

typescript/schemas/src/internal/generateProto.ts
indirectlylit: updated to make them independent, so we don't need to start validating inputs

typescript/schemas/src/internal/generateOmgIdl.test.ts
indirectlylit: updated to use `@optional`

snosenzo (approved):

PR #868, opened 2026-02-11
playground: add support for symbol highlighting & renaming

jtbandes:
### Changelog

playground: improved ability to highlight and rename variables in the editor.

### Docs

None

### Description

Related: https://github.com/davidhalter/jedi/issues/2086

All occurrences of text were highlighted (default behavior of monaco). The F2 key was not available for renaming a symbol (only cmd+F2 which renames all textual occurrences).

image

</td><td>

- Clicking a symbol highlights other occurrences of that symbol
- F2 now works to rename the symbol
- Pressing F2 on something like a numeric constant displays an error message (like VS Code).

image

image

</td></tr></table>

### DIFF
--- a/playground/src/Runner.ts
+++ b/playground/src/Runner.ts
@@ -62,11 +62,7 @@ export class Runner extends EventEmitter<EventMap> {
     return await this.#remote.getHover(code, line, col);
   }
 
-  async getReferenceRanges(
-    code: string,
-    line: number,
-    col: number,
-  ): Promise<monaco.IRange[] | undefined> {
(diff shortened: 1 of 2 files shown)


defunctzombie (approved):
I'm somewhat surprised we have to do so much here and its not part of some "language server" package or behavior.

jtbandes:
We could try to set up a proper language server off the shelf, but it [seems more complicated](https://github.com/TypeFox/monaco-languageclient/discussions/561) to try and run a separate server process thingy (and some [packages are unmaintained](https://www.npmjs.com/package/@typefox/pyright-browser)) vs just running some functions in the existing pyodide context.

PR #870, opened 2026-02-11
Add optional shape enums in sdk

thswansonfxg:
### Changelog

Add point_style optional enum field to LocationFix schema, allowing users to customize the marker style (dot, diamond, square, plus, cross, pin) when visualizing locations on a map. 

### Docs

The generated schemas/README.md includes the new PointStyle enum  and point_style field documentation. No additional docs PR is needed.                                                           
                  

### Description

This PR adds a new PointStyle enum and optional point_style field to the LocationFix schema across all SDK bindings (Rust,
  Python, C++, C, TypeScript, ROS, Protobuf, FlatBuffer, JSON Schema, OMG IDL).

  Changes:

  - New PointStyle enum with values: DOT (default, 0), DIAMOND (1), SQUARE (2), PLUS (3), CROSS (4), PIN (5)
  - point_style field on LocationFix — optional in all bindings (Option<i32> in Rust, std::optional in C++, *const pointer in C, Option in Python)
  - Generator fix (generatePyclass.ts) — Python SDK codegen was not handling optional enum fields, generating field as i32 instead of field.map(|s| s as i32)
  - Generator fix (generateSdkCpp.ts) — Support multiple enums per parent schema, handle optional enum to conversion with nullptr for absent values
  - Generator fix (generateSdkRustCTypes.ts) — Optional enum fields in C FFI now use pointer types for proper nullability
  - Serde support (schemas.rs) — Added serde_enum_mod_optional! macro for serializing/deserializing Option<i32> enum fields as string names in JSON and raw i32 in binary formats, with round-trip tests
  - Proto gen fix (foxglove_proto_gen) — Detects proto3_optional enum fields and emits the optional serde macro variant

https://linear.app/foxglove/issue/ERT-1612/add-optional-attribute-to-control-point-shape-in-locationfix-messages


rust/foxglove/src/schemas.rs
```diff
@@ -130,6 +130,58 @@ macro_rules! serde_enum_mod {
 #[cfg(feature = "serde")]
 pub(crate) use serde_enum_mod;
 
+/// Generates a serde module for an optional protobuf enum field.
+///
+/// Uses string names for human-readable formats (JSON) and i32 for binary formats.
+#[cfg(feature = "serde")]
+macro_rules! serde_enum_mod_optional {
```
jtbandes, line 137: @eloff @ericmlujan could you folks review this part? I'm not familiar with this code at all and @gasmith is out -- I do wonder if there's a way we could combine this with the non-optional macro rather than having two whole macros for it?  but not sure how feasible that suggestion is

typescript/schemas/src/internal/generateSdkRustCTypes.ts
```diff
@@ -213,6 +216,9 @@ impl BorrowToNative for ${name} {
             }
             return `${fieldName}: self.${fieldName}`;
           case "enum":
+            if (field.optional) {
+              return `${fieldName}: unsafe { self.${fieldName}.as_ref() }.map(|&v| v as i32)`;
```
jtbandes, line 220: is this actually `unsafe`? I think this is for someone who knows more Rust than me to answer

typescript/schemas/src/internal/generateSdkCpp.ts
```diff
@@ -329,6 +326,10 @@ function cppToC(schema: FoxgloveMessageSchema, copyTypes: Set<string>): string[]
         }
         return `dest.${dstName} = src.${srcName};`;
       case "enum":
+        if (field.optional) {
+          const cEnumType = `foxglove_${toSnakeCase(field.type.enum.name)}`;
+          return `if (src.${srcName}) { auto* ptr = arena.alloc<${cEnumType}>(1); *ptr = static_cast<${cEnumType}>(*src.${srcName}); dest.${dstName} = ptr; } else { dest.${dstName} = nullptr; }`;
```
jtbandes, line 331: I am wondering if this whole thing could possibly be replaced with `dest.${dstName} = reinterpret_cast<const ${cEnumType}*>(src.${srcName})`?

I'm not sure if that would actually work or be safe - it just came to mind since the generated code looked a bit complicated and I _think_ these types are compatible.

If that doesn't work, it might be nice to actually put some line breaks in this string so it's more readable here  (though it does get auto-formatted in the generated files)

typescript/schemas/src/internal/schemas.ts
```diff
@@ -1514,6 +1514,22 @@ const PositionCovarianceType: FoxgloveEnumSchema = {
   ],
 };
 
+const PointStyle: FoxgloveEnumSchema = {
+  type: "enum",
+  name: "PointStyle",
+  description: "Style of point used for map visualization",
+  parentSchemaName: "LocationFix",
+  protobufEnumName: "PointStyle",
```
jtbandes, line 1522: I wonder if we should do something like
```suggestion
  name: "MapPointStyle",  //(or LocationFixPointStyle?)
  description: "Style of point used for map visualization",
  parentSchemaName: "LocationFix",
  protobufEnumName: "PointStyle",
```
?

Although it's also making me realize that what is currently called "protobufEnumName" should maybe be used for any language where we nest enums inside the schemas (which it seems includes c++ too)...

typescript/schemas/src/internal/generateSdkRustCTypes.ts
thswansonfxg: I think yes because self.point_style is a *const FoxglovePointStyle — a raw pointer in the C FFI struct. 

But we also talked about this in person. Someone who knows rust better might need to stamp this.

typescript/schemas/src/internal/generateSdkCpp.ts
thswansonfxg: That was implemented because of this cursor comment here: https://github.com/foxglove/foxglove-sdk/pull/870#discussion_r2801413580 

Further reading says that maybe its totally safe? I am happy to change this, but its being flagged because of garbage bytes and the potential of `undefined behavior`

typescript/schemas/src/internal/schemas.ts
thswansonfxg: I think we want to have PointStyle as eventually the goal might be to have these point styles shared across a few different panels. I agree that maybe we need to update protobufEnumName to just be enumName. Also happy to make those changes

rust/foxglove/src/schemas.rs
thswansonfxg: I'm certainly not confident enough in Rust macros to combine them nicely, especially for SDK level code. This was done because it was easier for me to understand and easier for me to read. 

Super happy to change this if someone better at rust has some ideas.

typescript/schemas/src/internal/generateSdkCpp.ts
jtbandes: ah ok, I missed that the sizes are different, I guess that makes sense

PR #872, opened 2026-02-13
C++ data provider example for remote data loaders

Muon:
### Changelog

Adds an example C++ implementation of a data provider for remote data loaders.

### Docs
None

### Description

This PR introduces a new C++ example, `cpp/examples/data-provider`. This is a direct port of the Rust `data_provider` example.

**Design**
- **HTTP framework**: `cpp-httplib` was chosen because it is header-only.
- **Helper types**: Native C++ structs with `nlohmann/json` serialization are used for manifest types, matching existing practice. No bindings from Rust to C++ are added. (`ChannelSet` is very small and not worth the trouble.)

A `no_seek_fn()` helper has been added to `mcap.hpp` to better support `disable_seeking = true`. It is used in the example.

Testing done using conformance test suite for Rust example. The conformance crate now has a `bin` target for invoking on arbitrary servers. 

Closes FIRE-288.


cpp/foxglove/include/foxglove/data_provider.hpp
```diff
@@ -0,0 +1,240 @@
+#pragma once
```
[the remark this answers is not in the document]
Muon: This is intended as part of the SDK surface.

cpp/examples/data-provider/src/main.cpp
```diff
@@ -0,0 +1,353 @@
...
+}
+
+// ============================================================================
+// Auth
+// ============================================================================
+
+/// Validate the bearer token from the Authorization header.
+/// Returns true if the request is authorized; sets a 401 response and returns false otherwise.
+///
+/// Replace this with real token validation (e.g. JWT verification).
```
bennetthardwick: Might be worth calling out extra loud that this currently doesn't do any validation / authentication? It does say replace with "real token validation" but if you don't read that far it reads like it does some kind of validation. 🤷🏻

cpp/foxglove/include/foxglove/data_provider.hpp
```diff
@@ -0,0 +1,240 @@
...
+/// source.topics = std::move(channels.topics);
+/// source.schemas = std::move(channels.schemas);
+/// source.start_time = "2024-01-01T00:00:00Z";
+/// source.end_time = "2024-01-02T00:00:00Z";
+///
+/// dp::Manifest manifest;
+/// manifest.name = "Flight ABC123";
+/// manifest.sources = {std::move(source)};
+///
+/// nlohmann::json j = manifest;
```
james-rms: why expose nlohmann as part of our API surface? i'm very hesitant to do this, would rather expose a free function `std::string serialize(const &Manifest manifest)`.

cpp/examples/data-provider/src/main.cpp
```diff
@@ -0,0 +1,353 @@
...
+
+      // Track bytes written for position queries from the MCAP writer.
+      uint64_t write_position = 0;
+      bool write_ok = true;
+
+      // The CustomWriter sends MCAP data directly to the HTTP socket. The MCAP writer
+      // buffers internally (up to chunk_size bytes) before calling write, so each call
+      // here corresponds to one MCAP chunk being flushed.
+      foxglove::CustomWriter custom_writer;
+      custom_writer.write =
```
james-rms, line 224: awkward that we require the user to write out this implementation. binrw has that NoSeek helper wrapper type that does exactly this - is there a way we can provide that for others as well? maybe not even specific to remote data loaders.

james-rms: Sorry, this comment was meant to be on the seek impl.

cpp/foxglove/include/foxglove/data_provider.hpp
Muon: What's the specific concern? `nlohmann/json` is pretty good. And even if there's a free function, it becomes a transitive include anyway.

james-rms: I for one think nlohmann/json is pretty easy to use but not my favorite JSON serialization dependency, i don't love how many tiny objects end up being created.

Even if we do like it now, I don't like including it as part of our documented API surface. To move away from it, we'd need to make breaking changes to function signatures that we've documented and encouraged people to use. 

I see your point that if it's a transitive include, people are free to use `nlohmann` through that transitive dependency, and taking it out will break their build. However I feel less sympathetic to those users - we didn't commit to that as apublic API, and they have a trivially easy way to make their code work again.

PR #874, opened 2026-02-13
derive: Support deriving Encode for newtypes

gasmith:
### Changelog
- rust: Added #[derive(Encode)] support for newtypes

### Docs
None

### Description
Building on recent improvements to derive support, this PR introduces support for deriving the Encode trait for newtype structs. This allows for using newtypes as standalone messages (in which case it is represented as a message with a single field `value` at slot 1), or as nested fields in a struct that derives Encode (in which case the newtype wrapper is transparent).

For example, it is now possible to do something like this:

```rust
/// A newtype wrapping a String. As a standalone message, this is serialized as `{ value: "..." }`.
#[derive(Encode)]
struct MyString(String);

/// A struct that uses a newtype field, serialized as `{ label: "..." }`.
#[derive(Encode)]
struct MessageWithNewtypeField {
    label: MyString,
}
```


james-rms:
In this example:
```rust
/// A newtype wrapping a String. As a standalone message, this is serialized as `{ value: "..." }`.
#[derive(Encode)]
struct MyString(String);

/// A struct that uses a newtype field, serialized as `{ label: "..." }`.
#[derive(Encode)]
struct MessageWithNewtypeField {
    label: MyString,
}
```

Is `serialized as { label: "..." }` really correct? I would have assumed MessageWithNewtypeField would look like this when serialized using protobuf introspection:
```typescript
{ label: { value: "..." } }
```

I bring this up because this affects what foxglove users are going to see in the raw message panel.

There's also some interesting backwards compatibility questions that need documented answers here. Since the newtype struct (actually a single-valued tuple struct) is a protobuf message with a value in slot 1, that means this is a backwards-compatible schema evolution:

```rust
#[derive(Encode)]
struct MyString {
   value: String,
   // I converted the tuple struct to a normal struct and added a field here in slot 2
   // this doesn't feel intuitively safe, but it is under this scheme
   generation: u32,
}
```
but this isn't:

```rust
/// A struct that uses a newtype field, serialized as `{ label: "..." }`.
#[derive(Encode)]
struct MessageWithNewtypeField {
    // I "unwrapped" the string newtype here
    // this feels safe for those used to C memory layout, but it isn't in this scheme
    label: String,
}
```

In general I'm pretty iffy on this direction. Protobuf serialization is a leaky abstraction, trying to bring "intuitive" behavior for native rust types brings weirdness beyond a certain point.

gasmith:
> In this example:
> 
> ```rust
> /// A newtype wrapping a String. As a standalone message, this is serialized as `{ value: "..." }`.
> #[derive(Encode)]
> struct MyString(String);
> 
> /// A struct that uses a newtype field, serialized as `{ label: "..." }`.
> #[derive(Encode)]
> struct MessageWithNewtypeField {
>     label: MyString,
> }
> ```
> 
> Is `serialized as { label: "..." }` really correct? I would have assumed MessageWithNewtypeField would look like this when serialized using protobuf introspection:
> 
> ```ts
> { label: { value: "..." } }
> ```
> 
> I bring this up because this affects what foxglove users are going to see in the raw message panel.

My original comment is accurate. The newtype is transparent when it is a struct field; we do not wrap it in another nested message with a "value" key, because that isn't necessary.

The reason we need the message framing for a standalone newtype, is that the newtype's inner value might not be serializable as a message. If it's a primitive like a u64 for example, that's not directly serializable as a protobuf message - it needs to be wrapped.

> There's also some interesting backwards compatibility questions that need documented answers here.
> [...]
> In general I'm pretty iffy on this direction. Protobuf serialization is a leaky abstraction, trying to bring "intuitive" behavior for native rust types brings weirdness beyond a certain point.

Agreed, but I feel strongly that the `Encode` macro is not intended to address these concerns. It's a quick-and-dirty way to get data into Foxglove. Folks that care about compatibility should use a proper IDL like protobuf. Also discussed [here](https://github.com/foxglove/foxglove-sdk/pull/889#issuecomment-3915943598).

james-rms:
@gasmith can i suggest a convention for all tuple structswhere the protobuf representation of 
```rust
struct TupleStruct(String, i32, ...)
```
is

```proto
syntax = "proto3";

message TupleStruct {
  string value_0 = 1;
  int32 value_1 = 2;
  ...
}
 ```

in all contexts (as a field and as a toplevel message). That way we're not boxed out of supporting higher-cardinality tuple structs in an inconsistent way.

The tradeoff is that people expecting to use this as a "newtype" and see this value_0 stuff might be a bit put off by that. IMO it's worth it, but I don't feel super strongly about it.

gasmith:
> @gasmith can i suggest a convention for all tuple structswhere the protobuf representation of
> 
> ```rust
> struct TupleStruct(String, i32, ...)
> ```
> 
> is
> 
> ```proto
> syntax = "proto3";
> 
> message TupleStruct {
>   string value_0 = 1;
>   int32 value_1 = 2;
>   ...
> }
> ```
> 
> in all contexts (as a field and as a toplevel message). That way we're not boxed out of supporting higher-cardinality tuple structs in an inconsistent way.
> 
> The tradeoff is that people expecting to use this as a "newtype" and see this value_0 stuff might be a bit put off by that. IMO it's worth it, but I don't feel super strongly about it.

I like the idea of supporting higher-cardinality tuple structs in that way, but I don't think it's the right move for newtype structs. The "value_0" indirection is too obtrusive. I'd be more in favor of the consistency argument if this were a proper IDL, and we were concerned about compatibility across schema evolution. Since this is purely an ergonomic convenience, I'm more inclined towards representing newtype values transparently.

eloff:
> > @gasmith can i suggest a convention for all tuple structswhere the protobuf representation of
> > ```rust
> > struct TupleStruct(String, i32, ...)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is
> > ```proto
> > syntax = "proto3";
> > 
> > message TupleStruct {
> >   string value_0 = 1;
> >   int32 value_1 = 2;
> >   ...
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > in all contexts (as a field and as a toplevel message). That way we're not boxed out of supporting higher-cardinality tuple structs in an inconsistent way.
> > The tradeoff is that people expecting to use this as a "newtype" and see this value_0 stuff might be a bit put off by that. IMO it's worth it, but I don't feel super strongly about it.
> 
> I like the idea of supporting higher-cardinality tuple structs in that way, but I don't think it's the right move for newtype structs. The "value_0" indirection is too obtrusive. I'd be more in favor of the consistency argument if this were a proper IDL, and we were concerned about compatibility across schema evolution. Since this is purely an ergonomic convenience, I'm more inclined towards representing newtype values transparently.

So what's the example look like for a tuple newtype? Or is that not supported in this PR? In that case how would it look in the future if we supported that? I'm trying to understand if this boxes us in or not when supporting tuple newtypes.

james-rms:
Since newtype structs are just single-valued tuple structs in Rust, IMO this puts us in a slightly odd place for when we support tuple structs, since the rules are different for single-valued structs. That's not the end of the world though, I leave the value judgement up to you.

gasmith:
> Since newtype structs are just single-valued tuple structs in Rust, IMO this puts us in a slightly odd place for when we support tuple structs, since the rules are different for single-valued structs. That's not the end of the world though, I leave the value judgement up to you.

I'll argue that's justified in the sense that single-valued tuple structs are so different from multi-valued tuple structs in rust that they have their own special name ("newtypes"). I strongly feel that the ergonomics are much better for visualization.

gasmith:
> So what's the example look like for a tuple newtype? Or is that not supported in this PR? In that case how would it look in the future if we supported that? I'm trying to understand if this boxes us in or not when supporting tuple newtypes.

The newtype (single-value tuple struct) examples are in the PR description. This PR doesn't implement multi-value tuple structs. I don't think we're boxing ourselves in wrt multi-value tuple structs, because:
- I think they're a fundamentally different shape in a Rust developer's mind
- The `Encode` macro is a convenience for a _predictable_ serialization, not a _stable_ serialization

PR #878, opened 2026-02-13
Remove v1 deps from common tests

clalancette:
### Changelog

None.

### Docs

None.

### Description

The tests in rust/foxglove/src/protocol/common contained some v1-specific things.  In particular, those tests were using some structs from the v1 protocol.

Because we are switching to the v2 protocol, we want to make sure these tests are actually generic and don't depend on v1 at all.  This PR does exactly that.

Particular note for reviewers: The one thing I'm not completely sure about is the changes to the `.snap` files.  Previously, we were using `to_bytes` (from the v1 protocol `ClientMessage`), which includes the opcode.  Now we are just using `write_payload`, which doesn't include the opcode.  So the change to the snaps is needed to keep the tests working, but I'm not 100% sure if we are still testing what we want to.

Assuming we are happy with this, I'll have a follow-up PR to add in additional tests specifically for the v1 and v2 protocols.

Fixes: [FLE-171](https://linear.app/foxglove/issue/FLE-171/remove-dependency-on-v1-encodedecode-logic-in-common-tests)

### DIFF
--- a/rust/foxglove/src/protocol/v1/client.rs
+++ b/rust/foxglove/src/protocol/v1/client.rs
@@ -74,7 +74,10 @@ mod tests {
     #[test]
     fn test_message_data_encode() {
         let message = MessageData::new(30, br#"{"key": "value"}"#);
-        insta::assert_snapshot!(format!("{:#04x?}", message.to_bytes()));
+        let buf = message.to_bytes();
+        insta::assert_snapshot!(format!("{:#04x?}", buf));
+        let parsed = ClientMessage::parse_binary(&buf).unwrap();
+        assert_eq!(parsed, ClientMessage::MessageData(message));
     }
 
     #[test]
@@ -85,7 +88,10 @@ mod tests {
             seek_time: None,
             request_id: "some-id".to_string(),
         };
-        insta::assert_snapshot!(format!("{:#04x?}", message.to_bytes()));
+        let buf = message.to_bytes();
+        insta::assert_snapshot!(format!("{:#04x?}", buf));
+        let parsed = ClientMessage::parse_binary(&buf).unwrap();
+        assert_eq!(parsed, ClientMessage::PlaybackControlRequest(message));
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


gasmith:
> Particular note for reviewers: The one thing I'm not completely sure about is the changes to the `.snap` files. Previously, we were using `to_bytes` (from the v1 protocol `ClientMessage`), which includes the opcode. Now we are just using `write_payload`, which doesn't include the opcode. So the change to the snaps is needed to keep the tests working, but I'm not 100% sure if we are still testing what we want to.

In thinking about this, I feel the value of the snaps is an assurance that the our v1 implementation remains wire-compatible over time as we make changes. By stripping some of the framing, I worry that we might be relaxing that validation a little too far. For JSON messages, serde is doing all the framing, so I suppose we can leave those tests in common. For binary messages, I'm mildly inclined towards keeping the v1- and v2-specific snapshot tests, and forego the common snapshot tests (since they'd be entirely redundant).

clalancette:
> In thinking about this, I feel the value of the snaps is an assurance that the our v1 implementation remains wire-compatible over time as we make changes. By stripping some of the framing, I worry that we might be relaxing that validation a little too far. For JSON messages, serde is doing all the framing, so I suppose we can leave those tests in common. For binary messages, I'm mildly inclined towards keeping the v1- and v2-specific snapshot tests, and forego the common snapshot tests (since they'd be entirely redundant).

Yeah, that's fair.  I can make the change to move these into the v1 files instead.

clalancette:
All right, checkout the latest state of this.  The snapshot tests have all been moved into the v1 directory.

rust/foxglove/src/protocol/v1/client.rs
```diff
@@ -67,6 +67,61 @@ impl BinaryMessage for PlaybackControlRequest {
     }
 }
 
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_message_data_encode() {
```
gasmith, line 75: Let's make these roundtrip tests, by adding a `ClientMessage::parse_binary`. Same for server tests below.

clalancette: Good idea.  Done in 263b0a4

PR #888, opened 2026-02-16
rust: Refactor remote access (and websocket) features

gasmith:
### Changelog
- rust: Renamed `live_visualization` feature as `websocket`

### Docs
None

### Description
This change is a bit of refactoring for the new remote access feature:
- Renames the `live_visualization` feature as `websocket`, leaving a deprecated alias behind
- Decouples `remote_access` from `websocket`, pulling common dependencies into an internal feature
- Creates new `Client` and `Capability` types for remote access, independent of the websocket equivalents
  - Just `ClientPublish` for now; we'll extend in the future.
  - Make `ClientPublish` explicitly user-configured, as it is for websocket
- Renames `RemoteAccessSink` to `remote_access::Gateway`
  - And rename `RemoteAccessSinkListener` to `remote_access::Listener`
  - If we agree, I think we should do the same for connection, session, etc, in a future change
- Reworked the server listener to use `ChannelDescriptor` instead of `ClientChannel` or `ChannelView`
  - The descriptor is an appropriate superset and it's cheaply clonable
  - The u32 client channel ID fits within the u64 channel ID

### DIFF
--- a/rust/foxglove/src/remote_access/capability.rs
+++ b/rust/foxglove/src/remote_access/capability.rs
@@ -1,6 +1,6 @@
 use crate::protocol::v2::server::server_info;
 
-/// A capability that can be advertised by a [`Server`](super::Server).
+/// A capability that can be advertised by a [`Gateway`](super::Gateway).
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum Capability {
     /// Allow clients to advertise channels to send data messages to the server.


rust/examples/remote_access/src/main.rs
```diff
@@ -27,17 +27,18 @@ async fn main() {
 
     // Open a connection for remote visualization and teleop.
     // This requires Foxglove Agent to be running on the same machine.
-    let handle = foxglove::RemoteAccessSink::new()
+    let handle = Server::new()
```
eloff: I'm not sure how I feel about the Server naming here. There's no server listening on any port here, it's more of a client connection to an intermediary server. The name seems misleading.

rust/foxglove/Cargo.toml
```diff
@@ -8,30 +8,33 @@ repository.workspace = true
...
-  "dep:tokio-tungstenite",
   "dep:serde_repr",
   "dep:serde_with",
 ]
+websocket = [
+  "_remote_common",
+  "dep:futures-util",
+  "dep:tokio-tungstenite",
+]
+live_visualization = ["websocket"]
```
eloff, line 27: Good, you covered the backwards compatibility. Seems fine then.

rust/foxglove/Cargo.toml
```diff
@@ -8,30 +8,33 @@ repository.workspace = true
...
   "dep:base64",
   "dep:flume",
-  "dep:futures-util",
   "dep:tokio",
   "dep:tokio-util",
-  "dep:tokio-tungstenite",
   "dep:serde_repr",
   "dep:serde_with",
 ]
+websocket = [
```
eloff, line 22: The reason this was live visualization to begin with is we weren't sure if we'd keep the web socket long term - it's more of an implementation detail.

We could swap it with web transport for example and we shouldn't have to change the name.

That only went skin deep though, since we still called the server a WebSocketServer.

I guess this is fine and if we ever did that it would probably be new type and feature names anyway.

eloff (approved):
I have some hangups around the naming, which maybe we can discuss briefly before merging this (or renamed it after if need be). But otherwise it's good to go.

rust/examples/remote_access/src/main.rs
gasmith: Yeah, let's talk about it tomorrow. I was thinking of it as a server in the senses that:
- Clients connect to it
- It provides a service that is only available while it is running
- Users of this interface probably don't care about the implementation details of how a connection through the intermediary is established - they care mostly about the relationship between this thing, and the viz client

PR #891, opened 2026-02-17
rust: Add basic integration tests for remote access

gasmith:
### Changelog
None

### Docs
None

### Description
This change adds an integration test suite for remote access functionality. It's split into two parts - one that depends on the Foxglove server (to validate device authentication), and one that depends on a local livekit server (to validate protocol interactions).

This change does expose `foxglove::protocol` as a public module via the undocumented `_protocol` feature. I don't think there's any great harm in this, considering the SDK is already open source. If someone wants to abuse knowledge of protocol details, they'll find a way.


rust/foxglove/src/lib.rs
```diff
@@ -385,7 +385,9 @@ pub(crate) use time::nanoseconds_since_epoch;
 
 #[cfg(feature = "remote_access")]
 mod api_client;
-#[cfg(feature = "_remote_common")]
+#[cfg(all(feature = "_remote_common", feature = "_protocol"))]
+pub mod protocol;
+#[cfg(all(feature = "_remote_common", not(feature = "_protocol")))]
```
gasmith, line 390: An alternative approach that occurs to me, is that we could re-export this from a new module (like `foxglove::testing` or `foxglove::hazmat`), with a corresponding leading-underscore feature flag.

eloff: I think the _ name and doc(hidden) is enough. If someone uses it, they know what risks they're taking.

rust/remote_access_tests/src/frame.rs
```diff
@@ -0,0 +1,64 @@
+use anyhow::{bail, ensure, Result};
```
eloff, line 1: This file mostly duplicates code (OpCode type and frame_binary_message) from session.rs.

I think that's not terrible, but we could also do a hidden export of it like for _protocol.

rust/remote_access_tests/tests/auth_test.rs
```diff
@@ -0,0 +1,250 @@
...
+        .name("auth-integration-test")
+        .device_token(token)
+        .foxglove_api_url(&config.foxglove_api_url)
+        .start()
+        .context("start Gateway")?;
+
+    // Give it time to connect and authenticate.
+    // The sink authenticates via device-info, then fetches RTC credentials.
+    // We wait long enough for the auth flow to complete.
+    tokio::time::sleep(Duration::from_secs(10)).await;
```
eloff, line 207: I wonder if we could expose (for the tests only) an awaitable something to block until auth has completed with a Result<()>.

This test runs the auth code, but just waits for it. If the auth flow fails, this test will pass.

rust/remote_access_tests/src/frame.rs
gasmith: It seemed small enough. I figured the copy was less obtrusive than the feature-gating would be.

rust/remote_access_tests/tests/auth_test.rs
gasmith: I'd like to think that one through. I think there's value outside of just tests. Developers probably want something similar - a way to poll or get notified about changes in the gateway health. I've filed https://linear.app/foxglove/issue/FLE-263/consider-adding-a-way-to-query-for-gateway-health.

PR #896, opened 2026-02-18
Add additional protocol tests

clalancette:
### Changelog

None.

### Docs

None.

### Description

This is a follow-up to #878 , to add tests for the v1 and v2 protocols to ensure we have good test coverage.

For v1, we add a few new tests and snapshots to the client.rs and server.rs.  The biggest addition is tests to tungstenite.rs, which previously had no tests.

For v2, we add tests to client.rs and server.rs, which both previously had no tests.


clalancette:
As suggested by @gasmith , I've now updated this PR to only add in high-value tests, which cuts the size in half.  Take a look when you get a chance.

PR #899, opened 2026-02-19
Implement channel advertisements and subscriptions

eloff:
### Changelog

None

### Docs

None

### Description

This PR adds support for reading messages from the client on the per-participant byte streams. It adds control plane and data plane queues for "buffering" messages to send to the participants, much like we have for the WebSocketServer. It's unclear if we'll need this for the lossy data packet based transport as well, but it makes sense for the "reliable" messages.

It then adds support for advertising channels, subscribe and unsubscribe, and sending log messages to subscribed participants. It implements the Sink trait.

### DIFF
--- a/rust/foxglove/src/remote_access/session.rs
+++ b/rust/foxglove/src/remote_access/session.rs
@@ -133,17 +133,20 @@ impl Sink for RemoteAccessSession {
         }
 
         let advertise_msg = advertise::advertise_channels(filtered.iter().copied());
-
-        // Track advertised channels
-        {
-            let mut cached = self.channels.write();
-            for &ch in &filtered {
-                cached.insert(ch.id(), ch.clone());
-            }
-        }
-
         if advertise_msg.channels.is_empty() {
             return None;
+        }
+
+        // Track advertised channels, excluding any that failed to encode (e.g. MissingSchema).
+        let advertised_ids: std::collections::HashSet<u64> =
+            advertise_msg.channels.iter().map(|ch| ch.id).collect();
+        {
+            let mut advertised_channels = self.channels.write();
+            for &ch in &filtered {
+                if advertised_ids.contains(&u64::from(ch.id())) {
+                    advertised_channels.insert(ch.id(), ch.clone());
+                }
+            }
         }
 
         let framed = frame_text_message(advertise_msg.to_string().as_bytes());


rust/foxglove/src/protocol/v1/message.rs
```diff
@@ -1,7 +1,19 @@
...
 pub trait BinaryMessage {
-    /// Encodes the message to bytes with the v1 opcode.
-    fn to_bytes(&self) -> Vec<u8>;
+    /// Returns the total encoded length of this message (opcode byte + payload).
+    fn encoded_len(&self) -> usize;
+
+    /// Encodes the message (opcode byte + payload) into the provided buffer.
+    fn encode(&self, buf: &mut impl BufMut);
+
+    /// Encodes the message to a new `Vec<u8>`.
```
dave-foxglove:
```suggestion
    /// Encodes the message (opcode byte + payload) to a new `Vec<u8>`.
```

rust/foxglove/src/protocol/v1/server.rs
```diff
@@ -52,47 +52,57 @@ impl BinaryOpcode {
 }
 
 impl BinaryMessage for MessageData<'_> {
-    fn to_bytes(&self) -> Vec<u8> {
-        let mut buf = Vec::with_capacity(1 + self.payload_size());
+    fn encoded_len(&self) -> usize {
+        1 + self.payload_size()
```
dave-foxglove: Can we defined an Opcode type and use the size of that data type here, instead of hardcoding `1`.
(applies to multiple places in the code).

Best would be a newtype (`struct Opcode(u8);`) (a type alias would be okay, but I'd ask why not a newtype :) ).

rust/foxglove/src/context/subscriptions/tests.rs
```diff
@@ -17,7 +17,7 @@ macro_rules! assert_subscribers_eq {
     };
 }
 
-fn chid(id: u64) -> ChannelId {
+fn chid(id: u32) -> ChannelId {
```
dave-foxglove: Can we defined a newtype and use that data type here for the data type of `id`.

Best would be a newtype (`struct NameForId(u8);`) (a type alias would be okay, but I'd ask why not a newtype :) ).

rust/foxglove/src/protocol/v1/message.rs
```diff
@@ -1,7 +1,19 @@
 //! Binary message encoding for Foxglove protocol v1.
 
-/// Trait a binary message with v1 protocol opcodes.
+use bytes::BufMut;
+
+/// Trait for a binary message with v1 protocol opcodes.
```
dave-foxglove: Is this still v1 specific (consider changing comment if it is not).

rust/foxglove/src/protocol/v2/message.rs
```diff
@@ -1,7 +1,19 @@
 //! Binary message encoding for Foxglove protocol v2.
 
-/// Trait a binary message with v2 protocol opcodes.
+use bytes::BufMut;
+
+/// Trait for a binary message with v2 protocol opcodes.
```
dave-foxglove: Is this the same as the v1 protocol? Should we consider "pulling this up" a directory level or similar (i.e. making this interface unversioned)

rust/foxglove/src/context/subscriptions/tests.rs
eloff: ChannelId is a newtype that wraps u32 (previously u64). Did you mean something else?

rust/foxglove/src/protocol/v1/server.rs
eloff: There is an opcode type, I think this is a good idea.

rust/foxglove/src/protocol/v1/message.rs
eloff: They're (currently) identical. This is the v1 copy, and there's a v2 copy. I believe the thinking is that's so we can feel free to diverge the implementations. But currently they are the same, and when I changed the v2 one in this PR, I also changed the v1 trait. It could maybe be moved to common and re-exported instead.

gasmith: Since the opcode isn't actually part of the trait, I agree on moving the trait definition to common.

rust/foxglove/src/protocol/v1/message.rs
```diff
@@ -1,7 +1,19 @@
...
 pub trait BinaryMessage {
-    /// Encodes the message to bytes with the v1 opcode.
-    fn to_bytes(&self) -> Vec<u8>;
+    /// Returns the total encoded length of this message (opcode byte + payload).
+    fn encoded_len(&self) -> usize;
+
+    /// Encodes the message (opcode byte + payload) into the provided buffer.
+    fn encode(&self, buf: &mut impl BufMut);
```
gasmith: Hm, what if:

```suggestion
pub trait BinaryMessage<'a>: BinaryPayload<'a> {
    /// The opcode for this message.
    const OPCODE: u8;

    /// Returns the total encoded length of this message (opcode byte + payload).
    fn encoded_len(&self) -> usize {
        1 + self.payload_size()
    }

    /// Encodes the message (opcode byte + payload) into the provided buffer.
    fn encode(&self, buf: &mut impl BufMut) {
        buf.put_u8(Self::OPCODE);
        self.write_payload(buf);
    }
```

rust/foxglove/src/remote_access/connection.rs
```diff
@@ -13,21 +16,41 @@ use tracing::{debug, error, info, warn};
...
+    ChannelId, Context, FoxgloveError, Metadata, RawChannel, RemoteAccessSinkListener, Sink,
+    SinkChannelFilter, SinkId,
 };
 
-use crate::protocol::v2::{server::ServerInfo, JsonMessage};
-
 type Result<T> = std::result::Result<T, RemoteAccessError>;
 
 const WS_PROTOCOL_TOPIC: &str = "ws-protocol";
+const MESSAGE_FRAME_SIZE: usize = 5;
```
gasmith:
```suggestion
const MESSAGE_FRAME_SIZE: usize = 5;  // 1 byte opcode + u32 LE length
```

rust/foxglove/src/remote_access/connection.rs
```diff
@@ -382,11 +428,374 @@ impl RemoteAccessConnection {
...
+}
+
+/// RemoteAccessSession tracks a connected LiveKit session (the Room)
+/// and any state that is specific to that session.
+/// We discard this state if we close or lose the connection.
+/// [`RemoteAccessConnection`] manages the current connected session (if any)
+///
+/// The Sink impl is at the RemoteAccessSession level (not per-participant)
+/// so that it can deliver messages via multi-cast to multiple participants.
+struct RemoteAccessSession {
```
gasmith: Nit: How would you feel about moving this to `crate::remote_access::session`?

rust/foxglove/src/remote_access/connection.rs
```diff
@@ -382,11 +428,374 @@ impl RemoteAccessConnection {
...
+                    }
+                }
+                msg = session.data_plane_rx.recv_async() => {
+                    let Ok(msg) = msg else { break };
+                    // Use a SmallVec to avoid extra allocations/free for the temporary participants vec
+                    let participants: SmallVec<[Arc<Participant>; 8]> = {
+                        let participants = session.participants.read();
+                        participants
+                            .values()
+                            .filter(|p| p.is_subscribed(msg.channel_id))
```
gasmith: We can save it for a future change, but I think it might be a bit more efficient to keep a map from channel ID to a set of participants instead, so that we can do a single lock & lookup, rather than one-per-participant. Might even be able to avoid the lock with an arcswap, since subscriptions change so infrequently.

rust/foxglove/src/remote_access/connection.rs
```diff
@@ -382,11 +428,374 @@ impl RemoteAccessConnection {
...
+                            .cloned()
+                            .collect()
+                    };
+                    for participant in &participants {
+                        if let Err(e) = participant.send(&msg.data).await {
```
gasmith: This is fine for now, but I don't like the idea of doing fan-out in the SDK. We should be leveraging the livekit SFU for that. Presuming we want a reliable data channel, I think it should be separate from our point-to-point reliable control plane channels.

rust/foxglove/src/remote_access/connection.rs
```diff
@@ -382,11 +428,374 @@ impl RemoteAccessConnection {
...
+            };
+
+            buffer.extend_from_slice(&chunk);
+
+            // Parse complete messages from buffer
+            while buffer.len() >= MESSAGE_FRAME_SIZE {
+                // Parse frame header: 1 byte OpCode + 4 bytes little-endian u32 length
+                let opcode = buffer[0];
+                let length =
+                    u32::from_le_bytes(buffer[1..MESSAGE_FRAME_SIZE].try_into().unwrap()) as usize;
```
gasmith: Nit: maybe `Buf::get_u32_le()`?

rust/foxglove/src/remote_access/connection.rs
```diff
@@ -382,11 +428,374 @@ impl RemoteAccessConnection {
...
+                        "message too large ({length} bytes) from client {:?}, disconnecting",
+                        participant_identity
+                    );
+                    return;
+                }
+
+                // Check if we have the complete message
+                if buffer.len() < MESSAGE_FRAME_SIZE + length {
+                    break; // Wait for more data
+                }
```
gasmith: This feels like it might be a good use tokio_util's [`StreamReader`](https://docs.rs/tokio-util/latest/tokio_util/io/struct.StreamReader.html), which implements `AsyncRead` and abstracts the chunk buffering. Then we can use [`read_exact`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncReadExt.html#method.read_exact) to read the 5-byte header, and the N-byte payload.

rust/foxglove/src/remote_access/connection.rs
```diff
@@ -382,11 +428,374 @@ impl RemoteAccessConnection {
...
+                return;
+            }
+        };
+
+        // Look up participant ID before the final match
+        let Some(participant) = ({
+            let participants = self.participants.read();
+            participants.get(participant_identity).cloned()
+        }) else {
+            error!("Unknown participant identity: {:?}", participant_identity);
```
gasmith: Doesn't seem like a user-actionable error. Maybe worth a debug log? I feel like this should also be a signal to the caller that it should exit the rx loop for this bytestream. Toward that end, maybe this lookup should happen in the caller, and this function should be invoked with a proper Participant reference.

rust/foxglove/src/remote_access/connection.rs
```diff
@@ -382,11 +428,374 @@ impl RemoteAccessConnection {
...
+            ClientMessage::Unsubscribe(msg) => {
+                self.handle_client_unsubscribe(&participant, msg);
+            }
+            // TODO: Implement other message handling branches
+            _ => {}
+        }
+    }
+
+    fn handle_client_subscribe(&self, participant: &Participant, msg: client::Subscribe) {
+        let channel_ids: Vec<ChannelId> = msg
```
gasmith: Probably a good use of a smallvec, since our one and only client only subscribes to a single channel at a time.

rust/foxglove/src/remote_access/connection.rs
```diff
@@ -382,11 +428,374 @@ impl RemoteAccessConnection {
...
+                    "Participant {:?} subscribed to channel {channel_id:?}",
+                    participant
+                );
+            } else {
+                warn!(
+                    "Participant {:?} is already subscribed to channel {channel_id:?}; ignoring",
+                    participant
+                );
+            }
+        }
```
gasmith: Maybe just move this into the Participant impl? Same for unsubscribe.

rust/foxglove/src/remote_access/connection.rs
```diff
@@ -382,11 +428,374 @@ impl RemoteAccessConnection {
...
+        let newly_subscribed = participant.subscribe(&channel_ids);
+
+        for &channel_id in &channel_ids {
+            if newly_subscribed.contains(&channel_id) {
+                debug!(
+                    "Participant {:?} subscribed to channel {channel_id:?}",
+                    participant
+                );
+            } else {
+                warn!(
```
gasmith: Warn feels awfully noisy. Same for unsubscribe.

rust/foxglove/src/remote_access/connection.rs
```diff
@@ -382,11 +428,374 @@ impl RemoteAccessConnection {
...
+
+        match client_msg {
+            ClientMessage::Subscribe(msg) => {
+                self.handle_client_subscribe(&participant, msg);
+            }
+            ClientMessage::Unsubscribe(msg) => {
+                self.handle_client_unsubscribe(&participant, msg);
+            }
+            // TODO: Implement other message handling branches
+            _ => {}
```
gasmith: Maybe warn about them being unimplemented for now.

rust/foxglove/src/channel.rs
```diff
@@ -22,30 +22,37 @@ const STACK_BUFFER_SIZE: usize = 256 * 1024;
 
 /// Uniquely identifies a channel in the context of this program.
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Debug, Deserialize, Serialize)]
-pub struct ChannelId(u64);
+pub struct ChannelId(u32);
```
gasmith: Why are we narrowing channel ID as part of this PR? That's an API-breaking change, and one that would also require changes for C/C++. If this is really something we want to do, let's do it in a separate PR and discuss the merits there.

rust/foxglove/src/websocket.rs
```diff
@@ -1,6 +1,6 @@
 //! Websocket functionality
 
-mod advertise;
+pub(crate) mod advertise;
```
gasmith: No longer needed, I presume.
```suggestion
mod advertise;
```

rust/foxglove/src/remote_access/connection.rs
```diff
@@ -382,11 +428,374 @@ impl RemoteAccessConnection {
...
+    }
+
+    fn handle_client_subscribe(&self, participant: &Participant, msg: client::Subscribe) {
+        let channel_ids: Vec<ChannelId> = msg
+            .channel_ids
+            .iter()
+            .map(|&id| ChannelId::new(id))
+            .collect();
+
+        let newly_subscribed = participant.subscribe(&channel_ids);
```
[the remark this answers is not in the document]
eloff: We subscribe dynamically, but the subscribe/unsubscribe handlers do need to propagate the subscription to the context

rust/foxglove/src/remote_access/connection.rs
eloff: Actually I changed the impl and moved the tracking to the session, so I can propagate the subscriptions to the context. So I think it should stay here now.

rust/foxglove/src/remote_access/connection.rs
gasmith: Oh! Actually, this might be critical... I just realized that we're not calling `context.subscribe_channels` and `context.unsubscribe_channels` anywhere. We're going to need to maintain a map of channel to participants in order to maintain those subscriptions with the Context. Otherwise, we'll never get a `log()` call.

rust/foxglove/src/remote_access/connection.rs
```diff
@@ -382,11 +428,374 @@ impl RemoteAccessConnection {
...
 impl RemoteAccessSession {
-    fn new(room: Room) -> Self {
+    fn new(
+        room: Room,
+        channel_filter: Option<Arc<dyn SinkChannelFilter>>,
+        cancellation_token: CancellationToken,
+        message_backlog_size: usize,
+    ) -> Self {
+        let (data_plane_tx, data_plane_rx) = flume::bounded(message_backlog_size);
+        let (control_plane_tx, control_plane_rx) = flume::bounded(message_backlog_size);
```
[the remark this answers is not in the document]
eloff: We do the same in the WebSocketServer, I think it's fine to keep it simpler. The user doesn't need to know there are two queues in the impl.

rust/foxglove/src/remote_access/connection.rs
eloff: Yeah, it's reached that threshold for sure

rust/foxglove/src/remote_access/connection.rs
eloff: If we get this, it means the user is on a version older than what we support (which probably fails earlier). I'll change the wording here, and exit the rx loop.

rust/foxglove/src/remote_access/connection.rs
eloff: This would only be for the reliable byte stream impl, which is not going to be our default once you've implemented the datagram version - and that lets us do multicast and push the fanout to livekit. This will then maybe only be used for specific opt-in "reliable" channels - if that.

rust/foxglove/src/channel.rs
eloff: Yeah you're absolutely right. As discussed offline, I'm undoing this change and widening the v2 message type channel ids to 64bit instead, with a check that we don't exceed 2^53 when creating channel ids so we don't need to deal with 64bit values in JSON on the app side.

rust/foxglove/src/remote_access/connection.rs
eloff: Yeah, realized this also. I implemented this as you'd expect with a single map on session instead.

rust/foxglove/src/protocol/v2/message.rs
eloff: Yeah, moved this to common

rust/foxglove/src/remote_access/connection.rs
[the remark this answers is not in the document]
eloff: Good point

rust/foxglove/src/remote_access/connection.rs
eloff: Which should be a smallvec?

rust/foxglove/src/protocol/v2/client/subscribe.rs
```diff
@@ -10,12 +10,12 @@ use crate::protocol::{BinaryPayload, ParseError};
 #[serde(tag = "op", rename = "subscribe", rename_all = "camelCase")]
 pub struct Subscribe {
     /// Channel IDs to subscribe to.
-    pub channel_ids: Vec<u32>,
+    pub channel_ids: Vec<u64>,
```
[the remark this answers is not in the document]
eloff: We're coordinating the app-side changes, and there are no users yet

rust/foxglove/src/remote_access/session.rs
```diff
@@ -0,0 +1,606 @@
...
+    fn remove_channel(&self, channel: &RawChannel) {
+        let channel_id = channel.id();
+        if self.channels.write().remove(&channel_id).is_none() {
+            return;
+        }
+
+        let unadvertise = Unadvertise::new([u64::from(channel_id)]);
+        let framed = frame_text_message(unadvertise.to_string().as_bytes());
+        self.broadcast_control(framed);
+    }
```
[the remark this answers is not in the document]
eloff: That appears to be an issue in the WebSocketServer as well. Did we overlook that, or does the app respond to an unadvertise message by sending an unsubscribe message? I'll investigate.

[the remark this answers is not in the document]
eloff: Yes, the client does send back an unsubscribe message for unadvertised channels. Here's the full flow when the client receives a server unadvertise message:

FoxgloveClient (transport layer) — FoxgloveClient.ts line 136–138:
Simply parses the message and emits a "unadvertise" event with the list of removed channel IDs. No wire message is sent at this layer.

FoxgloveWebSocketPlayer (player layer) — index.ts lines 690–715:
This is where the real work happens. For each unadvertised channel ID:
Unknown channel check: If the channel ID isn't in #channelsById, it checks #unsupportedChannelIds. If it's not there either, it records an error problem ("Server unadvertised channel X that was not advertised").
Unsubscribe from affected subscriptions: Iterates over all #resolvedSubscriptionsById. For any subscription that was on the unadvertised channel:
Removes it from #resolvedSubscriptionsById
Removes it from #resolvedSubscriptionsByTopic
Calls this.#client?.unsubscribe(subId) — which sends an {"op": "unsubscribe", "subscriptionIds": [subId]} message back to the server (per FoxgloveClient.ts line 200–202)
Adds the topic to #unresolvedSubscriptions (so it can be re-subscribed if the channel is re-advertised later)
Clean up channel tracking: Removes the channel from #channelsById and #channelsByTopic.
Updates topics/datatypes and queues an emit to notify the UI that the available topics have changed.

So to directly answer your question: yes, the client sends an unsubscribe message back to the server for each subscription that was on an unadvertised channel, and then moves those topics into #unresolvedSubscriptions so they can be re-subscribed automatically if the server re-advertises those channels in the future.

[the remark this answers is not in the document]
gasmith: Let's file a linear ticket and fix this for both implementations. Seems like a better choice to proactively drop subscriptions for unadvertised channels, rather than depending on client behavior.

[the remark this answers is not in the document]
gasmith: https://linear.app/foxglove/issue/FLE-240/proactively-drop-subscriptions-on-unsubscribe

PR #907, opened 2026-02-20
Remote access refactors for testing and unit tests

eloff:
### Changelog

None

### Docs

None

### Problem

The `RemoteAccessSession` in `rust/foxglove/src/remote_access/session.rs` previously mixed pure state management logic (tracking participants, channels, and subscriptions) with I/O operations and LiveKit interactions. This tight coupling made the state machine difficult to test in isolation and obscured the core state transitions.

### What has changed

This PR refactors the `RemoteAccessSession` to separate its state management from its I/O responsibilities.

1.  A new module, `rust/foxglove/src/remote_access/session_state.rs`, introduces the `SessionState` struct. This struct now exclusively manages the internal state, including `participants`, `channels`, and `subscriptions`.
2.  Methods within `SessionState` operate on `&self` and return explicit results (e.g., which channels gained/lost their first/last subscriber). This allows `RemoteAccessSession` to react to state changes by performing appropriate I/O actions.

### Motivation

This refactor improves the testability of the core state machine, enhances separation of concerns, and makes the `RemoteAccessSession` easier to understand and maintain. The `SessionState` is now fully unit-tested in isolation.

Fixes: [FLE-241](https://linear.app/foxglove/issue/FLE-241/refactor-and-unit-test-remoteaccesssession-state-machine)

---
<p><a href="https://cursor.com/agents?id=bc-eae5a1ec-8c15-464b-bcbc-4af2e34fd6bc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-eae5a1ec-8c15-464b-bcbc-4af2e34fd6bc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</p>

### DIFF
--- a/rust/foxglove/src/remote_access/session_state.rs
+++ b/rust/foxglove/src/remote_access/session_state.rs
@@ -56,7 +56,10 @@ impl SessionState {
     /// Removes a participant and all of its subscriptions.
     ///
     /// Returns channel IDs that lost their last subscriber (now have zero subscribers).
-    pub fn remove_participant(&mut self, identity: &ParticipantIdentity) -> SmallVec<[ChannelId; 4]> {
+    pub fn remove_participant(
+        &mut self,
+        identity: &ParticipantIdentity,
+    ) -> SmallVec<[ChannelId; 4]> {
         if self.participants.remove(identity).is_none() {
             return SmallVec::new();
         }
@@ -178,9 +181,7 @@ impl SessionState {
     /// Returns the number of subscribers for a channel.
     #[cfg(test)]
     pub fn get_subscriber_count(&self, channel_id: &ChannelId) -> usize {
-        self.subscriptions
-            .get(channel_id)
-            .map_or(0, |s| s.len())
+        self.subscriptions.get(channel_id).map_or(0, |s| s.len())
     }
 
     /// Returns the number of advertised channels.
--- a/rust/foxglove/src/remote_access/session.rs
+++ b/rust/foxglove/src/remote_access/session.rs
@@ -459,7 +459,10 @@ impl RemoteAccessSession {
             ParticipantWriter::Livekit(stream),
         ));
 
-        Ok(self.state.write().insert_participant(participant_id, participant))
+        Ok(self
+            .state
+            .write()
+            .insert_participant(participant_id, participant))
     }
 
     /// Remove a participant from the session, cleaning up its subscriptions.


rust/foxglove/src/remote_access/session_state.rs
```diff
@@ -0,0 +1,460 @@
...
+///
+/// Methods that modify subscriptions return the set of channel IDs whose subscription
+/// status changed (first subscriber added or last subscriber removed), so the caller
+/// can notify the Context.
+pub(crate) struct SessionState {
+    participants: RwLock<HashMap<ParticipantIdentity, Arc<Participant>>>,
+    /// Channels that have been advertised to participants.
+    channels: RwLock<HashMap<ChannelId, Arc<RawChannel>>>,
+    /// Maps channel ID to the participant identities subscribed to that channel.
+    subscriptions: RwLock<HashMap<ChannelId, SmallVec<[ParticipantIdentity; 1]>>>,
```
gasmith: I think a single RwLock around the entire SessionState might be preferable.

The only hot path we care about is `log`, which currently takes a read-lock on subscriptions & participants. Write locks are only taken infrequently for control plane operations, and aren't held very long. Having separate locks probably doesn't buy us very much.

Meanwhile, there's a small window of inconsistency in `remove_participant`, where `subscriptions` may refer to a participant that has been removed from `partcipants`. It's a minor concern, since existing callers handle this correctly, but I worry that this sort of thing might catch us (or an LLM) by surprise in the future.

gasmith (approved):
LG in general, with a suggestion.

rust/foxglove/src/remote_access/session_state.rs
eloff: That's a good point. Claude initially did this with a single RwLock, which I told it to undo. But it may have been on the right track. Log is the only hot path, and it's tough to beat a single read lock, especially when we don't hold the write lock often or for long. At any rate it's a safer, simpler option and we can start there and only change it if we get evidence that it's an issue.

PR #914, opened 2026-02-23
Foxglove SDK checkout composite action

Muon:
### Changelog

None

### Docs

None

### Description

This PR introduces a composite action at `.github/actions/internal-checkout/action.yml` which creates a symlink to a checkout of the SDK repo. This will be used in `data-platform` CI for a smoke test which will be automatically rolled forward by dependabot.

I have tested that this action works by pointing GitHub CI at it in the forthcoming `data-platform` CI PR.

### DIFF
--- a/.github/actions/internal-checkout/action.yml
+++ b/.github/actions/internal-checkout/action.yml
@@ -12,4 +12,6 @@ runs:
   using: composite
   steps:
     - shell: bash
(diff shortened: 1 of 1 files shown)


james-rms:
:+1: do you have any way to test that this works without merging it? can you point to your commit from the other repo's workflow?

james-rms:
This should be clearly indicated as for internal use with no guarantees of support, ideally in the action name.

james-rms (approved):

PR #924, opened 2026-02-24
Add optional timestamp to image annotations

aneuwald:
### Changelog
Add `timestamp` field as optional in `ImageAnnotations` messages

### Docs
None

### Description
This change introduces an optional timestamp field `ImageAnnotations` messages to improve temporal consistency and enable more robust time-aware features.

The primary motivation is to facilitate the development of new features that depend on explicit message timestamps.

The field is optional to maintain backward compatibility and avoid breaking existing producers and consumers

ERT-1119

### DIFF
--- a/typescript/schemas/package.json
+++ b/typescript/schemas/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@foxglove/schemas",
-  "version": "1.10.0",
+  "version": "1.10.1",
   "description": "Foxglove-defined message schemas for ROS, Protobuf, FlatBuffers, OMG IDL, and JSON",
   "license": "MIT",
   "author": {


typescript/schemas/src/internal/schemas.ts
```diff
@@ -764,6 +764,13 @@ const SceneUpdate: FoxgloveMessageSchema = {
       array: true,
       description: "Scene entities to add or replace",
     },
+    {
+      name: "timestamp",
+      type: { type: "nested", schema: Timestamp },
+      description:
+        "Timestamp for the scene update. Some Foxglove features will use this timestamp when set. It is preferable that entity and deletion timestamps match this value when provided.",
```
snosenzo: Actually I don't think this is necessary for `SceneUpdates` since the `SceneEntityDeletion` has a timestamp on it already. In the annotation case there's no way to hide the annotations without emitting an empty image annotation message. 
Sorry I didn't realize this earlier.

typescript/schemas/src/internal/schemas.ts
```diff
@@ -1500,6 +1507,13 @@ const ImageAnnotations: FoxgloveMessageSchema = {
         "Additional user-provided metadata associated with the image annotations. Keys must be unique.",
       array: true,
     },
+    {
+      name: "timestamp",
```
snosenzo, line 1505: I know this is a bit tangential, but why does each Annotation type have it's own timestamp, when in all likelyhood all of these are probably coming from the same perception process? and each perception process would emit annotations on a separate topic rather than adding them onto the same ImageAnnotation message. In other words, should we make the internal annotation timestamps optional as well, and tell people to use this instead?

aneuwald: I would say so @snosenzo. Now we are more focused in the sync feature so this will help a lot. But I believe in the future we will walk to a direction where we even remove the internal ones. But for now, we left there for backwards compatibiliy

typescript/schemas/src/internal/schemas.ts
aneuwald: And if the deletions in an empty array? Also, for now this change is more focused in the sync feature that would be great for us internally if there is an external timestamp. Thoughts @achim-k ?

snosenzo: If the deletions and additions are an empty array that means that nothing should change, unless I'm missing something and `SceneEntityDeletions` should be empty when clearing all scene entities.

aneuwald: it can be, but for example, if I have an empty deletions but have 10 additions? Wouldn't be better to have one timestamp to all these entities? I believe usually it's what happens, but then users will not have to fill all entities with the same timestamp (in the future)

typescript/schemas/src/internal/schemas.ts
achim-k: You can still publish individual Annotation messages (CircleAnnotation, TextAnnotation, ..) and in that case you want the timestamp. So I'm not sure if it's a good idea to making it optional for these

typescript/schemas/src/internal/schemas.ts
achim-k: Instead of adding a new timestamp field here, maybe we can update the description of the deletions and entities fields above saying that these should all have the same timestamp?

aneuwald: @achim-k maybe we just not add this, and write this on the sync feature page? Because normally they can have different timestamps.

schemas/README.md
```diff
@@ -1251,6 +1251,19 @@ Text annotations
...
+<tr>
+<td><code>timestamp</code> (optional)</td>
+<td>
+
+[Timestamp](#timestamp)
+
+</td>
+<td>
+
+Timestamp for the image annotations. Some Foxglove features will use this timestamp when set. It is preferable that individual annotation timestamps match this value when provided.
```
defunctzombie: This language feels/sounds "weak" to me. "Some foxglove features" - what features / is that relevant here?

"It is preferable" ... is it required or not? What breaks if it is not the case that these match?

Our users need guidance and clarity on what these fields do and the semantics they imply.

snosenzo: Yeah I agree. I think it's fair to say that this value will be ignored if there are annotations present, and that it will only be used if annotations list is empty.

snosenzo:
The description and title of this PR should be updated to reflect that we're not changing SceneUpdates

typescript/schemas/src/internal/schemas.ts
aneuwald: Actually, our image panel does not accept individual Annotations, users must publish an `foxglove.ImageAnnotations` message, but yeah, for some reason if the user wants to publish it individually, he could always save it there

schemas/README.md
aneuwald: totally agreed, but it won't be ignored. The idea is to always use this timestamp as the reference for synchronization of all annotations. I would say the best thing is say that this is the timestamp used and the internal ones will be ignored? Maybe mark then as deprecated for a while?

aneuwald: Ok changed, to:

> Timestamp of the image annotations. When set, individual annotation timestamps will be ignored.

In case user doesn't set, than we handle the individual ones depending of the feature.

typescript/schemas/src/internal/schemas.ts
```diff
@@ -1501,6 +1501,13 @@ const ImageAnnotations: FoxgloveMessageSchema = {
       array: true,
       optional: true,
     },
+    {
+      name: "timestamp",
+      type: { type: "nested", schema: Timestamp },
+      description:
+        "Timestamp of the image annotations. When set, individual annotation timestamps will be ignored.",
```
snosenzo, line 1508: is it worth saying something about how this should be set for when the annotations are empty and the desire is to clear the previous annotations?

aneuwald: I'm not sure if we should specify how it's going to be used. Is it a normal approach? I've look around and it doesn't seem that in other schemas we explain how properties will be used in Foxglove.

aneuwald: and this is kinda specific for syncing right? if the user doesn't set this timestamp and send empty annotations while not syncing messages, it will be processed and existing ones will be cleared, even without the timestamp set, right?

PR #927, opened 2026-02-25
remote_access: Transcode image channels into video tracks

gasmith:
### Changelog
None

### Docs
None

### Description
This change updates the remote access gateway to support transcoding image channels into video tracks.

Part of the work involves decoding image messages, transcoding them to YUV/I420, and then passing that over to livekit's video encoder. That's pretty straightforward, and more or less lifted from the prototype implementation we had in the agent.

A more substantial part of the change involves some protocol changes. We do not want to be transcoding video if no one is subscribed, but livekit doesn't give us the signals we'd need to determine whether someone is
subscribed. So instead, we use ws-protocol for managing subscription explicitly. The `Advertise` message now includes channel metadata, with a `foxglove.hasVideoTrack` key to indicate that a channel is available
as a livekit video track. The client uses `Subscribe` and `Unsubscribe` messages in the usual way, and the gateway spawns video tracks on demand. The client is responsible for interpreting the metadata in the advertisement, and waiting for a video track to appear after subscribing.

### Testing
I tested this change against https://github.com/foxglove/app/pull/12751

Fixes: FLE-177
Fixes: FLE-239

### DIFF
--- a/rust/examples/remote_access/src/main.rs
+++ b/rust/examples/remote_access/src/main.rs
@@ -76,7 +76,7 @@ async fn camera_loop() {
 
     let calibration = CameraCalibration {
         timestamp: None,
-        frame_id: "camera".into(),
+        frame_id: "".into(),
         width: width as u32,
         height: height as u32,
         distortion_model: String::new(),
@@ -118,7 +118,7 @@ async fn camera_loop() {
         let mono_row = &mono_gradient[offset..offset + width];
         let mono_img = RawImage {
             timestamp: Some(Timestamp::now()),
-            frame_id: "camera".into(),
+            frame_id: "".into(),
             width: width as u32,
             height: height as u32,
             encoding: "mono8".into(),
@@ -131,7 +131,7 @@ async fn camera_loop() {
         let rgb_row = &rgb_gradient[offset * 3..(offset + width) * 3];
         let rgb_img = RawImage {
             timestamp: Some(Timestamp::now()),
-            frame_id: "camera".into(),
+            frame_id: "".into(),
             width: width as u32,
             height: height as u32,
             encoding: "rgb8".into(),


rust/foxglove/src/remote_access/session/video_track.rs
```diff
@@ -0,0 +1,289 @@
...
+        if let Err(tokio::sync::mpsc::error::TrySendError::Closed(_)) =
+            self.tx.try_send((data, log_time_ns))
+        {
+            warn!("video publisher channel closed");
+        }
+        // Full channel: frame dropped silently (acceptable for live video)
+    }
+}
+
+/// Decode the image message and publish it as a video frame.
```
clalancette: I'm a little confused by this comment, and the following code.  The comment says "decode and publish", the function is called "encode_and_publish", and then the very first line of code in the function is "decode_image_message".

Let me explain what I think is happening here, and you can correct me if I'm wrong.  This part of the foxglove-sdk is being handed an Image from whatever application is using the SDK.  That image is one of `CompressedImage`, `RawImage`, `Ros1CompressedImage`, `Ros1RawImage`, `Ros2CompressedImage`, or `Ros2RawImage`.  We are unwrapping that data, converting it to YUV420, then passing it into the LiveKit machinery via `capture_frame` to be sent to the room (and eventually to other participants).  Does that accurately describe what is happening?

If so, I might suggest calling this `decode_and_publish`, and add a comment right above `capture_frame` saying that this is the `publish` step, since it is somewhat unclear.

gasmith: Yeah, that's a good call. Since we're doing decode _and_ encode, I'm going to call it "transcode".

rust/foxglove/src/remote_access/session/video_track.rs
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    /// Returns a reference to the underlying video source.
+    #[allow(dead_code)]
+    pub fn video_source(&self) -> &NativeVideoSource {
+        &self.video_source
+    }
+
+    /// Send a frame for encoding. Non-blocking: if the channel is full, the frame is
+    /// silently dropped (back-pressure for live video).
```
eloff: It would be better to drop the oldest buffered frame like we do for data channels in lvie viz and remote viz?

rust/foxglove/src/remote_access/session/video_track.rs
```diff
@@ -0,0 +1,294 @@
...
+    video_source: NativeVideoSource,
+}
+
+impl VideoPublisher {
+    /// The bounded channel capacity for frame back-pressure.
+    const CHANNEL_CAPACITY: usize = 2;
+
+    /// Creates a new video publisher and spawns the background processing task.
+    pub fn new(video_source: NativeVideoSource, input_schema: VideoInputSchema) -> Self {
+        let (tx, mut rx) = tokio::sync::mpsc::channel::<(Bytes, u64)>(Self::CHANNEL_CAPACITY);
```
eloff: Since we use flume and it's generally more flexible and more efficient, probably better to use it here than the tokio one

rust/foxglove/src/remote_access/session.rs
```diff
@@ -103,6 +116,14 @@ impl Sink for RemoteAccessSession {
         metadata: &Metadata,
     ) -> std::result::Result<(), FoxgloveError> {
         let channel_id = channel.id();
+
+        // If a video publisher is active for this channel, send the frame to it
+        // and skip the raw data plane.
+        if let Some(publisher) = self.state.read().get_video_publisher(&channel_id) {
```
eloff, line 122: Is there a more efficient way we can tell this is a video publisher from the Channel itself? Avoid the read lock and lookup (which is paid by all log calls, not just the video ones)?

eloff (commented):
This looks pretty good, I left some comments, but not blockers

rust/foxglove/src/remote_access/session/video_track.rs
gasmith: Maybe. What's different in this case is that the buffer is extremely shallow - only two elements. So it wouldn't make much of a difference.

rust/foxglove/src/remote_access/session.rs
gasmith: We could put an extra guard there for `get_video_input_schema(channel).is_some()`, but I guess I'd like to measure whether that's actually less expensive than just grabbing the read lock.

gasmith: My gut feels like it isn't going to be worth it... multiple pointer dereferences & string comparisons are probably going to be in the same ballpark, or possibly more expensive than, an uncontended read lock. We'll come back to it if it shows up on a profiler.

rust/foxglove/src/remote_access/session/video_track.rs
gasmith: Are there any recent evaluations of flume's efficiency compared to tokio? Flume has been in maintenance mode for two years at this point, and the benchmarks they published in their repo are five years old. I feel like we should probably default to using tokio's mpsc queue unless it is either lacking a feature we require, or we've actually measured that it is significantly less efficient than an alternative like flume, in a way that matters. For 30 or even 60 Hz video transcoding, we probably don't care much about queueing overheads.

rust/foxglove/src/remote_access/session.rs
eloff: Yeah you're right, it's something we'll need to measure. I'm sure we could improve on it (e.g. have a bool on the channel would definitely be faster). But it's probably not the most impactful performance improvement either.

rust/foxglove/src/remote_access/session/video_track.rs
eloff: It's slightly better, especially if we end up dropping multiple frames in a row. If we drop the oldest, the buffered frame will at least be recent. If we drop the newest the buffered frame will be N frames out of date, for N drops.

rust/foxglove/src/remote_access/session/video_track.rs
eloff: That's better than no signal. It's probably silly, but it feels bad to me to use two implementations of a data structure in the same program when one would do. Does it matter? Probably not, it's just a preference I carry from bygone times. Feel free to ignore.

PR #969, opened 2026-03-04
Feat/mcap events

hcwiley:
### Changelog                                                                                
                                                                                               
Add `foxglove.Event` and `foxglove.EventProperty` message schemas for publishing discrete    
timestamped events to the Foxglove platform.
                                                                                               
### Docs        
                                                                                             
None.           

### Description

Adds two new message schemas, `foxglove.Event` and `foxglove.EventProperty`, to support
publishing discrete events from the SDK that can be ingested and displayed on the Foxglove
timeline.

`foxglove.Event` represents a discrete occurrence at a point in time (or over a duration)
with:
- A required `timestamp` and optional `duration` (zero-duration = instantaneous marker)
- `event_type` for filtering/grouping against platform-defined event categories
- `event_properties` — typed structured properties (`EventPropertyType`: TEXT,
MULTILINE_TEXT, BOOLEAN, NUMBER, SINGLE_SELECT, MULTI_SELECT)
- `metadata` — unstructured key-value pairs complementary to structured properties
- `display_name`, `color`, `id`, and `device_id` for presentation and deduplication

All schema formats are covered: Flatbuffers, JSON Schema, Protobuf, OMG IDL, ROS 1/2.
Generated bindings are updated for Rust, C, C++, and Python SDKs, as well as `foxglove_msgs`
for ROS.


typescript/schemas/src/internal/schemas.ts
```diff
@@ -297,6 +297,135 @@ const KeyValuePair: FoxgloveMessageSchema = {
...
+        "Stable identity for deduplication during data platform ingestion.",
+      protobufFieldNumber: 5,
+    },
+    {
+      name: "device_id",
+      type: { type: "primitive", name: "string" },
+      description:
+        "Device ID this event is associated with. Use the platform device ID when known, or a local identifier (e.g. hostname, serial number). Required so consumers always know the source device.",
+      protobufFieldNumber: 6,
+    },
```
gasmith, line 425: An MCAP file is always in the context of a particular device, so you don't need to embed the device ID here.

typescript/schemas/src/internal/schemas.ts
```diff
@@ -297,6 +297,135 @@ const KeyValuePair: FoxgloveMessageSchema = {
...
+      optional: true,
+      protobufFieldNumber: 4,
+    },
+    {
+      name: "id",
+      type: { type: "primitive", name: "string" },
+      description:
+        "Stable identity for deduplication during data platform ingestion.",
+      protobufFieldNumber: 5,
+    },
```
gasmith, line 418: I feel like, at most, this should be an optional user-provided key that they can use as a stable reference for retrieving this event from the data platform. But it's certainly not a primary key. We shouldn't depend on the user to provide us with a unique ID.

typescript/schemas/src/internal/schemas.ts
```diff
@@ -297,6 +297,135 @@ const KeyValuePair: FoxgloveMessageSchema = {
...
+    {
+      name: "event_properties",
+      type: { type: "nested", schema: EventProperty },
+      array: true,
+      description:
+        "Typed property values matching the platform's structured properties model.",
+      optional: true,
+      protobufFieldNumber: 4,
+    },
```
gasmith, line 411: How would these be reconciled with event types/properties defined in the app? If an event of type X doesn't match the schema for X that the user has defined in the app, what's the expected behavior?

typescript/schemas/src/internal/schemas.ts
```diff
@@ -297,6 +297,135 @@ const KeyValuePair: FoxgloveMessageSchema = {
...
+  description: "Type of an event property value",
+  parentSchemaName: "EventProperty",
+  protobufEnumName: "Type",
+  values: [
+    { value: 0, name: "TEXT", description: "Single-line text" },
+    { value: 1, name: "MULTILINE_TEXT", description: "Multi-line text" },
+    { value: 2, name: "BOOLEAN", description: "True/false" },
+    { value: 3, name: "NUMBER", description: "Numeric value" },
+    { value: 4, name: "SINGLE_SELECT", description: "Single choice from options" },
+    { value: 5, name: "MULTI_SELECT", description: "Multiple choices from options" },
```
gasmith, line 312: What does it mean for a static and immutable event in an mcap file to have a "multi-select" or "single-select" property type?

typescript/schemas/src/internal/schemas.ts
```diff
@@ -297,6 +297,135 @@ const KeyValuePair: FoxgloveMessageSchema = {
...
+    {
+      name: "type",
+      type: { type: "enum", enum: EventPropertyType },
+      description: "Value type",
+      protobufFieldNumber: 2,
+    },
+    {
+      name: "value",
+      type: { type: "primitive", name: "string" },
+      description: "String-encoded value. Interpretation depends on type.",
```
gasmith, line 337: I wonder whether it might be better to have separate fields for the possible value types (a string value, a bool value, a numeric value), instead of adding another layer of encoding.

typescript/schemas/src/internal/schemas.ts
hcwiley: oh! how do we get/enforce that? is the device id stored in the MCAP file metadata?

gasmith: When you import an mcap file into the data platform, you can either import it against a particular device, or as a deviceless recording. And, notably, we don't support events for deviceless recordings.

Most recordings are associated with a device. You can define that association with MCAP metadata, object store metadata, or through the upload API (if you're uploading to foxglove-hosted storage).

typescript/schemas/src/internal/schemas.ts
hcwiley: i originally had it as optional, but after a few conversations i changed it to required. my thinking is it is required the same way that a frame_id is required for an image:
https://github.com/foxglove/foxglove-sdk/blob/9681ac7faef743469cafe6dda128ada1f30ed762/typescript/schemas/src/internal/schemas.ts#L1010-L1014
you can put in junk or "" and that meets the schema requirement, but if we provide good example code + docs on why this id is really important to be unique (use uuid?) it will make customers lives better immediately

typescript/schemas/src/internal/schemas.ts
hcwiley: If an event type doesn't exist by ID or matching name (which would map to name on the cloud platform. so maybe i should key to the foxglove.EventType MCAP too?), then the cloud platform will make a new one and mark it as pending, waiting an admin approval. If the admin approves, it shows up in the UI and works like all the other events. If an admin doesn't approve, it is marked as invalid or disallowed. And so we don't keep doing that loop over and over again, but is not included in the org wide event listing.

typescript/schemas/src/internal/schemas.ts
hcwiley: That is for correlating to the cloud platform so we can render the properties in the app correctly.

typescript/schemas/src/internal/schemas.ts
hcwiley: that's a good idea. something like

`EventPropertyNumber: EventProperty` where the value is then a number?

PR #1125, opened 2026-04-07
feat(schemas): add foxglove.Event schema

hcwiley:
### Changelog

Add `foxglove.Event` schema for embedding timeline events in MCAP recordings.

### Docs

Schema descriptions are auto-generated into `schemas/README.md`. No additional docs needed.

### Description

Adds a new `foxglove.Event` message schema for embedding discrete timeline events in MCAP files. The schema was simplified from the original proposal after team feedback:

```
foxglove.Event {
  start_time: Time        // required — event position on timeline
  label: string           // required — display name (e.g. "FAULT", "STATE_CHANGE")
  end_time?: Time         // optional — omit for point events
  color?: Color           // optional — foxglove.Color (RGBA) for visualization
  metadata?: KeyValuePair[] // optional — untyped key-value string pairs
}
```

Reuses existing schema types (`Timestamp`, `Color`, `KeyValuePair`) — no new sub-schemas needed. This matches the precedent set by `ImageAnnotations`, `SceneEntity`, and `LocationFix` which all use `KeyValuePair[]` for metadata.

**Design decisions:**
- `label` is required (not optional) to ensure every event has a display name
- `color` uses `foxglove.Color` (RGBA floats) rather than a hex string, consistent with other visualization schemas
- `metadata` uses `KeyValuePair[]` (untyped string pairs) instead of typed property values, avoiding the ROS optional-field ambiguity that the previous `EventProperty` approach had
- No `EventType` or `EventProperty` sub-schemas — flattened for simplicity in v1

Verified locally:
- `make generate` — all bindings regenerated successfully
- `yarn test` — 226/226 tests pass, 24/24 snapshots pass
- `make build-cpp` — C++ compiles clean (Docker)

Fixes: [FG-399](https://linear.app/foxglove/issue/FG-399/ingest-events-from-foxgloveevent-message-in-recordings)

### Screenshots
They end up looking like this in app:
Screenshot 2026-04-14 at 11 05 14 AM


typescript/schemas/src/internal/schemas.ts
```diff
@@ -323,6 +323,103 @@ const KeyValuePair: FoxgloveMessageSchema = {
...
+  fields: [
+    {
+      name: "name",
+      type: { type: "primitive", name: "string" },
+      description: "Human-readable category name",
+    },
+    {
+      name: "color",
+      type: { type: "primitive", name: "string" },
+      description: "Hex color string (e.g. RRGGBB or RRGGBBAA, with optional leading #)",
```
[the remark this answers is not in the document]
hcwiley: this is a good callout, and i would consider using the `foxglove.Color` directly instead of string. i think it might create more confusion for our users than alleviate potential misalignment cloud platform.

typescript/schemas/src/internal/generateSdkCpp.ts
```diff
@@ -353,6 +353,9 @@ function cppToC(schema: FoxgloveMessageSchema, copyTypes: Set<string>): string[]
     switch (field.type.type) {
       case "primitive":
         if (field.type.name === "string") {
+          if (field.optional) {
+            return `dest.${dstName} = src.${srcName} ? arena.mapOne<foxglove_string>(*src.${srcName}, [](foxglove_string& dest, const std::string& src, Arena&) { dest = {src.data(), src.size()}; }) : nullptr;`;
+          }
```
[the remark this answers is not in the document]
hcwiley: yea... not sure how we want to handle this. it's kind of the crux of this way of doing it though @gasmith 🤔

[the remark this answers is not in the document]
gasmith: Regarding testing... I was thinking it might be nice to add some examples anyway. Although I don't think we run examples in CI, so maybe a C++ unit test would be a good idea.

typescript/schemas/src/internal/schemas.ts
```diff
@@ -292,6 +292,103 @@ const KeyValuePair: FoxgloveMessageSchema = {
   ],
 };
 
+const EventProperty: FoxgloveMessageSchema = {
+  type: "message",
+  name: "EventProperty",
+  description:
+    "A typed property value for an event. Exactly one of string_value, bool_value, or number_value should be set.",
```
gasmith: I think you'll want to add a discriminator, because not all serialization formats (notably ROS 1 & 2) will allow you to distinguish an unset value from a default value.

typescript/schemas/src/internal/schemas.ts
```diff
@@ -292,6 +292,43 @@ const KeyValuePair: FoxgloveMessageSchema = {
   ],
 };
 
+const Event: FoxgloveMessageSchema = {
```
aneuwald, line 295: this is a bit different from our data platform events: https://docs.foxglove.dev/api#tag/Events/paths/~1events~1%7Bid%7D/get

should we have things like:
- createdAt
- properties ( we have metadata, but there we have both)
- eventTypeId

aneuwald: sorry if this was discussed before and I missed that, just had this on my mind. I know the schemas on MCAP will have a different / reduced usability, but those maybe can be useful somehow.

hcwiley: Good question — and yeah there was a bunch of back and forth on this that's easy to miss! Here's a summary of where we landed and why:

**TL;DR:** We intentionally simplified the MCAP schema to avoid implying parity with the data platform API that we can't actually enforce from a file.

**On `eventTypeId` / `properties`:** Devon, Richard, Ritwik, Mike and I had a [long thread](https://foxglove.slack.com/archives/C0469RNBPEW/p1775680892129029) about whether MCAP events should carry typed event types and properties. The consensus was:

- Calling something an "event_type" in the MCAP implies it maps 1:1 to a platform-defined type — but we can't enforce that from a file (types are mutable, admin-defined, and org-specific). Devon: *"we shouldn't call it a 'type' unless we are saying that it really maps 1:1 with the concept of types in the data platform."*
- Typed `properties` (number, boolean, enum, etc.) create reconciliation headaches during cloud import if the type definition has changed since the MCAP was recorded.
- The cloud API already supports events *without* an `eventTypeId` — untyped metadata is searchable/filterable in the dashboard.

So we went with `label` (display-only name, not a "type") + `metadata: KeyValuePair[]` (untyped string pairs, same as other Foxglove schemas like ImageAnnotations and SceneEntity).

**On `createdAt`:** Not needed in the MCAP — the message's `log_time` in the MCAP container serves this purpose. The platform assigns its own `createdAt` during ingestion.

**Future path:** When the data platform team has bandwidth to build the import pipeline, `label` can optionally be matched to an existing EventType by name within the org. Typed properties and `eventTypeId` can be added to the schema later as an additive change if we find we need them. [Design notes here](https://github.com/foxglove/ai-notes_hcwiley/blob/main/projects/mcap-events/notes/label-to-eventtype-mapping.md) if you want the full picture.

aneuwald: thanks man! I was really not aware of all those discussions, thanks for summarizing it for me!

typescript/schemas/src/internal/schemas.ts
```diff
@@ -292,6 +292,43 @@ const KeyValuePair: FoxgloveMessageSchema = {
...
+  type: "message",
+  name: "Event",
+  description: "A discrete event that occurred at a specific time or over a time range",
+  fields: [
+    {
+      name: "start_time",
+      type: { type: "nested", schema: Timestamp },
+      description: "Event start time",
+    },
```
indirectlylit, line 304: can this be deleted because the message itself already has a log time?

if not, what's the distinction?

```suggestion
```

typescript/schemas/src/internal/schemas.ts
```diff
@@ -292,6 +292,43 @@ const KeyValuePair: FoxgloveMessageSchema = {
...
+    {
+      name: "end_time",
+      type: { type: "nested", schema: Timestamp },
+      description: "Event end time. Omit for point events.",
+      optional: true,
+    },
```
indirectlylit, line 310: I forget if this had some discussion already, but what drove the decision to use end time instead of duration?

```suggestion
      name: "duration",
      type: { type: "nested", schema: Timestamp },
      description: "Event duration",
    },
```

indirectlylit (commented):
Hey, sorry finally got around to looking at this.

Question: in the work trial I recall some folks advocating for pairs of EventStart and EventEnd messages. Did we explicitly decide against that?

typescript/schemas/src/internal/schemas.ts
hcwiley: my thinking on this:
the timestamp of event message might NOT be the time of the event you are attempting to describe, it's the time the event was logged. those are not necessarily the same thing.

typescript/schemas/src/internal/schemas.ts
hcwiley: there was discussion of whether duration or end made more sense. i think during my wt i had duration but i think end actually is better and duration is a derived property from that. i'm not strongly opinionated on this though and could be swayed by a strong feeling by someone for duration over end.

hcwiley:
> Hey, sorry finally got around to looking at this.
> 
> Question: in the work trial I recall some folks advocating for pairs of EventStart and EventEnd messages. Did we explicitly decide against that?

i don't love the idea of separate start and end messages because it feels like something i would forget to do as an engineer. 

> did you remember to turn off the stove?

personally, i would rather the Event started/stopped than "something is happening" "oh that thing ended" and having to remember which thing started that you now need to stop.

indirectlylit (approved):
I feel satisfied with these changes, but I would appreciate maybe @defunctzombie, @ritwik, and/or @gasmith to also weigh in

typescript/schemas/src/internal/schemas.ts
ritwik: I agree with that - we should let you log an event after some processing, which might be a few cycles after it happened.

ritwik (approved):
I didn't review each language (I take it they are all generated?) but the ts changes and the schema as described lgtm with the context of all the discussion we've had on it.

typescript/schemas/src/internal/__snapshots__/exportTypeScriptSchemas.test.ts.snap
```diff
@@ -326,6 +326,31 @@ export type Duration = {
...
+/** A discrete event that occurred at a specific time or over a time range */
+export type Event = {
+  /** Event start time */
+  start_time: Time;
+
+  /** Event end time. Omit for point events. */
+  end_time?: Time;
+
+  /** Display name for the event */
+  label: string;
```
ritwik, line 346: some good feedback that has come up is that this entity is meant to map to untyped events but `label` is already divergent from that, so we should either remove that or come up with the plan for that so we can discuss it.

indirectlylit:
@hcwiley has been pulled on to other work, so putting this in draft until someone is able to pick up the overall project and run with it

defunctzombie:
I'm going to take over landing this schema work based on latest conversations. Transferring PR here: https://github.com/foxglove/foxglove-sdk/pull/1368

PR #1140, opened 2026-04-09
cpp: Build remote access for 22.04

gasmith:
### Changelog
None

### Docs
None

### Description
It turns out we can build remote access for 22.04, which will be instrumental in enabling humble builds of the foxglove bridge on the ROS build farm.

In a subsequent change, we're gonna move the ros jobs into the c/cpp workflow, so that we can re-use the build artifacts there.

### DIFF
--- a/.github/workflows/c_cpp.yml
+++ b/.github/workflows/c_cpp.yml
@@ -38,10 +38,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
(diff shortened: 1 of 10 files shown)


eloff (approved):
Nice work! This really simplifies the CI for C++ and gets us Ubuntu 22.04 support.

Not as part of this work, but I wonder if we can get Linux clang support working if we build with glibc and not libc++. I think that's supported.

gasmith:
> Not as part of this work, but I wonder if we can get Linux clang support working if we build with glibc and not libc++. I think that's supported.

Worth a shot, I filed FLE-420

PR #1142, opened 2026-04-10
ci: Update ros workflow to use remote-access build

gasmith:
### Changelog
None

### Docs
None

### Description
The idea here is to lay the groundwork for building the bridge with
remote access support. Rather than USE_LOCAL_SDK, we introduce
FOXGLOVE_CPP_SDK_DIR, and construct an artifact that matches what
we would typically download from the github releases page.


eloff (approved):
LGTM

USE_LOCAL_SDK always felt like a bit of a hack, this is nicer

PR #1144, opened 2026-04-10
cpp: Add ChannelDescriptor::id()

gasmith:
### Changelog
- cpp: Add ChannelDescriptor::id()

### Docs
None

### Description
We forgot to add an ID accessor method on the ChannelDescriptor when we
plumbed it through the C/C++ bindings.


eloff (approved):
LGTM

PR #1145, opened 2026-04-10
foxglove_bridge: Add support for remote access

gasmith:
### Changelog
- foxglove_bridge: Added support for remote access

### Docs
Forthcoming; we'll update docs.foxglove.dev with the new configuration.

### Description
This change adds support for remote access to the foxglove bridge.

Since the remote access functionality is only exposed via the shared
library, this change updates cmake to link against that instead of the
static library.

Other changes here:
- Plumb the assets capability, for symmetry with rust API
- Embed the soname in the cdylib so that things linking against it don't link by absolute path
- Add deb-{dist} targets to ros makefile for packaging debs


ros/src/foxglove_bridge/CMakeLists.txt
```diff
@@ -74,16 +76,25 @@ FetchContent_Declare(
...
   PUBLIC
     $<BUILD_INTERFACE:${foxglove_sdk_SOURCE_DIR}/include>
     $<INSTALL_INTERFACE:include>
 )
-file(GLOB_RECURSE FOXGLOVE_SDK_SOURCES CONFIGURE_DEPENDS "${foxglove_sdk_SOURCE_DIR}/src/*.cpp")
-target_sources(foxglove_cpp_static PRIVATE ${FOXGLOVE_SDK_SOURCES})
-set_target_properties(foxglove_cpp_static PROPERTIES POSITION_INDEPENDENT_CODE ON)
-target_link_libraries(foxglove_cpp_static PRIVATE ${foxglove_sdk_SOURCE_DIR}/lib/libfoxglove.a)
+target_sources(foxglove_cpp_shared PRIVATE ${FOXGLOVE_SDK_SOURCES})
+# Import the prebuilt C shared library from the fetched SDK
```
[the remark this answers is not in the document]
gasmith: We're gonna ship remote access in all distributions, so it's pretty reasonable.

ros/src/foxglove_bridge/CMakeLists.txt
```diff
@@ -19,6 +19,8 @@ if (POLICY CMP0135)
   cmake_policy(SET CMP0135 NEW)
 endif()
 
+option(FOXGLOVE_BRIDGE_REMOTE_ACCESS "Build with remote access gateway support" ON)
```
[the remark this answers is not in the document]
gasmith: We're gonna ship remote access in all distributions, so it's pretty reasonable.

ros/src/foxglove_bridge/include/foxglove_bridge/ros2_foxglove_bridge.hpp
```diff
@@ -32,7 +36,6 @@ using MapOfSets = std::unordered_map<std::string, std::unordered_set<std::string
 using ServicesByType = std::unordered_map<std::string, std::string>;
 
 using ClientId = uint32_t;
-using SinkId = uint64_t;
```
clalancette, line 35: I think this removal is premature.  Both `createOrIncrementSubscription` and `createOrIncrementSubscriptionLocked` take a sinkId parameter (which is `uint64_t`).  Leaving this line and then using `SinkId` as the type would be more self-documenting.

ros/Makefile
```diff
@@ -64,6 +74,47 @@ docker-test-$(1):
 docker-sh-$(1):
 	docker run --rm -it $(DOCKER_RUN_ARGS) -v $(PWD)/..:/sdk foxglove-sdk-ros-$(1)
 
+.PHONY: deb-bridge-$(1)
```
clalancette: We shouldn't need hand-rolled debian packaging stuff for ROS packaging (all of that is handled by bloom).  Given that, who is the target audience for this debian-related changes?

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -224,6 +224,111 @@ FoxgloveBridge::FoxgloveBridge(const rclcpp::NodeOptions& options)
         _server->broadcastTime(static_cast<uint64_t>(timestamp));
       });
   }
+
+#ifndef FOXGLOVE_REMOTE_ACCESS
+  if (this->get_parameter(PARAM_REMOTE_ACCESS).as_bool()) {
+    RCLCPP_ERROR(this->get_logger(),
+                 "remote_access is set to true but the bridge was not built with "
+                 "FOXGLOVE_BRIDGE_REMOTE_ACCESS=ON. Remote access is not available.");
```
clalancette: This kind of feels like it should be a hard-error (throw an exception).  This is a configuration that can't possibly work, so I think we should just get out and let the user configure it correctly.

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -591,92 +748,214 @@ void FoxgloveBridge::subscribeConnectionGraph(bool subscribe) {
...
+    {
+      std::lock_guard<std::mutex> lock(_clientAdvertisementsMutex);
+      auto parserIt = _jsonParsers.find(schemaName);
+      if (parserIt != _jsonParsers.end()) {
+        parser = parserIt->second;
+      }
+    }
+    if (!parser) {
+      throw std::runtime_error("no JSON parser found for schema \"" + schemaName + "\"");
+    }
```
clalancette: I struggled quite a bit with this locking/lookup, and at the end of the day I don't like it.  It seems unnecessary that we have to lookup the jsonParser for every message we send.  We already know which parser we need at `createClientPublisher` time.  So the suggestion from Claude is to store the `jsonParser` inside of `struct ClientAdvertisement`, and then change the API for `publishClientData` to:

```
void publishClientData(const ClientAdvertisement& ad, const std::byte* data, size_t dataLen);
```

Then we can lookup the correct `jsonParser` directly from the ClientAdvertisement passed in.  What do you think?

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -591,92 +748,214 @@ void FoxgloveBridge::subscribeConnectionGraph(bool subscribe) {
...
+                topic.c_str(), datatype.c_str(), channelId);
+  }
+
+  // For transient_local topics, replay the cached message to the new client before adding
+  // them to the broadcast set, so they don't miss the latched value.
+  if (!isNewSubscription && subIt->second.hasCachedMessage && sinkId.has_value()) {
+    channel.log(reinterpret_cast<const std::byte*>(subIt->second.cachedMessageData.data()),
+                subIt->second.cachedMessageData.size(), subIt->second.cachedMessageTimestamp,
+                sinkId.value());
+  }
```
clalancette: This seems potentially problematic (or, at the least, it is surprising in at least one case).  If the ROS side has a topic with transient_local, and a history depth > 1, then what happens for a late-joining subscriber is this:

1.  First subscriber - skips this (it is a new subscription), but the ROS side (DDS) replays all of the data to the subscriber, so the downstream consumer gets all of the data through the normal delivery mechanism.
2. Subsequent subscribers - goes in here, and gets only the latest message.

This is honestly a corner-case; transient_local with a history depth > 1 aren't *that* common.  But they do exist, and they wouldn't work properly here, which could be surprising.

The obvious fix would be to maintain a cache, but then we have to worry about cache size, etc.  So I'm not sure what we should do here, but I did want to point this out.

clalancette (commented):
I've left some things to think about here.

ros/Makefile
gasmith: I wanted a makefile target that I could use to generate a .deb locally from unreleased sources. Is there a bloom command I can use for that?

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
gasmith: This was just carried forward from the original client message handling, but I'm not against the optimization. Let's do it.

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
gasmith: Mmmh, yeah, we probably need to fix this. We learn the depth in determineQoS, and there's a param for max queue depth, so at least it's not an unbounded cache.

ros/Makefile
clalancette: Yes, see https://docs.ros.org/en/rolling/How-To-Guides/Building-a-Custom-Deb-Package.html .  It's a little fiddly (it only supports building one package at a time), but it does work.  Note that there is a caveat when building `foxglove_bridge`, because it pulls from an external source and Ubuntu builds disallow that by default (ROS turns this on by default on the buildfarm).  So you need to do something like this:

```
bloom-generate rosdebian && sed -i 's/dh_auto_configure --/dh_auto_configure -- -DFETCHCONTENT_FULLY_DISCONNECTED=OFF/' debian/rules && fakeroot debian/rules binary
```

ros/src/foxglove_bridge/include/foxglove_bridge/ros2_foxglove_bridge.hpp
```diff
@@ -76,9 +84,36 @@ class FoxgloveBridge : public rclcpp::Node {
...
+    size_t maxMessages = 1;
+  };
+  struct ChannelSubscription {
+    Subscription rosSubscription;
+    std::unordered_set<ClientId> wsClientIds;
+    std::unordered_set<ClientId> gatewayClientIds;
+    rclcpp::QoS qos{10};
+    // Per-publisher message cache for transient_local topics, replayed to late subscribers.
+    std::map<Gid, PublisherCache> publisherCaches;
+    bool isTransientLocal = false;
```
clalancette: This is redundant; you can get the same information by checking whether `qos.durability() == rclcpp::DurabilityPolicy::TransientLocal`.  That latter is a mouthful, but we only use it in one place, so I'd save the space in the structure and use the qos field.

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -918,24 +1097,62 @@ void FoxgloveBridge::unsubscribeParameters(const std::vector<std::string_view>&
...
+        }
+      }
+    }
+    pubCache.messages.push_back(CachedMessage{
+      {rclSerializedMsg.buffer, rclSerializedMsg.buffer + rclSerializedMsg.buffer_length},
+      static_cast<uint64_t>(timestamp),
+    });
+    while (pubCache.messages.size() > pubCache.maxMessages) {
+      pubCache.messages.pop_front();
+    }
```
clalancette: This is minor, but pushing before popping allows the pubCache to temporarily grow beyond the maxMessage depth.  If We are already at the maxMessage size, then on the next message we grow to maxMessage + 1, before popping one off of the front.  It's not the biggest deal, but it is probably worth popping first if we are going to overflow.

Also, the while loop seems unnecessary; we know that we will always pop off either 0 or 1 items (since this handles one message at a time).

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -918,24 +1097,62 @@ void FoxgloveBridge::unsubscribeParameters(const std::vector<std::string_view>&
...
+    Gid gid;
+    const auto& rawGid = messageInfo.get_rmw_message_info().publisher_gid;
+    std::copy(rawGid.data, rawGid.data + RMW_GID_STORAGE_SIZE, gid.begin());
+
+    auto& pubCache = subIt->second.publisherCaches[gid];
+    if (pubCache.messages.empty()) {
+      // New publisher — look up its QoS depth
+      pubCache.maxMessages = 1;
+      const std::string topic(channelIt->second.topic());
+      for (const auto& pub : this->get_publishers_info_by_topic(topic)) {
```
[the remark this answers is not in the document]
clalancette: This really shouldn't be much of a problem; this is a local lookup, and we only do this on the first message from a particular publisher.  So I don't think any wacky locking schemes are worth it here.

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -933,9 +1122,21 @@ void FoxgloveBridge::rosMessageHandler(ChannelId channelId, SinkId sinkId,
...
 
+  // Cache the message for transient_local subscriptions so late subscribers can receive it.
+  auto subIt = _subscriptions.find(channelId);
+  if (subIt != _subscriptions.end() &&
+      subIt->second.qos.durability() == rclcpp::DurabilityPolicy::TransientLocal) {
+    subIt->second.cachedMessageData.assign(
+      rclSerializedMsg.buffer, rclSerializedMsg.buffer + rclSerializedMsg.buffer_length);
+    subIt->second.cachedMessageTimestamp = timestamp;
+    subIt->second.hasCachedMessage = true;
+  }
```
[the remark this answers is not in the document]
clalancette: This is a pre-existing problem, and is honestly pretty minor.  I guess a *very* busy topic could starve out an subscribe/unsubscribe, but I think there are probably other problems like that in here as well.  I'd not worry about this one.

clalancette (commented):
A couple more (minor!) issues, then I think this is ready.

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -1113,6 +1375,143 @@ void FoxgloveBridge::publishClientCount() {
...
+  RCLCPP_INFO(this->get_logger(), "Remote access gateway status: %s", label);
+}
+
+void FoxgloveBridge::gatewaySubscribe(uint32_t clientId,
+                                      const foxglove::ChannelDescriptor& channel) {
+  RCLCPP_INFO(this->get_logger(),
+              "Gateway: received subscribe request for channel %lu (\"%s\") from client %u",
+              channel.id(), std::string(channel.topic()).c_str(), clientId);
+  createOrIncrementSubscription(channel.id(), clientId, true, _gateway->sinkId());
```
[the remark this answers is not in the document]
gasmith: We'll add a warning for this particular case, to be defensive.

Note that the gateway _does_ fire `onUnsubscribe` callbacks when the session is disconnected.

ros/src/foxglove_bridge/CMakeLists.txt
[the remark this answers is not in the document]
gasmith: ROS is Linux only, and this CMakeLists.txt already panics for non-Linux platforms.

Dockerfile.ros
```diff
@@ -26,7 +26,8 @@ RUN if [ "$USE_ROS_TESTING" = "true" ]; then \
 # That ensures that we have the latest versions of the
 # ROS packages in the underlay and for what is installed
```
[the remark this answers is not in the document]
gasmith: Meh.

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -1113,6 +1378,151 @@ void FoxgloveBridge::publishClientCount() {
...
+      label = "connected";
+      break;
+    case foxglove::RemoteAccessConnectionStatus::ShuttingDown:
+      label = "shutting down";
+      break;
+    case foxglove::RemoteAccessConnectionStatus::Shutdown:
+      label = "shutdown";
+      break;
+  }
+  RCLCPP_INFO(this->get_logger(), "Remote access gateway status: %s", label);
```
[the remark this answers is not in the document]
gasmith: Yes, `onClientUnadvertise` will fire.

ros/README.md
```diff
@@ -0,0 +1,65 @@
...
+### Run tests
+
+```sh
+make docker-test          # rolling
+make docker-test-humble   # specific distro
+```
+
+### Build .deb packages
+
+Uses [bloom](https://wiki.ros.org/bloom) to generate Debian packaging from
```
clalancette, line 33: Nit: The ROS wiki is now read-only and will eventually be shutdown.  Most of the documentation for bloom was ported over to docs.ros.org: https://docs.ros.org/en/rolling/How-To-Guides/Releasing/Releasing-a-Package.html

cpp/foxglove/include/foxglove/remote_access.hpp
```diff
@@ -229,6 +233,13 @@ class RemoteAccessGateway final {
   /// @param graph The connection graph to publish.
   [[nodiscard]] FoxgloveError publishConnectionGraph(const ConnectionGraph& graph) const;
 
+  /// @cond foxglove_internal
+  /// @brief Get the sink ID of the gateway's current session.
+  ///
+  /// Returns std::nullopt if no session is currently active.
+  [[nodiscard]] std::optional<uint64_t> sinkId() const;
```
clalancette, line 240: It occurs to me that since we kept the `SinkId` type in `ros2_foxglove_bridge.hpp`, we could probably also use it here.  The right fix would be to move the `using SinkId = uint64_t` here into remote_access.hpp, then use it from `ros2_foxglove_bridge.hpp`.  Unfortunately it isn't that straightforward, as this header is only optionally included.

I still think we should fix this, but I'm fine if we file a ticket and do it as a follow-up.

ros/src/foxglove_bridge/include/foxglove_bridge/ros2_foxglove_bridge.hpp
```diff
@@ -136,8 +170,28 @@ class FoxgloveBridge : public rclcpp::Node {
...
+  void removeOrDecrementSubscription(ChannelId channelId, ClientId clientId, bool isGateway);
+  void removeOrDecrementSubscriptionLocked(ChannelId channelId, ClientId clientId, bool isGateway);
+
+  // Shared helpers for client publish (used by both WebSocket and gateway paths).
+  // Must be called with _clientAdvertisementsMutex held. May throw.
+  ClientAdvertisement createClientPublisher(const std::string& topicName,
+                                            const std::string& topicType,
+                                            const std::string& encoding,
+                                            const std::byte* schemaData, size_t schemaLen);
+  // Must NOT be called with _clientAdvertisementsMutex held. May throw.
```
clalancette: I don't know that this comment is strictly true anymore.  At least, as far as I can tell `publishClientData` no longer takes the mutex, so in theory it would be OK to call it with the mutex held (even though none of the current callers do this).  I'll suggest updating this comment (or maybe removing it altogether).

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -296,85 +439,101 @@ void FoxgloveBridge::updateAdvertisedTopics(
...
+      _channels.insert({channelId, std::move(channelResult.value())});
     }
+  }
 
-    const ChannelId channelId = channelResult.value().id();
-    RCLCPP_INFO(this->get_logger(), "Advertising new channel %lu for topic \"%s\"", channelId,
-                topic.c_str());
-    _channels.insert({channelId, std::move(channelResult.value())});
+  // Close channels after releasing _subscriptionsMutex, since close() may fire
+  // onUnsubscribe callbacks that need to acquire _subscriptionsMutex.
```
clalancette, line 534: So this is a bit racy without the lock held; it can't race with itself since this is only called from `rosgraphPollThread`, which is single-threaded, but it can race with `createOrIncrementSubscriptionLocked` and `rosMessageHandler`.   In practice this works fine because we've removed them from `_channels` and thus both of those can't find the subscription.  I think it is worth adding a comment here about this, something like:

```suggestion
  // Close channels after releasing _subscriptionsMutex, since close() may fire
  // onUnsubscribe callbacks that need to acquire _subscriptionsMutex.
  // This is safe to do outside of the lock because this function is only called
  // single-threaded from rosgraphPollThread, and the removal of the channel
  // from _channels mean both createOrIncrementSubscriptionLocked and
  // rosMessageHandler gracefully ignore channels they can't find.
```

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -580,103 +765,254 @@ void FoxgloveBridge::updateConnectionGraph(
 
   RCLCPP_INFO(this->get_logger(), "publishing connection graph");
   _server->publishConnectionGraph(connectionGraph);
+#ifdef FOXGLOVE_REMOTE_ACCESS
+  if (_gateway) {
+    (void)_gateway->publishConnectionGraph(connectionGraph);
```
clalancette, line 775:
```suggestion
    _gateway->publishConnectionGraph(connectionGraph);
```

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -918,24 +1155,52 @@ void FoxgloveBridge::unsubscribeParameters(const std::vector<std::string_view>&
...
   // NOTE: Do not call any RCLCPP_* logging functions from this function. Otherwise, subscribing
   // to `/rosout` will cause a feedback loop
   const auto timestamp = this->now().nanoseconds();
   assert(timestamp >= 0 && "Timestamp is negative");
   const auto rclSerializedMsg = msg->get_rcl_serialized_message();
 
   std::lock_guard<std::mutex> lock(_subscriptionsMutex);
-  if (_channels.find(channelId) == _channels.end()) {
+  auto channelIt = _channels.find(channelId);
+  if (channelIt == _channels.end()) {
```
clalancette, line 1181: Nit: this can be:


```suggestion
  if (_channels.count() == 0) {
```

clalancette (approved):
I've left a bunch of nitpicky stuff in here; on the whole, this is fine.  If you'd like to defer the nitpicks to another PR I'm fine with that, so I'll approve anyway.

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
gasmith: Looks like the remote access gateway propagates the error from the C FFI, whereas the websocket server swallows it. We should make those consistent. Filed https://linear.app/foxglove/issue/FLE-430/publishconnectiongraph-error-propagation-should-be-consistent

cpp/foxglove/include/foxglove/remote_access.hpp
gasmith: Yeah, that change would involve updating all the codegen'd channel classes and stuff too. I'll punt: https://linear.app/foxglove/issue/FLE-434/add-sinkid-to-c-headers

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
gasmith: Mmh, yeah... but then we have a redundant channels.at() down below... I'm going to keep this as it is, but remove the redundant channels.at() to reuse the iterator from here.

PR #1148, opened 2026-04-11
cpp: Restrict log levels to foxglove crate

gasmith:
### Changelog
None

### Docs
None

### Description
When a user sets a debug log level, that should be scoped to the
foxglove crate, not its transitive dependencies. This is especially
important with libwebrtc, which is insanely noisy at the debug level.


eloff (approved):
LGTM

PR #1149, opened 2026-04-11
remote_access: Reduce log level for unhandled room events

gasmith:
### Changelog
None

### Docs
None

### Description
Room events are somewhat noisy, and their debug representation tends to
be quite large, often including raw data as a sequence of byte values.
That's not very helpful.


eloff (approved):
LGTM

PR #1150, opened 2026-04-11
remote_access: Accept client channels without schema data

gasmith:
### Changelog
None

### Docs
None

### Description
We need to accept client-advertised channels that use well-known (e.g.
ROS) schemas without schema data. Websocket already does this.


eloff (approved):
Good catch. LGTM

PR #1151, opened 2026-04-11
ci: Build C libs separately for C++ lint

gasmith:
### Changelog
None

### Docs
None

### Description
Build C libs outside of corrosion, so that we can take advantage of the
setup-rust-toolchain cache, and get some speedups.


eloff (approved):
LGTM

PR #1153, opened 2026-04-11
remote access: data tracks prototype for sdk -> app data channels

eloff:
### Changelog

None

### Docs

None

### Description

App side: https://github.com/foxglove/app/pull/13790

This is a prototype implementation of implementing sdk data channels over data tracks instead of over bytestreams.

Like with the previous implementation, each channel gets its own track and they're published eagerly when the device connects to the room and not closed.

It's a lossy protocol. Both sides have a fixed size buffer per side in terms of frames. A frame is not a packet, but a whole message. If you exceed the configured frame buffer (currently 64 in app, and 16 in sdk) then the next message to be sent or received will be dropped. Losing a packet of a message results in the message dropping. Messages seem to be re-ordered on the receiving end based on timestamp, its unclear to me if out-of-order messages are possible, that would need further investigation.

I managed to avoid the MessageData frame since the channel id is in the track name, and we can set a u64 timestamp with each message. That's enough to construct the message with timestamp and channel id on the app side without extra framing.

This does not handle client publish from the app to the sdk.

### Testing

- [x] Run remote access example locally and confirm data channel messages are delivered to app


rust/foxglove/src/remote_access/session.rs
```diff
@@ -235,11 +215,12 @@ impl Sink for RemoteAccessSession {
...
-        // Send to data subscribers.
-        if state.has_data_subscribers(&channel_id) {
-            drop(state);
-            let message = ServerMessageData::new(u64::from(channel_id), metadata.log_time, msg);
-            self.send_data_lossy(ChannelMessage::binary(channel_id, &message));
+        // Send to data subscribers via data track.
+        if let Some(track) = state.get_data_track(&channel_id) {
+            let frame = DataTrackFrame::new(Vec::from(msg)).with_user_timestamp(metadata.log_time);
+            if let Err(e) = track.try_push(frame) {
+                info!("data track frame dropped for channel {channel_id:?}: {e:?}");
```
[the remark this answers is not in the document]
gasmith: Agree; this should probably be a throttled `debug!` message.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -876,6 +778,32 @@ impl RemoteAccessSession {
...
+                    .await
+                {
+                    Ok(track) => {
+                        session.state.write().insert_data_track(channel_id, track);
+                    }
+                    Err(e) => {
+                        error!("failed to publish data track for channel {channel_id:?}: {e:?}");
+                    }
+                }
+            });
```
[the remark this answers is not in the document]
gasmith: Fully agree. We should buffer message until the track is published. I wouldn't be surprised if the bridge depends on this behavior, e.g. to send a latched messages immediately upon subscription.

Do we have any sense for how long it takes to register a data track?

rust/foxglove/src/remote_access/session_state.rs
```diff
@@ -14,13 +15,18 @@ use crate::remote_common::ClientId;
 use crate::{ChannelDescriptor, ChannelId, RawChannel};
 
 /// Channels and parameters that lost their last subscriber when a participant was removed.
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
gasmith: Yeah, seems like we can drop this and `last_data_unsubscribed`.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -876,6 +778,32 @@ impl RemoteAccessSession {
...
+            tokio::spawn(async move {
+                let topic = format!("{CHANNEL_TOPIC_PREFIX}{}", u64::from(channel_id));
+                match session
+                    .room
+                    .local_participant()
+                    .publish_data_track(topic)
+                    .await
+                {
+                    Ok(track) => {
+                        session.state.write().insert_data_track(channel_id, track);
```
gasmith: This looks racy with respect to concurrent subscribe/unsubscribe calls. I think we need some kind of wrapper that we can synchronously stash in SessionState, like we do with VideoPublisher.

gasmith (commented):
This generally looks pretty good. My biggest concern is about the async process of publishing a data track. Feels like we want to reuse some of the ideas we had for the VideoPublisher.

rust/foxglove/src/remote_access/protocol_version.rs
```diff
@@ -9,16 +9,14 @@ use tracing::error;
...
 /// The minimum remote access protocol version this SDK will accept from a connecting participant.
 pub(crate) const REMOTE_ACCESS_MIN_SUPPORTED_PROTOCOL_VERSION: semver::Version =
-    semver::Version::new(2, 0, 0);
+    semver::Version::new(2, 1, 0);
 
 /// The protocol version assumed when a participant does not advertise one.
-///
-/// This is the version that was in use before version advertisement was introduced.
-pub(crate) const DEFAULT_PROTOCOL_VERSION: semver::Version = semver::Version::new(2, 0, 0);
+pub(crate) const DEFAULT_PROTOCOL_VERSION: semver::Version = semver::Version::new(2, 1, 0);
```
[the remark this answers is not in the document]
eloff: This isn't deployed yet, so we can do this

docker-compose.yaml
```diff
@@ -1,8 +1,8 @@
 services:
   livekit:
-    image: livekit/livekit-server:v1.9.11
+    image: livekit/livekit-server:latest
```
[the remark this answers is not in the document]
eloff: It also means if livekit ships a version that breaks something, we don't find out about it. I'd rather CI breaks and alerts us than production breaks and alerts us

rust/remote_access_tests/tests/netem_test.rs
```diff
@@ -302,29 +302,26 @@ async fn netem_burst_delivery_under_impairment() -> Result<()> {
...
-            other => anyhow::bail!("expected MessageData, got: {other:?}"),
-        }
+        assert_eq!(msg.channel_id, channel_id);
+        assert_eq!(
+            msg.data.as_ref(),
+            expected.as_bytes(),
+            "message {i} out of order or missing"
+        );
     }
     info!("all {count} messages delivered in order under impairment");
```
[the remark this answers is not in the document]
eloff: This test will have to be deleted or disabled until such a time as we implement reliable data channels.

Maybe it could be reworked to test control channel messages like connection graph diffs.

docker-compose.yaml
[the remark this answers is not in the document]
dave-foxglove: sadly that would mean that someone potentially trying to do something unrelated starts having their pr ci fail. Is there another way to handle this? Maybe something which separately runs a job to run latest, but isn't every pr trying to land?

If we decide to continue with "latest", please add a # comment describing why this is done (because someone could think this is a whoops and 'fix it').

rust/foxglove/src/remote_access/protocol_version.rs
```diff
@@ -9,16 +9,14 @@ use tracing::error;
 pub(crate) const PROTOCOL_VERSION_ATTRIBUTE: &str = "protocolVersion";
 
 /// The remote access protocol version supported by this SDK build.
-pub(crate) const REMOTE_ACCESS_PROTOCOL_VERSION: semver::Version = semver::Version::new(2, 0, 1);
+pub(crate) const REMOTE_ACCESS_PROTOCOL_VERSION: semver::Version = semver::Version::new(2, 1, 0);
 
 /// The minimum remote access protocol version this SDK will accept from a connecting participant.
 pub(crate) const REMOTE_ACCESS_MIN_SUPPORTED_PROTOCOL_VERSION: semver::Version =
-    semver::Version::new(2, 0, 0);
+    semver::Version::new(2, 1, 0);
```
dave-foxglove, line 16: Would this be worthy of mentioning in the changelog section of the PR description

rust/foxglove/src/remote_access/session/data_track_publisher.rs
```diff
@@ -0,0 +1,160 @@
...
+    pub fn take_track(&self) -> Option<LocalDataTrack> {
+        let mut inner = self.inner.lock();
+        inner.buffer.clear();
+        inner.track.take()
+    }
+
+    fn drain_buffer(buffer: &mut Vec<DataTrackFrame>, track: &LocalDataTrack) {
+        for frame in buffer.drain(..) {
+            if track.try_push(frame).is_err() {
+                break;
```
dave-foxglove: If this is dropping frames (I may be misunderstanding, but if it is) would this be a good place to add a counter metric or similar for number of frames dropped.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -2257,110 +2194,132 @@ impl RemoteAccessSession {
...
+                }
+
+                let topic = format!("{CHANNEL_TOPIC_PREFIX}{}", u64::from(channel_id));
+                let track = {
+                    const INITIAL_BACKOFF: Duration = Duration::from_millis(100);
+                    const MAX_BACKOFF: Duration = Duration::from_secs(3);
+                    const TIMEOUT: Duration = Duration::from_secs(60);
+                    let deadline = tokio::time::Instant::now() + TIMEOUT;
+                    let mut backoff = INITIAL_BACKOFF;
+                    loop {
```
dave-foxglove: Should this respect a cancellation? like will this potentially just keep trying for 60 seconds or so. Would it help to use tokio::select! and look for a cancellation token?

rust/foxglove/src/remote_access/session/data_track_publisher.rs
```diff
@@ -0,0 +1,160 @@
+use std::fmt;
+
```
dave-foxglove: Should/could this file be called data_track_publisher.rs
Or  maybe a //! doc comment explaining that the file is about

rust/foxglove/src/remote_access/session/data_track_publisher.rs
```diff
@@ -0,0 +1,160 @@
...
+use livekit::prelude::{DataTrackFrame, LocalDataTrack, PushFrameError};
+
+/// Buffers data track frames while a LiveKit data track is being published.
+///
+/// Inserted synchronously into [`SessionState`] before the async `publish_data_track` call,
+/// so that [`Sink::log`] can push frames immediately. Frames are buffered internally
```
dave-foxglove: Is a data track lossy? 

(Feels odd buffering output unless we expect or need some data to arrive, which would conflict with lossy-ness).

gasmith: I suggested this approach, so I'll defend it. In ROS, there's the notion of a "latched" topic - publishers publish once (or very infrequently), and when a subscriber subscribes, they immediately get the last message published to that topic, without having to wait for the next one to be published.

The SDK does not have this concept, and so the foxglove bridge (as the glue between ROS and the SDK) needs to emulate it. The emulation isn't particularly interesting for this discussion... but imagine what happens when a client subscribes to a latched topic for the first time. The bridge gets an Listener::on_subscribe() callback, it subscribes to the corresponding ROS topic, and very soon afterwards receives the latched message from ROS, which it now must forward through the SDK. If this Data Track isn't set up yet, the SDK will drop the message, and we'll never deliver it. In the worst (and unfortunately common, cf. `/tf_static`) case, there are no updates published to the latched topic, and so the client never gets this data.

While we do agree that the Data Track (and, by extension, SDK logging over Remote Access) are lossy, I think it's worth handling this particular case in the name of "best effort".

rust/foxglove/src/remote_access/session/data_track_publisher.rs
```diff
@@ -0,0 +1,160 @@
...
+/// until [`set_track`](Self::set_track) installs the published track, at which point the
+/// buffer is drained. After that, frames are forwarded directly to the track.
+pub(crate) struct DataTrackPublisher {
+    inner: parking_lot::Mutex<DataTrackInner>,
+}
+
+struct DataTrackInner {
+    track: Option<LocalDataTrack>,
+    buffer: Vec<DataTrackFrame>,
+}
```
gasmith: Nit: Maybe an enum?

rust/foxglove/src/remote_access/session/data_track_publisher.rs
```diff
@@ -0,0 +1,160 @@
...
+                buffer: Vec::new(),
+            }),
+        }
+    }
+
+    /// Push a frame to the data track, buffering if the track is not yet published.
+    ///
+    /// When the track is available, any buffered frames are drained first.
+    /// When the pre-publish buffer is full, new frames are dropped to preserve
+    /// the earliest (potentially most important) frames.
```
gasmith: Nit: I'd drop the "potentially most important" bit; I don't think that'll age well. The real reason we drop the newest, is that this is the same semantics as DataTrack::try_push. I think if we had our druthers, we'd keep the semantics we previously had with send_data_lossy, and drop the oldest.

rust/foxglove/src/remote_access/protocol_version.rs
eloff: Not since it's not public yet. The changelog is for our customers, not us. At least that's how I understand it.

rust/foxglove/src/remote_access/session/data_track_publisher.rs
eloff: It shouldn't drop frames since both buffers are sized the same (from the same const). It's also just a one-time thing when setting up the data track. The frame dropping will happen on the log path.

rust/foxglove/src/remote_access/session.rs
eloff: Yes, good catch

rust/foxglove/src/remote_access/session.rs
```diff
@@ -235,11 +220,14 @@ impl Sink for RemoteAccessSession {
...
-        // Send to data subscribers.
-        if state.has_data_subscribers(&channel_id) {
-            drop(state);
-            let message = ServerMessageData::new(u64::from(channel_id), metadata.log_time, msg);
-            self.send_data_lossy(ChannelMessage::binary(channel_id, &message));
+        // Send to data subscribers via data track publisher.
+        if let Some(publisher) = state.get_data_track_publisher(&channel_id) {
+            let frame = DataTrackFrame::new(Vec::from(msg)).with_user_timestamp(metadata.log_time);
+            if let Err(e) = publisher.try_push(frame) {
+                if self.drop_throttler.lock().try_acquire() {
```
gasmith: Deferrable nit, but we might want to throttle per channel.

rust/foxglove/src/remote_access/session.rs
gasmith: I wish this loop wasn't necessary. What if we didn't unpublish the data track on unsubscribe? Create it  lazily, but just keep it around and don't log to it until there's at least one subscribe present? Of course, we'd still tear it down when the channel is removed.

Alternatively, I feel like we should be able to build a tighter first-class state machine around the data track lifecycle. I'm still trying to wrap my head around the oneshot callback chaining, but it seems complex and risky.

rust/foxglove/src/remote_access/session/data_track_publisher.rs
gasmith: I guess my suggestion doesn't really help, because the client still has to _subscribe to the actual data track_, and it's gonna miss stuff until that subscription is complete. Which means latched topics are not going to work well with this scheme.

gasmith: Actually, maybe it's not so bad. We just need to defer the `Listener::on_subscribe` callback to happen when we get the `RoomEvent::TrackSubscribed` event to notify us that the subscriber is ready.

rust/foxglove/src/remote_access/session.rs
gasmith: The more I think about it, the more I feel like we should just keep the data track published. I doubt it's much overhead for the SFU, it reduces complexity in the SDK considerably, and it also reduces the latency for setting up a client subscription for a data track that has been published previously.

rust/foxglove/src/remote_access/session/data_track_publisher.rs
gasmith: Ugh, `TrackSubscribed` is for remote tracks. `LocalTrackSubscribed` doesn't tell you which participant subscribed.

OK, what about this: We eagerly publish data tracks for all channels. When a client wants to subscribe, they first subscribe to the data track, wait for a `TrackSubscribed`, and then send a protocol Subscribe message. When a client wants to unsubscribe, they unsubscribe from the track and send a protocol Unsubscribe message (in any order, we don't care).

That gives us what we're looking for in the SDK. When the SDK receives the protocol Subscribe message, we know the client is ready to receive data, which means we can issue the `on_subscribe` callback, the bridge can immediately send a latched message, the SDK will certainly write it to the data track, and (if the link is good) the client is likely to receive it.

eloff: RoomEvent.TrackSubscribed is for audio/video tracks, I don't think it fires for data tracks (would have to verify that.)

There's an internal method in the client library openSubscriptionStream which returns A tuple of the ReadableStream and a Promise that resolves once the SFU subscription is fully established / the stream is ready to receive frames.

However,  the public RemoteDataTrack.subscribe() API discards the promise and only returns the stream.

But suppose we use the internal method for now and can convince livekit to make it public or expose similar functionality via the public API.

Then:

1. Eagerly publish data tracks for all channels (SDK side)
2. When client wants to subscribe, call openSubscriptionStream (or modify subscribe() to also return the promise)
3. Await sfuSubscriptionComplete — the SFU has now acknowledged and will forward frames
4. Then send the protocol Subscribe message
5. SDK fires on_subscribe, bridge sends latched message, client receives it

rust/foxglove/src/remote_access/session.rs
eloff: I think that's better, but we still might run into trouble when the device disconnects uncleanly and then reconnects and tries to republish the data tracks.

I'll do some testing to see if it may be safe to drop this or not.

rust/foxglove/src/remote_access/session_state.rs
```diff
@@ -52,16 +55,22 @@ pub(crate) struct UnsubscribeResult {
...
     subscriptions: HashMap<ChannelId, SmallVec<[ParticipantIdentity; 1]>>,
-    /// Data subscriber identities and version counters per channel.
+    /// Data subscriber identities per channel.
     data_subscriptions: HashMap<ChannelId, ChannelSubscription>,
+    /// Published LiveKit data tracks, keyed by channel ID.
+    /// Lifecycle follows channel advertise/unadvertise, not subscribe/unsubscribe.
+    data_tracks: HashMap<ChannelId, LocalDataTrack>,
+    /// Completion receivers for in-flight data track lifecycle tasks (publish or unpublish).
+    /// Used to serialize publish/unpublish operations per channel and avoid DuplicateName errors.
+    data_track_lifecycle: HashMap<ChannelId, oneshot::Receiver<()>>,
```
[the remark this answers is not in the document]
eloff: That's fine, it's eventually cleaned up with the session.

rust/foxglove/src/remote_access/session/data_track_publisher.rs
dave-foxglove: On the OP: the goal of doing a mimic of a latched message is ok with me. Please consider commenting about why the data is stored and sent (on a lossy connection). like "Even though this is a lossy connection we blah because foo"

rust/foxglove/src/remote_access/session_state.rs
```diff
@@ -50,18 +52,24 @@ pub(crate) struct UnsubscribeResult {
...
     /// All subscriber identities per channel, regardless of subscription type.
     subscriptions: HashMap<ChannelId, SmallVec<[ParticipantIdentity; 1]>>,
-    /// Data subscriber identities and version counters per channel.
-    data_subscriptions: HashMap<ChannelId, ChannelSubscription>,
+    /// Published LiveKit data tracks, keyed by channel ID.
+    /// Lifecycle follows channel advertise/unadvertise, not subscribe/unsubscribe.
+    data_tracks: HashMap<ChannelId, LocalDataTrack>,
+    /// Completion receivers for in-flight data track lifecycle tasks (publish or unpublish).
+    /// Used to serialize publish/unpublish operations per channel and avoid DuplicateName errors.
+    data_track_lifecycle: HashMap<ChannelId, oneshot::Receiver<()>>,
```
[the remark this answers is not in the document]
eloff: I've explained multiple times why this is not an issue.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -2257,110 +2226,139 @@ impl RemoteAccessSession {
...
+                if let Some(prev) = prev_rx {
+                    _ = prev.await;
+                }
+                let track = {
+                    const INITIAL_BACKOFF: Duration = Duration::from_millis(100);
+                    const MAX_BACKOFF: Duration = Duration::from_secs(3);
+                    const TIMEOUT: Duration = Duration::from_secs(60);
+                    let deadline = tokio::time::Instant::now() + TIMEOUT;
+                    let mut backoff = INITIAL_BACKOFF;
+                    loop {
```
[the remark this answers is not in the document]
eloff: I'm fine with this. It's a problem if we have to give up here, so trying as long as we can is worth it.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -269,22 +310,27 @@ impl Sink for RemoteAccessSession {
...
             state.add_metadata_to_advertisement(&mut advertise_msg);
-        }
+            topics
+        };
 
         self.broadcast_control(encode_json_message(&advertise_msg));
 
-        // Clients subscribe asynchronously.
+        // Eagerly publish a data track for each newly advertised channel.
+        self.publish_data_tracks(&advertised_topics);
```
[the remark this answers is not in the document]
eloff: I assume at some point we'll reach a limit, but for now we will ignore this.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -59,27 +59,15 @@ pub(crate) struct SessionStats {
...
-    fn binary<'a>(channel_id: ChannelId, message: &impl BinaryMessage<'a>) -> Self {
-        Self {
-            channel_id,
-            data: encode_binary_message(message),
-        }
-    }
-}
+// Note: we may want to track a per-channel sequence in the future, depending how we use this.
+/// Global monotonic sequence number for data track frames, used for packet loss detection.
+static DATA_TRACK_SEQUENCE: AtomicU32 = AtomicU32::new(0);
```
gasmith: A global sequence is useless, unless every viewer subscribes to exactly the same set of data tracks at exactly the same time. This needs to be per-data-track.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -2257,110 +2226,139 @@ impl RemoteAccessSession {
...
-        }
-        None => {
-            channel_writers.remove(channel_id);
-            None
+            let local_participant = self.room.local_participant();
+            let cancel = self.cancellation_token.clone();
+            let state = Arc::clone(&self.state);
+            let control_plane_tx = self.control_plane_tx.clone();
+            let listener = self.listener.clone();
+            tokio::spawn(async move {
```
gasmith: Careful. This is now being called from `Sink::add_channel` which is not necessarily running in a tokio runtime. We need to plumb the configured runtime handle from the connection, and use it here to spawn the task.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -2257,110 +2226,139 @@ impl RemoteAccessSession {
...
+                                    );
+                                    _ = done_tx.send(());
+                                    return;
+                                }
+                                backoff = (backoff * 2).min(MAX_BACKOFF);
+                                if tokio::time::Instant::now() >= deadline {
+                                    error!(
+                                        "giving up publishing data track {topic} \
+                                         after {TIMEOUT:?} of DuplicateName retries"
+                                    );
```
gasmith: I'm not sure we ever want to time out permanently. There's no harm in retrying forever, and it's more visible to users (why can't I visualize this topic? oh, the logs are full of these errors...). If there's some bug in livekit that eventually resolves, users shouldn't have to bounce their foxglove_bridge service to recover.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -2257,110 +2226,139 @@ impl RemoteAccessSession {
...
+    /// Tear down the data track for a channel. Chains an async task that awaits any in-flight
+    /// publish, then removes and unpublishes the track.
+    fn teardown_data_track(&self, channel_id: ChannelId) {
+        let (prev_rx, done_tx) = {
+            let mut state = self.state.write();
+            Self::chain_data_track_task(&mut state, channel_id)
+        };
+
+        let state = Arc::clone(&self.state);
+        tokio::spawn(async move {
```
gasmith: Same story as above - this is called from `Sink::remove_channel`, which may not be a tokio runtime.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -2257,110 +2226,139 @@ impl RemoteAccessSession {
...
-            None
+            let local_participant = self.room.local_participant();
+            let cancel = self.cancellation_token.clone();
+            let state = Arc::clone(&self.state);
+            let control_plane_tx = self.control_plane_tx.clone();
+            let listener = self.listener.clone();
+            tokio::spawn(async move {
+                if let Some(prev) = prev_rx {
+                    _ = prev.await;
+                }
```
gasmith: This _shouldn't_ be necessary. Note that `Context::remove_channel` is only `pub(crate)` - the only way to remove a channel from the context is via `RawChannel::close`, and so there's no way to re-use a channel ID within the lifetime of a `Context`. I don't remember if we have other things depending on this, but I would be amenable to making this a clearly-stated (and enforced?) rule.

rust/foxglove/src/remote_access/session.rs
gasmith: Ah, I see, it's because there's no cancellation. We should fix that I think, and we can clean up the task-chaining while we're at it. Imagine we had an abstraction that we could stash in SessionState:

```rust
pub struct DataTrack {
    /// Set by the publish task on success; read lock-free from the hot path.
    track: Arc<OnceLock<LocalDataTrack>>,
    /// Cancels the pending publish task, if it is running.
    cancel: CancellationToken,
    /// The spawned publish task.
    task: JoinHandle<()>,
}

impl DataTrack {
    /// Returns the track if it has been published.
    fn get(&self) -> Option<&LocalDataTrack> {
        self.track.get()
    }

    /// Cancel any in-flight publish, wait for it to settle, then unpublish if needed.
    async fn close(self) {
        self.cancel.cancel();
        _ = self.task.await;
        if let Some(track) = self.track.get() {
            track.unpublish();
        }
    }
}
```

gasmith (commented):
Still reading, but I had a few things worth sharing early.

rust/foxglove/src/remote_access/session.rs
gasmith: Oo, and the cancellation token can be a child of the session's cancellation token, so that it is automatically cancelled when we tear down the session.

rust/foxglove/src/remote_access/session.rs
eloff: Hmm, good point. I didn't realize it's private.

One could reuse the topic name I guess, so I could remove this if I went back to naming the tracks by channel id.

rust/foxglove/src/remote_access/session.rs
eloff: I'm onboard with indefinate retries, I was debating this while implementing it

rust/foxglove/src/remote_access/session.rs
eloff: Implemented

rust/foxglove/src/remote_access/session/data_track.rs
```diff
@@ -0,0 +1,127 @@
...
+                    Err(e) => {
+                        error!(
+                            "failed to publish data track {name}: {e:?}, \
+                             retrying in {backoff:?}"
+                        );
+                    }
+                }
+                tokio::select! {
+                    () = cancel_clone.cancelled() => return,
+                    () = tokio::time::sleep(backoff) => {}
```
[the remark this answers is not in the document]
gasmith: I was thinking about this too. The fact is, there are two reasons for cancellation.

When we're closing the track, we MUST NOT cancel an in-flight publish_data_track() call, because that could end up leaking a data track on the SFU.

When we're shutting down the session, however, we COULD be more aggressive, because the SFU will automatically clear out the data track when our local participant leaves the room. This is a nice-to-have, and something that we can defer as future work.

rust/foxglove/src/remote_access/session/data_track.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /// Returns the underlying LiveKit track if publishing has completed successfully.
+    pub fn get(&self) -> Option<&LocalDataTrack> {
+        self.track.get()
+    }
+
+    /// Build a sequenced frame and push it to the data track.
+    /// Drops with a throttled debug log if the track is not ready or full.
+    pub fn log(&self, channel_id: ChannelId, msg: &[u8], metadata: &Metadata) {
```
gasmith, line 107: This is a nice encapsulation!

rust/foxglove/src/remote_access/session/data_track.rs
```diff
@@ -0,0 +1,127 @@
...
+            drop_throttler: parking_lot::Mutex::new(crate::throttler::Throttler::new(
+                Duration::from_secs(30),
+            )),
+        }
+    }
+
+    /// Returns the underlying LiveKit track if publishing has completed successfully.
+    pub fn get(&self) -> Option<&LocalDataTrack> {
+        self.track.get()
+    }
```
gasmith: Nit: Looking at the only callsite, I think we should get rid of this entirely,and fall through to DataTrack::log, where we have the throttled debug log about the track not being ready.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -174,14 +175,13 @@ pub(crate) struct RemoteAccessSession {
     room: Room,
     context: Weak<Context>,
     remote_access_session_id: Option<String>,
-    state: RwLock<SessionState>,
+    state: Arc<RwLock<SessionState>>,
```
gasmith: Does this still need the Arc?

rust/foxglove/src/remote_access/session.rs
```diff
@@ -235,11 +235,9 @@ impl Sink for RemoteAccessSession {
             publisher.send(Bytes::copy_from_slice(msg), metadata.log_time);
         }
 
-        // Send to data subscribers.
-        if state.has_data_subscribers(&channel_id) {
-            drop(state);
-            let message = ServerMessageData::new(u64::from(channel_id), metadata.log_time, msg);
-            self.send_data_lossy(ChannelMessage::binary(channel_id, &message));
+        // Send to data subscribers via the eagerly-published data track.
```
gasmith: Nit: I think we can drop the "eagerly-published" bit, that's not relevant here.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -269,22 +267,27 @@ impl Sink for RemoteAccessSession {
         // Track advertised channels and detect video-capable ones.
         let advertised_ids: std::collections::HashSet<u64> =
             advertise_msg.channels.iter().map(|ch| ch.id).collect();
-        {
+        let advertised_channel_ids: SmallVec<[ChannelId; 4]> = {
```
gasmith, line 270: I think `advertised_channel_ids` will always be the same set as `advertised_ids`. Unless we care about ordering, I think we can just use `advertised_ids`.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -269,22 +267,27 @@ impl Sink for RemoteAccessSession {
...
                 }
             }
             state.add_metadata_to_advertisement(&mut advertise_msg);
-        }
+            ids
+        };
 
         self.broadcast_control(encode_json_message(&advertise_msg));
 
-        // Clients subscribe asynchronously.
```
gasmith, line 287: Let's keep this comment about the return value.

rust/foxglove/src/remote_access/session_state.rs
```diff
@@ -583,15 +554,37 @@ impl SessionState {
...
+        if !self.has_data_subscribers(channel_id) {
+            return None;
+        }
+        let track = self.data_tracks.get(channel_id)?;
+        track.get()?;
+        Some(track)
+    }
+
+    pub fn insert_data_track(&mut self, channel_id: ChannelId, track: DataTrack) {
+        self.data_tracks.insert(channel_id, track);
```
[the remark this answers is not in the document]
gasmith: Yes, let's do that.

gasmith (commented):
All my comments are fairly minor, I think this is just about ready to go.

rust/foxglove/src/remote_access/session_state.rs
```diff
@@ -583,15 +554,42 @@ impl SessionState {
...
+        if !self.has_data_subscribers(channel_id) {
+            return None;
+        }
+        let track = self.data_tracks.get(channel_id)?;
+        track.get()?;
+        Some(track)
+    }
+
+    pub fn insert_data_track(&mut self, channel_id: ChannelId, track: DataTrack) {
+        let old = self.data_tracks.insert(channel_id, track);
```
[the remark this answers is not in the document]
eloff: This shouldn't happen. We could upgrade it to an assert, but I don't know that I like that any better

rust/remote_access_tests/tests/netem_test.rs
```diff
@@ -264,10 +260,19 @@ async fn netem_message_delivery_under_impairment() -> Result<()> {
     viewer.subscribe_and_wait(&[channel_id], &channel).await?;
 
     let payload = b"netem-hello";
-    channel.log(payload);
+    let sender_channel = channel.clone();
+    let sender = tokio::spawn(async move {
+        let mut interval = tokio::time::interval(Duration::from_millis(1));
+        loop {
+            interval.tick().await;
+            sender_channel.log(payload);
```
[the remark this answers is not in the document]
eloff: I want the test to complete as fast as possible, noisy logs are fine.

rust/foxglove/src/remote_access/session_state.rs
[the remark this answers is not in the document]
gasmith: Agreed, debug assert is fine with me.

rust/foxglove/src/remote_access/session.rs
eloff: advertised_ids is unfiltered channel ids, so I think is correct as-is

gasmith: Disagree - it is filtered.

channels -> filtered -> advertise_msg -> advertised_ids

rust/foxglove/src/remote_access/session.rs
eloff: Good catch, I think it doesn't need to be and Arc now

rust/foxglove/src/remote_access/session/data_track.rs
eloff: I was pretty proud of that little win, because I spotted it and not Claude.

gasmith: Take _that_, AI overlords!

rust/foxglove/src/remote_access/session/data_track.rs
```diff
@@ -115,13 +128,23 @@ impl DataTrack {
...
 
-    /// Close the data track: cancel any in-flight publish, wait for it to settle,
-    /// then unpublish the track if it was successfully published.
-    pub async fn close(self) {
-        self.cancel.cancel();
-        _ = self.task.await;
+    /// Close the data track: stop retrying, wait for any in-flight publish to
+    /// complete, then unpublish the track if it was successfully published.
+    pub async fn close(&mut self) {
+        debug!("closing data track");
```
gasmith: Could probably drop this, so we don't end up with two log messages per explicitly-closed channel.

rust/foxglove/src/remote_access/session_state.rs
```diff
@@ -575,12 +575,16 @@ impl SessionState {
             return None;
         }
         let track = self.data_tracks.get(channel_id)?;
-        track.get()?;
         Some(track)
```
gasmith: Ultra-nit: could just return self.data_tracks.get(channel_id) directly.

rust/foxglove/src/remote_access/session/data_track.rs
```diff
@@ -0,0 +1,140 @@
...
+        payload.extend_from_slice(&(FRAME_HEADER_SIZE as u16).to_le_bytes()); // data_offset
+        payload.extend_from_slice(&seq.to_le_bytes());
+        payload.extend_from_slice(msg);
+        let frame = DataTrackFrame::new(payload).with_user_timestamp(metadata.log_time);
+        if let Err(e) = track.try_push(frame) {
+            if self.drop_throttler.lock().try_acquire() {
+                debug!("data track message dropped for channel {channel_id:?}: {e:?}");
+            }
+        }
+    }
```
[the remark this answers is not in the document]
gasmith: I'm not overly concerned about this. We can optimize later.

gasmith (approved):
Last nits

PR #1155, opened 2026-04-13
remote_access: Receive client data messages over control bytestream

gasmith:
### Changelog
None

### Docs
None

### Description
It turns out that LiveKit multiplexes all ByteStreams on top of the same
WebRTC data channel, which means that we're not actually benefiting from
giving each client-advertised channel its own ByteStream. On the flip
side, there's considerable risk and complexity in managing per-channel
ByteStreams. We've decided to remove support for them, and switch back
to just sending client data messages over the control channel.

Since this is not public yet, we are not concerned about backward
compatibility at the protocol level.

### Testing
- [x] Tested against foxglove/app#13835

### Links
Fixes: FLE-426

### DIFF
--- a/rust/foxglove/src/remote_access/session.rs
+++ b/rust/foxglove/src/remote_access/session.rs
@@ -60,10 +60,8 @@ pub(crate) struct SessionStats {
 }
 
 const CONTROL_CHANNEL_TOPIC: &str = "control";
-const CLIENT_CHANNEL_TOPIC_PREFIX: &str = "client-ch-";
 const MESSAGE_FRAME_SIZE: usize = 5; // 1 byte opcode + u32 LE length
 const MAX_MESSAGE_SIZE: usize = 16 * 1024 * 1024; // 16 MiB
-pub(crate) const DEFAULT_PENDING_CLIENT_READER_TIMEOUT: Duration = Duration::from_secs(15);
 
 /// A control plane message queued for delivery to a specific participant.
 pub(super) struct ControlPlaneMessage {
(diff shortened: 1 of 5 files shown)


rust/foxglove/src/remote_access/session.rs
```diff
@@ -467,17 +456,11 @@ impl RemoteAccessSession {
         }
     }
 
-    /// Read framed messages from a client byte stream.
-    ///
-    /// `expected_channel_id` identifies a `client-ch-{channelId}` stream: the channel ID parsed from
-    /// the topic name. Every `MessageData` frame on this stream must carry the same channel ID;
-    /// mismatches are considered a protocol violation (debug-asserted). Pass `None` for the
-    /// `"control"` control stream.
+    /// Read framed messages from a client byte stream on the control channel.
```
clalancette, line 459: It might be worth a comment here saying that we are explicitly using a single data channel because the underlying LiveKit ByteStreams are all on one channel anyway.

rust/remote_access_tests/src/test_helpers.rs
```diff
@@ -509,32 +509,12 @@ impl ViewerConnection {
         self.send_framed_message(&framed).await
     }
 
-    /// Sends a binary-framed `ClientMessageData` on a per-channel topic `"client-ch-{channelId}"`.
-    ///
-    /// This tests the new per-channel delivery path for client publish message data.
+    /// Sends a binary-framed `ClientMessageData` on the control channel.
```
clalancette, line 512: Same here.

clalancette (approved):
Two very minor comments, but approving anyway.

eloff (approved):
LGTM

PR #1160, opened 2026-04-14
remote_access: Per-participant control plane flush tasks

dave-foxglove:
### Changelog

None

### Docs

None

### Description

**Problem:** The control plane in `run_sender` processed messages serially — each
`participant.send().await` blocked until the receiver acked. This created
head-of-line blocking where a slow participant stalled control messages for all
other participants.

**Approach:** Replace the shared control plane queue (`control_plane_tx/rx`) with
a per-participant `flume::bounded` channel (sized by `message_backlog_size`,
default 1024), each drained by a dedicated tokio flush task. Each participant now
has its own mailbox and worker. When a participant's queue is full, it is
disconnected (reset requested) rather than exerting backpressure — the entire
send path is non-blocking and sync.

Key changes:
- `Participant::spawn()` encapsulates channel creation, cancellation token,
  writer ownership, and flush task spawning. The flush task uses biased select
  and cancel-aware write (writer.write wrapped in a cancel select).
- `try_queue_control()` (#[must_use], non-blocking try_send) and
  `send_control()` (try_send + cancel token + reset on full).
- Per-participant `CancellationToken` (child of session token) — cancelled on
  queue overflow so the flush task stops immediately.
- Participant reset requests use a shared `HashSet` + `Notify` instead of an
  unbounded mpsc channel, deduplicating multiple resets for the same participant.
- `run_sender` renamed to `run_video_metadata_watcher` — control plane arm
  removed entirely.
- `ResponseSender` (service.rs) changed to `Weak<Participant>` so in-flight
  service calls don't extend participant lifetime past removal.
- `RemoteAccessSession::close()` encapsulates teardown: cancels all participants
  via `take_participants()`, awaits flush handles, closes room.
- `remove_participant` cancels the participant's token and detaches the flush
  handle.

Fixes: [FLE-427](https://linear.app/foxglove/issue/FLE-427)

### DIFF
--- a/rust/foxglove/src/remote_access/connection.rs
+++ b/rust/foxglove/src/remote_access/connection.rs
@@ -32,7 +32,7 @@ use crate::{
 type Result<T> = std::result::Result<T, Box<RemoteAccessError>>;
 
 const AUTH_RETRY_PERIOD: Duration = Duration::from_secs(30);
-use super::session::DEFAULT_CONTROL_QUEUE_SIZE;
+use super::session::DEFAULT_MESSAGE_BACKLOG_SIZE;
 
 /// The status of the remote access gateway connection.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -294,7 +294,7 @@ impl RemoteAccessConnection {
                         cancellation_token: self.cancellation_token.clone(),
                         message_backlog_size: self
                             .message_backlog_size
-                            .unwrap_or(DEFAULT_CONTROL_QUEUE_SIZE),
+                            .unwrap_or(DEFAULT_MESSAGE_BACKLOG_SIZE),
                         services: self.services.clone(),
                         connection_graph: self.connection_graph.clone(),
                         remote_access_session_id: self
--- a/rust/foxglove/src/remote_access/connection.rs
+++ b/rust/foxglove/src/remote_access/connection.rs
@@ -32,7 +32,7 @@ use crate::{
 type Result<T> = std::result::Result<T, Box<RemoteAccessError>>;
 
 const AUTH_RETRY_PERIOD: Duration = Duration::from_secs(30);
-const DEFAULT_MESSAGE_BACKLOG_SIZE: usize = 1024;
+use super::session::DEFAULT_CONTROL_QUEUE_SIZE;
 
 /// The status of the remote access gateway connection.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -294,7 +294,7 @@ impl RemoteAccessConnection {
                         cancellation_token: self.cancellation_token.clone(),
                         message_backlog_size: self
                             .message_backlog_size
-                            .unwrap_or(DEFAULT_MESSAGE_BACKLOG_SIZE),
+                            .unwrap_or(DEFAULT_CONTROL_QUEUE_SIZE),
                         services: self.services.clone(),
                         connection_graph: self.connection_graph.clone(),
                         remote_access_session_id: self
@@ -372,7 +372,9 @@ impl RemoteAccessConnection {
         context.add_sink(session.clone());
 
         // We can use spawn here because we're already running on self.runtime
-        let sender_task = tokio::spawn(RemoteAccessSession::run_sender(session.clone()));
+        let video_metadata_task = tokio::spawn(RemoteAccessSession::run_video_metadata_watcher(
+            session.clone(),
+        ));
 
         // Send ServerInfo and channel advertisements to participants already in the room.
         // ParticipantConnected events only fire for participants joining after us.
@@ -430,16 +432,14 @@ impl RemoteAccessConnection {
             }
         }
         context.remove_sink(session.sink_id());
-        sender_task.abort();
-        // Wait for the sender task to fully stop so no callbacks are in flight.
-        let _ = sender_task.await;
+        video_metadata_task.abort();
+        let _ = video_metadata_task.await;
 
         info!(remote_access_session_id, "disconnecting from room");
-        // Close the room (disconnect) on shutdown.
-        // If we don't do that, there's a 15s delay before this device is removed from the participants
-        if let Err(e) = session.room().close().await {
-            error!(remote_access_session_id, error = %e, "failed to close room: {e}");
-        }
+        // handle_room_events was one arm of the select! above — when the select
+        // exits, that future is dropped, so no more remove_participant calls can
+        // race with the flush handle drain inside close().
+        session.close().await;
     }
 
     /// Connect to the room, retrying indefinitely.
--- a/rust/foxglove/src/remote_access/gateway.rs
+++ b/rust/foxglove/src/remote_access/gateway.rs
@@ -304,12 +304,13 @@ impl Gateway {
         self
     }
 
-    /// Set the message backlog size.
+    /// Set the per-participant control plane message queue size.
     ///
-    /// The sink buffers outgoing log entries into a queue. If the backlog size is exceeded, the
-    /// oldest entries will be dropped.
+    /// Each participant gets an independent queue of this size. If a participant's
+    /// queue fills up (because it is not reading fast enough), it will be disconnected
+    /// and asked to reconnect.
     ///
-    /// By default, the sink will buffer 1024 messages.
+    /// By default, each participant gets a queue of 256 messages.
     pub fn message_backlog_size(mut self, size: usize) -> Self {
         self.message_backlog_size = Some(size);
         self
--- a/rust/foxglove/src/remote_access/gateway.rs
+++ b/rust/foxglove/src/remote_access/gateway.rs
@@ -310,7 +310,7 @@ impl Gateway {
     /// queue fills up (because it is not reading fast enough), it will be disconnected
     /// and asked to reconnect.
     ///
-    /// By default, each participant gets a queue of 256 messages.
+    /// By default, each participant gets a queue of 1024 messages.
     pub fn message_backlog_size(mut self, size: usize) -> Self {
         self.message_backlog_size = Some(size);
         self
--- a/rust/foxglove/src/remote_access/participant.rs
+++ b/rust/foxglove/src/remote_access/participant.rs
@@ -84,6 +84,7 @@ impl Participant {
     ///
     /// When this returns `false`, the caller should trigger a participant reset
     /// (disconnect + reconnect) — a full queue means the client is not keeping up.
+    #[must_use]
     pub(crate) fn try_queue_control(&self, data: Bytes) -> bool {
         match self.control_tx.try_send(data) {
             Ok(()) => true,
@@ -113,14 +114,14 @@ impl Participant {
     /// protocol messages via `send_control` flow regularly and will trigger the
     /// reset if the queue stays full.
     pub(crate) fn send_asset_response(&self, data: &[u8], request_id: u32) {
-        self.try_queue_control(encode_binary_message(&FetchAssetResponse::asset_data(
+        let _ = self.try_queue_control(encode_binary_message(&FetchAssetResponse::asset_data(
             request_id, data,
         )));
     }
 
     /// Send a fetch asset error to the participant via the control plane queue.
     pub(crate) fn send_asset_error(&self, error: &str, request_id: u32) {
-        self.try_queue_control(encode_binary_message(&FetchAssetResponse::error_message(
+        let _ = self.try_queue_control(encode_binary_message(&FetchAssetResponse::error_message(
             request_id, error,
         )));
     }
--- a/rust/foxglove/src/remote_access/participant.rs
+++ b/rust/foxglove/src/remote_access/participant.rs
@@ -49,7 +49,65 @@ pub(crate) struct Participant {
 }
 
 impl Participant {
-    /// Creates a new participant.
+    /// Creates a new participant with its own control plane channel and flush task.
+    ///
+    /// The flush task drains the bounded channel into the `writer`. It exits when
+    /// the per-participant cancellation token fires (queue overflow or session
+    /// shutdown) or when all `control_tx` senders are dropped.
+    ///
+    /// Returns the participant (wrapped in `Arc` for shared ownership) and the
+    /// flush task's `JoinHandle` (for teardown awaiting).
+    pub fn spawn(
+        identity: ParticipantIdentity,
+        protocol_version: Version,
+        writer: ParticipantWriter,
+        queue_size: usize,
+        reset_tx: tokio::sync::mpsc::UnboundedSender<ParticipantIdentity>,
+        session_cancel: &CancellationToken,
+    ) -> (std::sync::Arc<Self>, tokio::task::JoinHandle<()>) {
+        let (control_tx, control_rx) = flume::bounded::<Bytes>(queue_size);
+        let cancel = session_cancel.child_token();
+        let cancel_for_task = cancel.clone();
+        let identity_for_task = identity.clone();
+        let reset_tx_for_task = reset_tx.clone();
+
+        let flush_handle = tokio::spawn(async move {
+            loop {
+                let data = tokio::select! {
+                    () = cancel_for_task.cancelled() => break,
+                    msg = control_rx.recv_async() => match msg {
+                        Ok(data) => data,
+                        Err(_) => break,
+                    },
+                };
+                if let Err(e) = writer.write(&data).await {
+                    tracing::warn!(
+                        "control write failed for {:?}, requesting reset: {e:?}",
+                        identity_for_task,
+                    );
+                    let _ = reset_tx_for_task.send(identity_for_task.clone());
+                    break;
+                }
+            }
+        });
+
+        let participant = std::sync::Arc::new(Self {
+            client_id: ClientId::next(),
+            participant_id: identity,
+            protocol_version,
+            control_tx,
+            reset_tx,
+            cancel,
+            service_call_sem: Semaphore::new(DEFAULT_SERVICE_CALLS_PER_PARTICIPANT),
+            fetch_asset_sem: Semaphore::new(DEFAULT_FETCH_ASSET_PER_PARTICIPANT),
+        });
+        (participant, flush_handle)
+    }
+
+    /// Creates a new participant without spawning a flush task.
+    ///
+    /// For use in tests that only need a participant with a pre-created channel.
+    #[cfg(test)]
     pub fn new(
         identity: ParticipantIdentity,
         protocol_version: Version,
@@ -68,7 +126,6 @@ impl Participant {
             fetch_asset_sem: Semaphore::new(DEFAULT_FETCH_ASSET_PER_PARTICIPANT),
         }
     }
-
 
     /// Returns the locally-significant client ID.
     pub fn client_id(&self) -> ClientId {
--- a/rust/foxglove/src/remote_access/participant.rs
+++ b/rust/foxglove/src/remote_access/participant.rs
@@ -78,13 +78,21 @@ impl Participant {
         let flush_handle = tokio::spawn(async move {
             loop {
                 let data = tokio::select! {
+                    biased;
                     () = cancel_for_task.cancelled() => break,
                     msg = control_rx.recv_async() => match msg {
                         Ok(data) => data,
                         Err(_) => break,
                     },
                 };
-                if let Err(e) = writer.write(&data).await {
+                // Wrap the write in a cancel-aware select with a generous timeout
+                // as a safeguard against writer.write() blocking indefinitely.
+                let write_result = tokio::select! {
+                    biased;
+                    () = cancel_for_task.cancelled() => break,
+                    result = writer.write(&data) => result,
+                };
+                if let Err(e) = write_result {
                     tracing::warn!(
                         "control write failed for {:?}, requesting reset: {e:?}",
                         identity_for_task,
@@ -163,9 +171,8 @@ impl Participant {
         &self.participant_id
     }
 
-    /// Try to queue a control plane message. Returns `true` if enqueued or if
-    /// the queue is already disconnected (no reset needed). Returns `false` if
-    /// the queue is full (caller should trigger a participant reset).
+    /// Try to queue a control plane message. Returns `false` if the queue is
+    /// full and the caller should trigger a participant reset.
     #[must_use]
     pub(crate) fn try_queue_control(&self, data: Bytes) -> bool {
         match self.control_tx.try_send(data) {
--- a/rust/foxglove/src/remote_access/participant.rs
+++ b/rust/foxglove/src/remote_access/participant.rs
@@ -35,6 +35,9 @@ pub(crate) struct Participant {
     /// Per-participant control plane queue. The receiving end is owned by the
     /// flush task spawned in `add_participant`.
     control_tx: flume::Sender<Bytes>,
+    /// Channel to request a participant reset (disconnect + reconnect) when the
+    /// control queue is full.
+    reset_tx: tokio::sync::mpsc::UnboundedSender<ParticipantIdentity>,
     /// Limits concurrent service calls from this participant.
     service_call_sem: Semaphore,
     /// Limits concurrent fetch asset requests from this participant.
@@ -47,12 +50,14 @@ impl Participant {
         identity: ParticipantIdentity,
         protocol_version: Version,
         control_tx: flume::Sender<Bytes>,
+        reset_tx: tokio::sync::mpsc::UnboundedSender<ParticipantIdentity>,
     ) -> Self {
         Self {
             client_id: ClientId::next(),
             participant_id: identity,
             protocol_version,
             control_tx,
+            reset_tx,
             service_call_sem: Semaphore::new(DEFAULT_SERVICE_CALLS_PER_PARTICIPANT),
             fetch_asset_sem: Semaphore::new(DEFAULT_FETCH_ASSET_PER_PARTICIPANT),
         }
@@ -81,18 +86,12 @@ impl Participant {
     /// Try to queue a control plane message. Returns `true` if enqueued or if
     /// the queue is already disconnected (no reset needed). Returns `false` if
     /// the queue is full (caller should trigger a participant reset).
-    ///
-    /// When this returns `false`, the caller should trigger a participant reset
-    /// (disconnect + reconnect) — a full queue means the client is not keeping up.
     #[must_use]
     pub(crate) fn try_queue_control(&self, data: Bytes) -> bool {
         match self.control_tx.try_send(data) {
             Ok(()) => true,
             Err(flume::TrySendError::Full(_)) => {
-                tracing::warn!(
-                    "control queue full for {}",
-                    self.participant_id
-                );
+                tracing::warn!("control queue full for {}", self.participant_id);
                 false
             }
             Err(flume::TrySendError::Disconnected(_)) => {
@@ -107,21 +106,24 @@ impl Participant {
         }
     }
 
+    /// Queue a control plane message, requesting a participant reset if the
+    /// queue is full.
+    pub(crate) fn send_control(&self, data: Bytes) {
+        if !self.try_queue_control(data) {
+            let _ = self.reset_tx.send(self.participant_id.clone());
+        }
+    }
+
     /// Send a fetch asset response to the participant via the control plane queue.
-    ///
-    /// The `try_queue_control` return is intentionally ignored here: these methods
-    /// don't have access to `participant_reset_tx` to trigger a reset. In practice,
-    /// protocol messages via `send_control` flow regularly and will trigger the
-    /// reset if the queue stays full.
     pub(crate) fn send_asset_response(&self, data: &[u8], request_id: u32) {
-        let _ = self.try_queue_control(encode_binary_message(&FetchAssetResponse::asset_data(
+        self.send_control(encode_binary_message(&FetchAssetResponse::asset_data(
             request_id, data,
         )));
     }
 
     /// Send a fetch asset error to the participant via the control plane queue.
     pub(crate) fn send_asset_error(&self, error: &str, request_id: u32) {
-        let _ = self.try_queue_control(encode_binary_message(&FetchAssetResponse::error_message(
+        self.send_control(encode_binary_message(&FetchAssetResponse::error_message(
             request_id, error,
         )));
     }
--- a/rust/foxglove/src/remote_access/participant.rs
+++ b/rust/foxglove/src/remote_access/participant.rs
@@ -78,8 +78,9 @@ impl Participant {
         &self.participant_id
     }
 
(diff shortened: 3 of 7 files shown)


rust/foxglove/src/remote_access/gateway.rs
```diff
@@ -309,12 +308,12 @@ impl Gateway {
...
-    ///
-    /// By default, the sink will buffer 1024 messages.
-    pub fn message_backlog_size(mut self, size: usize) -> Self {
-        self.message_backlog_size = Some(size);
+    /// **Deprecated:** This parameter no longer has any effect. The data plane
+    /// now uses per-channel data tracks, and the control plane uses per-participant
+    /// bounded queues with a fixed size.
+    #[deprecated(note = "This parameter no longer has any effect.")]
+    pub fn message_backlog_size(mut self, _size: usize) -> Self {
+        self.message_backlog_size = Some(_size);
```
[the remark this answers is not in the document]
dave-foxglove: [AI] Fixed in db62ea2 — removed the `message_backlog_size` field, Default init, and Debug output entirely. The deprecated setter now just returns `self`. Added `#[allow(deprecated)]` at the C bindings call site for backward compatibility.

rust/foxglove/src/remote_access/session_state.rs
```diff
@@ -30,6 +31,10 @@ pub(crate) struct RemovedSubscriptions {
     pub client_channels: Vec<ChannelDescriptor>,
     /// Parameter names that lost their last subscriber.
     pub last_param_unsubscribed: Vec<String>,
+    /// The flush task handle for the removed participant, if one was registered.
+    /// Intentionally dropped by callers — the task exits naturally when `control_tx` drops.
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
dave-foxglove: [AI] Good catch — updated the comment to be more precise: the flush task drains when all `Arc<Participant>` refs drop (which drops `control_tx`), with the session-level `CancellationToken` as the backstop for prompt shutdown. Fixed in db62ea2.

dave-foxglove:
> The core change — moving from a shared control plane queue to per-participant `flume::bounded(64)` channels with dedicated flush tasks — looks structurally sound. The `parking_lot::MutexGuard` / `!Send` restructuring (collecting deferred messages under the lock, sending after release) is the right p
> ...

[resolved] [AI] Both items addressed in db62ea2: removed dead `message_backlog_size` field entirely, and clarified the `flush_handle` drop comment to explain the `Arc<Participant>` lifetime + `CancellationToken` backstop.

dave-foxglove:
> Previous concerns addressed — `message_backlog_size` field removed cleanly, `flush_handle` comment updated.
> 
> The per-participant flush task pattern is sound. The `parking_lot::MutexGuard` / `!Send` restructuring — collecting deferred messages under the lock, sending after release — is consistently a
> ...

[resolved] [AI] Acknowledged — this is a known tradeoff documented in the PR description. The sync `broadcast_control` path uses blocking `flume::send`, which could stall a tokio worker if a participant's 64-slot queue fills. This is a strict improvement over the old shared queue (slow participant blocked ALL sends), and async callers (handler chain, video metadata) use `queue_control_async` which yields. The sync callers (`Sink` trait, public API) block the caller's thread, which is the correct backpressure behavior.

c/src/gateway.rs
```diff
@@ -624,7 +624,9 @@ unsafe fn do_foxglove_gateway_start(
         gateway = gateway.foxglove_api_timeout(Duration::from_secs(timeout_secs));
     }
 
-    // Message backlog size
+    // Message backlog size (deprecated — no longer has any effect, but preserved
+    // for C API backward compatibility).
+    #[allow(deprecated)]
     if let Some(&backlog_size) = options.message_backlog_size {
         gateway = gateway.message_backlog_size(backlog_size);
     }
```
gasmith: We don't need to maintain backwards compat for remote-access interfaces yet. Just nuke it.

rust/foxglove/src/remote_access/connection.rs
```diff
@@ -435,9 +430,19 @@ impl RemoteAccessConnection {
             }
         }
         context.remove_sink(session.sink_id());
-        sender_task.abort();
-        // Wait for the sender task to fully stop so no callbacks are in flight.
-        let _ = sender_task.await;
+        video_metadata_task.abort();
+        let _ = video_metadata_task.await;
+
+        // SAFETY: handle_room_events was one arm of the select! above. When the
```
gasmith: Nit: "SAFETY:" comments are usually for annotating `unsafe` blocks.

At a higher level, I had to dig a bit to see how `SessionState::remove_participant` (for example) is connected to `RemoteAccessSession::take_flush_handles`. I get it now, but I feel like we might want to document that function a bit more clearly.

I think, ideally, I'd prefer to see a `async fn RemoteAccessSession::close(self)` that encapsulates these ideas. In practice, it's probably impossible to write a `self` receiver that takes ownership, because of all the `Arc<RemoteAccessSession>` instances floating around. It'd be fine to do `&self` or `self: Arc<Self>`.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -69,29 +68,16 @@ pub(crate) const DEFAULT_PENDING_CLIENT_READER_TIMEOUT: Duration = Duration::fro
...
 
-/// A control plane message queued for delivery to a specific participant.
-pub(super) struct ControlPlaneMessage {
-    participant: Arc<Participant>,
-    data: Bytes,
-}
+/// Size of the per-participant control plane queue. Large enough to buffer the
+/// initial burst from `add_participant` (server info + channel/service advertisements)
+/// without blocking, small enough to exert backpressure on a slow participant.
+const CONTROL_QUEUE_SIZE: usize = 64;
```
gasmith: Why do we want to make this a fixed constant? I feel like this should be configurable with a sane default. Perhaps there's some legitimate use case that involves a burst of more than 64 messages on the control plane queue - we can't predict that. Especially with #1159 on the horizon. I think it's fine to keep the existing `message_backlog_size` for this purpose - it is indeed the backlog for reliable messages.

rust/foxglove/src/remote_access/participant.rs
```diff
@@ -79,38 +78,44 @@ impl Participant {
         &self.participant_id
     }
 
-    /// Returns the control plane sender for this participant.
-    pub(crate) fn control_plane_tx(&self) -> &flume::Sender<ControlPlaneMessage> {
-        &self.control_plane_tx
+    /// Queue a control plane message (blocking). Use from non-tokio threads.
```
gasmith: I'm moderately uncomfortable with a blocking interface here. When a client is too slow to keep up with this queue, we need to disconnect the client, not exert backpressure on the rest of the program.

rust/foxglove/src/remote_access/service.rs
```diff
@@ -15,55 +15,62 @@ pub use crate::remote_common::service::{
 };
 
 /// Sends service call responses over the remote access control plane.
+///
+/// Holds a `Weak<Participant>` so that in-flight service calls do not extend the
+/// participant's lifetime past removal. If the participant has been removed by
+/// the time the handler responds, the response is logged and dropped.
```
gasmith, line 21: Nice.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -576,19 +541,21 @@ impl RemoteAccessSession {
...
                     return;
                 }
-            } else if !self.handle_client_control_message(
-                &participant_identity,
-                opcode,
-                Bytes::from(payload),
-            ) {
+            } else if !self
+                .handle_client_control_message(&participant_identity, opcode, Bytes::from(payload))
+                .await
```
gasmith: We definitely don't want these to be async or blocking.

gasmith (commented):
Did a first pass. Generally this looks quite good. I think my biggest comment is that when the queue is full, we need to disconnect the client, not exert backpressure. I think that should simplify the diff in session.rs considerably.

rust/foxglove/src/remote_access/session.rs
dave-foxglove: I wondered about repurposing that, but I was concerned that would be confusing (that in one version of the software it's shared among all participants and in this change it's per participant - e.g. "wow, my ram usage suddenly went up!" - but I did wrestle with that,  so I see the connection. If you up for it, I'll repurpose it.

gasmith: Actually in the websocket, it is also per-participant. ;)

Let's repurpose it.

dave-foxglove:
> Prior concerns (dead `message_backlog_size` field, imprecise flush-handle comment) are resolved.
> 
> The core refactor — per-participant `flume::bounded(64)` channels, dedicated flush tasks, `Weak<Participant>` in `ResponseSender` — reads correctly to me. The lock restructuring (collecting deferred mes
> ...

[resolved] [AI] Both inline issues are moot after the rework to try_send +
  disconnect.

dave-foxglove:
> Two prior unresolved threads still apply (doc wording on `queue_control` and the `broadcast_control` HOL question) — no new issues beyond those.
> 
> The lock restructuring looks correct: `parking_lot::MutexGuard` never crosses a yield point (`handle_client_subscribe` drops the guard before the deferred
> ...

[resolved] [AI] Both threads resolved by the rework to try_send + disconnect.

rust/foxglove/src/remote_access/participant.rs
```diff
@@ -79,38 +78,44 @@ impl Participant {
...
+    ///
+    /// Blocks the calling thread if the per-participant buffer is full,
+    /// preserving backpressure — control messages must not be dropped.
```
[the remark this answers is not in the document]
dave-foxglove: [resolved] [AI] Moot — queue_control/queue_control_async
  replaced with try_queue_control (non-blocking try_send).

rust/foxglove/src/remote_access/session.rs
```diff
@@ -450,69 +412,70 @@ impl RemoteAccessSession {
...
+    /// Enqueue a control plane message for all currently connected participants (blocking).
+    ///
+    /// For sync callers (Sink trait methods, public API). Blocks the calling thread
+    /// if a participant's per-participant buffer is full.
     fn broadcast_control(&self, data: Bytes) {
         let participants = self.state.read().collect_participants();
-        send_control_to_all(&self.control_plane_tx, &participants, data);
+        for participant in participants {
+            participant.queue_control(data.clone());
+        }
```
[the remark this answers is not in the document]
dave-foxglove: [resolved] [AI] Moot — broadcast_control now uses try_send.
  Full queue disconnects the slow participant instead of blocking.

dave-foxglove:
> Did a first pass. Generally this looks quite good. I think my biggest comment is that when the queue is full, we need to disconnect the client, not exert backpressure. I think that should simplify the diff in session.rs considerably.

[resolved] Yes, that's simpler

c/src/gateway.rs
dave-foxglove: [declined] keeping this since we are reusing it

rust/foxglove/src/remote_access/participant.rs
dave-foxglove: [resolved] when the queue is full, they will be disconnected

rust/foxglove/src/remote_access/session.rs
dave-foxglove: [resolved] Got it. Reworked it.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -173,15 +135,13 @@ pub(crate) struct RemoteAccessSession {
     room: Room,
     context: Weak<Context>,
     remote_access_session_id: Option<String>,
-    state: RwLock<SessionState>,
+    state: Arc<RwLock<SessionState>>,
```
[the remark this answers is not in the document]
gasmith: Looks like rebase damage; we can probably remove the Arc.

dave-foxglove:
> Previous concerns all resolved — the rework to `try_send` with disconnect-on-full is clean.
> 
> Main concern: `close()` looks like it hangs on the room-disconnect path. The flush tasks are awaited, but nothing drops the `control_tx` senders or cancels a token they're watching, so `recv_async` blocks fo
> ...

[resolved] Acknowledged

dave-foxglove:
> All three prior concerns resolved: `close()` hang fixed via `clear_participants()` before awaiting flush handles, doc wording on `try_queue_control` is now accurate, and the stale `Arc` wrapper on `state` is removed.
> 
> Traced the `close()` teardown path end-to-end. The ordering in `connection.rs::run
> ...

[resolved] Acknowledged

dave-foxglove:
> All prior bot threads are resolved. The rework looks solid — `try_send` with disconnect-on-full, per-participant flush tasks, `Weak<Participant>` in `ResponseSender`, and the `close()` ordering fix all land cleanly.
> 
> Two non-blocking items inline:
> - The `try_queue_control` return value is silently i
> ...

[resolved] Acknowledged

dave-foxglove:
> All prior threads resolved. The rework from blocking shared queue to per-participant `try_send` with disconnect-on-full addresses gasmith's feedback cleanly.
> 
> Traced the `close()` shutdown path: `clear_participants()` drops the state-map `Arc<Participant>` refs, unblocking flush tasks via channel cl
> ...

[resolved] Acknowledged

rust/foxglove/src/remote_access/participant.rs
```diff
@@ -79,38 +78,51 @@ impl Participant {
...
-    pub(crate) async fn send(&self, bytes: &[u8]) -> Result<()> {
-        self.writer.write(bytes).await
+    /// When this returns `false`, the caller should trigger a participant reset
+    /// (disconnect + reconnect) — a full queue means the client is not keeping up.
+    pub(crate) fn try_queue_control(&self, data: Bytes) -> bool {
+        match self.control_tx.try_send(data) {
+            Ok(()) => true,
+            Err(flume::TrySendError::Full(_)) => {
+                tracing::warn!(
+                    "control queue full for {}, disconnecting slow client",
```
gasmith: Nit: This function is not actually performing the disconnection.

rust/foxglove/src/remote_access/participant.rs
```diff
@@ -79,38 +78,51 @@ impl Participant {
...
+    /// Try to queue a control plane message. Returns `true` if enqueued or if
+    /// the queue is already disconnected (no reset needed). Returns `false` if
+    /// the queue is full (caller should trigger a participant reset).
     ///
-    /// The message is serialized and framed already and provided as a slice of bytes.
-    pub(crate) async fn send(&self, bytes: &[u8]) -> Result<()> {
-        self.writer.write(bytes).await
+    /// When this returns `false`, the caller should trigger a participant reset
+    /// (disconnect + reconnect) — a full queue means the client is not keeping up.
+    pub(crate) fn try_queue_control(&self, data: Bytes) -> bool {
```
gasmith, line 177: I think this ought to have a `#[must_use]`, if we expect callers to do something with the result.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -198,13 +158,15 @@ pub(crate) struct RemoteAccessSession {
...
-    /// stream write fails. `handle_room_events` receives identities and calls
-    /// `reset_participant`. Using `mpsc` rather than `Notify` because `Notify` is
-    /// not cancel-safe in `select!`.
+    /// Channel used by per-participant flush tasks to request participant resets
+    /// when a control stream write fails. `handle_room_events` receives identities
+    /// and calls `reset_participant`. Using `mpsc` rather than `Notify` because
+    /// `Notify` is not cancel-safe in `select!`.
     participant_reset_tx: tokio::sync::mpsc::UnboundedSender<ParticipantIdentity>,
     participant_reset_rx:
         tokio::sync::Mutex<tokio::sync::mpsc::UnboundedReceiver<ParticipantIdentity>>,
```
gasmith: Can you help me remember why these are based on `ParticipantIdentity` (which a client re-uses across reconnects) and not `ParticipantSid` or `ClientId`?

rust/foxglove/src/remote_access/participant.rs
```diff
@@ -79,38 +78,51 @@ impl Participant {
...
-        let msg = ControlPlaneMessage::binary(
-            self.clone(),
-            &FetchAssetResponse::asset_data(request_id, data),
-        );
-        if let Err(e) = self.control_plane_tx.send(msg) {
-            tracing::warn!("control plane queue disconnected, dropping asset response: {e}");
-        }
+    ///
+    /// The `try_queue_control` return is intentionally ignored here: these methods
+    /// don't have access to `participant_reset_tx` to trigger a reset. In practice,
```
gasmith: Should they have access to `participant_reset_tx`? Might be worthwhile.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -393,69 +353,90 @@ impl RemoteAccessSession {
...
-        let msg = ControlPlaneMessage { participant, data };
-        if let Err(e) = self.control_plane_tx.send(msg) {
-            warn!("control plane queue disconnected, dropping message: {e}");
+    /// Send a control plane message to a participant. If the queue is full,
+    /// the participant is disconnected (reset requested).
+    fn send_control(&self, participant: &Participant, data: Bytes) {
+        if !participant.try_queue_control(data) {
+            let _ = self
+                .participant_reset_tx
+                .send(participant.participant_id().clone());
```
gasmith: It's a little unfortunate that this grows unbounded. We used to have a mechanism to avoid that (the poisoned_participants set), but that's gone now.

What we really want is to store the participant sid (or client id) in a set and wake the task responsible for performing resets (which should drain that set, or swap it for an empty one and process its contents).

rust/foxglove/src/remote_access/session.rs
```diff
@@ -971,30 +952,56 @@ impl RemoteAccessSession {
...
+        let (control_tx, control_rx) = flume::bounded::<Bytes>(self.message_backlog_size);
+        let writer = ParticipantWriter::Livekit(stream);
+        let participant_id_for_task = participant_id.clone();
+        let reset_tx = self.participant_reset_tx.clone();
+        let cancel = self.cancellation_token.clone();
+
+        let flush_handle = tokio::spawn(async move {
+            loop {
+                let data = tokio::select! {
+                    () = cancel.cancelled() => break,
```
gasmith: I also think this should be a child cancellation token, which gets set as soon as the queue overflows. Once the queue overflows, we know the participant is going to be disconnected. We shouldn't waste time trying to send its contents to the participant.

gasmith: Moreover, I have a gut feeling that these mechanisms (queue, drain, reset) should be better encapsulated in the participant. Obviously reset needs to span the boundary, because we need to handle that from the room event loop. But there's a lot of other machinery bleeding into the Session implementation that doesn't seem to belong here.

rust/foxglove/src/remote_access/participant.rs
dave-foxglove: added reset_tx to Participant

rust/foxglove/src/remote_access/participant.rs
dave-foxglove: added

rust/foxglove/src/remote_access/participant.rs
dave-foxglove: fixed

rust/foxglove/src/remote_access/session.rs
dave-foxglove: The cancellation token is straightforward. The encapsulation will be more effort (doable, but more than a minute).

rust/foxglove/src/remote_access/session.rs
dave-foxglove: The cancellation token (added in response to another of your comments) reduces the impact here (doesn't eliminate it entirely). I'll look for more improvement, but wanted to mention that the cancellation token is something.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -393,69 +350,76 @@ impl RemoteAccessSession {
...
-    fn send_control(&self, participant: Arc<Participant>, data: Bytes) {
-        let msg = ControlPlaneMessage { participant, data };
-        if let Err(e) = self.control_plane_tx.send(msg) {
-            warn!("control plane queue disconnected, dropping message: {e}");
-        }
-    }
-
+    /// Send a control plane message to a participant. If the queue is full,
+    /// the participant is disconnected (reset requested).
     /// Send an error status message to a participant.
```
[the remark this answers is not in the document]
dave-foxglove: [resolved] [AI] Fixed — removed stale doc comment lines.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -393,69 +350,76 @@ impl RemoteAccessSession {
...
+    /// The caller must ensure that `handle_room_events` has stopped so no new
+    /// `remove_participant` / `reset_participant` calls can race with us.
+    pub(crate) async fn close(&self) {
+        let flush_handles = {
+            let mut state = self.state.write();
+            // Clear participants first — this drops `Arc<Participant>` which drops
+            // `control_tx`, causing each flush task's `recv_async` to return `Err`
+            // and exit. Without this, flush tasks hang if the `CancellationToken`
+            // hasn't been fired (e.g., on room disconnect without cancellation).
+            state.clear_participants();
```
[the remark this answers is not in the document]
dave-foxglove: [resolved] [AI] Addressed — per-participant CancellationToken (child of
  session token) means flush tasks exit via the cancel arm even if Arc<Participant> clones persist in
  handler tasks.

dave-foxglove:
> All prior threads resolved. Reviewed the full diff end-to-end against `main`.
> 
> The `close()` teardown ordering is sound: `clear_participants()` drops state-map `Arc<Participant>` refs → `control_tx` drops → flush tasks' `recv_async` returns `Err` and exits → `take_flush_handles()` awaits them → `roo
> ...

[resolved] [AI] Acknowledged — no new issues.

dave-foxglove:
> All prior bot threads are resolved. Reviewed the full diff end-to-end against `main`.
> 
> Traced the `close()` shutdown path: `clear_participants()` drops state-map `Arc<Participant>` refs → `control_tx` drops → flush tasks' `recv_async` returns `Err` and exits → `take_flush_handles()` awaits completio
> ...

[resolved] [AI] Acknowledged — no new issues.

rust/foxglove/src/remote_access/session.rs
dave-foxglove: [resolved] [AI] participant_reset_tx uses ParticipantIdentity because
  that's what LiveKit's room.remote_participants() is keyed by. reset_participant looks up the
  participant there to decide whether to re-add. ParticipantSid changes per session and ClientId is
  local — neither works for the LiveKit lookup. This predates this PR.

rust/foxglove/src/remote_access/participant.rs
```diff
@@ -30,30 +33,105 @@ pub(crate) struct Participant {
...
+        let flush_handle = tokio::spawn(async move {
+            loop {
+                let data = tokio::select! {
+                    () = cancel_for_task.cancelled() => break,
+                    msg = control_rx.recv_async() => match msg {
+                        Ok(data) => data,
+                        Err(_) => break,
+                    },
+                };
+                if let Err(e) = writer.write(&data).await {
```
gasmith: We need to check for cancellation here too:

```suggestion
                let write_result = tokio::select! {
                    () = cancel_for_task.cancelled() => break,
                    result = writer.write(&data) => result,
                };
                if let Err(e) = write_result {
```

We should probably go further than that, and add a timeout, which could be as generous as a minute or two, just as a safeguard. I'm fairly certain I discovered a deadlock in writer.write() yesterday evening while testing the bridge.

rust/foxglove/src/remote_access/participant.rs
```diff
@@ -74,43 +152,65 @@ impl Participant {
...
     }
 
-    /// Returns the control plane sender for this participant.
-    pub(crate) fn control_plane_tx(&self) -> &flume::Sender<ControlPlaneMessage> {
-        &self.control_plane_tx
+    /// Try to queue a control plane message. Returns `true` if enqueued or if
+    /// the queue is already disconnected (no reset needed). Returns `false` if
+    /// the queue is full (caller should trigger a participant reset).
+    #[must_use]
+    pub(crate) fn try_queue_control(&self, data: Bytes) -> bool {
```
gasmith, line 177: Seems this no longer needs to be pub(crate)

rust/foxglove/src/remote_access/session.rs
gasmith: Sure, it predates this PR, but we're also radically reworking this stuff, and this bug needs to be fixed. We need to have map for participants by SID or ClientId, and that's the right thing to use for requesting resets.

gasmith (commented):
Need to jump for a meeting, but I'm really liking the direction this is going.

rust/foxglove/src/remote_access/participant.rs
```diff
@@ -74,43 +152,65 @@ impl Participant {
...
     pub fn participant_id(&self) -> &ParticipantIdentity {
         &self.participant_id
     }
 
-    /// Returns the control plane sender for this participant.
-    pub(crate) fn control_plane_tx(&self) -> &flume::Sender<ControlPlaneMessage> {
-        &self.control_plane_tx
+    /// Try to queue a control plane message. Returns `true` if enqueued or if
+    /// the queue is already disconnected (no reset needed). Returns `false` if
+    /// the queue is full (caller should trigger a participant reset).
```
eloff: nit: I think this comment on the return value could be simplified to "return `false` if the queue is full and caller should trigger a participant reset"

rust/foxglove/src/remote_access/session_state.rs
```diff
@@ -95,9 +98,39 @@ impl SessionState {
...
+        self.flush_handles.remove(identity)
+    }
+
+    /// Removes all participants, dropping their `Arc<Participant>` references.
+    /// This causes per-participant `control_tx` senders to drop, which signals
+    /// flush tasks to exit.
+    ///
+    /// For use during session teardown only — does not clean up subscriptions,
+    /// video subscribers, client channels, or parameter subscriptions (the room
+    /// is closing immediately after).
```
eloff: Is this the behavior we want? specifically we don't call any of the listener methods here like on_unsubscribe or on_client_unadvertise. If users are using those to do cleanup they might be surprised they never fire in the shutdown case. I'm not sure that's wrong necessarily, but maybe worth documenting if we leave it like this.

rust/foxglove/src/remote_access/participant.rs
```diff
@@ -30,30 +33,105 @@ pub(crate) struct Participant {
...
+        let (control_tx, control_rx) = flume::bounded::<Bytes>(queue_size);
+        let cancel = session_cancel.child_token();
+        let cancel_for_task = cancel.clone();
+        let identity_for_task = identity.clone();
+        let pending_resets_for_task = pending_resets.clone();
+        let reset_notify_for_task = reset_notify.clone();
+
+        let flush_handle = tokio::spawn(async move {
+            loop {
+                let data = tokio::select! {
```
eloff, line 80: I think we should use a biased select here to prefer the cancelled arm if both are ready

eloff (commented):
This looks great. Some minor comments inline, and Greg left some good feedback that should be addressed.

Here's some feedback from the AI as well, which you can take or leave:

- In send_control: "messages already in the queue may be dropped" (i.e. the behavior from item 2 above).
- In handle_room_events drain loop: add a comment that Notify::notified() is documented as not cancel-safe, which is why the drain is at the top, not in the select arm. The commit message says this but the code does not.
- There is no test that calls Participant::send_control on a full queue and asserts that pending_resets contains the identity and reset_notify was woken. The try_queue_control_returns_false_when_full test covers the bool, but not the side-effects, which are the actual contract of send_control.

rust/foxglove/src/remote_access/session_state.rs
```diff
@@ -95,9 +98,39 @@ impl SessionState {
...
+    pub fn remove_flush_handle(
+        &mut self,
+        identity: &ParticipantIdentity,
+    ) -> Option<JoinHandle<()>> {
+        self.flush_handles.remove(identity)
+    }
+
+    /// Removes all participants, dropping their `Arc<Participant>` references.
+    /// This causes per-participant `control_tx` senders to drop, which signals
+    /// flush tasks to exit.
```
gasmith: This is a bit misleading - we're dropping one reference on the Arc, but there may be others. This is only safe to use in conjunction with Participant::cancel().

Considering that the only callsite is Session::close(), perhaps it'd be better to have a take_participants() method... imagine:

```rust
let (participants, flush_handles) = self.state.take_participants();
for participant in participants {
    participant.cancel();
}
for handle in flush_handles {
    let _ = handle.await;
}
```

Maybe the same story for insert/remove participant: a participant always travels with its flush handle during setup/teardown. Makes it easier to keep the maps in sync, and harder to misuse.

rust/foxglove/src/remote_access/session_state.rs
gasmith: That's a good point - we should make sure we are invoking those callbacks. I'm OK with deferring that, though... I think that's a pre-existing issue on the session teardown path.

rust/foxglove/src/remote_access/session.rs
dave-foxglove: ok the encapsulation is added, Participant::spawn() encapsulates channel
    creation, cancellation token, writer ownership, and flush task spawning.

rust/foxglove/src/remote_access/participant.rs
dave-foxglove: [resolved] [AI] Added biased; to both select! arms in the flush task.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -1004,7 +974,13 @@ impl RemoteAccessSession {
     pub(crate) fn remove_participant(self: &Arc<Self>, participant_id: &ParticipantIdentity) {
         let _guard = self.subscription_lock.lock();
 
-        let removed = self.state.write().remove_participant(participant_id);
+        let removed = {
+            let mut state = self.state.write();
+            let removed = state.remove_participant(participant_id);
+            // Detach the flush task — it exits when control_tx drops.
+            drop(state.remove_flush_handle(participant_id));
```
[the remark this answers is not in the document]
dave-foxglove: [resolved] [AI] Done — remove_participant now calls participant.cancel()
   before removing, consistent with close().

rust/foxglove/src/remote_access/participant.rs
dave-foxglove: [resolved] done

gasmith (approved):
I still think we should be keying pending_resets by client id or SID, but we can do that in a follow-on change. This looks good.

PR #1163, opened 2026-04-16
remote_access: Plumb QoS for C++ and Python

gasmith:
### Changelog
None

### Docs
None

### Description
This change plumbs the QoS configuration through the C++ and Python
bindings.


python/foxglove-sdk/python/foxglove/__init__.py
```diff
@@ -199,6 +201,9 @@ def start_gateway(
         :param channel_filter: A ``Callable`` that determines whether a channel should be logged
             to. Return ``True`` to log the channel, or ``False`` to skip it. By default, all
             channels will be logged.
+        :param qos_classifier: A ``Callable`` that returns the
+            :py:class:`foxglove.remote_access.QosProfile` to use for a given channel. If not set,
+            the default profile is used for all channels.
```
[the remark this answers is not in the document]
clalancette: I agree that stating the default is helpful, but I don't agree with the wording since it exposes some internal details.  I'd propose something like:

```
:param qos_classifier: A ``Callable`` that returns the
    :py:class:`foxglove.remote_access.QosProfile` to use for a given channel. If not set,
    all channels use the default Lossy profile.
```

c/src/gateway.rs
```diff
@@ -458,6 +523,20 @@ pub struct FoxgloveGatewayOptions<'a> {
         ) -> bool,
     >,
 
+    /// Context provided to the `qos_classifier` callback.
+    pub qos_classifier_context: *const c_void,
+
+    /// A QoS classifier for channels.
+    ///
+    /// Returns a [`FoxgloveQosProfile`] for the given channel, determining how data is delivered.
+    /// If not set, all channels default to lossy (data track delivery).
```
clalancette:
```suggestion
    /// Returns a [`FoxgloveQosProfile`] for the given channel, determining how data is delivered.
    /// If not set, all channels use the default Lossy profile.
```

cpp/foxglove/include/foxglove/remote_access.hpp
```diff
@@ -166,6 +185,11 @@ struct RemoteAccessGatewayOptions {
   FetchAssetHandler fetch_asset;
   /// @brief A sink channel filter callback.
   SinkChannelFilterFn sink_channel_filter;
+  /// @brief A QoS classifier callback.
+  ///
+  /// If set, this callback is invoked for each channel to determine its quality-of-service
+  /// profile. If not set, all channels default to lossy (data track delivery).
```
clalancette:
```suggestion
  /// @brief A QoS classifier callback.
  ///
  /// If set, this callback is invoked for each channel to determine its quality-of-service
  /// profile. If not set, all channels use the default Lossy profile.
```

clalancette (commented):
I've left a few minor comment fixes.  Besides that, it might be nice to have a few unit tests to exercise setting the profiles.

eloff (approved):

gasmith:
> I've left a few minor comment fixes. Besides that, it might be nice to have a few unit tests to exercise setting the profiles.

Unfortunately it's hard to test this with unit tests, because the actual QoS profile is stored on an internal structure buried within the remote access gateway. The only way to observe the effect of configuring a QoS profile is to hook up to livekit and examine the channel advertisements. We have that in rust's integration test suite, and we'll add it to the C++/python integration test suites when they land.

PR #1165, opened 2026-04-16
remote_access: Enable ros1/ros2 message support by default

gasmith:
### Changelog
None

### Docs
None

### Description
Might be part of the reason our foxglove bridge experiments have been struggling...


eloff (approved):
I wonder if that was the only issue

PR #1166, opened 2026-04-16
python: conditionally import wasm-unavailable types

jtbandes:
### Changelog

None

### Docs

None

### Description

Some recently added types are disabled in wasm builds, however they were unconditionally imported in `__init__.py`. This makes the imports (and inclusion in `__all__`) conditional so that `from foxglove import *` still works in the playground.


defunctzombie (approved):

python/foxglove-sdk/python/foxglove/__init__.py
```diff
@@ -17,81 +17,94 @@
...
+            "ParameterType",
+            "ParameterValue",
+            "Service",
+            "ServiceHandler",
+            "ServiceRequest",
+            "ServiceSchema",
+            "StatusLevel",
+        ]
+    )
+except ImportError:
```
defunctzombie, line 105: In non-wasm environments this can make it hard to debug actual errors where we would expect these imports to be present

jtbandes: Yeah perhaps. FWIW we already have this for some websocket-related imports. What alternative would you recommend? Logging a warning every time you import this module would also not be wonderful...

One approach is maybe to allow the types/functions/classes to be defined, but have them throw errors at runtime when they are used?

defunctzombie: Is there any indicator we have from the wasm build that this file could read to decide if its running in wasm vs not and throw an error when in python where we expect these files to be present?

Sort of like the `process.main` or whatnot from nodejs?

gasmith: Claude suggests using `sys.platform == "emscripten"`, which sounds fairly reasonable. Could use that to decide whether to swallow or re-throw the ImportError.

gasmith (approved):

python/foxglove-sdk/python/foxglove/__init__.py
jtbandes: Done

PR #1170, opened 2026-04-17
ci: Add draft_ros_release.yml

gasmith:
### Changelog
None

### Docs
None

### Description
This change adds a CI workflow for building debian packages for ROS
releases. It runs automatically when new ros-v* tags are published, and
drafts a release from the package changelogs. When run in this mode, it
builds against the version of the SDK configured in CMakeLists.txt.

As an extra feature, it can also be used to generate packages against
the version of an SDK on an arbitrary branch. In this mode, it builds a
C++ distribution against the branch, and does not draft a release. This
maybe useful when we need to distribute an ad-hoc build quickly, without
cutting a release or waiting for it to percolate through the standard
ROS distribution channels.

### Testing
See #1171: 
 - https://github.com/foxglove/foxglove-sdk/actions/runs/24587931589
 - https://github.com/foxglove/foxglove-sdk/releases/tag/untagged-408474d29960ab059ed2

### DIFF
--- a/.github/workflows/draft-ros-release.yml
+++ b/.github/workflows/draft-ros-release.yml
@@ -75,9 +75,11 @@ jobs:
 
       - name: Build .deb
         working-directory: ros
-        env:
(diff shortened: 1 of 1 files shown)


eloff (approved):
sudo apt-get install ros-jazzy-foxglove-bridge_3.2.6-0noble_amd64.deb

PR #1172, opened 2026-04-17
remote_access: Key pending_resets by ClientId

dave-foxglove:
### Changelog

None

### Docs

None

### Description

**Problem.** The `pending_resets` set in the remote-access session was keyed by `ParticipantIdentity`, which LiveKit reuses across disconnect/reconnect for the same logical client. A reset request queued for one physical connection could fire against a different, healthy connection that reconnected under the same identity before the drain ran — spuriously tearing down the replacement.

Flagged as a post-merge follow-up by @gasmith in the review of #1160:
https://github.com/foxglove/foxglove-sdk/pull/1160#pullrequestreview-4124702297

**Approach.** Rekey `pending_resets` by `ClientId` — a crate-local `u32` assigned once per physical connection and never reused. The drain in `handle_room_events` looks up each `ClientId` via `SessionState::get_participant_by_client_id`; when the id is no longer registered the request is stale (either the participant was removed, or it was already replaced by a reconnection with a fresh `ClientId`) and the drain skips it.

To support `ClientId` lookups alongside the existing identity-keyed access, participant storage on `SessionState` is encapsulated behind a new `Participants` type that maintains both indexes in sync by construction. The invariant is local to that type rather than spread across every mutator of `SessionState`.

The three commits are separated by responsibility:

1. `Encapsulate participant storage in Participants type` — pure refactor behind the dual-key wrapper. No semantic change.
2. `Key pending_resets by ClientId instead of ParticipantIdentity` — the actual fix. Adds a test `get_participant_by_client_id_does_not_match_replaced_participant` that locks in the staleness-rejection property.
3. `Tighten insert_participant signature and dedupe pending_resets rationale` — post-review cleanup: drops a redundant `identity` parameter, consolidates the `ClientId`-vs-identity rationale (previously duplicated across three doc/comment locations), and adds a drain-site comment capturing the single-task invariant the lookup relies on.

**Tradeoffs considered.**

- **Tuple `(ClientId, ParticipantIdentity)` in the set** vs. `HashSet<ClientId>` + secondary index: the tuple approach has a single load-bearing staleness check at the drain site and no permanent secondary index. The secondary-index approach (chosen here) pays a small always-on memory cost (8 bytes per participant for the extra `Arc` pointer) but makes `ClientId` → `Participant` a first-class access path, which fits the codebase's direction — `connection_graph` already keys by `ClientId`. Encapsulating both indexes in a single `Participants` type keeps the sync invariant local, removing 1b's main downside (distributed invariant across mutators).
- **Other identity-keyed maps on `SessionState`** (`subscriptions`, `video_subscribers`, `client_channels`, `subscribed_parameters`, `flush_handles`) were audited and kept as-is. None share `pending_resets`' producer/consumer decoupling — they're mutated synchronously with `add_participant` / `remove_participant` under `subscription_lock`, and `SessionState::remove_participant` sweeps all of them atomically. No stale-key window exists.

**Testing.**

Unit tests in `participants.rs` cover every method on the new type, including the dual-identity / distinct-`ClientId` rejection case. The core staleness-rejection property is tested at `SessionState` level (`get_participant_by_client_id_does_not_match_replaced_participant`), which exercises remove + replace under the same identity and asserts the original `ClientId` no longer resolves. Full-suite `cargo test -p foxglove --all-features` (476 pass) and `cargo test -p foxglove --no-default-features` (56 pass) both clean. `cargo clippy --no-deps --all-targets --tests -- -D warnings` clean. `cargo doc` clean.

No live LiveKit testing required — the staleness behavior is deterministic at `SessionState` level and exercised by the unit tests.

Fixes: [FLE-427](https://linear.app/foxglove/issue/FLE-427)

### DIFF
--- a/rust/foxglove/src/remote_access/participants.rs
+++ b/rust/foxglove/src/remote_access/participants.rs
@@ -1,0 +1,195 @@
+//! Two-key lookup table for connected participants.
+//!
+//! Maintains both `ParticipantIdentity` → `Arc<Participant>` and
+//! `ClientId` → `Arc<Participant>` indexes. Both reference the same
+//! `Arc<Participant>` values and are kept in sync by construction —
+//! mutation is only possible through the inherent methods on [`Participants`],
+//! all of which update both indexes together.
+
+use std::collections::HashMap;
+use std::collections::hash_map::Entry;
+use std::sync::Arc;
+
+use livekit::id::ParticipantIdentity;
+
+use crate::remote_access::participant::Participant;
+use crate::remote_common::ClientId;
+
+/// Collection of connected participants, indexed by both `ParticipantIdentity`
+/// and `ClientId`.
+#[derive(Default)]
+pub(crate) struct Participants {
+    by_identity: HashMap<ParticipantIdentity, Arc<Participant>>,
+    by_client_id: HashMap<ClientId, Arc<Participant>>,
+}
+
+impl Participants {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Inserts `participant` if no participant with the same identity is present.
+    ///
+    /// Returns `true` on insert, `false` if the identity was already occupied —
+    /// in the latter case neither index is modified.
+    pub fn insert(&mut self, participant: Arc<Participant>) -> bool {
+        let identity = participant.participant_id().clone();
+        let Entry::Vacant(v) = self.by_identity.entry(identity) else {
+            return false;
+        };
+        self.by_client_id
+            .insert(participant.client_id(), participant.clone());
+        v.insert(participant);
+        true
+    }
+
+    /// Removes and returns the participant for the given identity, if present.
+    pub fn remove_by_identity(
+        &mut self,
+        identity: &ParticipantIdentity,
+    ) -> Option<Arc<Participant>> {
+        let participant = self.by_identity.remove(identity)?;
+        self.by_client_id.remove(&participant.client_id());
+        Some(participant)
+    }
+
+    /// Returns the participant for the given identity, if present.
+    pub fn get_by_identity(&self, identity: &ParticipantIdentity) -> Option<&Arc<Participant>> {
+        self.by_identity.get(identity)
+    }
+
+    /// Returns the participant for the given `client_id`, if present.
+    #[allow(dead_code)]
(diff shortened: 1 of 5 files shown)


dave-foxglove:
> Traced the drain loop, the `Participants` dual-index type, and both reset-insertion sites (`send_control` and the flush task). The staleness-rejection logic at drain time is straightforward: `ClientId` lookup misses → skip, and the single-task invariant on `handle_room_events` means no TOCTOU betwee
> ...

[resolved] [AI] Acknowledged — review was a summary with no concerns.

rust/foxglove/src/remote_access/participants.rs
```diff
@@ -0,0 +1,194 @@
...
+        self.by_identity.values()
+    }
+
+    /// Returns the number of registered participants.
+    pub fn len(&self) -> usize {
+        self.by_identity.len()
+    }
+
+    /// Removes all participants and returns them.
+    pub fn drain(&mut self) -> Vec<Arc<Participant>> {
```
gasmith: Nit: "take" sounds more appropriate here; "drain" usually implies an iterator.

gasmith (approved):
LG, just a nit

rust/foxglove/src/remote_access/participants.rs
dave-foxglove: [resolved] [AI] Renamed `drain` to `take` and updated the one caller (`session_state.rs:take_participants`) and the unit test.

dave-foxglove:
> LG, just a nit

[resolved] [AI] Nit addressed — renamed `Participants::drain` to `take` (plus caller and test).

dave-foxglove:
> Traced the fix end-to-end: `ClientId` is allocated once in `spawn`, shared by copy into the flush task closure and the `Participant` struct, so both sides always insert the same id into `pending_resets`. The drain loop in `handle_room_events` does a `read()` lock that drops before `reset_participant
> ...

[resolved] [AI] Acknowledged — trace and invariant analysis match the approach.

dave-foxglove:
> Re-reviewed after the `drain` → `take` rename and `insert_participant` signature tightening.
> 
> Traced the full path again on the latest commits: `ClientId` allocated in `spawn` before the flush task closure captures it, both `send_control` and the flush task insert the same copy into `pending_resets`
> ...

[resolved] [AI] Acknowledged — post-rename re-trace found no new concerns.

PR #1175, opened 2026-04-17
remote_access: Prefer H.264

gasmith:
### Changelog
None

### Docs
None

### Description
Eventually we want to enable NVENC hardware offloading, but it's going
to take some work to bake CUDA into the toolchain. For now we can take
advantage of VAAPI on the platforms where it is supported simply by
preferring H.264. Sadly, this only makes a difference for Linux x86_64,
because webrtc-sys does not include support for VAAPI on aarch64.

When we eventually support NVENC, we can either stick with H.264, or
offer an opt-in for H.265.

### Testing
Ran the example. Before this change, `intel_gpu_top` does not show the
example. After this change, it does.


eloff (approved):
LGTM. Really disappointing it only has hardware encoding support on x86, but what can you do.

PR #1177, opened 2026-04-17
remote_access: Use session-scoped CancellationToken for watcher teardown

dave-foxglove:
### Changelog

None

### Docs

None

### Description

Follow-up from #1160: the `video_metadata_task.abort()` + `let _ = task.await` (silently discarding the result) in the session teardown was only necessary because the watcher's break arm listened on the same `CancellationToken` as the outer `select!` in `run()`. That token is the connection-level token, so when `handle_room_events` returned from a room disconnect (without the connection being cancelled), nothing fired the break — hence the explicit `abort()`.

This PR makes the session's `CancellationToken` a child of the connection's, via `child_token()` instead of `clone()`, and adds a `RemoteAccessSession::cancel()` method to fire it. The teardown path now calls `session.cancel()` and `await`s the task normally, logging any `JoinError` instead of silently discarding it.

Side effect: the byte-stream readers (`handle_byte_stream_from_client`), data tracks (`publish_data_tracks`), and per-participant flush tasks (via `Participant::spawn`'s child token) also now observe session-scoped cancellation. All three were already semantically session-scoped — the previous behavior only happened to work because the token was cancelled on connection shutdown. The new behavior also cancels them when the session ends mid-connection, which is desirable.

No new tests — the watcher/cancel interaction isn't unit-testable without a real `livekit::Room` (no mock constructor). Regression coverage comes from `remote_access_tests/livekit_test.rs`, whose session lifecycle tests would time out if teardown hung. A follow-up note tracks adding a targeted integration test when we next expand that crate.

Fixes: [FLE-441](https://linear.app/foxglove/issue/FLE-441/remote-access-use-session-scoped-cancellationtoken-for-watcher)

### DIFF
--- a/rust/foxglove/src/remote_access/connection.rs
+++ b/rust/foxglove/src/remote_access/connection.rs
@@ -301,7 +301,7 @@ impl RemoteAccessConnection {
                         capabilities: self.capabilities.clone(),
                         supported_encodings: self.supported_encodings.clone().unwrap_or_default(),
                         runtime: self.runtime.clone(),
-                        cancellation_token: self.cancellation_token.clone(),
+                        cancellation_token: self.cancellation_token.child_token(),
                         message_backlog_size: self
                             .message_backlog_size
                             .unwrap_or(DEFAULT_MESSAGE_BACKLOG_SIZE),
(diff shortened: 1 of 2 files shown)


dave-foxglove:
> Small, focused change — traced through the token hierarchy and teardown path, no issues found.
> 
> `child_token()` vs `clone()` is the right fix: it makes the session token independently cancellable while still inheriting connection-level cancellation. The downstream consumers (`handle_byte_stream_from
> ...

[resolved] [AI] Acknowledged — clean review summary, no inline items to address.

gasmith (approved):

PR #1179, opened 2026-04-21
Add sysinfo publishing cpu and memory stats

eloff:
### Changelog

Foxglove Bridge 2 now logs CPU and memory stats on /foxglove_bridge/sysinfo topic by default. Can be disabled by setting sysinfo param to false.

### Docs

None

### Description

Adds a SystemInfoPublisher builder that creates a channel and starts a background task to periodically log CPU and memory usage stats.

This publishes stats:

process_memory
process_virtual_memory
process_cpu_percent
total_cpu_percent
num_cpus,
total_memory
used_memory
total_swap
used_swap
kernel_version
os_version

On the /sysinfo topic (by default; /foxglove_bridge/sysinfo in ros bridge)

Also added it to the C++ and Python constructors and enabled it in the foxglove bridge.

Screenshot from 2026-04-20 18-40-21

### Testing

- [x] Test with plot panel, raw message panel, and websocket server
- [x] Same test with remote access gateway


ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -159,6 +159,8 @@ FoxgloveBridge::FoxgloveBridge(const rclcpp::NodeOptions& options)
       sdkServerOptions.capabilities | foxglove::WebSocketServerCapabilities::Time;
   }
 
+  sdkServerOptions.sysinfo_refresh_interval = 200ms;
```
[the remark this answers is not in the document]
eloff: It's a small message, 5hz is fine

cpp/foxglove/include/foxglove/websocket.hpp
```diff
@@ -277,6 +278,17 @@ struct WebSocketServerOptions {
...
+  /// @brief Optional refresh interval for publishing process and system statistics to the
+  /// `/sysinfo` topic.
+  ///
+  /// When set, the server publishes a `SystemInfo` message at the given interval for the
+  /// duration of the server's lifetime. The interval is clamped to a minimum of
+  /// `sysinfo::MINIMUM_CPU_UPDATE_INTERVAL`, since CPU usage samples taken more frequently than
+  /// that are not refreshed by the underlying crate.
+  ///
+  /// Defaults to `std::nullopt` (disabled).
+  std::optional<std::chrono::milliseconds> sysinfo_refresh_interval = std::nullopt;
```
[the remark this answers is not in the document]
eloff: I'll fix the docs

rust/foxglove/src/lib.rs
```diff
@@ -348,6 +350,12 @@
 #![warn(missing_docs)]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
+// The `#[derive(foxglove::Encode)]` macro emits code that refers to
+// `::foxglove::...`. Expose the current crate under that name so derived
+// implementations compile when used from within this crate.
+#[cfg(feature = "derive")]
+extern crate self as foxglove;
```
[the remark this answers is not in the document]
eloff: I think I'm going to switch to JSON to eliminate the dependency on derive

rust/foxglove/src/remote_access/gateway.rs
```diff
@@ -272,6 +276,19 @@ impl Gateway {
...
 
+    /// Enables or disables publishing process and system statistics to the
+    /// `/sysinfo` topic.
+    ///
+    /// When set to `Some(interval)`, the server publishes a
+    /// SystemInfo message at the given `interval` with cpu and memory
+    /// stats for the process and the system. Clamped to a minimum of 200ms.
+    ///
+    /// Defaults to `None` (disabled).
+    pub fn sysinfo(mut self, refresh_interval: Option<Duration>) -> Self {
```
gasmith: I think we should definitely allow overriding the topic. Especially if we want to enable this in the bridge by default (which I do), because users typically think of the bridge as _transparent_ - they're not expect it to publish topics of its own. The bridge's sysinfo topic could be `/foxglove_bridge/sysinfo` to dispel that confusion, but that default doesn't make sense for folks integrating against the SDK directly.

I wonder whether we want a full-blown builder pattern for the SysInfo configuration. I could imagine there being other tunables we want to set in the future.

rust/foxglove/src/remote_access/gateway.rs
```diff
@@ -217,8 +220,9 @@ impl std::fmt::Debug for Gateway {
             .field("has_qos_classifier", &self.qos_classifier.is_some())
             .field("server_info", &self.server_info)
             .field("message_backlog_size", &self.message_backlog_size)
-            .field("has_context", &(self.context.strong_count() > 0))
-            .finish()
+            .field("has_context", &(self.context.strong_count() > 0));
+        dbg.field("sysinfo", &self.sysinfo);
```
gasmith: Nit: Seems a bit odd

rust/foxglove/src/lib.rs
```diff
@@ -446,6 +446,13 @@ mod api_client;
 #[cfg(feature = "remote-access")]
 pub mod remote_access;
 
+#[cfg(feature = "_remote-common")]
+#[cfg_attr(
+    docsrs,
+    doc(cfg(any(feature = "remote-access", feature = "websocket")))
```
gasmith: Could this be decoupled from remote access/websocket completely? Configuring this on the sink seems kind of strange, now that I'm thinking about it. This is a _source_ of data. Maybe you want to write it to an MCAP file as well.

rust/foxglove/src/system_info.rs
```diff
@@ -0,0 +1,293 @@
...
+    let channel = ChannelBuilder::new(SYSINFO_TOPIC)
+        .context(&ctx)
+        .build::<SystemInfo>();
+    // We don't need the context anymore, don't keep it alive longer than needed
+    drop(ctx);
+    drop(context);
+
+    let pid = Pid::from_u32(std::process::id());
+    let kernel_version = System::kernel_version().unwrap_or_default();
+    let os_version = System::os_version().unwrap_or_default();
```
gasmith, line 288: I wonder if it's worthwhile to publish these at a different frequency? I suppose they're rather small, so it's not a huge deal.

gasmith (commented):
I'm very excited about this feature. The core implementation looks good, but I think we should make the configuration a bit more extensible, and decouple it entirely from the sink configurations.

rust/foxglove/src/lib.rs
eloff: I didn't think about that, but it makes sense. I'll revisit the design

rust/foxglove/src/system_info.rs
eloff: I though about putting them maybe on another channel or something, but it's probably micro-optimizing

rust/foxglove/src/system_info.rs
```diff
@@ -0,0 +1,294 @@
...
+            process_cpu_percent: process_cpu_percent as f64,
+            total_cpu_percent: system.global_cpu_usage() as f64,
+            num_cpus,
+            total_memory: system.total_memory() as f64,
+            used_memory: system.used_memory() as f64,
+            total_swap: system.total_swap() as f64,
+            used_swap: system.used_swap() as f64,
+            kernel_version: kernel_version.clone(),
+            os_version: os_version.clone(),
+        };
```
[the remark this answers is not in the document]
eloff: I'm not concerned about this

rust/foxglove/src/system_info.rs
```diff
@@ -0,0 +1,294 @@
...
+    "total_cpu_percent",
+    "num_cpus",
+    "total_memory",
+    "used_memory",
+    "total_swap",
+    "used_swap",
+    "kernel_version",
+    "os_version"
+  ]
+}"#;
```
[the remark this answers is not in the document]
eloff: I think it would be better to drop the required part from the schema, it's not needed to read the messages

rust/foxglove/Cargo.toml
```diff
@@ -8,9 +8,10 @@ repository.workspace = true
 license.workspace = true
 
 [features]
-default = ["derive", "websocket", "lz4", "zstd"]
+default = ["derive", "websocket", "sysinfo", "lz4", "zstd"]
```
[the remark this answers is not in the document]
eloff: I suppose It's fine to leave it out of default, it only affects rust users as you say, and they can opt into it.

rust/foxglove/src/system_info.rs
```diff
@@ -0,0 +1,410 @@
...
+            total_memory: system.total_memory() as f64,
+            used_memory: system.used_memory() as f64,
+            total_swap: system.total_swap() as f64,
+            used_swap: system.used_swap() as f64,
+            kernel_version: kernel_version.clone(),
+            os_version: os_version.clone(),
+        };
+
+        channel.log(&info);
+    }
```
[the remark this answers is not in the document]
eloff: I missed some comments, I'll fix that. If the user wants to stop it they can call abort() on the handle, otherwise it is meant to run forever.

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -229,6 +229,22 @@ FoxgloveBridge::FoxgloveBridge(const rclcpp::NodeOptions& options)
   this->set_parameter(rclcpp::Parameter{PARAM_PORT, _server->port()});
   RCLCPP_INFO(this->get_logger(), "Server listening on port %d", _server->port());
 
+  if (this->get_parameter(PARAM_SYSINFO).as_bool()) {
```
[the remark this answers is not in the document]
eloff: That's the intent. Added to the changelog.

rust/foxglove/src/system_info.rs
```diff
@@ -0,0 +1,419 @@
...
+    pub fn refresh_interval(mut self, interval: Duration) -> Self {
+        self.refresh_interval = Some(interval);
+        self
+    }
+
+    /// Sets the [`Context`] on which the publisher creates its channel.
+    ///
+    /// Defaults to the global default context.
+    pub fn context(mut self, ctx: &Arc<Context>) -> Self {
+        self.context = Some(Arc::downgrade(ctx));
```
gasmith: Nit: Since the context is only held for the lifetime of the builder, I don't think we need to downgrade it.

rust/foxglove/src/system_info.rs
```diff
@@ -0,0 +1,419 @@
...
+use crate::{Channel, ChannelBuilder, Context, Encode, Schema, runtime::get_runtime_handle};
+
+/// The default topic the [`SystemInfoPublisher`] publishes to.
+pub const DEFAULT_SYSINFO_TOPIC: &str = "/sysinfo";
+
+/// The default refresh interval for [`SystemInfoPublisher`].
+pub const DEFAULT_SYSINFO_REFRESH_INTERVAL: Duration = Duration::from_secs(1);
+
+/// JSON Schema (draft 2020-12) describing [`SystemInfo`] for consumers of the `/sysinfo` topic.
+const SYSINFO_JSON_SCHEMA: &str = r#"{
```
gasmith, line 35: Let's not do it just yet, but I could imagine this being promoted to a first class foxglove message type in the near future.

ros/src/foxglove_bridge/include/foxglove_bridge/param_utils.hpp
```diff
@@ -39,6 +42,8 @@ constexpr char DEFAULT_ADDRESS[] = "0.0.0.0";
 constexpr int64_t DEFAULT_SEND_BUFFER_LIMIT = 10000000;
 constexpr int64_t DEFAULT_MIN_QOS_DEPTH = 1;
 constexpr int64_t DEFAULT_MAX_QOS_DEPTH = 25;
+constexpr char DEFAULT_SYSINFO_TOPIC[] = "/foxglove_bridge/sysinfo";
+constexpr int64_t DEFAULT_SYSINFO_REFRESH_INTERVAL_MS = 500;
```
gasmith, line 46: Why did we pick a different default for the bridge?

python/foxglove-sdk/src/system_info.rs
```diff
@@ -0,0 +1,76 @@
...
+            handle.abort();
+        }
+    }
+}
+
+impl Drop for PySystemInfoPublisher {
+    fn drop(&mut self) {
+        self.stop();
+    }
+}
```
gasmith: This is inconsistent with the rust implementation.

gasmith (commented):
Looks good. I've got a few minor comments, but otherwise I think this is good to go.

rust/foxglove/src/system_info.rs
eloff: Yeah, that's a holdover from earlier code I ripped out

rust/foxglove/src/system_info.rs
eloff: Yeah, it's possible

python/foxglove-sdk/src/system_info.rs
eloff: True, good catch

ros/src/foxglove_bridge/include/foxglove_bridge/param_utils.hpp
eloff: No good reason, I'll make it 500ms across the board. It seems like a reasonable default value.

rust/foxglove/src/system_info.rs
```diff
@@ -198,8 +198,8 @@ impl SystemInfoPublisher {
     /// Sets the [`Context`] on which the publisher creates its channel.
     ///
     /// Defaults to the global default context.
-    pub fn context(mut self, ctx: &Arc<Context>) -> Self {
-        self.context = Some(Arc::downgrade(ctx));
+    pub fn context(mut self, ctx: Arc<Context>) -> Self {
```
gasmith: Nit: Other APIs that take a context take a reference and clone it internally.

gasmith (approved):

rust/foxglove/src/system_info.rs
```diff
@@ -0,0 +1,408 @@
...
+    "kernel_version": {
+      "type": "string",
+      "description": "Kernel version string, or empty if unavailable on this platform."
+    },
+    "os_version": {
+      "type": "string",
+      "description": "OS version string, or empty if unavailable on this platform."
+    }
+  }
+}"#;
```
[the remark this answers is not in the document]
eloff: I'd rather not do that, it's not needed by the app and it might make it harder to evolve the type over time

rust/foxglove/src/system_info.rs
```diff
@@ -0,0 +1,450 @@
...
+            .unwrap_or_default();
+
+        let info = SystemInfo {
+            process_memory: process_memory as f64,
+            process_virtual_memory: process_virtual_memory as f64,
+            process_cpu_percent: process_cpu_percent as f64,
+            total_cpu_percent: system.global_cpu_usage() as f64,
+            num_cpus,
+            total_memory: system.total_memory() as f64,
+            used_memory: system.used_memory() as f64,
```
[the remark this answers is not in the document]
eloff: That's fine

clippy.toml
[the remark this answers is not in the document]
eloff: Intentional

rust/foxglove/src/system_info.rs
```diff
@@ -0,0 +1,450 @@
...
+    ///
+    /// The task is intended to run until [`SystemInfoHandle::abort`] is called on the
+    /// returned handle.
+    ///
+    /// The publisher creates its channel and registers it with the context
+    /// synchronously before spawning the background task. The channel is
+    /// closed when the background task exits (for example after `abort`).
+    pub fn start(self) -> SystemInfoHandle {
+        // If we refresh too quickly we'll get invalid values for cpu usage.
+        // sysinfo crate exports a platform dependent MINIMUM_CPU_UPDATE_INTERVAL
```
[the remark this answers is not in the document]
eloff: I fixed this, and I don't see this in the code. Odd

PR #1182, opened 2026-04-21
ros: Release version 3.3.0

gasmith:
### Changelog
foxglove_bridge:
- Add support for remote access
- Update Foxglove SDK version to 0.22.0

foxglove_msgs:
- Add Odometry schema
- Remove Velocity3 schema; LocationFix.velocity now uses Vector3

### Docs
None

### Description
This change is to release the next version of the bridge, which includes
support for remote access.

### Testing
Mock release build: https://github.com/foxglove/foxglove-sdk/actions/runs/24732192556

### DIFF
--- a/ros/src/foxglove_bridge/CHANGELOG.rst
+++ b/ros/src/foxglove_bridge/CHANGELOG.rst
@@ -2,6 +2,11 @@
 Changelog for package foxglove_bridge
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+3.3.0 (2026-04-21)
+------------------
+* Add support for remote access
+* Update Foxglove SDK version to 0.22.0
+
 3.2.6 (2026-04-02)
 ------------------
 * Fix ROS rolling build for rclcpp 31.0.0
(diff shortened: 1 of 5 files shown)


gasmith:
> One question: the foxglove_msgs changelog calls out "Remove Velocity3 schema; LocationFix.velocity now uses Vector3" — that's a breaking change for anyone currently using `Velocity3`. Is that covered by the major version bump in foxglove_msgs, or does it warrant a migration note somewhere? Docs says "None" in the PR description.

This field was only recently added to LocationFix (in 3.2.6, April 3), and in all likelihood is not actually being consumed by anyone. We're deliberately choosing to keep this as a minor version bump, even though it technically is a breaking change.

The only maintained package that depends on `foxglove_msgs` is `foxglove_compressed_video_transport` (https://index.ros.org/p/foxglove_msgs/#rolling), and that hasn't seen a new release [since 2025-09-01](https://github.com/ros-misc-utilities/foxglove_compressed_video_transport/blob/release/CHANGELOG.rst#302-2025-09-01).

gasmith:
We're going to postpone this release. This PR is superseded by #1183.

PR #1185, opened 2026-04-21
docs: Fix remote data loader example

gasmith:
### Changelog
None

### Docs
None

### Description
The remote data loader backend example was [not rendering properly](https://foxglove-sdk-api-docs.pages.dev/cpp/generated/api/file_foxglove_include_foxglove_remote_data_loader_backend.hpp#detailed-description), due
to a quirk in exhale, which seems to be related to the `@file` section
and `## Example` subheading somehow, but I couldn't find a reasonable
way to fix it. Moved the example to the `Manifest` struct instead, with a
link from the header doc.

I also noticed that I was unable to build the C++ documentation locally
with the build container, because the build container was using doxygen
1.9.1 instead of the 1.13.2 that we're using in CI. Updated the
Dockefile to pin doxygen at 1.13.2.

New docs experience:
- [remote_data_loader_backend.hpp](https://f24432f8.foxglove-sdk-api-docs.pages.dev/cpp/generated/api/file_foxglove_include_foxglove_remote_data_loader_backend.hpp)
- [Example](https://f24432f8.foxglove-sdk-api-docs.pages.dev/cpp/generated/api/structfoxglove_1_1remote__data__loader__backend_1_1Manifest#struct-documentation)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,6 @@ RUN apt-get update \
     && apt-get install -y \
         cmake \
         curl \
-        doxygen \
         gcc \
         g++ \
         git \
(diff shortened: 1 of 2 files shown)


Muon:
This is a bit odd, if there is a bug with the renderer, can we fix that instead?

gasmith:
> This is a bit odd, if there is a bug with the renderer, can we fix that instead?

You're welcome to take a crack at it, if the positioning of the example is that important to you. I took this approach because it felt like the most bang for the buck.

Muon (approved):

Muon:
Well, I took a look and it seems `exhale` is unmaintained with no code changes for 2 years now.

I think accepting this change is OK for the moment, but we should either fork it or find a different documentation generator.

gasmith:
> Well, I took a look and it seems `exhale` is unmaintained with no code changes for 2 years now.
> 
> I think accepting this change is OK for the moment, but we should either fork it or find a different documentation generator.

Fair. Filed https://linear.app/foxglove/issue/FLE-453/consider-forking-or-replacing-exhale-for-c-docs for future investigation, though I think I'd personally need a bit stronger motivation than this particular issue to pick it up.

PR #1193, opened 2026-04-24
Dockerfile.ros: Swap to ros2-testing via ros2-testing-apt-source

clalancette:
### Changelog

None.

### Docs

None.

### Description

The prior sed-based swap editing /etc/apt/sources.list.d/ros2.sources was silently reverted during `apt-get dist-upgrade` when ros2-apt-source was upgraded 1.1.0 -> 1.2.0: the new package replaces the sed'd regular file with a symlink back into /usr/share, pointing at the default ros2/ubuntu URL. The subsequent `apt-get install python3-bloom` then failed with "Unable to locate package" because apt had only fetched the ros2-testing index.

Install the official ros2-testing-apt-source package instead, which Conflicts with ros2-apt-source (so dpkg cleanly swaps them) and configures ros2-testing as the sole ROS source in a way that survives dist-upgrade.

This should fix our currently failing Rolling builds, like in https://github.com/foxglove/foxglove-sdk/actions/runs/24850793861/job/72875771799?pr=1191

### DIFF
--- a/.github/workflows/ros.yml
+++ b/.github/workflows/ros.yml
@@ -24,6 +24,7 @@ jobs:
               - 'c/**'
               - 'cpp/**'
               - 'ros/**'
+              - 'Dockerfile.ros'
               - '.github/workflows/ros.yml'
 
   build-cpp-dist:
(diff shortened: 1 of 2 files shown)


gasmith (approved):
🚢

PR #1194, opened 2026-04-24
remote_access: Fix same-identity reconnect race

clalancette:
### Changelog

None.

### Docs

None.

### Description

This saga started when a few days ago I noticed that the netem test on one of my PRs failed.  That has happened (very) sporadically before, so I decided to dig into it.  I ended up finding a race condition.

Under impairment the netem integration test can fail with "byte stream ended unexpectedly" when the viewer's outer connect-retry timer fires (`ViewerConnection::connect_with_timeout` in `test_helpers.rs`, 5s). The test reuses a hardcoded `"viewer-1"` identity across retries, so the retry joins the room under the same identity as the attempt the gateway is still holding state for. The gateway's reset path (triggered by attempt 1's flush task failing a write) re-opens a stream against that identity, and a queued `ParticipantDisconnected` for attempt 1 then tears down the freshly re-registered attempt 2 — because `remove_participant` keyed on `ParticipantIdentity` alone.
    
Each `Participant` now stores LiveKit's per-connection `ParticipantSid` at registration time, and `remove_participant` takes both the identity and the expected SID; a removal whose SID doesn't match the currently stored one is a no-op. A stale `ParticipantDisconnected` for a prior instance can't remove its replacement. `reset_participant` captures
    the stored participant's SID and `protocol_version` before removing: the stored SID is what's passed to `remove_participant` (identifying which specific instance to tear down), and the `protocol_version` is reused on re-add to avoid a `remote_participants()` lookup that would race with reconnection. The re-add SID itself is still queried fresh from `remote_participants()` — if a same-identity reconnect has already landed, we want to register the *new* instance's SID, not the just-removed one's, so that a later stale `ParticipantDisconnected` for the old SID can't tear down the replacement.
    
Participant membership lives in a new `ParticipantRegistry` that owns the participant map, per-participant flush-task `JoinHandle`s, and the `pending_resets` / `reset_notify` signalling surfaces. It has no LiveKit dependency, so the race is reproducible in a unit test without a real `Room`: the session opens the control-plane stream and passes the resulting `ParticipantWriter` into `registry.register_participant`; tests build writers directly. `SessionState` is now purely session-level state (channels, subscriptions, video publishers, client channels, parameters); subscribe/unsubscribe methods take `&ParticipantIdentity` instead of `&Participant`, so `SessionState` has no dependency on `Participant` at all. Participant-removal cleanup (listener callbacks, context unsubscribe, video track teardown, connection-graph update) runs after registry removal via `run_participant_removal_cleanup`. `Sink::log` and `Sink::remove_channel` resolve subscriber identities through `ParticipantRegistry::resolve_identities`, which takes the registry read lock once per broadcast rather than per subscriber.

Regression test:
`remote_access::participant_registry::tests::stale_disconnect_must_not_remove_reconnected_participant`.

I tested locally with the remote_access example and connecting to it through the app, and it seems to work fine (both before and after this PR)

Coincidentally, I *believe* that this:

Fixes FLE-417

But I'm not certain of that because I don't have exact details on how that failure manifested.

### DIFF
--- a/rust/foxglove/src/remote_access.rs
+++ b/rust/foxglove/src/remote_access.rs
@@ -7,6 +7,7 @@ mod credentials_provider;
 mod gateway;
 mod listener;
 mod participant;
+mod participant_registry;
 mod participants;
 pub(super) mod protocol_version;
 mod qos;
--- a/rust/foxglove/src/remote_access/participant.rs
+++ b/rust/foxglove/src/remote_access/participant.rs
@@ -8,7 +8,6 @@ use livekit::{
     ByteStreamWriter, StreamWriter,
     id::{ParticipantIdentity, ParticipantSid},
 };
-use semver::Version;
 use tokio_util::sync::CancellationToken;
 
 use crate::protocol::v2::server::FetchAssetResponse;
@@ -39,11 +38,6 @@ pub(crate) struct Participant {
     /// (for a prior instance) apart from a legitimate disconnect of the
     /// current instance.
     participant_sid: ParticipantSid,
-    /// The remote access protocol version advertised by this participant.
-    /// Captured at `add_participant` time and reused on reset so we don't have to
-    /// re-query the LiveKit room (which would race with reconnection under the
-    /// same identity).
-    protocol_version: Version,
     /// Per-participant control plane queue. The receiving end is owned by the
     /// flush-task.
     control_tx: flume::Sender<Bytes>,
@@ -81,7 +75,6 @@ impl Participant {
     pub fn spawn(
         identity: ParticipantIdentity,
         participant_sid: ParticipantSid,
-        protocol_version: Version,
         writer: ParticipantWriter,
         queue_size: usize,
         pending_resets: Arc<parking_lot::Mutex<HashSet<ParticipantSid>>>,
@@ -130,7 +123,6 @@ impl Participant {
             client_id,
             participant_id: identity,
             participant_sid,
-            protocol_version,
             control_tx,
             pending_resets,
             reset_notify,
@@ -152,7 +144,6 @@ impl Participant {
     pub fn new(
         identity: ParticipantIdentity,
         participant_sid: ParticipantSid,
-        protocol_version: Version,
         control_tx: flume::Sender<Bytes>,
         pending_resets: Arc<parking_lot::Mutex<HashSet<ParticipantSid>>>,
         reset_notify: Arc<tokio::sync::Notify>,
@@ -162,7 +153,6 @@ impl Participant {
             client_id: ClientId::next(),
             participant_id: identity,
             participant_sid,
-            protocol_version,
             control_tx,
             pending_resets,
             reset_notify,
@@ -196,11 +186,6 @@ impl Participant {
     /// Returns the participant's identity.
     pub fn participant_id(&self) -> &ParticipantIdentity {
         &self.participant_id
-    }
-
-    /// Returns the protocol version advertised by this participant.
-    pub(crate) fn protocol_version(&self) -> &Version {
-        &self.protocol_version
     }
 
     /// Returns the LiveKit session ID this participant was added with. Unique
--- a/rust/foxglove/src/remote_access/participant.rs
+++ b/rust/foxglove/src/remote_access/participant.rs
@@ -8,6 +8,7 @@ use livekit::{
     ByteStreamWriter, StreamWriter,
     id::{ParticipantIdentity, ParticipantSid},
 };
+use semver::Version;
 use tokio_util::sync::CancellationToken;
 
 use crate::protocol::v2::server::FetchAssetResponse;
@@ -38,6 +39,15 @@ pub(crate) struct Participant {
     /// one specific instance: `ParticipantDisconnected` event handling,
     /// SID-keyed `remove_participant`, and `pending_resets` bookkeeping.
     participant_sid: ParticipantSid,
+    /// The remote access protocol version advertised by this participant.
+    /// Stored for future use when branching protocol behavior based on the
+    /// participant's version. Captured at registration and refreshed on
+    /// reset: `reset_participant` re-validates the freshly-queried
+    /// attributes and the new participant is registered with that fresh
+    /// version, so the stored value reflects the current connection
+    /// instance's advertised version even across same-identity reconnects.
+    #[expect(dead_code)]
+    protocol_version: Version,
     /// Per-participant control plane queue. The receiving end is owned by the
     /// flush-task.
     control_tx: flume::Sender<Bytes>,
@@ -76,6 +86,7 @@ impl Participant {
     pub fn spawn(
         identity: ParticipantIdentity,
         participant_sid: ParticipantSid,
+        protocol_version: Version,
         writer: ParticipantWriter,
         queue_size: usize,
         pending_resets: Arc<parking_lot::Mutex<HashSet<ParticipantSid>>>,
@@ -124,6 +135,7 @@ impl Participant {
             client_id,
             participant_id: identity,
             participant_sid,
+            protocol_version,
             control_tx,
             pending_resets,
             reset_notify,
@@ -145,6 +157,7 @@ impl Participant {
     pub fn new(
         identity: ParticipantIdentity,
         participant_sid: ParticipantSid,
+        protocol_version: Version,
         control_tx: flume::Sender<Bytes>,
         pending_resets: Arc<parking_lot::Mutex<HashSet<ParticipantSid>>>,
         reset_notify: Arc<tokio::sync::Notify>,
@@ -154,6 +167,7 @@ impl Participant {
             client_id: ClientId::next(),
             participant_id: identity,
             participant_sid,
+            protocol_version,
             control_tx,
             pending_resets,
             reset_notify,
--- a/rust/foxglove/src/remote_access/participant.rs
+++ b/rust/foxglove/src/remote_access/participant.rs
@@ -34,9 +34,9 @@ pub(crate) struct Participant {
     /// LiveKit session ID for this specific connection instance. Unique per
     /// physical connection — unlike `participant_id`, it changes when a
     /// participant disconnects and reconnects under the same identity. Used
-    /// by the `ParticipantDisconnected` handler to tell a stale disconnect
-    /// (for a prior instance) apart from a legitimate disconnect of the
-    /// current instance.
+    /// to disambiguate connection instances on every operation that targets
+    /// one specific instance: `ParticipantDisconnected` event handling,
+    /// SID-keyed `remove_participant`, and `pending_resets` bookkeeping.
     participant_sid: ParticipantSid,
     /// Per-participant control plane queue. The receiving end is owned by the
     /// flush-task.
@@ -46,10 +46,11 @@ pub(crate) struct Participant {
     /// us. Keyed by `ParticipantSid` (unique per physical connection) rather
     /// than `ParticipantIdentity` (stable across disconnect/reconnect) so a
     /// stale reset request doesn't fire against a reconnected participant
-    /// that happens to reuse the same identity. `ClientId` would also work
-    /// (per-instance), but keeping internal bookkeeping on `ParticipantSid`
-    /// matches the SID-keyed `remove_participant` path and reserves
-    /// `ClientId` for the user-facing `Listener` API.
+    /// that happens to reuse the same identity. `ClientId` is also
+    /// per-instance and would work for staleness disambiguation, but keying
+    /// on `ParticipantSid` matches the SID-keyed `remove_participant` path
+    /// directly; `ClientId` stays the identifier exposed through `Listener`
+    /// callbacks and the connection-graph subscriber index.
     pending_resets: Arc<parking_lot::Mutex<HashSet<ParticipantSid>>>,
     /// Wakes `handle_room_events` when we add ourselves to `pending_resets`.
     reset_notify: Arc<tokio::sync::Notify>,
--- a/rust/foxglove/src/remote_access/participant.rs
+++ b/rust/foxglove/src/remote_access/participant.rs
@@ -4,7 +4,10 @@ use std::collections::HashSet;
 use std::sync::Arc;
 
 use bytes::Bytes;
-use livekit::{ByteStreamWriter, StreamWriter, id::ParticipantIdentity};
+use livekit::{
+    ByteStreamWriter, StreamWriter,
+    id::{ParticipantIdentity, ParticipantSid},
+};
 use semver::Version;
 use tokio_util::sync::CancellationToken;
 
@@ -27,11 +30,19 @@ const DEFAULT_FETCH_ASSET_PER_PARTICIPANT: usize = 32;
 pub(crate) struct Participant {
     /// Locally-significant identifier for this particular instance of this participant.
     client_id: ClientId,
-    /// LiveKit participant ID.
+    /// LiveKit participant identity (stable across disconnect + reconnect).
     participant_id: ParticipantIdentity,
+    /// LiveKit session ID for this specific connection instance. Unique per
+    /// physical connection — unlike `participant_id`, it changes when a
+    /// participant disconnects and reconnects under the same identity. Used
+    /// by the `ParticipantDisconnected` handler to tell a stale disconnect
+    /// (for a prior instance) apart from a legitimate disconnect of the
+    /// current instance.
+    livekit_sid: ParticipantSid,
     /// The remote access protocol version advertised by this participant.
-    /// Stored for future use when branching protocol behavior based on the participant's version.
-    #[expect(dead_code)]
+    /// Captured at `add_participant` time and reused on reset so we don't have to
+    /// re-query the LiveKit room (which would race with reconnection under the
+    /// same identity).
     protocol_version: Version,
     /// Per-participant control plane queue. The receiving end is owned by the
     /// flush task.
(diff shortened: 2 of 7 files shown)


rust/foxglove/src/remote_access/session.rs
```diff
@@ -1334,18 +1369,41 @@ impl RemoteAccessSession {
...
         // (e.g., because the underlying WebRTC connection dropped). In that case, the
         // `ParticipantDisconnected` event is already queued and a future reconnect will
         // go through the normal `ParticipantConnected` → `add_participant` path.
-        let remote_participant = self
+        //
+        // If a new instance has already reconnected under the same identity,
+        // its SID is what we re-register with — so that a later stale
+        // `ParticipantDisconnected` for the *old* instance's SID won't match
+        // and tear down this re-registration.
+        let Some(sid) = self
```
eloff: In the PR description you mentioned:

reset_participant captures the stored participant's SID alongside its protocol_version before removing, _reusing both on re-add instead of re-querying room.remote_participants()_

But we still query .remote_participants() here. And that seems like the correct thing to do. Just want to clarify that the code is doing what you think it's doing here.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -964,61 +982,68 @@ impl RemoteAccessSession {
...
+            participant_id,
+            livekit_sid,
             protocol_version,
             ParticipantWriter::Livekit(stream),
-            self.message_backlog_size,
-            self.pending_resets.clone(),
-            self.reset_notify.clone(),
             &self.cancellation_token,
+            initial_messages,
         );
```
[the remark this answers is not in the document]
eloff: Maybe a #[must_use] on register_participant, and debug_assert!(self.registry.register_participant(...)) at the call site would help protect against future misuse here.

rust/foxglove/src/remote_access/participant.rs
```diff
@@ -63,8 +74,10 @@ impl Participant {
     ///
     /// Returns the participant (wrapped in `Arc` for shared ownership) and the
     /// flush task's `JoinHandle` (for teardown awaiting).
+    #[allow(clippy::too_many_arguments)]
     pub fn spawn(
         identity: ParticipantIdentity,
+        livekit_sid: ParticipantSid,
```
dave-foxglove: nit: about naming this livekit_sid, would you be up for naming this participant_sid or just sid or session _id or session_identity. It appears inconsistent with patterns like "identity: ParticipantIdentity" and using "livekit" in a vary name seems unusual unless this is very livekit specific (which I would not expect a session id to be livekit specific.

rust/foxglove/src/remote_access/participant.rs
```diff
@@ -138,6 +154,8 @@ impl Participant {
         Self {
             client_id: ClientId::next(),
             participant_id: identity,
+            livekit_sid: ParticipantSid::try_from("PA_test".to_string())
+                .expect("valid LiveKit SID prefix"),
```
eloff: nit: tests that build two participants with the same identity to simulate reconnect would silently get matching SIDs, which could cause issues depending how they're used

eloff (commented):
This is excellent. Great job tracking down that race and fixing it. Also refactoring things out into the registry is a really nice move to make things testable and adding a reproduction test case was the icing on the cake. Really well done!

I think there's a similar pre-existing race in the other direction:

1) Attempt 1 registered (SID S1).
2) LiveKit emits ParticipantActive(S2) (or ParticipantConnected then Active) for attempt 2 before the disconnect event for attempt 1.
3) add_participant runs, sees self.registry.has_participant(&id) == true (attempt 1 is still there) → returns Ok(()) early.
4) LiveKit emits ParticipantDisconnected(S1) for attempt 1.
5) remove_participant(id, S1) → SID matches stored → attempt 1 is correctly removed.

Net result: nobody is registered for id, but attempt 2 is alive in the LiveKit room. We never come back to register it.

I added a test and fix for this [here](https://github.com/foxglove/foxglove-sdk/pull/1196), since with the new registry and sid tracking that you added, it's now much easier to detect and handle this case as well.

rust/foxglove/src/remote_access/participant_registry.rs
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    /// Spawns a [`Participant`] and its flush task against the supplied
+    /// `writer` and inserts it into the registry.
+    ///
+    /// Each byte slice in `initial_messages` is queued on the participant's
+    /// control-plane channel after the flush task is spawned but before the
+    /// participant is visible in the registry. This preserves the invariant
+    /// that these bytes (typically `ServerInfo` + channel/service
+    /// advertisements) are the first the flush task delivers to the viewer,
```
dave-foxglove: ? "delivers" -> "delivered"

eloff (approved):

rust/foxglove/src/remote_access/participant_registry.rs
```diff
@@ -0,0 +1,355 @@
...
+use semver::Version;
+use tokio::sync::Notify;
+use tokio_util::sync::CancellationToken;
+
+use crate::remote_access::participant::{Participant, ParticipantWriter};
+use crate::remote_access::participants::Participants;
+use crate::remote_common::ClientId;
+
+/// Owns the participant membership state machine: add / remove / lookup, plus
+/// the `pending_resets` channel that lets a flush task request its own reset.
```
dave-foxglove: nit: In some comments "flush task" is written as "flush-task" which helps make this a *thing*. Writing it as two words was difficult for me to read/parse until I got used to it. Consider some kind of group indicator (such as the joining hyphen) or capitalization or quotes or something.

dave-foxglove (approved):
I made some minor nits, but overall LG.

rust/foxglove/src/remote_access.rs
```diff
@@ -7,6 +7,7 @@ mod credentials_provider;
 mod gateway;
 mod listener;
 mod participant;
+mod participant_registry;
 mod participants;
```
gasmith, line 11: Nit: I feel like there's an opportunity for a better organization here. Maybe just a matter of coming up with a better name for "participants", and/or nesting some of the collections under the "participant" module. Not terribly important since all this stuff is private, just something I noticed.

rust/foxglove/src/remote_access/participants.rs
```diff
@@ -1,55 +1,66 @@
...
+    /// handle. Returns the participant; the flush handle is detached and
+    /// dropped so the task exits when its control-plane channel senders are
+    /// all gone (or earlier, if the caller cancels the participant).
     pub fn remove_by_identity(
         &mut self,
         identity: &ParticipantIdentity,
     ) -> Option<Arc<Participant>> {
         let participant = self.by_identity.remove(identity)?;
         self.by_client_id.remove(&participant.client_id());
+        drop(self.flush_handles.remove(identity));
```
gasmith, line 68: Nit: Considering that we're dropping the flush handle here, it probably also makes sense to also call `participant.cancel()`, rather than making that the caller's responsibility. Simplifies the API contract a little. I can't imagine why we'd want to call this method without calling cancel.

rust/foxglove/src/remote_access/session.rs
gasmith: This doesn't feel right. We're tracking pending resets with ClientId (which has the same scope as a ParticipantSid), looking up the participant by ClientId, and then... requesting a reset by ParticipantIdentity. Oops. 

Do we really need the participants-by-ClientId map at all? What if, we tracked pending resets by ParticipantSid, and then reset participant by ParticipantSid? Wouldn't that be nicer?

We still need ClientId on the participant, since that's the identifier we expose to users through the Listener trait, but I don't think we need to use it for anything else internally. We should use the ParticipantSid.

rust/foxglove/src/remote_access/participant.rs
gasmith: I fully agree. Let's call it `participant_sid`, for parallelism with `participant_id`.

rust/foxglove/src/remote_access/participant.rs
gasmith: Yeah, I'd mildly prefer that tests pass in the sid, but that could wait until we actually have tests that care about distinguishing sids.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -134,6 +140,9 @@ pub(crate) struct RemoteAccessSession {
     room: Room,
     context: Weak<Context>,
     remote_access_session_id: Option<String>,
+    /// Session-level state: channels, subscriptions, video publishers, client
+    /// channels, parameter subscriptions. Participant membership lives on
+    /// [`registry`] instead.
     state: RwLock<SessionState>,
```
gasmith, line 146: I'm beginning to feel that this should be renamed `SubscriptionState` or `ChannelRegistry` or similar. Feel free to defer.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -158,15 +167,8 @@ pub(crate) struct RemoteAccessSession {
...
-    /// Populated by `Participant::send_control` on queue overflow and by flush
-    /// tasks on write failure. Drained by `handle_room_events`. See
-    /// [`Participant::pending_resets`] for why this is keyed by `ClientId`.
-    pending_resets: Arc<parking_lot::Mutex<HashSet<ClientId>>>,
-    /// Wakes `handle_room_events` when a new reset is added to `pending_resets`.
-    reset_notify: Arc<tokio::sync::Notify>,
-    /// Size of the per-participant control plane queue.
-    message_backlog_size: usize,
+    /// Participant membership and flush-task lifecycle.
+    registry: ParticipantRegistry,
```
gasmith: Nit: Let's call the field `participants` or `participant_registry` perhaps?

rust/foxglove/src/remote_access/session.rs
```diff
@@ -1334,18 +1369,41 @@ impl RemoteAccessSession {
...
+            Some(p) => (p.protocol_version().clone(), p.livekit_sid().clone()),
+            None => {
+                info!(
+                    remote_access_session_id,
+                    participant_identity = %participant_id,
+                    "reset requested for already-removed participant; skipping",
+                );
+                return;
+            }
+        };
```
gasmith: This gunk also gets a lot cleaner if we're resetting participants by SID directly, as I suggested above:

```rust
let Some(participant) = self.participants.remove_by_sid(sid) else { return };
let participant_id = participant.participant_id().clone();
let version = participant.protocol_version().clone();
drop(participant);
```

rust/foxglove/src/remote_access/session.rs
```diff
@@ -1354,25 +1412,13 @@ impl RemoteAccessSession {
             return;
         };
 
-        let Some(version) = protocol_version::check_participant_protocol_version(
-            &participant_id,
-            &remote_participant.attributes(),
-            remote_access_session_id,
-        ) else {
```
gasmith, line 1361: Should we keep this perhaps? Otherwise we're assuming that the new participant (which may have a new SID) is still running the same protocol version as the old participant. As written, this is _probably still safe_, since viewers change ParticipantIdentity when they reload to pick up code changes. But it might be nice to keep the defensive check.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -964,61 +982,68 @@ impl RemoteAccessSession {
...
-    pub(crate) fn remove_participant(self: &Arc<Self>, participant_id: &ParticipantIdentity) {
+    /// Callers: `handle_room_event::ParticipantDisconnected` passes the
+    /// event's SID; `reset_participant` passes the stored participant's SID.
+    /// Requiring the SID filters out stale disconnects for a prior connection
+    /// instance that has already been superseded by a same-identity reconnect.
+    pub(crate) fn remove_participant(
+        self: &Arc<Self>,
+        participant_id: &ParticipantIdentity,
+        expected_sid: &ParticipantSid,
+    ) {
```
gasmith: Seems a little silly to pass two required identifiers, when we only care about one of them. Let's just pass ParticipantSid?

gasmith (commented):
Generally looks great, and I'm supportive of getting this in before beta. However, I do have a few things I'd like to see cleaned up before we merge.

rust/foxglove/src/remote_access/session.rs
clalancette: Yeah, the code is correct, the PR description was wrong.  I've updated the PR description to be correct now.

rust/foxglove/src/remote_access.rs
clalancette: Claude suggested that we could rename "participants" to "participant_map" to disambiguate a little more.  We could make a sub-module "participant" and nest all of this stuff underneath, which is mechanical but would also add a lot of churn to paths and things like that.  What do you think we should do?

rust/foxglove/src/remote_access/participant_registry.rs
clalancette: I think "delivers" is correct here; the sentence is in present tense.  But see below about "flush-task".

rust/foxglove/src/remote_access/participant_registry.rs
```diff
@@ -0,0 +1,355 @@
...
+            self.reset_notify.clone(),
+            session_cancel,
+        );
+
+        for msg in initial_messages {
+            participant.send_control(msg);
+        }
+
+        self.participants.write().insert(participant, flush_handle)
+    }
```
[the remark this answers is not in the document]
clalancette: I went with Option 1 in the latest push.

rust/foxglove/src/remote_access/session.rs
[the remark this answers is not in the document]
clalancette: I went with @eloff's suggestion here.

rust/foxglove/src/remote_access/participant.rs
clalancette: Changed to `participant_sid` everywhere.

rust/foxglove/src/remote_access/participant.rs
clalancette: I fixed up the tests to pass in the sid, which was a relatively small change.

rust/foxglove/src/remote_access/participant_registry.rs
clalancette: I went with "flush-task" everywhere to make it clearer.

rust/foxglove/src/remote_access/participants.rs
clalancette: Yeah, it's a good callout.  I went back and forth a bit on this myself, and ended up keeping it separate for separation-of-concerns, but it is probably more correct (less likely for the caller to mess up) the way you suggested.  I made the change.

rust/foxglove/src/remote_access/session.rs
clalancette: This is a really good call.  What you suggested is way cleaner, and I made that change in the latest push.

rust/foxglove/src/remote_access/session.rs
clalancette: I put this in a deferred ticket https://linear.app/foxglove/issue/FLE-475/followup-to-httpsgithubcomfoxglovefoxglove-sdkpull1194

rust/foxglove/src/remote_access/session.rs
clalancette: I changed it to `participant_registry`.

rust/foxglove/src/remote_access/session.rs
clalancette: Yes, it cleaned up nicely.

rust/foxglove/src/remote_access/session.rs
clalancette: This was actually a bug; I never should have removed this.  Restored in the latest push.

rust/foxglove/src/remote_access/session.rs
clalancette: Yeah, this was also cleaned up in the latest push.

rust/foxglove/src/remote_access.rs
clalancette: I ended up opening https://linear.app/foxglove/issue/FLE-475/followup-to-httpsgithubcomfoxglovefoxglove-sdkpull1194 for some of the deferred things in here.  I think we should probably defer and consider a follow-up.

gasmith (approved):

eloff (approved):
I see one potential issue in Sink::log now in sesssion.rs, but I'm not convinced that it's worth fixing.

The new code captures ParticipantIdentity then re-resolves via the registry.

In the rare window:

1) Sink::log captures [viewer-1] and drops state.read().
2) Session::remove_participant(S1) runs (registry remove + cleanup_for_removed_identity).
3) add_participant(viewer-1, S2, ...) runs.
4) resolve_identities([viewer-1]) resolves to attempt 2.
5) Attempt 2 receives a MessageData for a channel it never subscribed to.

The comment at lines 188-190 acknowledges the removed-without-replacement case ("benign miss") but not the resolved-to-replacement case.

I think the app/viewer just drops the message it didn't subscribe to, so it's not harmful and why I don't think it needs fixing. If anything maybe just update the comment to acknowledge that this race case is also possible and benign.

clalancette:
> 5\. Attempt 2 receives a MessageData for a channel it never subscribed to.

Hm, yeah, this is definitely a race.  It seems *generally* benign, but I don't love it.  For now, I updated the comment in the latest push here, but I also added this to https://linear.app/foxglove/issue/FLE-475/followup-to-httpsgithubcomfoxglovefoxglove-sdkpull1194 to re-review.

PR #1195, opened 2026-04-27
remote_access: Only connect to room on-demand

gasmith:
### Changelog
None

### Docs
None

### Description
This change updates the SDK to use the new control plane APIs to only
create remote access sessions on-demand, when at least one viewer is
interested in joining.

### Testing
Tested against foxglove/app#14194

Note that `auth_remote_access_connection` is expected to fail in CI until foxglove/app#14191 reaches party.

### Links
Fixes: FLE-458
Fixes: FLE-440


rust/foxglove/src/remote_access/connection.rs
```diff
@@ -352,15 +363,269 @@ impl RemoteAccessConnection {
...
+        let Some(mut watch) = self.connect_watch(device_context, retry).await else {
+            return WatchLoopAction::Stop;
+        };
+        info!(
+            watch_lease_id = watch.lease_id(),
+            heartbeat_interval_ms = watch.hello().heartbeat_interval_ms,
+            device_wait_for_viewer_ms = watch.hello().device_wait_for_viewer_ms,
+            "watch stream established; dormant until wake"
+        );
         self.set_status(ConnectionStatus::Connected);
```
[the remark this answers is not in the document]
gasmith: Meh. We're deliberately keeping this a bit vague, but I'd argue that the semantics haven't meaningfully changed from a user's point of view. "Connected" did not previously mean that there were active participants. It meant exactly what we documented: that the gateway is "connected". Previously, that meant connected over WebRTC to an SFU. Now it means either that, or that we're idling waiting for a wake-up from the control plane. Either way, the meaning we intend to expose to users is the same: the gateway is authorized and ready to serve viewers.

rust/foxglove/src/remote_access/watch.rs
```diff
@@ -0,0 +1,621 @@
...
+                let hello: WatchHelloEvent =
+                    serde_json::from_str(&event.data).map_err(|e| WatchError::MalformedEvent {
+                        event: "hello",
+                        source: e,
+                    })?;
+                break hello;
+            }
+            debug!(event = %event.event, "ignoring unexpected event before hello");
+        };
+        if hello.heartbeat_interval_ms == 0 {
```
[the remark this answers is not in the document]
gasmith: We own and trust the API server.

rust/foxglove/src/remote_access/connection.rs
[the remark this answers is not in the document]
eloff: I did a double take as well, but we are connected, just to a different endpoint. Seems fine.

rust/foxglove/src/remote_access/watch.rs
eloff: We do some minimal validation of heartbeat_interval_ms here, but not device_wait_for_viewer_ms. We do own the server, but maybe a debug assert or check that it's a reasonable value would make the code more defensive, ditto with heartbeat_interval_ms. I'd hate to think we set it to 5 in the API and forget it's milliseconds or something silly like that and ddos ourselves.

rust/foxglove/src/remote_access/sse.rs
```diff
@@ -0,0 +1,279 @@
...
+            _ => {} // ignore id/retry and any unknown field
+        }
+    }
+
+    fn dispatch(&mut self) {
+        // Per the SSE spec, dispatch only if we have observed at least one data line in the
+        // current frame. This also ensures consecutive blank lines don't emit empty events.
+        if !self.has_data {
+            self.pending = SseEvent::default();
+            self.has_data = false;
```
eloff: nit: it's already false

rust/foxglove/src/remote_access/connection.rs
```diff
@@ -92,6 +102,30 @@ pub(crate) struct ConnectionParams {
...
 
+/// Pair of device metadata (fetched once via `fetch_device_info`) and the authenticated API
+/// client used for subsequent calls (watch stream, heartbeat). Initialized lazily on the first
+/// successful call to [`RemoteAccessConnection::get_or_init_device_context`].
+struct DeviceContext {
+    device: DeviceResponse,
+    client: Arc<FoxgloveApiClient<DeviceToken>>,
+}
+
+struct WatchRetryState {
```
eloff: nit: I feel like these 3 types belong better in watch.rs

eloff (approved):
Looks good, some nits only.

I think the watch loop behavior has some test gaps, but nothing worth holding up this PR.

rust/foxglove/src/remote_access/watch.rs
gasmith: Good call. I'll log & clamp both.

rust/foxglove/src/remote_access/connection.rs
gasmith: So... they don't quite belong in watch.rs... but I agree it'd be nice to get them out of connection.rs. I ended up adding watch_loop.rs to encapsulate the policy decisions about how various branches affect retries and so forth. I think it's an improvement, and also gives us a nice abstraction for tests.

rust/foxglove/src/remote_access/watch.rs
```diff
@@ -0,0 +1,691 @@
...
+            Duration::from_millis(hello.heartbeat_interval_ms).max(MIN_HEARTBEAT_INTERVAL);
+        let device_wait_for_viewer =
+            Duration::from_millis(hello.device_wait_for_viewer_ms).max(MIN_DEVICE_WAIT_FOR_VIEWER);
+
+        info!(
+            watch_lease_id = &hello.watch_lease_id,
+            ?heartbeat_interval,
+            ?device_wait_for_viewer,
+            "watch stream established"
+        );
```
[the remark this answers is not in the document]
gasmith: Decided not to log after all.. It's not like users care about a bogus advertisement from our API server. This certainly isn't something that's actionable for them.

rust/foxglove/src/remote_access/watch.rs
```diff
@@ -571,19 +637,19 @@ mod tests {
 
     #[tokio::test]
     async fn next_outcome_reports_stream_end() {
```
eloff: nit: The test names still reference next_outcome, but that method is gone and they call run now. Could be updated

eloff (commented):
Changes look good, one nit about test names

PR #1196, opened 2026-04-27
remote access: fix race in sdk if ParticipantActive arrives before Disconnected

eloff:
### Changelog

None

### Docs

None

### Description

Chris fixed a race in [clalancette/fix-failing-netem-test](https://github.com/foxglove/foxglove-sdk/pull/1194)

Which handles the case where Disconnected(S1) arrives after the replacement is already installed by the reset path. There's a similar race when ParticipantActive(S2) arrives before Disconnected(S1), so no reset has yet been triggered.

1) Attempt 1 registered (SID S1).
2) LiveKit emits ParticipantActive(S2) (or ParticipantConnected then Active) for attempt 2 before the disconnect event for attempt 1.
3) add_participant runs, sees self.registry.has_participant(&id) == true (attempt 1 is still there) → returns Ok(()) early.
4) LiveKit emits ParticipantDisconnected(S1) for attempt 1.
5) remove_participant(id, S1) → SID matches stored → attempt 1 is correctly removed.

Net result: nobody is registered for id, but attempt 2 is alive in the LiveKit room. We never come back to register it.

I added a test same_identity_reconnect_must_replace_prior_registration() which reproduces the issue, and then fixed it.

I'm not 100% sure this can happen in real life, it's possible livekit would always send disconnected first. But I have my doubts they do that or that it's even possible to always do that.

### DIFF
--- a/rust/foxglove/src/remote_access/participant_registry.rs
+++ b/rust/foxglove/src/remote_access/participant_registry.rs
@@ -513,9 +513,9 @@ mod tests {
         assert!(registry.remove_participant(&sid_1).is_none());
 
         // Step 4: attempt 2 must still be registered.
-        let current = registry
-            .get_participant(&id)
-            .expect("attempt 2 must remain registered after a late stale Disconnected for the prior SID");
+        let current = registry.get_participant(&id).expect(
+            "attempt 2 must remain registered after a late stale Disconnected for the prior SID",
+        );
         assert_eq!(current.participant_sid(), &sid_2);
         assert_ne!(current.client_id(), client_id_1);
     }
--- a/rust/foxglove/src/remote_access/participant_registry.rs
+++ b/rust/foxglove/src/remote_access/participant_registry.rs
@@ -49,14 +49,17 @@ impl ParticipantRegistry {
     }
 
     /// Spawns a [`Participant`] and its flush-task against the supplied
-    /// `writer` and inserts it into the registry.
+    /// `writer` and inserts it into the registry, atomically replacing any
+    /// prior registration for the same identity that has a **different**
+    /// `ParticipantSid`.
     ///
-    /// Each byte slice in `initial_messages` is queued on the participant's
-    /// control-plane channel after the flush-task is spawned but before the
-    /// participant is visible in the registry. This preserves the invariant
-    /// that these bytes (typically `ServerInfo` + channel/service
-    /// advertisements) are the first the flush-task delivers to the viewer,
-    /// ahead of any broadcast that reaches the participant after registration.
+    /// Each byte slice in `initial_messages` is queued on the new
+    /// participant's control-plane channel after the flush-task is spawned
+    /// but before the participant is visible in the registry. This preserves
+    /// the invariant that these bytes (typically `ServerInfo` +
+    /// channel/service advertisements) are the first the flush-task delivers
+    /// to the viewer, ahead of any broadcast that reaches the participant
+    /// after registration.
     ///
     /// `participant_sid` is stored on the new [`Participant`] and indexed in
     /// the registry so any later operation that targets a specific connection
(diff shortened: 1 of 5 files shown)


rust/foxglove/src/remote_access/participant_registry.rs
```diff
@@ -97,7 +117,35 @@ impl ParticipantRegistry {
             participant.send_control(msg);
         }
 
-        self.participants.write().insert(participant, flush_handle)
+        let mut participants = self.participants.write();
+
+        // Same-identity reconnect (different SID): atomically remove the prior
```
dave-foxglove, line 140: IIUC: the connections may arrive in any order. 

Imagine this: a connection is started (call it S1) and then connection S2 is started, but S2 arrives first then S1; if we replace S2 with S1 we will have covered/stomped the wanted connection (S2) with the unwanted connection (s1) because S1 arrived later. i.e. arriving later would not mean it's the correct one to keep.

a connection is started (call it S1) and then
  connection S2 is started, but s2 arrives first then s1; if we replace s2 with s1 we will have
  covered/stomped the wanted connection (s2) with the unwanted connection (s1) because s1 arrived later

dave-foxglove (changes requested):

rust/foxglove/src/remote_access/participant_registry.rs
dave-foxglove: Hoping this is helpful: 

Maybe this timestamp: https://github.com/livekit/protocol/blob/main/protobufs/livekit_models.proto#L179-L180

Something like:
```
if participant.joined_at() <= existing.joined_at() {
              // Stale session arrived late — drop it, keep the newer one.
              participant.cancel();
              return None;
          }
```

eloff: I think the difference is the same event should come ordered from the SFU, but two different events (the case this PR addresses: probably are not ordered.) That's making some assumptions though, and the defensive fix you recommend makes sense to me, so I've implemented it

rust/foxglove/src/remote_access/participant_registry.rs
```diff
@@ -49,50 +49,80 @@ impl ParticipantRegistry {
...
+    ///   registration is left untouched, the freshly-spawned participant
+    ///   is cancelled and dropped, and `None` is returned. This protects
+    ///   against an out-of-order `ParticipantActive` for a superseded
+    ///   instance overwriting a fresher registration.
+    /// - If a participant is registered with the **same** `participant_sid`,
+    ///   the existing registration is left untouched, the freshly-spawned
+    ///   participant is cancelled and dropped, and `None` is returned. This
+    ///   path is defensive: in practice LiveKit never reuses a SID across
+    ///   instances, so re-registering an identical `(identity, sid)` pair
+    ///   indicates a redundant call rather than a true reconnect.
```
clalancette, line 107: The way that this is currently done here seems to correctly avoid TOCTTOU problems, but on 2 of the 4 paths we spawn a Participant just to later drop it.  This isn't hugely expensive, but is kind of ugly.  I'm wondering if we can move the `let mut participants = self.participants.write();` to earlier in the function, do the checks to get out early in cases 3 and 4, and only do the Participant spawn in cases 1 and 2.

But this seems like a performance improvement more than anything else, so I'm not going to block on it.  I am going to add it to the follow-up ticket as something to consider.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -965,18 +965,58 @@ impl RemoteAccessSession {
...
+                    participant_identity = %participant_id,
+                    existing_sid = %existing.participant_sid(),
+                    existing_joined_at = existing.joined_at(),
+                    incoming_sid = %participant_sid,
+                    incoming_joined_at = joined_at,
+                    "skipping add_participant for stale instance (incoming joined_at precedes registered)",
+                );
+                return Ok(());
+            }
         }
```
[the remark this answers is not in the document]
eloff: I'm ok with this, the source of truth is in the registry under the writer lock.

Keeping this as-is is an optimization that doesn't matter, and makes the code a little more defensive. Not worth the churn to fix.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -975,18 +965,58 @@ impl RemoteAccessSession {
...
-        if self.participant_registry.has_participant(&participant_id) {
-            return Ok(());
+        // Gate on the registry *before* opening the stream: `stream_bytes`
+        // is an RPC that should not be wasted on an already-registered
+        // (identity, sid) pair, and equally not wasted on a stale incoming
+        // instance the registry would reject. A different-SID hit with a
+        // newer-or-equal `joined_at` means the registered instance is
+        // older and we must fall through to open a stream for the new
+        // instance.
+        if let Some(existing) = self.participant_registry.get_participant(&participant_id) {
```
clalancette, line 1014: There's still a TOCTTOU race between the `get_participant` here and `register_participant` below.  It ends up being benign because *if* something raced here, we would erroneously send it bytes via `stream_bytes` (which it would drop), and then `register_participant` would redo the checks anyway.  So I would say that this reduces the chance of, but does not eliminate, sending `stream_bytes` to an instance that we should not send it to.

clalancette (approved):

rust/foxglove/src/remote_access/session.rs
eloff: I think that's ok, but worth following up on later. I see you added it to the followup ticket.

rust/foxglove/src/remote_access/participant_registry.rs
eloff: Fixed

dave-foxglove (approved):

PR #1201, opened 2026-04-30
Tighten visibility in the remote_access module

clalancette:
### Changelog

None.

### Docs

None.

### Description

This came out of review of #1194 .  While doing that I noticed that we had a bunch of unnecessary visibility in these modules.  In particular:

1.  There were a few structs and functions which were never used outside of their file, yet marked as `pub` or `pub(crate)`.  Remove the visibility symbols to make them private.
2. There were a few structs where the type was `pub(crate)`, but the field were marked `pub`.  That doesn't make too much sense; something outside of the crate wouldn't be able to get to the type, so it can't get to the fields.  Change those to `pub(crate)`.
3. There were a number of places marked `pub(crate)` that can be tightened up to `pub(super)`.  Do that as well.

The end result here is basically no change to the binary, but just cleaner scope.

Part of FLE-475

### DIFF
--- a/rust/foxglove/src/remote_access/connection.rs
+++ b/rust/foxglove/src/remote_access/connection.rs
@@ -84,20 +84,20 @@ impl ConnectionStatus {
 ///
 /// This should be constructed from the [`crate::remote_access::Gateway`] builder.
 pub(crate) struct ConnectionParams {
-    pub name: Option<String>,
-    pub device_token: String,
-    pub foxglove_api_url: Option<String>,
-    pub foxglove_api_timeout: Option<Duration>,
-    pub listener: Option<Arc<dyn super::Listener>>,
-    pub capabilities: Vec<Capability>,
-    pub supported_encodings: Option<IndexSet<String>>,
-    pub fetch_asset_handler: Option<Arc<dyn AssetHandler<Client>>>,
-    pub runtime: Handle,
-    pub channel_filter: Option<Arc<dyn SinkChannelFilter>>,
-    pub qos_classifier: Option<Arc<dyn QosClassifier>>,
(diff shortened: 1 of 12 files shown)


rust/foxglove/src/remote_access/connection.rs
```diff
@@ -83,21 +83,21 @@ impl ConnectionStatus {
...
-    pub supported_encodings: Option<IndexSet<String>>,
-    pub fetch_asset_handler: Option<Arc<dyn AssetHandler<Client>>>,
-    pub runtime: Handle,
-    pub channel_filter: Option<Arc<dyn SinkChannelFilter>>,
-    pub qos_classifier: Option<Arc<dyn QosClassifier>>,
-    pub server_info: Option<HashMap<String, String>>,
-    pub message_backlog_size: Option<usize>,
-    pub context: Weak<Context>,
+pub(super) struct ConnectionParams {
+    pub(super) name: Option<String>,
```
gasmith, line 87: Nit: For struct fields in particular I feel like the reiteration of the struct's pub scope for each field is redundant, but I suppose that's really just a cosmetic wart.

gasmith (approved):
It's super!

rust/foxglove/src/remote_access/connection.rs
clalancette: My understanding (which may be wrong) is that the field scope is a separate concept from the struct scope.  That is, if you just had `pub(super) struct ConnectionParams` (with all fields just "default" visibility), consumers would know about the opaque type `struct ConnectionParams`, but would not be able to access any fields.  You need the visibility declaration for each field to have consumers access them.

PR #1206, opened 2026-04-30
remote_access: Split parameter subscriptions out of SessionState; rename to ChannelRegistry

clalancette:
### Changelog

None.

### Docs

None.

### Description

The SessionState struct mixed channel-keyed state (advertised channels, subscriptions, video tracks, QoS, client-advertised channels) with parameter-name-keyed subscription tracking. The two have independent lifecycles and the bundling made the struct hard to name.

Splits the parameter-subscription bookkeeping into a new ParameterSubscriptions struct (in parameter_subscriptions.rs) and renames the rest to ChannelRegistry (in channel_registry.rs). The struct's docstring is also corrected — it was previously labeled a "state machine" despite having no states or transitions.

RemoteAccessSession now holds two RwLocks (channel_registry and parameter_subscriptions) instead of one. No call site holds both at once, so there is no lock-ordering or deadlock concern.

@gasmith Take particular note, since you suggested this renaming.  I'm not 100% sure that this split is better than what we have today, so opinions definitely welcome (including: close this, we don't need it).

Part of FLE-475

### DIFF
--- a/rust/foxglove/src/remote_access.rs
+++ b/rust/foxglove/src/remote_access.rs
@@ -1,10 +1,12 @@
 //! Remote access implementation.
 
 mod capability;
+mod channel_registry;
 mod client;
 mod connection;
 mod gateway;
 mod listener;
+mod parameter_subscriptions;
 mod participant;
 mod participant_registry;
 mod participants;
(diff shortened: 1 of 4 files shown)


gasmith (approved):
I think this is worthwhile, though I'm open to other opinions.

clalancette:
> I think this is worthwhile, though I'm open to other opinions.

There have been no dissenting opinions, so I'm going to rebase this onto main one more time to run CI, then go ahead and merge.

PR #1207, opened 2026-04-30
python: Update remote access example to depend on v0.23.0

gasmith:
### Changelog
None

### Docs
None

### Description
Update the python remote access example to use the latest SDK release.

Also fix an AttributeError in the example.

### Testing
`uv run main.py`

### Links
Fixes: FLE-470

### DIFF
--- a/python/foxglove-sdk-examples/remote-access/main.py
+++ b/python/foxglove-sdk-examples/remote-access/main.py
@@ -28,7 +28,7 @@ class Listener(RemoteAccessListener):
     def on_connection_status_changed(
         self, status: RemoteAccessConnectionStatus
     ) -> None:
-        logging.info(f"Connection status: {status.name}")
+        logging.info(f"Connection status: {status}")
 
     def on_message_data(
         self,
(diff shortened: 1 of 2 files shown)


gasmith:
> Two-line change is straightforward. Confirmed why `.name` was broken: `RemoteAccessConnectionStatus` is generated via pyo3 `#[pyclass(eq, eq_int)]` (python/foxglove-sdk/src/remote_access.rs:40-57), which produces an int-backed class, not a `enum.Enum` subclass — so `.name` raises `AttributeError` at runtime. Dropping it and relying on `__repr__` is the right call.
> 
> Side note (not for this PR): the stub at python/foxglove-sdk/python/foxglove/_foxglove_py/remote_access.pyi:56 declares `class RemoteAccessConnectionStatus(Enum)`, which is what tricked the example into using `.name` in the first place. Worth fixing the stub (or the runtime class) separately so static type-checkers stop encouraging the same mistake.

Filed https://linear.app/foxglove/issue/FLE-486/add-name-and-value-properties-to-enum-classes-in-python

dave-foxglove (approved):
IIUC this is a sdk version update and a more verbose log message. lgtm.

PR #1210, opened 2026-05-05
remote access: don't publish data tracks for video channels

eloff:
Fixes: [FLE-503](https://linear.app/foxglove/issue/FLE-503)


eloff:
Actually this is not an issue. The app can subscribe to video tracks via request_video_track: false, and we publish data tracks eagerly to simplify the implementation.

We could maybe complicate life by trying to lazily publish data tracks (or do that just for video channels - seems the worst of both worlds though because it still needs all the complexity, but rarely exercised and likely to break)

PR #1211, opened 2026-05-05
foxglove_bridge: Disable remote access for glibc<2.35

gasmith:
### Changelog
- foxglove_bridge: Disable remote access for platforms with glibc<2.35

### Docs
None

### Description
Apparently the ROS build farm also builds the bridge for RHEL9, which
has glibc 2.34, and is not compatible with the remote-access-enabled
builds of the SDK.

This change updates the CMakeLists.txt for the bridge to introspect the
version of glibc and use that as the basis for choosing the default
value for the FOXGLOVE_BRIDGE_REMOTE_ACCESS option. When remote access
is disabled, we link the bridge against the staticlib.

### Links
Fixes: FLE-507

### Testing
- Built the bridge in a rockylinux:9 container against ROS jazzy,
  validated that the build shows the "Detected glibc 2.34..." message,
  and builds successfully against libfoxglove.a.
- Built the bridge in the foxglove-sdk-ros-humble container without
  specifying FOXGLOVE_BRIDGE_REMOTE_ACCESS, validated that the default
  value is ON, and that the bridge links against libfoxglove.so.

### DIFF
--- a/ros/src/foxglove_bridge/CMakeLists.txt
+++ b/ros/src/foxglove_bridge/CMakeLists.txt
@@ -247,10 +247,9 @@ install(TARGETS foxglove_bridge_component
   LIBRARY DESTINATION lib
   RUNTIME DESTINATION bin
 )
-# Install the SDK shared libraries alongside the bridge component library so
-# the runtime linker can find them via RPATH ($ORIGIN). When remote access is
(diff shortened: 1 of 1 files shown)


clalancette (approved):

clalancette:
By the way, if you rebase onto "main", CI will now be green for you.

PR #1221, opened 2026-05-07
Series of small cmake cleanups

clalancette:
### Changelog

None.

### Docs

None.

### Description

This is a rollup of a bunch of small CMake cleanup things I noticed as I was looking through the code.  None of these are required, but they do make our use of CMake a bit more idiomatic.  If we don't want the tradeoffs these expose, then closing this is also perfectly fine.

1.  Replace the source glob in ros/src/foxglove_bridge/CMakeLists.txt with an explicit list.  The glob is nice because it is somewhat lower maintenance, but it comes with a cost; adding or removing files will *not* automatically re-run CMake for you.  That means that when you add a new file, you have to clean out the build directory and re-run from scratch.  It is a tradeoff, but it is more idiomatic to list things out.
2. Same thing for cpp/CMakeLists.txt, with the same tradeoff.
3. Drop headers from cpp/CMakeLists.txt targets.  Headers are not directly involved in building an object file, so they shouldn't be listed for them.  Changes to headers will still be picked up through the `-MD` rules generated by the compiler.  The tradeoff here is that these won't show up as part of the target in an IDE, but I think that is less relevant nowadays.
4. Get rid of an unnecessary variable now that it only has one item.
5. Factor the SDK source prefix out of the bridge.  This is just a little cleaner.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -235,7 +235,6 @@ endif()
 
 # Generated code (excluded from clang-tidy)
 set(foxglove_generated_srcs
-    "foxglove/include/foxglove/messages.hpp"
     "foxglove/src/messages.cpp"
 )
 add_library(foxglove_generated_obj OBJECT "${foxglove_generated_srcs}")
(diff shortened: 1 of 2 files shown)


clalancette:
After discussion, we decided to somewhat change how we are distributing our C++ library.  While this will still be relevant in that new way of doing things, it will change.  So marking this as draft for now.

clalancette:
I've been working on a more substantial rewrite here, so I'm actually going to close this out completely and go with that one instead.

PR #1222, opened 2026-05-07
ci: Add msg-schema-stability check to flag breaking ROS .msg edits

clalancette:
### Changelog

None.

### Docs

None.

### Description

We recently had a problem where we somewhat unintentionally broke the message schema (see #965).  This is problematic particular for stable ROS distributions, which usually expect messages to be stable throughout their lifetime (this isn't a hard requirement, but it is nicer to your users).

To make sure we really, really mean to do this, this PR adds a new workflow.  It detects modifications or deletions to schemas/ros1/*.msg and schemas/ros2/*.msg in PRs. Pure additions don't trigger the check. The 'breaking:msg-schema' label overrides the gate when the change is intentional and approved.

If we approve of this direction, then there are some things we need to do:
1.  We'll need to agree on the name of the override label.  I currently have it as `breaking:msg-schema`, but I really don't care what it is called.
2.  We'll need to create a new label with that name so we can use it as an override.
3.  We need to decide whether we only want this to trigger on the ROS message schemas, or all of them (like protobuf).  I chose ROS only here because it is basically a canary for everything else.  But if someone thinks we should trigger on all schema changes, that is easy to add.
4. We'll need to add this workflow to the "required-checks" for the repository.

Opinions welcome!

Fixes FLE-491

### DIFF
--- a/.github/workflows/schema-stability.yml
+++ b/.github/workflows/schema-stability.yml
@@ -1,0 +1,61 @@
+name: Schema stability
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+
+jobs:
+  msg-schema-check:
+    name: msg-schema-check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Check for modified or deleted ROS .msg files
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+          HAS_OVERRIDE: ${{ contains(github.event.pull_request.labels.*.name, 'breaking:msg-schema') }}
+        run: |
+          set -euo pipefail
+
+          # --diff-filter=MDR: Modified, Deleted, or Renamed. Pure additions of
+          # new .msg files don't break existing downstream consumers, but renames
+          # do (a consumer importing Foo.msg breaks when it becomes Bar.msg).
+          # --name-status prints M/D/R prefixes so the summary distinguishes
+          # them, and renames show both old and new paths on one line.
+          changed=$(git diff --name-status --diff-filter=MDR \
+            "$BASE_SHA"..."$HEAD_SHA" \
+            -- 'schemas/ros1/*.msg' 'schemas/ros2/*.msg')
+
+          if [ -z "$changed" ]; then
+            echo "No modified or deleted ROS .msg files."
+            exit 0
+          fi
+
+          {
+            echo "## Modified or deleted ROS .msg files"
+            echo
+            echo '```'
+            echo "$changed"
+            echo '```'
+          } >> "$GITHUB_STEP_SUMMARY"
+
+          if [ "$HAS_OVERRIDE" = "true" ]; then
+            {
+              echo
+              echo "Override label \`breaking:msg-schema\` is present — allowing intentional schema break."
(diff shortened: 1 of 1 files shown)


.github/workflows/schema-stability.yml
```diff
@@ -0,0 +1,59 @@
...
+        run: |
+          set -euo pipefail
+
+          # --diff-filter=MD: only Modified or Deleted files. Pure additions of
+          # new .msg files don't break existing downstream consumers and don't
+          # need an override.
+          changed=$(git diff --name-only --diff-filter=MD \
+            "$BASE_SHA"..."$HEAD_SHA" \
+            -- 'schemas/ros1/*.msg' 'schemas/ros2/*.msg')
+
```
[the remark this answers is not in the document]
clalancette: This can't be a problem, since ROS does not support subdirectories for msgs.

.github/workflows/schema-stability.yml
```diff
@@ -0,0 +1,59 @@
+name: Schema stability
```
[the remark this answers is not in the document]
clalancette: I'm inclined to leave it as it is.  It is narrowly scoped, but broadly named, which fits the "ROS-is-a-canary" in my opening description.

.github/workflows/schema-stability.yml
```diff
@@ -0,0 +1,61 @@
...
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Check for modified or deleted ROS .msg files
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+          HAS_OVERRIDE: ${{ contains(github.event.pull_request.labels.*.name, 'breaking:msg-schema') }}
+        run: |
```
gasmith, line 21: Normally, I'd suggest that we make this a script that we can run outside of CI, so that we can test it independently. In this case, since it's really just a minimal wrapper around `git diff`, I'm not necessarily opposed to leaving it as it is.

.github/workflows/schema-stability.yml
```diff
@@ -0,0 +1,61 @@
+name: Schema stability
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
```
gasmith, line 5: Might also consider adding `paths: [ 'schemas/ros/**' ]` so that this workflow only runs for PRs that touch files under those paths.

gasmith (approved):
LG, and the choice of label seems fine. A few minor comments, but nothing blocking.

.github/workflows/schema-stability.yml
clalancette: Done in 4f761b9

.github/workflows/schema-stability.yml
clalancette: Yeah, I'm inclined to leave this as-is since you can just do it with git diff.

.github/workflows/schema-stability.yml
gasmith: Hm, this probably isn't worth it. This action is basically just doing a git diff, so it's kinda redundant. My suggestion was to add the one-liner in the hopes we could prevent the workflow from even running in the first place. But I suppose that's incompatible with status checks... Whoops.

Let's go back to what you had. We don't need the path gate. Sorry for the misguidance!

clalancette: Reverted now.

PR #1226, opened 2026-05-08
Add a flush method to MCAP writer

gasmith:
### Changelog
- Added a flush method to MCAP writer

### Docs
None

### Description
This change adds a flush method to the MCAP writer, which can be used to
finish the current chunk and flush it to disk. The lack of this method was
simply oversight; it should've been there from the beginning.

### Testing
Added unit tests.

### Links
Fixes: FLE-514


clalancette (approved):
Looks reasonable to me.

PR #1229, opened 2026-05-11
python: Package for abi3-py310

gasmith:
### Changelog
- Python wheels are now packaged against abi3-py310

### Docs
None

### Description
We've run out of space in our pypi project. Our releases are currently
~650MiB, and we should reduce that.

This change updates CI to build a single abi3-py310 wheel per platform,
rather than building one for each specific version of python (cp310,
cp311, cp312, etc.). This change alone reduces our release size by
~500 MiB.

### Testing
Ran the remote-access SDK example using the abi3 wheel built in CI.

### Links
Fixes: FLE-524

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -259,7 +259,7 @@ jobs:
         uses: PyO3/maturin-action@v1
         with:
           target: ${{ matrix.platform.target }}
-          args: --release --out dist --find-interpreter ${{ matrix.platform.extra-args }}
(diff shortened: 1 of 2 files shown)


dave-foxglove (approved):
I'm not an expert on the syntax here, but the intent I glean from the code LGTM, approving based on direction (not whether the syntax is correct).

eloff (approved):
Pleasantly surprised by how simple the change was, looks good

PR #1230, opened 2026-05-11
python: Don't package for atypical/legacy arches

gasmith:
### Changelog
- Python wheels are no longer built for ppc64le/s390x/x86

### Docs
None

### Description
We've run out of space in our pypi project. Our releases are currently
~650MiB, and we should reduce that.

This change updates CI to stop packaging for architectures that our
users are unlikely to use:

- ppc64le: Use for HPC
- s390x: IBM mainframes
- x86: 32-bit is so last century

Taken in addition to [the previous change](https://github.com/foxglove/foxglove-sdk/pull/1229) to target abi3, this change
reduces our release size by another ~80 MiB, bringing us down to about
~100 MiB per release.

### Links
Fixes: FLE-525

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -259,7 +259,7 @@ jobs:
         uses: PyO3/maturin-action@v1
         with:
           target: ${{ matrix.platform.target }}
-          args: --release --out dist --find-interpreter ${{ matrix.platform.extra-args }}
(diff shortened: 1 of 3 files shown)


dave-foxglove (approved):
Reducing built platforms to focus on the most used lgtm.

eloff (approved):

PR #1237, opened 2026-05-12
Fix broken link and markdown formatting in C++ README

kimjune01:
## Summary

- Fixed the relative link to the rgb-camera-visualization example in `cpp/README.md`. The link used `cpp/examples/...` but since it is already inside the `cpp/` directory, the correct path is `examples/...`.
- Fixed a markdown formatting typo in the example's README where `*Ubuntu/Debian:**` was missing the opening double asterisk, rendering as literal asterisks instead of bold text.

## Test plan

- [ ] `python3 test_docs_links.py` passes (verifies link target exists and formatting is correct)
- [ ] Rendered markdown displays the link and bold text correctly on GitHub

### DIFF
--- a/cpp/README.md
+++ b/cpp/README.md
@@ -95,7 +95,7 @@ The gateway-related C declarations in `foxglove-c/foxglove-c.h` are guarded by `
 
 ### RGB Camera Visualization Example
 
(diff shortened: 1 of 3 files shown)


kimjune01:
Cannot validate on current setup, so drafting. Please close or take it over.

sofuture:
Took over in #1281

PR #1240, opened 2026-05-13
[cpp]: Easy cleanups on the CMakeLists.txt

clalancette:
### Changelog

None.

### Docs

None.

### Description

As part of trying to cleanup how we build and install the C++ library with remote-access, Claude came up with a bunch of fixes.  The two in here should be uncontroversial:

1.  Fix Windows DLL and import install location - I honestly don't care much about this, but anytime I talk to Claude about this CMakeLists.txt it really wants to fix it.  So fine, we can fix Windows.
2. Skip the fetch of Corrosion when we are using a prebuilt library directory - just saves downloading a dependency that will never be used when we are using a prebuilt library.

Part of FLE-520

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -549,24 +549,22 @@ else()
   )
 endif()
 
-# Install the Rust/C shared library. POSIX (.so/.dylib) goes to LIBDIR; Windows (.dll) goes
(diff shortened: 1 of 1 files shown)


gasmith (approved):
I'm kinda tempted to bury a phone-home in here for the first user that attempts to actually build this on windows...

PR #1241, opened 2026-05-14
Update the foxglove_bridge dependencies

clalancette:
### Changelog

None.

### Docs

None.

### Description

While I was looking at revamping how the C++ installation works along with ROS, I noticed a number of small problems in the ROS foxglove_bridge package.  This PR cleans them up.  In particular:

1.  Adds in the `--no-warn-unused-cli` flag when calling `colcon build`, just to quiet down some warnings.
2.  Moves the `nlohmann_json` dependency from a build_depend to a test_depend, reflecting its actual use in the foxglove_bridge package.
3.  Removes the Boost ASIO configuration option.  This was broken (it never linked in Boost::Asio), and nobody has complained about it.
4.  Adds dependencies on `ament_index_cpp`, `rcl_interfaces`, `rcpputils`, `rcutils`, and `rosidl_typesupport_introspection_cpp`.  These were all transitively included before, this makes them explicit.
5.  Removes the dependency on OpenSSL by switching to a no-TLS websocketpp variant.  This is fine because this is only used for tests.

Part of FLE-520


gasmith (approved):
LG!

PR #1242, opened 2026-05-14
cpp+ros: Ship CMake package config in the dist; consume it from the bridge

clalancette:
### Changelog

None.

### Docs

None.

### Description

It's important to realize that the C++ package can have two artifacts; the "install", and the "dist".  The "install" is what happens when you run "cmake build/cmake install" locally, and you end up with a set of header files, C .so/.a files, C++ .so/.a files, and CMake glue to find it.  Unfortunately, that doesn't work for our distribution, since we want that to work cross-platform and the C++ ABI isn't guaranteed to be stable.  Prior to this PR the "dist" consists of the header files, C .so/.a, and C++ source files that are compiled on the target platform.

The dist now includes lib/cmake/foxglove-sdk/foxglove-sdkConfig.cmake so consumers can wire it up via `find_package(foxglove-sdk CONFIG REQUIRED HINTS /path/to/dist)` instead of FetchContent + GLOB + hand-rolled cmake. The ROS bridge is the first consumer.

The config file (cpp/cmake/foxglove-sdk-dist-config.cmake) is hand-written (not generated by configure_package_config_file) and Container.mk's build-cpp-dist cp's it verbatim into the dist tree. It is separate from cpp/cmake/foxglove-sdkConfig.cmake.in (used by `cmake --install` for local installs) because the two products ship different surfaces — the local install ships a prebuilt foxglove_cpp_shared/_static via install(EXPORT); the dist ships C++ wrapper sources and asks consumers to compile them against their own toolchain.

The dist config defines:
- foxglove-static IMPORTED target for libfoxglove.a (always non-RA), carrying the platform link libs the C lib's symbol surface depends on (Bcrypt etc. on Windows, Security/CoreFoundation on macOS, pthread/dl /m for older glibc on Linux).
- foxglove-shared IMPORTED target for libfoxglove.so (always RA), with IMPORTED_IMPLIB pointing at foxglove.dll.lib on Windows.
- FOXGLOVE_SDK_CPP_SOURCES + FOXGLOVE_SDK_CPP_REMOTE_ACCESS_SOURCE variables listing the wrapper sources shipped under src/.
- A foxglove_sdk_add_cpp_library(<name> [TYPE STATIC|SHARED] [REMOTE_ACCESS ON|OFF]) helper that compiles the wrapper sources and links the appropriate C library. The include dirs are propagated with SYSTEM PUBLIC so consumers' strict-warnings builds don't flag SDK headers.

Supported (TYPE, REMOTE_ACCESS) combinations:
- SHARED + ON  → dynamic-links libfoxglove.so (RA), adds remote_access.cpp, defines FOXGLOVE_REMOTE_ACCESS
- SHARED + OFF → statically bundles libfoxglove.a (non-RA) into the SHARED wrapper (the bridge's non-RA case)
- STATIC + OFF → static wrapper, propagates libfoxglove.a to consumers
- STATIC + ON  → unsupported (no static RA lib by design; would leak LiveKit/WebRTC symbols into a static consumer's binary)

The bridge (ros/src/foxglove_bridge/CMakeLists.txt) drops its FetchContent + GLOB + manual add_library/target_sources/target_link block in favor of `find_package(foxglove-sdk CONFIG)` + foxglove_sdk_add_cpp_library(foxglove_cpp_shared TYPE SHARED REMOTE_ACCESS ON|OFF). The FETCHCONTENT_SOURCE_DIR_FOXGLOVE_SDK cmake variable overrides the SDK source dir for development against locally-built dists; otherwise the existing FetchContent path is preserved (currently broken until a release ships with the cmake glue — flagged in a comment).

As preparation, cpp/CMakeLists.txt's three file(GLOB) / file(GLOB_RECURSE) calls are replaced with explicit set() lists for foxglove_cpp_srcs, foxglove_test_srcs, and foxglove_integration_test_srcs. New source files have to be added to the list deliberately. The same explicit list is mirrored in the dist config's FOXGLOVE_SDK_CPP_SOURCES variable.

No version file (foxglove-sdkConfigVersion.cmake) is shipped yet; consumers omit the VERSION arg to find_package. Adding it is a follow-up if downstream wants to version-pin via cmake rather than the existing release-zip URL.

Fixes FLE-520


cpp/CMakeLists.txt
```diff
@@ -344,12 +348,15 @@ endif()
...
-file(GLOB_RECURSE foxglove_test_srcs CONFIGURE_DEPENDS
-    "foxglove/tests/*.hpp"
-    "foxglove/tests/*.cpp"
+set(foxglove_test_srcs
+    "foxglove/tests/test_arena.cpp"
+    "foxglove/tests/test_channel.cpp"
+    "foxglove/tests/test_mcap.cpp"
+    "foxglove/tests/test_messages.cpp"
+    "foxglove/tests/test_parameter.cpp"
+    "foxglove/tests/test_system_info.cpp"
```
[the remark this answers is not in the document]
clalancette: This is on purpose, so I'm not going to fix it.

ros/src/foxglove_bridge/CMakeLists.txt
```diff
@@ -82,6 +82,14 @@ if (ENDIAN)
...
 
+# Fetch the released SDK zip from GitHub. To build against a locally-modified SDK,
+# override FETCHCONTENT_SOURCE_DIR_FOXGLOVE_SDK (cmake's standard FetchContent override)
+# to point at a `cpp/dist`-shaped directory — the ros/Makefile's FOXGLOVE_CPP_SDK_DIR
+# variable does this for you (see ros/README.md). The dist directory needs to contain
+# lib/cmake/foxglove-sdk/foxglove-sdkConfig.cmake; release zips built before the SDK's
+# cmake-glue change don't yet, so the FetchContent download path is broken until a
+# release ships with that glue. For local development, point the override at a
+# `make build-cpp-dist`-produced tree.
 set(FOXGLOVE_SDK_VERSION "0.23.1")
```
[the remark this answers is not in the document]
gasmith: I'm on the fence. While I don't think anyone depends on being able to build untagged versions of the bridge this way, it wouldn't hurt much to split this into two changes, and do an SDK release in between.

cpp/cmake/foxglove-sdk-dist-config.cmake
```diff
@@ -0,0 +1,168 @@
...
+
+# Records the actual remote-access flavor of the shipped cdylib. Container.mk's
+# build-cpp-dist writes this file alongside this config; it sets
+# FOXGLOVE_SDK_CDYLIB_REMOTE_ACCESS to ON or OFF to reflect whether libfoxglove.so
+# was built with --features remote-access. The helper below uses it to reject
+# REMOTE_ACCESS=ON requests against a non-RA dist instead of silently producing
+# a wrapper that links the wrong cdylib.
+include("${CMAKE_CURRENT_LIST_DIR}/dist-flavor.cmake" OPTIONAL RESULT_VARIABLE _foxglove_flavor_path)
+if(NOT _foxglove_flavor_path)
+  # Older dists predate the marker; assume RA, matching the historical default.
```
gasmith: This comment feels a bit odd. Is it possible that this cmake file would be used against an old build?

Container.mk
```diff
@@ -135,3 +135,17 @@ build-cpp-dist:
 	cp -R c/include/foxglove-c $(CPP_SDK_DIR)/include/
 	cp -R cpp/foxglove/include/foxglove $(CPP_SDK_DIR)/include/
 	cp -R cpp/foxglove/src/* $(CPP_SDK_DIR)/src/
+	# CMake glue. The config file is hand-written (not generated by cmake) and shipped
```
gasmith, line 138: This makefile target is getting pretty complicated. How would you feel about pulling all of this (not just the new additions) out into a bash script?

ros/src/foxglove_bridge/CMakeLists.txt
```diff
@@ -102,34 +110,23 @@ FetchContent_Declare(
...
+# The dist's CMake package config defines IMPORTED targets for the prebuilt C libraries
+# (foxglove-static, foxglove-shared) and the foxglove_sdk_add_cpp_library() helper, which
+# compiles the SDK's C++ wrapper sources against our toolchain. find_package looks for
+# the config at <sdk>/lib/cmake/foxglove-sdk/foxglove-sdkConfig.cmake.
+find_package(foxglove-sdk CONFIG REQUIRED HINTS "${foxglove_sdk_SOURCE_DIR}")
+
+# Build the C++ wrapper as a SHARED library either way: with RA, dynamic-link the RA
+# cdylib (libfoxglove.so); without RA, statically bundle the non-RA staticlib
+# (libfoxglove.a) into the wrapper so the C symbols travel inside the wrapper.so and the
+# bridge component doesn't need a separate runtime dep.
```
gasmith: The "so the C symbols travel inside the wrapper.so and the bridge component doesn't need a separate runtime dep" part of this comment surprised me a little. Do we care about this detail?

cpp/cmake/foxglove-sdk-dist-config.cmake
```diff
@@ -0,0 +1,168 @@
...
+#   - SHARED + OFF → statically bundles libfoxglove.a (non-RA) into the SHARED wrapper
+#   - STATIC + OFF → static wrapper, propagates libfoxglove.a (non-RA) to consumers
+#   - STATIC + ON  → unsupported; the dist does not ship a static RA library because
+#                    LiveKit/WebRTC has strict ABI requirements that would leak into
+#                    static consumers' binaries.
+function(foxglove_sdk_add_cpp_library name)
+  cmake_parse_arguments(_arg "" "TYPE;REMOTE_ACCESS" "" ${ARGN})
+  if(NOT _arg_TYPE)
+    set(_arg_TYPE SHARED)
+  endif()
```
gasmith, line 93: When `TYPE` is provided, we should probably validate that it is either "SHARED" or "STATIC".

cpp/cmake/foxglove-sdk-dist-config.cmake
```diff
@@ -0,0 +1,168 @@
...
+# C++ wrapper source list, derived from the shared foxglove-sources.cmake file
+# (also used by the in-tree build). The dist concatenates handwritten + generated
+# because, unlike the in-tree build, consumers compile them all into one wrapper.
+include("${CMAKE_CURRENT_LIST_DIR}/foxglove-sources.cmake")
+set(FOXGLOVE_SDK_CPP_SOURCES
+  ${FOXGLOVE_CPP_HANDWRITTEN_SOURCES}
+  ${FOXGLOVE_CPP_GENERATED_SOURCES}
+)
+list(TRANSFORM FOXGLOVE_SDK_CPP_SOURCES PREPEND "${_foxglove_src_dir}/")
+set(FOXGLOVE_SDK_CPP_REMOTE_ACCESS_SOURCE ${FOXGLOVE_CPP_REMOTE_ACCESS_SOURCES})
```
gasmith: Nit: "SOURCES" plural?

ros/src/foxglove_bridge/CMakeLists.txt
```diff
@@ -102,34 +110,23 @@ FetchContent_Declare(
...
-    $<BUILD_INTERFACE:${foxglove_sdk_SOURCE_DIR}/include>
-    $<INSTALL_INTERFACE:include>
-)
-target_sources(foxglove_cpp_shared PRIVATE ${FOXGLOVE_SDK_SOURCES})
-# Import the prebuilt C library from the fetched SDK.
+# The dist's CMake package config defines IMPORTED targets for the prebuilt C libraries
+# (foxglove-static, foxglove-shared) and the foxglove_sdk_add_cpp_library() helper, which
+# compiles the SDK's C++ wrapper sources against our toolchain. find_package looks for
+# the config at <sdk>/lib/cmake/foxglove-sdk/foxglove-sdkConfig.cmake.
+find_package(foxglove-sdk CONFIG REQUIRED HINTS "${foxglove_sdk_SOURCE_DIR}")
```
gasmith, line 118: I'm noticing that we've named `foxglove-sdk` with kebab case, whereas other packages below are named with snake case (`ament_cmake`, `ament_index_cpp`, etc.). Should we also name our component with snake case for consistency?

cpp/CMakeLists.txt
```diff
@@ -223,25 +223,17 @@ else()
...
+include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/foxglove-sources.cmake")
+
 # Generated code (excluded from clang-tidy)
-set(foxglove_generated_srcs
-    "foxglove/include/foxglove/messages.hpp"
-    "foxglove/src/messages.cpp"
-)
+set(foxglove_generated_srcs ${FOXGLOVE_CPP_GENERATED_SOURCES})
+list(TRANSFORM foxglove_generated_srcs PREPEND "foxglove/src/")
+list(APPEND foxglove_generated_srcs "foxglove/include/foxglove/messages.hpp")
```
gasmith: Nit: Having this header (and remote_access.hpp below) called out explicitly here feels a bit at-odds with the explicit source file listings in foxglove-sources.cmake.

cpp/README.md
```diff
@@ -91,6 +91,55 @@ Link against `foxglove_cpp_shared` and include the same C and C++ headers as the
...
+src/                                               # C++ wrapper sources
+```
+
+### Consuming the dist via CMake
+
+```cmake
+find_package(foxglove-sdk CONFIG REQUIRED HINTS /path/to/dist)
+foxglove_sdk_add_cpp_library(my_wrapper
+    TYPE SHARED            # or STATIC
+    REMOTE_ACCESS ON       # or OFF
```
gasmith: Might want to document the defaults here.

gasmith (approved):
I tested using [the quickstart example](https://docs.foxglove.dev/docs/sdk/example?lang=cpp), and it works great. This is a huge improvement!

```cmake
FetchContent_Declare(
  foxglove_sdk
  ...
)
FetchContent_MakeAvailable(foxglove_sdk)

find_package(foxglove-sdk CONFIG REQUIRED HINTS "${foxglove_sdk_SOURCE_DIR}")
foxglove_sdk_add_cpp_library(foxglove TYPE STATIC)

add_executable(quickstart main.cpp)
target_link_libraries(quickstart PRIVATE foxglove)
```

cpp/cmake/foxglove-sdk-dist-config.cmake
clalancette: Yeah, good point.  I let Claude get away from me here.  Actually simplified and fixed in 5647877e6f89675cda5080eb3c9c311710ac3d7e

ros/src/foxglove_bridge/CMakeLists.txt
clalancette: Yeah, this was too complicated.  I simplified it in 5647877e6f89675cda5080eb3c9c311710ac3d7e

cpp/cmake/foxglove-sdk-dist-config.cmake
clalancette: Added a check in 5647877e6f89675cda5080eb3c9c311710ac3d7e

cpp/cmake/foxglove-sdk-dist-config.cmake
clalancette: Changed to "SOURCES" in 5647877e6f89675cda5080eb3c9c311710ac3d7e

cpp/CMakeLists.txt
clalancette: Yeah, I missed this.  Removed in 5647877e6f89675cda5080eb3c9c311710ac3d7e to be consistent.

cpp/README.md
clalancette: Done in 5647877e6f89675cda5080eb3c9c311710ac3d7e

Container.mk
clalancette: My personal take is that it is currently long, but not too complicated.  So I wouldn't bother.  That said, it *might* make more sense to put the meat of this in cpp/Makefile, since it is, after all, C++ specific.  But I'd defer that to another PR since this one is getting pretty large already.

ros/src/foxglove_bridge/CMakeLists.txt
clalancette: I'm on the fence.  The ROS convention is indeed snake case.  But CMake doesn't care which one you use, and there are plenty of packages in the ecosystem that use conventions different from ROS.  It won't look odd in a ROS package.  Given that, and given that we largely refer to it as "foxglove-sdk", I'd stick with what we have.

ros/src/foxglove_bridge/CMakeLists.txt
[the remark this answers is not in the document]
clalancette: Up to you, I'm fine either way.  If we want to split, then I'd make this PR the changes to cpp/, and then open a new PR for the ROS changes.  Let's discuss it a little later today.

ros/src/foxglove_bridge/CMakeLists.txt
gasmith: I'm fine with kebabs!

Container.mk
gasmith: Right, I meant "long". It's a super minor thing, also. I don't mind if we keep it as it is.

ros/src/foxglove_bridge/CMakeLists.txt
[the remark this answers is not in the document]
gasmith: The one thing I'll mention is that we're gonna need a new PR anyway when we do the SDK version bump here. But either way is fine with me.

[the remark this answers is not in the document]
clalancette: We decided to just merge this as-is, do the SDK release, then bump the ROS version.  So I'm going to go ahead and merge this.

PR #1244, opened 2026-05-14
remote_access: Close Sink::log same-identity reconnect race

clalancette:
### Changelog

None.

### Docs

None.

### Description

Sink::log previously read channel subscriber identities under the channel_registry lock, dropped it, and then resolved those identities to Arc<Participant>s under participant_registry. The gap between the two operations let a complete remove-then-add cycle for the same identity slip in: cleanup of attempt 1 cleared its channel subscriptions, attempt 2 registered with the same identity, and the late resolve handed the message to attempt 2 — which had never subscribed to the channel.

The race was benign in practice since the viewer drops MessageData on unsubscribed channels.  But we can fix it completely.  LiveKit assigns a fresh ParticipantSid per connection instance, so re-keying the maps by SID makes a stale snapshot resolve to None instead of crossing over to a different connection:

* ChannelRegistry::subscriptions / video_subscribers are SID-keyed; lookup helpers renamed *_identities → *_sids. Sink::log resolves under a single channel_registry.read() and a separate participant_registry lookup.
* ChannelRegistry::client_channels is SID-keyed. cleanup_for_removed_participant no longer needs an identity argument.
* ParameterSubscriptions is SID-keyed; cleanup_for_removed_identity → cleanup_for_removed_participant to match.

SID-keying introduces a new failure mode in the reverse direction: if a reordered ParticipantActive for a same-identity reconnect runs add_participant's cleanup of SID1 between an in-flight client message's get_participant(identity) and its acquisition of subscription_lock, an insert path could reinsert SID1 entries that nothing will ever sweep again (until session shutdown). Four handlers (handle_client_subscribe, handle_client_advertise, handle_subscribe_parameter_updates, handle_client_message_data) re-check via a small participant_already_swept helper and early-return when the SID has been replaced.

Part of FLE-475

### DIFF
--- a/rust/foxglove/src/remote_access/session.rs
+++ b/rust/foxglove/src/remote_access/session.rs
@@ -850,7 +850,7 @@ impl RemoteAccessSession {
             let inserted = self
                 .channel_registry
                 .write()
-                .insert_client_channel(participant.participant_id(), descriptor.clone());
+                .insert_client_channel(participant.participant_sid(), descriptor.clone());
 
             if !inserted {
                 self.send_warning(
@@ -885,7 +885,7 @@ impl RemoteAccessSession {
             let removed = self
                 .channel_registry
                 .write()
-                .remove_client_channel(participant.participant_id(), channel_id);
+                .remove_client_channel(participant.participant_sid(), channel_id);
 
             match removed {
                 None => debug!(
@@ -964,7 +964,7 @@ impl RemoteAccessSession {
         let descriptor = {
             let state = self.channel_registry.read();
             state
-                .get_client_channel(participant.participant_id(), channel_id)
+                .get_client_channel(participant.participant_sid(), channel_id)
                 .cloned()
         };
         let Some(descriptor) = descriptor else {
@@ -1132,7 +1132,7 @@ impl RemoteAccessSession {
         let removed = self
             .channel_registry
             .write()
-            .cleanup_for_removed_participant(participant_sid, participant_id);
+            .cleanup_for_removed_participant(participant_sid);
         let last_param_unsubscribed = self
             .parameter_subscriptions
             .write()
--- a/rust/foxglove/src/remote_access/session.rs
+++ b/rust/foxglove/src/remote_access/session.rs
@@ -204,7 +204,7 @@ impl Sink for RemoteAccessSession {
         // release it before resolving against the participant registry. A
         // same-identity reconnect arrives with a *different* `ParticipantSid`,
         // so a stale snapshotted SID resolves to `None` in the participant
-        // registry rather than the new attempt
+        // registry rather than the new attempt.
         let reliable_sids = {
             let state = self.channel_registry.read();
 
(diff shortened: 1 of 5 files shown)


dave-foxglove (approved):
Approval based on: at worst, this doesn't look harmful - the resolve_identities call should be fast. I'm not confirming that this fixes the race condition mentioned (lmk if you'd like me to dig into that).

rust/foxglove/src/remote_access/session.rs
```diff
@@ -224,26 +219,25 @@ impl Sink for RemoteAccessSession {
             }
 
             if !state.has_data_subscribers(&channel_id) {
-                None
```
dave-foxglove, line 227: Trivia: I don't 100% grok the change from Option to checking if the list is empty. I'm curious about it, but this isn't even a nit-level concern. i.e. I'd take "I like this better" as reason enough, but I wanted to ask.

gasmith: An `Option<Vec<T>>` is a bit of an anti-pattern in situations where there's no semantic difference between `None` and an empty list.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -224,26 +219,25 @@ impl Sink for RemoteAccessSession {
...
+                Vec::new()
             } else if state.qos_profile(&channel_id).reliability == Reliability::Reliable {
-                Some(state.data_subscriber_identities(&channel_id))
+                let identities = state.data_subscriber_identities(&channel_id);
```
gasmith: Could we avoid the nested locking by returning ParticipantSids (instead of ParticipantIdentities), and then resolving participants by Sid below?

gasmith (commented):
I have a suggestion that might sidestep the nested locking (which always makes me a bit nervous), but it might also mean a bigger change (to track subscribers by ParticipantSid instead of ParticipantIdentity).

rust/foxglove/src/remote_access/session.rs
clalancette: This is a great suggestion, and simplifies this immensely.  The next push will change to this completely.  It ends up being a larger PR, but it is largely mechanical and fixes this problem without nested locking.

rust/foxglove/src/remote_access/session.rs
dave-foxglove: Cool.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -681,6 +669,17 @@ impl RemoteAccessSession {
         true
     }
 
+    /// Returns `true` if `participant`'s SID has already been replaced in
+    /// the participant registry by a reordered same-identity reconnect.
+    /// Insert paths must call this under `subscription_lock`; otherwise
+    /// state keyed by the swept SID can be reinserted post-cleanup and
+    /// leak until session shutdown.
+    fn participant_already_swept(&self, participant: &Participant) -> bool {
```
dave-foxglove: Nit: This returns a bool, so it seems consistent to use a "is_" prefix on the name. 

Suggestion: Also "swept" feels less clear. How about "is_participant_stale"?

dave-foxglove (approved):
New version LGTM as well (just a nit on naming, but not a big deal).

rust/foxglove/src/remote_access/session.rs
gasmith: Or even better yet, negate the semantics and call it: `is_participant_registered`?

rust/foxglove/src/remote_access/session.rs
```diff
@@ -972,11 +977,18 @@ impl RemoteAccessSession {
             );
             return;
         }
+
+        // Avoid surfacing a misleading "Client has not advertised channel"
+        // error for a channel the prior attempt legitimately advertised.
+        if self.participant_already_swept(participant) {
+            return;
+        }
```
gasmith: This feels like the wrong place for this check. We're not holding the lock, so there's still a TOCTOU problem. What if we moved it to L995 to suppress the warning when the participant is no longer registered? Probably could trim down the long explanatory comment in that case too.

gasmith (approved):
I think this looks good, but I agree with Dave's suggestion about the naming of that method.

PR #1251, opened 2026-05-15
Add optional require-cuda feature to fail build if building remote access without nvenc support

eloff:
### Changelog

None

### Docs

See lib.rs changes

### Description

There's a reason webrtc-sys doesn't warn about this - they actually do, but cargo hides the warning unless the verbosity is increased. So there's no point to a warning.

The solution here is a new require-cuda feature. If passed with remote-access for a target that supports nvenc, we'll fail the build with an error that links to our docs. It gives users (and us!) a way to indicate their preference for nvenc. It runs the exact same check logic webrtc-sys uses to find cuda.h and enable nvenc. That catch is that could potentially drift if they change how it works in the future.

There's a [follow-up PR](https://github.com/foxglove/foxglove-sdk/pull/1260) that enables this in our CI and makefiles etc so that we don't build remote access binaries without nvenc support.


rust/foxglove/build.rs
```diff
@@ -0,0 +1,94 @@
...
+    Required,
+    Warn,
+    Off,
+}
+
+fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-env-changed=CUDA_HOME");
+    println!("cargo:rerun-if-env-changed=FOXGLOVE_REMOTE_ACCESS_NVENC");
+    println!("cargo:rerun-if-env-changed=PROFILE");
```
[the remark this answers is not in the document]
eloff: I'll remove it

rust/foxglove/build.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    let nvenc_requirement = match env::var("FOXGLOVE_REMOTE_ACCESS_NVENC")
+        .unwrap_or_else(|_| "warn".to_string())
+        .to_ascii_lowercase()
+        .as_str()
+    {
+        "required" => NvencRequirement::Required,
+        "off" => NvencRequirement::Off,
+        _ => NvencRequirement::Warn,
+    };
```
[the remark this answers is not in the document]
eloff: I think this is fine

rust/foxglove/build.rs
```diff
@@ -0,0 +1,94 @@
...
+    println!("cargo:rerun-if-env-changed=FOXGLOVE_REMOTE_ACCESS_NVENC");
+    println!("cargo:rerun-if-env-changed=PROFILE");
+
+    // We are only interested in nvenv support if we're compiling in remote access support
+    if env::var_os("CARGO_FEATURE_REMOTE_ACCESS").is_none() {
+        return;
+    }
+
+    // Don't surface warnings if we're building docs.
+    if env::var_os("DOCS_RS").is_some() {
```
gasmith: TIL. I wonder if we need to set that in CI when we build docs.

rust/foxglove/build.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-env-changed=CUDA_HOME");
+    println!("cargo:rerun-if-env-changed=FOXGLOVE_REMOTE_ACCESS_NVENC");
+    println!("cargo:rerun-if-env-changed=PROFILE");
+
+    // We are only interested in nvenv support if we're compiling in remote access support
+    if env::var_os("CARGO_FEATURE_REMOTE_ACCESS").is_none() {
```
dave-foxglove: consider detecting if it is falsey like "0" or "false"

rust/foxglove/build.rs
dave-foxglove: should/could we handle DOCS_RS being "0" or "false" (which would be 'some' and still trigger this).

rust/foxglove/build.rs
```diff
@@ -0,0 +1,94 @@
...
+        return;
+    }
+
+    let nvenc_requirement = match env::var("FOXGLOVE_REMOTE_ACCESS_NVENC")
+        .unwrap_or_else(|_| "warn".to_string())
+        .to_ascii_lowercase()
+        .as_str()
+    {
+        "required" => NvencRequirement::Required,
+        "off" => NvencRequirement::Off,
```
dave-foxglove: Dunno if it's helpful to consider but IIUC some folks use a yes|no|auto with an env var styled as ENABLE_*.

dave-foxglove: (or a WITH_* prefix, or maybe an *_ON suffix - not suggesting doing more than one of these, I favor ENABLE_* - just saying there is prior art). The FOXGLOVE_REMOTE_ACCESS_NVENC name is a touch unclear about what it is, like maybe it is a path string or  version number etc.

rust/foxglove/build.rs
```diff
@@ -0,0 +1,94 @@
...
+        _ => NvencRequirement::Warn,
+    };
+    if matches!(nvenc_requirement, NvencRequirement::Off) {
+        return;
+    }
+
+    let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap_or_default();
+    let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap_or_default();
+
+    // webrtc-sys only compiles NVENC support on Linux for x86 / aarch64 / arm
```
dave-foxglove, line 46: HOw about moving this earlier in the code. Seems like an easy out and skips doing stuff that isn't needed anyway.

rust/foxglove/build.rs
eloff: It's a a flag-style env var that's either present or not, the value doesn't matter

rust/foxglove/build.rs
```diff
@@ -0,0 +1,81 @@
...
+//! missing on a target where webrtc-sys would otherwise have built NVENC
+//! support, the build fails with installation instructions.
+
+use std::env;
+use std::path::PathBuf;
+
+fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-env-changed=CUDA_HOME");
+    println!("cargo:rerun-if-env-changed=DOCS_RS");
```
gasmith: Do we still need this? We'll build docs with the `full` synthetic feature, so we shouldn't need to special case `DOCS_RS`.

rust/foxglove/build.rs
```diff
@@ -0,0 +1,81 @@
...
+            "The `require-cuda` feature is enabled, but the `remote-access` feature is not enabled.\n\
+            Enable the `remote-access` feature or disable the `require-cuda` feature.\n\
+            Learn more: https://docs.rs/foxglove/latest/foxglove/#nvenc-hardware-acceleration"
+        );
+    }
+
+    // Don't surface errors when building docs.
+    if env::var_os("DOCS_RS").is_some() {
+        return;
+    }
```
gasmith: I think this can go too.

rust/foxglove/build.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap_or_default();
+    let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap_or_default();
+
+    // webrtc-sys only compiles NVENC support on Linux for x86 / aarch64 / arm
+    // (see webrtc-sys/build.rs). Match the same set of targets, expressed in
+    // cargo's `CARGO_CFG_TARGET_ARCH` vocabulary (which uses "x86" rather
+    // than the "i686" upstream parses out of the target triple).
+    if target_os != "linux" {
```
gasmith: If someone has set `require-cuda` on an unsupported platform/arch, we should probably panic as we do when someone sets `require-cuda` without `remote-access`?

gasmith (approved):
A few minor comments, but otherwise LG.

rust/foxglove/build.rs
eloff: Yeah, that carried over, but we don't need it anymore

rust/foxglove/build.rs
eloff: Yeah, that's a good point, I agree

PR #1258, opened 2026-05-20
cpp: Introduce ParameterHandler for asynchronous parameter operations

gasmith:
### Changelog
- cpp: Added ParameterHandler with responder-based get/set.
- cpp: Deprecated the onGetParameters/onSetParameters callbacks.

### Docs
Covered by cpp docs.

### Description
This change plumbs the new ParameterHandler from the rust SDK, and
mirrors the AssetHandler pattern: get/set implementations receive
responders that the implementer can use to complete requests
asynchronously.

When a parameter handler is registered, it takes precedence over the
older onGetParameters/onSetParameters callbacks. Those legacy callbacks
continue to work, but they are marked deprecated.

This change also updates the param server example to use the new
interface, and extends it to work with remote access.

For now, we're just suppressing the deprecation warnings in the bridge;
we'll update the bridge to use the new interface in a separate PR.

### Testing
Validated param server example works for both websocket and remote
access.

### Links
Fixes: FLE-533


cpp/foxglove/include/foxglove/websocket.hpp
```diff
@@ -262,6 +263,12 @@ struct WebSocketServerOptions {
   FetchAssetHandler fetch_asset;
   /// @brief A sink channel filter callback.
   SinkChannelFilterFn sink_channel_filter;
+  /// @brief A parameter handler.
+  ///
+  /// When set, this handler takes precedence over the deprecated
+  /// `onGetParameters` / `onSetParameters` callbacks. Registering a handler
+  /// automatically enables the `Parameters` capability.
```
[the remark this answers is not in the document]
gasmith: This is pretty clear already. You're over-thinking it.

cpp/examples/param-server/src/main.cpp
```diff
@@ -36,116 +118,161 @@ int main() {
...
+          if constexpr (std::is_same_v<T, GetOp>) {
+            std::vector<foxglove::Parameter> result;
+            if (op.names.empty()) {
+              result.reserve(param_store.size());
+              for (const auto& it : param_store) {
+                result.push_back(it.second.clone());
+              }
+            } else {
+              for (const auto& name : op.names) {
+                if (auto it = param_store.find(name); it != param_store.end()) {
```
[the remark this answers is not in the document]
gasmith: The client application never tries to create new parameters. Even if it did... this is an example, Claude.

I don't think the sub/unsub illustration was particularly illuminating. We also removed it from the rust example.

gasmith:
> One thing still worth a sanity check on your end before un-drafting: the C-header docs for `on_get_parameters` / `on_set_parameters` on `foxglove_server_callbacks` and `foxglove_gateway_callbacks` (sourced from `c/src/server.rs` and `c/src/gateway.rs`) don't carry a `@deprecated` Doxygen tag, so non-C++ consumers reading the generated header won't see what the changelog promises. C++ `[[deprecated]]` is in place, so this is the last consumer surface where the deprecation is invisible. Easy to defer if you'd rather wait for a follow-up — flagging only because the PR description bills the deprecation as covered.

We don't support users building directly against the C FFI. C++ is the supported interface for users. It's all open source, so I guess we'd say C is "use at your own risk". There's no precedent for marking things as  deprecated in the C header.

cpp/examples/param-server/src/main.cpp
```diff
@@ -36,116 +122,188 @@ int main() {
...
+    ra_options.parameter_handler = handler;
+    auto gateway_result = foxglove::RemoteAccessGateway::create(std::move(ra_options));
+    if (!gateway_result.has_value()) {
+      std::cerr << "Failed to start remote-access gateway: "
+                << foxglove::strerror(gateway_result.error()) << '\n';
+      return 1;
+    }
+    gateway.emplace(std::move(gateway_result.value()));
+  }
+#endif
```
clalancette: I'm really on the fence about this whole block.  I know we want to show off remote access capability, but I feel like this block (and the #ifdef FOXGLOVE_REMOTE_ACCESS below) hurts the readability of this example, which is trying to show off parameters.  On balance, I'd remove this from here and maybe add in a second example which does remote access instead.

cpp/examples/param-server/src/main.cpp
```diff
@@ -36,116 +122,188 @@ int main() {
...
+
   std::atomic_bool done = false;
   sigint_handler = [&] {
     done = true;
+    queue.shutdown();
+  };
```
clalancette: I'm going to suggest we rearrange this a bit so all of the "signal handling" stuff gets done in one place.  I'd either move this up to the top (where we called `std::signal` (which would also require moving some of the other variables around), or move the `std::signal` registration to right above this (since up to this point, it doesn't do anything anyway).

cpp/examples/param-server/src/main.cpp
```diff
@@ -36,116 +122,188 @@ int main() {
...
+            }
+          }
+        },
+        *maybe_op
+      );
+    }
+
+    if (std::chrono::steady_clock::now() >= next_tick) {
+      auto elapsed_secs =
+        std::chrono::duration<double>(std::chrono::steady_clock::now() - start_time).count();
```
clalancette: Minor nit, but we are calling `std::chrono::steady_clock::now()` twice here.  We could just call it once (more for maintainability/readability than anything else) by doing:

```suggestion
    now = std::chrono::steady_clock::now();
    if (now >= next_tick) {
      auto elapsed_secs = std::chrono::duration<double>(now - start_time).count();
```

cpp/examples/param-server/src/main.cpp
```diff
@@ -36,116 +122,188 @@ int main() {
...
+      auto elapsed = foxglove::Parameter("elapsed", elapsed_secs);
+      param_store.insert_or_assign("elapsed", elapsed.clone());
+      std::vector<foxglove::Parameter> to_publish;
+      to_publish.emplace_back(std::move(elapsed));
+      publish(std::move(to_publish));
```
clalancette: If we get rid of the REMOTE_ACCESS stuff, then this just becomes a single method call to `server.publishParameterValues(std::move(to_publish));`

cpp/foxglove/include/foxglove/parameter_handler.hpp
```diff
@@ -0,0 +1,183 @@
...
+  /// @param params Parameter values to send.
+  void respond(std::vector<Parameter>&& params) && noexcept;
+
+  /// @brief Default destructor. If the responder has not been consumed by
+  /// `respond()`, sends a generic error status to the requesting client; this
+  /// path is reserved for unrecoverable internal errors.
+  ~GetParametersResponder() = default;
+  /// @brief Default move constructor.
+  GetParametersResponder(GetParametersResponder&&) noexcept = default;
+  /// @brief Default move assignment.
```
clalancette: I feel like saying `Default destructor`, `Default move constructor`, etc is unnecessary, given the modern `default` keyword in C++.  More than that, though, the whole statement:

```
  /// @brief Default destructor. If the responder has not been consumed by
  /// `respond()`, sends a generic error status to the requesting client; this
  /// path is reserved for unrecoverable internal errors.
```

Isn't really true here, is it?  It is not *this* destructor that does this; it is the entire stack that does it.  Or maybe I'm misunderstanding how the generic response is generated?

cpp/foxglove/include/foxglove/parameter_handler.hpp
```diff
@@ -0,0 +1,183 @@
...
+  /// @param params Parameter values that were applied.
+  void respond(std::vector<Parameter>&& params) && noexcept;
+
+  /// @brief Default destructor. If the responder has not been consumed by
+  /// `respond()`, sends a generic error status to the requesting client; this
+  /// path is reserved for unrecoverable internal errors.
+  ~SetParametersResponder() = default;
+  /// @brief Default move constructor.
+  SetParametersResponder(SetParametersResponder&&) noexcept = default;
+  /// @brief Default move assignment.
```
clalancette: Same comments here.

cpp/foxglove/src/remote_access.cpp
```diff
@@ -322,6 +334,74 @@ FoxgloveResult<RemoteAccessGateway> RemoteAccessGateway::create(
...
+    const bool has_on_set = bool(options.parameter_handler.onSet);
+    if (has_on_get != has_on_set) {
+      warn() << "ParameterHandler requires both onGet and onSet to be set";
+      return tl::unexpected(FoxgloveError::ValueError);
+    }
+  }
+  if (options.parameter_handler.onGet && options.parameter_handler.onSet) {
+    parameter_handler = std::make_unique<ParameterHandler>(std::move(options.parameter_handler));
+    c_parameter_handler.context = parameter_handler.get();
+    c_parameter_handler.get = [](
```
clalancette, line 349: This is purely an aesthetic concern, and not all the fault of this PR, but this one `create` method is ~420 lines long.  I wonder if it might be better to have the inline lambdas here each be a named private function on the class so this is more readable.  I'm not sure you should do it in this PR, but maybe we should consider it as a followup.

cpp/foxglove/src/websocket.cpp
```diff
@@ -317,6 +329,74 @@ FoxgloveResult<WebSocketServer> WebSocketServer::create(
...
+    const bool has_on_set = bool(options.parameter_handler.onSet);
+    if (has_on_get != has_on_set) {
+      warn() << "ParameterHandler requires both onGet and onSet to be set";
+      return tl::unexpected(FoxgloveError::ValueError);
+    }
+  }
+  if (options.parameter_handler.onGet && options.parameter_handler.onSet) {
+    parameter_handler = std::make_unique<ParameterHandler>(std::move(options.parameter_handler));
+    c_parameter_handler.context = parameter_handler.get();
+    c_parameter_handler.get = [](
```
clalancette, line 344: Same comment here.

clalancette (commented):
I've left a number of things to think about.  None of them are blocking.

cpp/examples/param-server/src/main.cpp
gasmith: I'm also on the fence. The idea was to demonstrate that you can wire up a common parameters handler to both transports, but I agree that the readability is not great. I'm fine with removing it.

cpp/examples/param-server/src/main.cpp
gasmith: Yeah, that's a good call.

cpp/foxglove/include/foxglove/parameter_handler.hpp
gasmith: Yeah, this is all just LLM slop commentary. Dropping it.

cpp/foxglove/src/remote_access.cpp
gasmith: Strong agree; I'll put up a separate PR for that.

cpp/foxglove/include/foxglove/parameter_handler.hpp
gasmith: I guess the move docs are required by our docs compiler.

cpp/foxglove/src/websocket.cpp
```diff
@@ -317,6 +329,74 @@ FoxgloveResult<WebSocketServer> WebSocketServer::create(
...
+        request_id.emplace(c_request_id->data, c_request_id->len);
+      }
+      std::vector<std::string_view> param_names;
+      if (c_param_names != nullptr) {
+        param_names.reserve(param_names_len);
+        for (size_t i = 0; i < param_names_len; ++i) {
+          param_names.emplace_back(c_param_names[i].data, c_param_names[i].len);
+        }
+      }
+      GetParametersResponder responder(c_responder);
```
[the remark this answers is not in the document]
gasmith: That's true. But I'm not going to fix it here. This PR is already enormous. I'll fix it in #1275.

cpp/examples/param-server/src/main.cpp
gasmith: It has been removed.

cpp/foxglove/src/remote_access.cpp
gasmith: https://github.com/foxglove/foxglove-sdk/pull/1275

eloff (approved):
This looks good.

One thing not related to this PR is: onParametersSubscribe / onParametersUnsubscribe can  deadlock if the callback calls publishParameterValues() on the WebSocket server. The server holds subscribed_parameters.write() while invoking the listener, and publish_parameter_values() calls subscribed_parameters.read().

We warn about this deadlock danger for connection graph callbacks, but these have the same issue (and user's are more likely to bump into it.) 

I created an issue to track it, it's not worth holding this PR up for that: FLE-542

PR #1260, opened 2026-05-20
Enable hardware offload for video transcoding (nvenc) in our builds and CI

eloff:
### Changelog

Foxglove SDK and ROS bridge remote access: C++ and Python builds now compile with nvenc hw acceleration for video encoding. Simulcast is now disabled.

### Docs

None

### Description

This is the second PR in the series to enable nvenc acceleration in our build. The[ previous PR](https://github.com/foxglove/foxglove-sdk/pull/1251
) added the require-cuda feature, the docs, and the build detection and error.

This PR installs cuda in our CI environment, sets CUDA_HOME, and  enables the require-cuda feature.

An additional make knob FOXGLOVE_REQUIRE_CUDA which defaults to OFF. We use this in the CI to enable require-cuda as needed for integration tests and dist builds.

SImuclast was disabled because in testing it doesn't work well with nvenc, and it seems like a bad tradeoff for us to be making anyway.


.github/workflows/python.yml
```diff
@@ -194,13 +201,22 @@ jobs:
...
               elif command -v yum; then
-                yum install -y libva-devel
+                yum install -y libva-devel yum-utils
+                arch=$(uname -m)
+                case "$arch" in
+                  aarch64) cuda_arch=sbsa ;;
+                  *) cuda_arch="$arch" ;;
+                esac
+                yum-config-manager --add-repo "https://developer.download.nvidia.com/compute/cuda/repos/rhel8/${cuda_arch}/cuda-rhel8.repo"
+                yum install -y cuda-toolkit-12-6
```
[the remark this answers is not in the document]
eloff: I'm going to consider that out-of-scope for this PR

.github/workflows/ci.yml
```diff
@@ -17,6 +17,12 @@ jobs:
       - name: Install common dependencies
         uses: ./.github/actions/common-deps
 
+      # Install CUDA headers so that builds can find `cuda.h` and compile with NVENC support.
+      - name: Install CUDA toolkit
+        run: |
+          sudo apt-get install -y --no-install-recommends nvidia-cuda-toolkit
+          echo "CUDA_HOME=/usr" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
eloff: I will try this

python/foxglove-sdk/Cargo.toml
```diff
@@ -18,10 +18,14 @@ default = ["aws-lc-rs"]
...
 ring = ["foxglove/ring"]
+# Opts into a build-time check that NVENC hardware acceleration for video
+# encoding will be available. Only meaningful in combination with
+# `remote-access`. See foxglove crate docs for details.
+require-cuda = ["foxglove/require-cuda"]
 
 # The full set of supported features, with opinionated choices for mutually-
 # exclusive feature sets.
-full = ["aws-lc-rs", "remote-access"]
+full = ["aws-lc-rs", "remote-access", "require-cuda"]
```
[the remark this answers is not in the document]
eloff: Fixed

Container.mk
```diff
@@ -118,6 +118,11 @@ test-cpp-sanitize:
 # FETCHCONTENT_SOURCE_DIR_FOXGLOVE_SDK in CMake.
 CPP_SDK_DIR ?= cpp/dist
 FOXGLOVE_REMOTE_ACCESS ?= ON
+# Opts into a build-time check that NVENC hardware acceleration for video
+# encoding will be available (cuda.h is present on supported targets). Only
+# meaningful when remote-access is also enabled, so it defaults to whatever
+# FOXGLOVE_REMOTE_ACCESS is set to.
+FOXGLOVE_REQUIRE_CUDA ?= $(FOXGLOVE_REMOTE_ACCESS)
```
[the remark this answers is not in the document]
eloff: This is intentional, I'll update the changelog

rust/foxglove/Cargo.toml
```diff
@@ -84,6 +84,7 @@ full = [
   "live_visualization",
   "lz4",
   "remote-access",
+  "require-cuda",
```
gasmith: I don't think this should be part of the `full` feature set. A user should have to opt-in explicitly.

Container.mk
```diff
@@ -4,9 +4,18 @@ generate:
...
+# meaningful when remote-access is also enabled, so it defaults to whatever
+# PYTHON_REMOTE_ACCESS is set to.
+PYTHON_REQUIRE_CUDA ?= $(PYTHON_REMOTE_ACCESS)
 
 ifeq ($(PYTHON_REMOTE_ACCESS),ON)
+ifeq ($(PYTHON_REQUIRE_CUDA),ON)
 MATURIN_PEP517_ARGS += --features full
+else
+MATURIN_PEP517_ARGS += --features remote-access
+endif
```
gasmith: We should split this out differntly:
```suggestion
MATURIN_PEP517_ARGS += --features full
ifeq ($(PYTHON_REQUIRE_CUDA),ON)
MATURIN_PEP517_ARGS += --features require-cuda
endif
```

Container.mk
```diff
@@ -4,9 +4,18 @@ generate:
 	yarn generate
 
 PYTHON_REMOTE_ACCESS ?= ON
+# Opts into a build-time check that NVENC hardware acceleration for video
+# encoding will be available (cuda.h is present on supported targets). Only
+# meaningful when remote-access is also enabled, so it defaults to whatever
+# PYTHON_REMOTE_ACCESS is set to.
+PYTHON_REQUIRE_CUDA ?= $(PYTHON_REMOTE_ACCESS)
```
gasmith: I don't think we should necessarily require that developers have CUDA installed. I think I'd prefer to keep this OFF, and toggle it on in CI in places where we expect CUDA to be available.

Container.mk
```diff
@@ -118,6 +127,11 @@ test-cpp-sanitize:
 # FETCHCONTENT_SOURCE_DIR_FOXGLOVE_SDK in CMake.
 CPP_SDK_DIR ?= cpp/dist
 FOXGLOVE_REMOTE_ACCESS ?= ON
+# Opts into a build-time check that NVENC hardware acceleration for video
+# encoding will be available (cuda.h is present on supported targets). Only
+# meaningful when remote-access is also enabled, so it defaults to whatever
+# FOXGLOVE_REMOTE_ACCESS is set to.
+FOXGLOVE_REQUIRE_CUDA ?= $(FOXGLOVE_REMOTE_ACCESS)
```
gasmith: Same comment as above: I think we should default-off, and opt-in from CI.

.github/workflows/python.yml
```diff
@@ -45,6 +45,11 @@ jobs:
       - name: Install common dependencies
         uses: ./.github/actions/common-deps
 
+      # Install CUDA headers so that `make test-python` (which uses
+      # `--features full`, including `require-cuda`) can find `cuda.h`.
+      - name: Install CUDA headers
+        uses: ./.github/actions/install-cuda-headers
```
gasmith: Seems excessive, unless we actually have tests that target cuda.

Container.mk
eloff: Ok, I'll make the changes

.github/workflows/python.yml
eloff: Not really in Python, it should be a enabled for integration style tests. It also doesn't cost anything, the headers are tiny, so I'm inclined to keep it on. It matches the binaries we ship.

rust/foxglove/Cargo.toml
eloff: Ok, I'll rework this

.github/workflows/python.yml
gasmith: Are we looking at the same package? Looks like 634.37 MB to download and 1.76 GB to install?

https://ubuntu.pkgs.org/22.04/ubuntu-multiverse-arm64/nvidia-cuda-dev_11.5.1-1ubuntu1_arm64.deb.html

On Arch linux it's [even worse](https://archlinux.org/packages/extra/x86_64/cuda/)...

eloff: That's too much, I'm going to remove this for some places where it's not likely to matter then

.github/workflows/docs.yml
```diff
@@ -39,7 +39,10 @@ jobs:
           node-version: 22
           cache: yarn
       - uses: astral-sh/setup-uv@v7
-      - run: make -f Container.mk docs-python
+      # Building docs doesn't need NVENC hardware acceleration to actually work
+      # at runtime, so skip the build-time `cuda.h` check rather than installing
+      # the CUDA toolkit on the runner.
+      - run: make -f Container.mk PYTHON_REQUIRE_CUDA=OFF docs-python
```
gasmith: We should be able to revert this hunk, now that cuda is opt-in.

rust/foxglove/src/lib.rs
```diff
@@ -327,8 +327,8 @@
 //! to find it.
 //!
 //! You can enable the `require-cuda` feature on this crate to make it a build error if
-//! `remote-access` is disabled, the target does not support NVENC, or `cuda.h` is not
-//! found on a target where NVENC would be built.
+//! `cuda.h` is not found and NVENC support won't be built. This only applies when
+//! remote access is enabled and the target supports NVENC.
```
gasmith: This seems like a step backwards - I think we want to keep the existing behavior.

rust/foxglove/build.rs
```diff
@@ -23,9 +23,10 @@ use std::path::PathBuf;
 fn main() {
     println!("cargo:rerun-if-changed=build.rs");
     println!("cargo:rerun-if-env-changed=CUDA_HOME");
+    println!("cargo:rerun-if-env-changed=DOCS_RS");
```
gasmith: Looks like some merge/rebase damage here too.

gasmith (commented):
LG, but I think there's some merge/rebase damage.

.github/workflows/remote-access-tests.yml
```diff
@@ -93,6 +103,11 @@ jobs:
       - name: Install common dependencies
         uses: ./.github/actions/common-deps
 
+      - name: Install CUDA headers for NVENC
+        run: |
+          sudo apt-get install -y --no-install-recommends nvidia-cuda-dev
+          echo "CUDA_HOME=/usr" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
eloff: I'm not adding that back, those are not likely to change

gasmith:
I have sad news. I tested with nvenc offloading enabled (nvidia-smi shows the process, and there's a "Nvidia Decoder is supported" message on stdout). When I attempt to connect to a remote access gateway for the first time, I end up with horrible encoding artifacts:

https://github.com/user-attachments/assets/31f37c3c-04c2-4dc7-ab3b-7e07d90839bd

Things never get better. Even if I spin up new clients concurrently, they all see the same thing.

If I have only a single client, and I reload (causing the gateway to unpublish and republish the video track), then I get a full-res stream.

This video was taken with a python build from your PR. I was not able to repro with a python build from main.  I've tried local builds and CI builds with nvenc, both rust and python. Always the same behavior.

Maybe it's a quirk on my machine? But either way, it gives me pause. Especially since there doesn't seem to be any way to turn of nvenc at runtime (or so Claude thinks, from perusing the LiveKit SDK). We might want to file a patch upstream to add an environment variable override.

gasmith:
> Especially since there doesn't seem to be any way to turn of nvenc at runtime (or so Claude thinks, from perusing the LiveKit SDK). We might want to file a patch upstream to add an environment variable override.

Actually there is a way to turn it off, we just need to upgrade webrtc-sys: https://github.com/livekit/rust-sdks/commit/b73cbde5

I just tested with #1272, and can confirm that setting the preferred encoder to vaapi works around the issue.

eloff:
> > Especially since there doesn't seem to be any way to turn of nvenc at runtime (or so Claude thinks, from perusing the LiveKit SDK). We might want to file a patch upstream to add an environment variable override.
> 
> Actually there is a way to turn it off, we just need to upgrade webrtc-sys: [livekit/rust-sdks@b73cbde5](https://github.com/livekit/rust-sdks/commit/b73cbde5)
> 
> I just tested with #1272, and can confirm that setting the preferred encoder to vaapi works around the issue.

Except none of our customers are likely using a VAAPI compatible device, which means nvenc would be used and there isn't a way to disable it.

This is sad. I'm going to try playing with the video encoding settings, I think it's trying to save bandwidth too aggressively for some reason (on localhost!) But if we can't fix this ourselves,  we can't merge this PR until livekit fixes this upstream and provides a way to disable nvenc.

gasmith:
> > > Especially since there doesn't seem to be any way to turn of nvenc at runtime (or so Claude thinks, from perusing the LiveKit SDK). We might want to file a patch upstream to add an environment variable override.
> > 
> > 
> > Actually there is a way to turn it off, we just need to upgrade webrtc-sys: [livekit/rust-sdks@b73cbde5](https://github.com/livekit/rust-sdks/commit/b73cbde5)
> > I just tested with #1272, and can confirm that setting the preferred encoder to vaapi works around the issue.
> 
> Except none of our customers are likely using a VAAPI compatible device, which means nvenc would be used and there isn't a way to disable it.

Ah, you're right. We need that env var to take a value that forces software encoding. We should file a patch for that upstream.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -2183,6 +2183,7 @@ impl RemoteAccessSession {
                 // are software-only in our builds, so H.264 is at worst parity elsewhere.
                 let publish_options = TrackPublishOptions {
                     video_codec: VideoCodec::H264,
+                    simulcast: false,
```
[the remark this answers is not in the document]
eloff: This doesn't play nicely with nvenc.

rust/examples/remote_access_services/Cargo.toml
```diff
@@ -9,7 +9,7 @@ workspace = true
 [dependencies]
 anyhow = "1.0"
 env_logger = "0.11"
-foxglove = { path = "../../foxglove", features = ["schemars", "remote-access"] }
+foxglove = { path = "../../foxglove", features = ["schemars", "remote-access", "require-cuda"] }
```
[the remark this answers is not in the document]
eloff: This was just for testing, it's been reverted

.github/workflows/c_cpp.yml
```diff
@@ -170,9 +174,11 @@ jobs:
         if: matrix.remote_access && runner.os == 'Linux'
         run: |
           sudo apt-get update
-          sudo apt-get install -y libglib2.0-dev libva-dev libwebsockets-dev
+          sudo apt-get install -y libglib2.0-dev libva-dev libwebsockets-dev \
+            nvidia-cuda-dev
```
gasmith: Nit: We could do this in a separate job to further gate it on require_cuda.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -2178,11 +2178,14 @@ impl RemoteAccessSession {
             let session = self.clone();
             tokio::spawn(async move {
                 let local_track = LocalTrack::Video(track);
-                // Prefer H.264 so that the libwebrtc VAAPI encoder (H.264-only) can be used
-                // on Linux hosts that have libva + a VA driver available. VP8/VP9/AV1 paths
+                // Prefer H.264 so that the libwebrtc nvenc encoder (H.264-only) can be used
+                // on Linux hosts that have nvenc available. VP8/VP9/AV1 paths
                 // are software-only in our builds, so H.264 is at worst parity elsewhere.
+                // Disable simulcast. We expect viewers will be mostly homogenous, and
+                // simulcast is a lot of work for the robot without much to gain.
```
gasmith, line 2185: Nit: Maybe also mention the problem we observed, where nvenc+simulcast seems to result in horribly low quality video.

gasmith (approved):
This looks good. In terms of merging: Should we wait until we have a way to disable it, in case it causes problems in some environments?

PR #1267, opened 2026-05-21
bridge: Handle parameter operations asynchronously

gasmith:
### Changelog
- Fixed an issue where the `parameters` implementation would
  occasionally stall message forwarding for several seconds.
- Fixed an issue where the bridge was not refcounting subscriptions
  properly when a websocket server and remote access gateway were
  simultaneously in use.
- Fixed an issue where the parameter whitelist would randomly prevent
  subscribing to or setting unrelated parameters.

### Docs
None

### Description
This change updates the bridge to use the new ParameterHandler interface
for handling parameter get/set requests. The new interfaces allows for
asynchronous handling, which is particularly useful in the bridge,
because get and set both require fan-out requests to other ROS nodes.

This change also fixes a related bug in the way that the bridge was
tracking parameter subscriptions. The websocket server and remote access
gateway each aggregate subscriptions independently. The bridge must
implement another aggregation layer on top of that.

The new interface has the benefit that it does not attempt to broadcast
parameter updates when responding to a set request. This means the
bridge no longer has to attempt to work around that behavior.

There are also two drive-by fixes here. A fix for the parameter whitelist
(two `return` statements that ought to have been `continue` statements),
and an initialization ordering problem for the rosgraphPollThread.

### Links
Fixes: FLE-534

### DIFF
--- a/ros/src/foxglove_bridge/include/foxglove_bridge/ros2_foxglove_bridge.hpp
+++ b/ros/src/foxglove_bridge/include/foxglove_bridge/ros2_foxglove_bridge.hpp
@@ -188,6 +188,14 @@ private:
   using ParameterOp =
     std::variant<GetParamsOp, SetParamsOp, SubscribeParamsOp, UnsubscribeParamsOp>;
 
+  // Wire the parameter-related callbacks/handler on a server or gateway options struct.
+  // Both options structs share these field types; this helper keeps the WS and gateway
+  // sites in sync.
+  void wireParameterCallbacks(
+    std::function<void(const std::vector<std::string_view>&)>& onSubscribe,
+    std::function<void(const std::vector<std::string_view>&)>& onUnsubscribe,
+    foxglove::ParameterHandler& handler);
+
   void enqueueParameterOp(ParameterOp&& op);
   void parameterWorkerLoop();
   void handleGetParams(GetParamsOp&& op);
--- a/ros/src/foxglove_bridge/include/foxglove_bridge/ros2_foxglove_bridge.hpp
+++ b/ros/src/foxglove_bridge/include/foxglove_bridge/ros2_foxglove_bridge.hpp
@@ -2,12 +2,16 @@
 
 #include <atomic>
 #include <chrono>
+#include <condition_variable>
 #include <deque>
 #include <map>
 #include <memory>
+#include <mutex>
+#include <queue>
 #include <regex>
 #include <thread>
 #include <unordered_set>
+#include <variant>
 
 #include <rclcpp/rclcpp.hpp>
 #include <rmw/types.h>
@@ -17,6 +21,7 @@
 
 #include <foxglove/fetch_asset.hpp>
 #include <foxglove/foxglove.hpp>
+#include <foxglove/parameter_handler.hpp>
 #include <foxglove/system_info.hpp>
 #include <foxglove/websocket.hpp>
 #ifdef FOXGLOVE_REMOTE_ACCESS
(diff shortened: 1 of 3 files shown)


ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -200,32 +200,55 @@ FoxgloveBridge::FoxgloveBridge(const rclcpp::NodeOptions& options)
...
 
     _paramInterface = std::make_shared<ParameterInterface>(this, paramWhitelistPatterns,
                                                            ignoreUnresponsiveParamNodes
                                                              ? UnresponsiveNodePolicy::Ignore
                                                              : UnresponsiveNodePolicy::Retry);
     _paramInterface->setParamUpdateCallback(std::bind(&FoxgloveBridge::parameterUpdates, this, _1));
+
+    // Start the worker thread that drains the queue and dispatches into _paramInterface.
+    _paramWorkerThread =
+      std::make_unique<std::thread>(std::bind(&FoxgloveBridge::parameterWorkerLoop, this));
```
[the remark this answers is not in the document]
gasmith: Hm, we should also move _rosgraphPollThread.

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -1158,55 +1212,125 @@ void FoxgloveBridge::clientMessage(ClientId clientId, ChannelId clientChannelId,
...
+    for (const auto& param : op.parameters) {
+      names.emplace_back(param.name());
+    }
+    auto updated = _paramInterface->getParams(names, std::chrono::seconds(5));
+    std::move(op.responder).respond(std::move(updated));
+  } catch (const std::exception& ex) {
+    RCLCPP_ERROR(this->get_logger(), "setParams failed: %s", ex.what());
+    // Dropping the responder sends the client an error status.
+  }
 }
```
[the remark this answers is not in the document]
gasmith: Disagree. Just because set "succeeded", that doesn't mean the parameter changes were actually applied. I'd rather surface an error - even a generic one - so someone thinks to go look at logs. The real fix here is first class error reporting via the responder, but that's not happening right now.

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -1158,55 +1212,125 @@ void FoxgloveBridge::clientMessage(ClientId clientId, ChannelId clientChannelId,
...
+  }
+  if (toForward.empty()) {
+    return;
+  }
+  try {
+    _paramInterface->subscribeParams(toForward);
+  } catch (const std::exception& ex) {
+    RCLCPP_ERROR(this->get_logger(), "subscribeParams failed: %s", ex.what());
+  }
 }
```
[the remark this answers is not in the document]
gasmith: Mm, the refcounting is perfectly fine: onParametersSubscribe is fire-and-forget, and so the SDK doesn't care whether we fail to apply some subset of the subscriptions. It won't call subscribe again.

But the bug you point out in ParameterInterface is real. And it's there for setParams too. Returned where we meant to continue.

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -1158,55 +1138,167 @@ void FoxgloveBridge::clientMessage(ClientId clientId, ChannelId clientChannelId,
...
+      RCLCPP_WARN(this->get_logger(), "Unsubscribe for untracked parameter '%s'", name.c_str());
+      continue;
+    }
+    if (--it->second == 0) {
+      toForward.emplace_back(name);  // `name` lives in `op.names` until this function returns
+      _paramSubscriberCount.erase(it);
+    }
+  }
+  if (toForward.empty()) {
+    return;
```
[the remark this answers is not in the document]
gasmith: The SDK isn't going to call subscribe with duplicate names.

ros/src/foxglove_bridge/include/foxglove_bridge/ros2_foxglove_bridge.hpp
```diff
@@ -163,20 +168,59 @@ class FoxgloveBridge : public rclcpp::Node {
...
+  void handleGetParams(GetParamsOp&& op);
+  void handleSetParams(SetParamsOp&& op);
+  void handleSubscribeParams(SubscribeParamsOp&& op);
+  void handleUnsubscribeParams(UnsubscribeParamsOp&& op);
 
   void parameterUpdates(const std::vector<foxglove::Parameter>& parameters);
 
+  std::mutex _paramOpMutex;
+  std::condition_variable _paramOpCv;
+  std::queue<ParameterOp> _paramOpQueue;
```
eloff, line 210: This is unbounded. I don't *think* that should be an issue, fwiw

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -1158,55 +1138,167 @@ void FoxgloveBridge::clientMessage(ClientId clientId, ChannelId clientChannelId,
...
+      return;
     }
+    auto op = std::move(_paramOpQueue.front());
+    _paramOpQueue.pop();
+    lock.unlock();
+
+    std::visit(
+      [this](auto& concrete) {
+        using T = std::decay_t<decltype(concrete)>;
+        if constexpr (std::is_same_v<T, GetParamsOp>) {
```
eloff, line 1213: nit: we're relaying on transitive includes for <type_traits> here, it would be better to add an include for it explicitly

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -1158,55 +1138,167 @@ void FoxgloveBridge::clientMessage(ClientId clientId, ChannelId clientChannelId,
...
+      toForward.emplace_back(name);
+    }
+  }
+  if (toForward.empty()) {
+    return;
+  }
+  try {
+    _paramInterface->subscribeParams(toForward);
+  } catch (const std::exception& ex) {
+    RCLCPP_ERROR(this->get_logger(), "subscribeParams failed: %s", ex.what());
```
eloff, line 1277: We already incremented the subscription count for the params here, but I think that's correct, despite the fact that we couldn't make good on it, and the app still thinks we're subscribed

In the future it'd be nice to surface this in the app, but that seems out of scope right now

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -427,6 +395,18 @@ FoxgloveBridge::~FoxgloveBridge() {
   }
 #endif
   _server->stop();
```
eloff, line 398: the bridge parameter worker may still own queued GetParametersResponder / SetParametersResponder objects after the SDK transports have stopped.

I think this is fine, send_parameter_values is a no-op if the transport was stopped.

ros/src/foxglove_bridge/src/parameter_interface.cpp
eloff: We're reading this without the mutex here, but unsubscribeParams() mutates and erases those same structures under the mutex.

That looks like a data race. This lambda is registered with ros as an event callback, not run before the mutex lock goes out of scope.

Might also want to use find() here instead of [], in case the nodeName is not found after locking the mutex.

eloff (changes requested):
Some minor comments, but the data race looks like a blocking issue, if I'm right about it.

ros/src/foxglove_bridge/include/foxglove_bridge/ros2_foxglove_bridge.hpp
gasmith: Yeah, we bound the callbacks within the SDK, with a semaphore that allows 32 outstanding requests per client. That number was picked for commonality with fetch asset and service calls, but now that I'm looking at it, I realize we can probably ratchet it down considerably.

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
gasmith: Yep, that was my thinking as well.

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
gasmith: Yeah, the idea behind holding onto them is so that we can just drop them without sending more status messages back

ros/src/foxglove_bridge/src/parameter_interface.cpp
gasmith: Oh wow, nice catch! Yep, that's definitely a data race.

eloff (approved):
Looks good now!

PR #1268, opened 2026-05-21
Use rustup minimal profile for local docker image

gasmith:
### Changelog
None

### Docs
None

### Description
The docker image we use for local development installs three rust
toolchains. Whenever this runs, it always thrashes my disk writing out
millions of little rust doc files that we won't actually use.

By setting the rustup profile to minimal, we reduce the image size by
1.5G, and reduce the docker build time by several minutes.


eloff (approved):
Nice improvement, I never knew about this option

PR #1275, opened 2026-05-22
cpp: Refactor RemoteAccessGateway::create and WebSocketServer::create

gasmith:
### Changelog
None

### Docs
None

### Description
This change refactors the create method for the RemoteAccessGateway and
WebSocketServer. These methods were fairly long and complicated, and
shared some commonality that we were able to factor out.

There's some ugliness here with the introduction of the new
foxglove::internal namespace and ForwarderAccess dummy struct, which we
use to grant friend access to the responders and ParameterArray.

### DIFF
--- a/cpp/foxglove/include/foxglove/parameter.hpp
+++ b/cpp/foxglove/include/foxglove/parameter.hpp
@@ -21,6 +21,12 @@ struct foxglove_parameter_array;
 
 namespace foxglove {
 
+/// @cond foxglove_internal
+namespace internal {
+struct ForwarderAccess;
+}
+/// @endcond
+
 /// @brief An optional type hint for a `Parameter`, used to disambiguate values whose intended
 /// type cannot be inferred from the wire representation alone.
 ///
@@ -720,6 +726,7 @@ private:
   friend class RemoteAccessGateway;
   friend class GetParametersResponder;
   friend class SetParametersResponder;
+  friend struct internal::ForwarderAccess;
 
   struct Deleter {
     void operator()(foxglove_parameter_array* ptr) const noexcept;
--- a/cpp/foxglove/src/callback_forwarders.cpp
+++ b/cpp/foxglove/src/callback_forwarders.cpp
@@ -1,0 +1,97 @@
+#include "callback_forwarders.hpp"
+
+#include <foxglove-c/foxglove-c.h>
+#include <foxglove/channel.hpp>
+#include <foxglove/error.hpp>
+#include <foxglove/fetch_asset.hpp>
+#include <foxglove/parameter.hpp>
+#include <foxglove/parameter_handler.hpp>
+
+#include <optional>
+#include <string_view>
+#include <utility>
+#include <vector>
+
+namespace foxglove::internal {
+
+void forwardFetchAsset(
+  const void* context, const foxglove_string* c_uri, foxglove_fetch_asset_responder* c_responder
+) {
+  const auto* handler = static_cast<const FetchAssetHandler*>(context);
+  std::string_view uri{c_uri->data, c_uri->len};
+  auto responder = ForwarderAccess::makeFetchAssetResponder(c_responder);
+  try {
+    (*handler)(uri, std::move(responder));
+  } catch (const std::exception& exc) {
+    warn() << "Fetch asset callback failed: " << exc.what();
+    constexpr std::string_view kErrMessage{"Fetch asset callback failed"};
+    if (auto* ptr = ForwarderAccess::releaseFetchAsset(responder)) {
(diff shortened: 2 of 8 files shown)


cpp/foxglove/src/callback_forwarders.cpp
```diff
@@ -0,0 +1,97 @@
...
+) {
+  // Take ownership of the responder before any throwable work, so its destructor
+  // sends the generic error to the client if anything below throws.
+  auto responder = ForwarderAccess::makeSetResponder(c_responder);
+  callbackGuard("ParameterHandler.onSet", [&] {
+    if (c_params == nullptr) {
+      // Should not happen; the C implementation never passes a null pointer.
+      std::move(responder).respond({});
+      return;
+    }
```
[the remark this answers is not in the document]
gasmith: Fair. I'll just drop the branch, and leave a comment as a reminder about the ABI contract.

cpp/foxglove/include/foxglove/parameter.hpp
```diff
@@ -21,6 +21,12 @@ struct foxglove_parameter_array;
 
 namespace foxglove {
 
+/// @cond foxglove_internal
+namespace internal {
+struct ForwarderAccess;
+}
+/// @endcond
+
```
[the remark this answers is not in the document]
gasmith: Let me get this straight. You want me to... document... that the `internal` namespace that we've _hidden from the docs_ with `@cond foxglove_internal` is... internal? I mean, maybe we should also call it `internal_internal_internal_do_not_use_please_because_its_internal`?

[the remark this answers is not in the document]
eloff: +1 for internal_internal_internal_do_not_use_please_because_its_internal

cpp/foxglove/src/callback_forwarders.cpp
```diff
@@ -0,0 +1,93 @@
...
+void forwardFetchAsset(
+  const void* context, const foxglove_string* c_uri, foxglove_fetch_asset_responder* c_responder
+) {
+  const auto* handler = static_cast<const FetchAssetHandler*>(context);
+  std::string_view uri{c_uri->data, c_uri->len};
+  auto responder = ForwarderAccess::makeFetchAssetResponder(c_responder);
+  try {
+    (*handler)(uri, std::move(responder));
+  } catch (const std::exception& exc) {
+    warn() << "Fetch asset callback failed: " << exc.what();
```
eloff, line 26: Could we build this message as a string, and then log it and send it to the app, so the user can see what exactly failed?

cpp/foxglove/src/callback_forwarders.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include <vector>
+
+namespace foxglove::internal {
+
+void forwardFetchAsset(
+  const void* context, const foxglove_string* c_uri, foxglove_fetch_asset_responder* c_responder
+) {
+  const auto* handler = static_cast<const FetchAssetHandler*>(context);
+  std::string_view uri{c_uri->data, c_uri->len};
+  auto responder = ForwarderAccess::makeFetchAssetResponder(c_responder);
```
eloff, line 22: I don't like this ForwarderAccess workaround, but I can't think of a better alternative.

I think it's either that or we make things like makeFetchAssetResponder public, but hide them from the docs. I don't like that either.

Having pub(crate) is so nice.

cpp/foxglove/src/remote_access.cpp
```diff
@@ -12,244 +12,194 @@
...
+// consumers, but this file is the internal C-ABI forward and must keep
+// reading them.
 #if defined(__GNUC__) || defined(__clang__)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 #elif defined(_MSC_VER)
 #pragma warning(push)
 #pragma warning(disable : 4996)
 #endif
+bool wireGatewayCallbacks(foxglove_gateway_callbacks& c, const RemoteAccessGatewayCallbacks& cb) {
```
eloff, line 117: This is so much easier to read now

eloff (approved):

cpp/foxglove/src/callback_forwarders.cpp
gasmith: We could. I originally had that, but then an LLM complained that I wasn't wrapping that string construction in another try/catch block to guard against the alloc failure, and I decided to punt.

We're already on the exceptional path here (if you'll pardon the pun)... a well-behaved handler implementation isn't supposed to throw an exception, they're supposed to respond with an error. This catch block is a last resort.

gasmith: Sure, but was that error response actually _load-bearing_, claude?

PR #1279, opened 2026-05-27
playground: upgrade jedi & remove some workarounds

jtbandes:
### Changelog

None

### Docs

None

### Description

We had some workarounds for two `jedi` issues that have recently been fixed:
- https://github.com/davidhalter/jedi/issues/2087
  - workaround by using `Script` and adjusting line numbers: https://github.com/foxglove/foxglove-sdk/pull/796/changes/782aac2644b7edee782b2dfbe1a22d3112ecc8cf#diff-17a42a2f6f9ff43e0fa84baf2785173ce7968eb99bf1542fc78caf37d8856c50L59-L60
- https://github.com/davidhalter/jedi/issues/2073
  - workaround by using `__init__` instead of `__new__` in type stubs: https://github.com/foxglove/foxglove-sdk/pull/657

Since these are now fixed, we are able to update jedi and undo the workarounds. This also requires bundling a version of jedi newer than the one that [comes with pyodide](https://pyodide.org/en/stable/usage/packages-in-pyodide.html). This was relatively straightforward because jedi is a python-only package, so we can simply fetch the wheel directly from PyPI and bundle it alongside the other ones we fetch from the pyodide CDN.


gasmith (approved):
Tested it out and everything seems to work properly.

PR #1288, opened 2026-05-29
ros: Add script to ease bloom-release

gasmith:
### Changelog
None

### Docs
Updated `CONTRIBUTING.md`

### Description
This change adds a wrapper script for bloom-release. The script does a
few things:

 - Ensures that the desired tag exists
 - Synchronizes LFS objects before running bloom-release
 - Runs bloom-release inside a humble container

Example usage:

```bash
ros/bloom-release.sh --tag ros-v3.4.0 --distro rolling
```

### Testing
I used this script to do the bloom-release for 3.4.0.

### Links
Fixes: FLE-492


eloff (approved):
I'm not the ROS expert, so I'm curious what @clalancette would think of this.

I know we've been using humble for this before, so this doesn't change anything, but I wonder if we should use a newer version? What concerns me is the python3-bloom package comes from the apt repo, so if the version of that is too old in humble it might not have the right templates or config for the newest ros versions.

I have no idea if that's likely to be an issue in the future. It clearly isn't right now, since that's what we currently do.

This does make the release process simpler and harder to get wrong, which is great.

ros/bloom-release.sh
```diff
@@ -0,0 +1,143 @@
...
+# ~/.config/bloom; override by setting the BLOOM_CONFIG environment variable.
+# The file should contain the following content:
+#
+#   {
+#     "github_user": "your-github-username",
+#     "oauth_token": "ghp_..."
+#   }
+#
+# For more information on how to set up the PAT, see:
+# https://wiki.ros.org/bloom/Tutorials/GithubManualAuthorization
```
clalancette: The wiki is deprecated, here is the updated docs:
```suggestion
# https://docs.ros.org/en/rolling/How-To-Guides/Releasing/First-Time-Release.html
```

ros/release-to-ros.bash
clalancette, line 1: I'm going to suggest that we do not call this script `bloom-release.sh`; it is way too close of a name to the upstream `bloom-release`.  Also, technically this is a bash script, not a sh script.  Maybe `foxglove-sdk-bloom-release.bash`, or something like that?

ros/Makefile
```diff
@@ -136,6 +139,30 @@ docker-deb-bridge: docker-deb-bridge-rolling
...
+	@test -n "$$GIT_USER_NAME" || { echo "error: GIT_USER_NAME is required" >&2; exit 1; }
+	@test -n "$$GIT_USER_EMAIL" || { echo "error: GIT_USER_EMAIL is required" >&2; exit 1; }
+	docker run --rm -it $(DOCKER_RUN_ARGS) \
+		-e HOME=/home/foxglove \
+		-e GIT_USER_NAME \
+		-e GIT_USER_EMAIL \
+		-e GITHUB_TOKEN \
+		-e BLOOM_RELEASE_DISTRO="$(BLOOM_RELEASE_DISTRO)" \
+		-e BLOOM_PRETEND_FLAG="$(BLOOM_PRETEND_FLAG)" \
+		-v $(BLOOM_CONFIG):/home/foxglove/.config/bloom \
```
clalancette, line 161: This is just a comment (nothing to necessarily change), but passing this in is *almost* redundant.  The `bloom-release` script already needs to have the `GITHUB_TOKEN` for LFS work, and that means we only need the GitHub user name to have the script generate its own valid `~/.config/bloom` file.  I guess that doesn't really work if you are running the script locally (since you don't want the script to overwrite your config), but you could have the "run" command at the bottom of this docker invocation create one inside the container (rather than mounting it with -v).

Dockerfile.ros
```diff
@@ -28,6 +28,11 @@ RUN if [ "$USE_ROS_TESTING" = "true" ]; then \
 RUN apt-get update && apt-get -y dist-upgrade && \
     apt-get install -y python3-bloom fakeroot dpkg-dev debhelper
```
clalancette, line 29: I'm actually a bit surprised we install these things into the Dockerfile; it means that these Dockerfiles are "dual-use": for actually building/running the foxglove_bridge, and for doing "release-y" things around it.  As long as these aren't meant to be minimal Docker images that we provide to end-users, that is totally fine (and how I do things locally as well).

clalancette (changes requested):
Overall, the idea here looks fine.  I've left a few things that I think we should change/discuss.

ros/release-to-ros.bash
gasmith: Mmh, that sounds very redundant, for a script that lives inside of the foxglove-sdk repository. I was tempted to call it just `release.bash`, but I wanted to distinguish this from the "release" that we do in GitHub via CI. Is the concern that one of us might confuse a bash wrapper for the real thing? I suppose we could rename the wrapper to describe the intent instead of the tools that it uses to get there... `release-to-ros.bash`?

ros/Makefile
gasmith: Is that true? Bloom's documentation suggests that it [requires a PAT](https://docs.ros.org/en/rolling/How-To-Guides/Releasing/First-Time-Release.html#set-up-a-personal-access-token), and that's poked in through the `~/.config/bloom` config file. If bloom can make do with a `GITHUB_TOKEN` alone, that would be a welcome improvement.

Dockerfile.ros
gasmith: Yeah, this image is only used internally, for developers and CI. The image we provide to customers is built from `ros/src/foxgxlove_bridge/Dockerfile`.

ros/release-to-ros.bash
clalancette: Yeah, the concern is that someone who isn't super familiar with the ROS release process confuses this "bloom-release.sh" and the upstream one.  I'm fine with `release-to-ros.bash`.

ros/Makefile
clalancette: I looked into it, and bloom absolutely requires the token to be in the `~/.config/bloom` file; it doesn't look at anything else on the filesystem or in the environment.  However, my comment was more along the lines that *we* are passing `GITHUB_TOKEN` through into the docker environment.  Instead of bind-mounting `~/.config/bloom` into the docker container, we could have the "command" we run at the end create the `~/.config/bloom` inside of itself.  It's kind of a "six-on-one-hand, half-a-dozen-on-the-other", though it would slightly insulate this script from an outdated/missing bloom config on the host.  Up to you whether you want to change it.

clalancette (approved):
Looks good to me!  Thanks for iterating.

ros/Makefile
gasmith: I think I'll leave it as is for now. Let's wait to see how it goes when someone besides me uses this tooling, and maybe at that point we'll have some more ideas for how to make it less error-prone.

PR #1298, opened 2026-06-06
Emit Encode schema definitions in deterministic order (fixes #1297)

### LINKED ISSUE #1297: derive(Encode) is non-deterministic: schema definitions emitted in HashMap order (changes consuming crate SVH)
### Summary

`#[derive(Encode)]` (in `foxglove_derive`) emits its generated schema-descriptor statements in `std::collections::HashMap` iteration order, which is randomized per process. The macro therefore expands to a **different token order on every compile**, changing the **crate-metadata hash (SVH)** of the deriving crate.

### Mechanism

In `foxglove_derive/src/lib.rs`, `derive_named_struct_impl` accumulates per-field-type definitions into hash maps and then splices them into the output:

```rust
let mut enum_defs:    HashMap<&syn::Type, proc_macro2::TokenStream> = HashMap::new();
let mut message_defs: HashMap<&syn::Type, proc_macro2::TokenStream> = HashMap::new();
let mut file_defs:    HashMap<&syn::Type, proc_macro2::TokenStream> = HashMap::new();
// … populated per field via .entry(field_type).or_insert_with(…) …
let enum_defs    = enum_defs.into_values().collect::<Vec<_>>();
let message_defs = message_defs.into_values().collect::<Vec<_>>();
let file_defs    = file_defs.into_values().collect::<Vec<_>>();
// … later spliced: #(#enum_defs)*  #(#message_defs)*  #(#file_defs)*
```

`HashMap::into_values()` yields per-process-random order, so the spliced descriptor statements are emitted in a different order each compile.

### Why it matters

Under a content-addressed build system with a shared binary cache (e.g. Nix/crate2nix), the same input produces different output bytes across builds/hosts, and linking a fresh crate against a **cached** dependency with a mismatched SVH fails with `error[E0463]: can't find crate`. This is the proc-macro variant of rust-lang/rust#89904.

### Reproduction

```rust
use foxglove::Encode;

#[derive(Encode)]
pub struct S {
    pub a: u32, pub b: f64, pub c: bool, pub d: i64,
    pub e: i32, pub f: u16, pub g: u8,  pub h: i16,
}
```

Build three times with a clean rebuild each time and hash the produced `.rmeta`: the three hashes differ. (Any struct with ≥2 distinct field types triggers it.)

### Fix (verified)

Emit the definitions in a deterministic order — e.g. sort the collected defs by their field-type tokens before splicing (an insertion-ordered map such as `IndexMap` would also work):

```rust
let mut enum_defs: Vec<_> = enum_defs.into_iter().collect();
enum_defs.sort_by_key(|(t, _)| quote::quote!(#t).to_string());
let enum_defs = enum_defs.into_iter().map(|(_, d)| d).collect::<Vec<_>>();
// … same for message_defs and file_defs
```

With this change the same crate rebuilds byte-identically across repeated clean builds. `0.25.1` and `main` have the same pattern. Happy to open a PR if you'd like.

pchikku:
Fixes #1297.

`#[derive(Encode)]` (`derive_named_struct_impl`) collects per-field-type schema definitions into three `HashMap<&syn::Type, TokenStream>` (`enum_defs`/`message_defs`/`file_defs`) and then splices them via `into_values()`:

```rust
let enum_defs = enum_defs.into_values().collect::<Vec<_>>();   // HashMap order = per-process random
// … #(#enum_defs)*  #(#message_defs)*  #(#file_defs)*
```

`HashMap` iteration order is randomized per process, so the generated descriptor statements come out in a different order on every compile, which changes the **consuming crate's metadata hash (SVH)**. Under content-addressed build systems with a shared cache that yields different artifacts for the same input, and mixing a fresh build with a cached dependency fails with `error[E0463]: can't find crate` (the proc-macro variant of rust-lang/rust#89904).

This PR sorts the collected definitions by their field-type tokens before splicing, making the order deterministic. (An insertion-ordered map such as `IndexMap` would also work.)

**Verified:** a struct with several distinct primitive field types, built three times with a clean rebuild each, produced three different `.rmeta` hashes before this change and three identical hashes after.

### DIFF
--- a/rust/foxglove_derive/src/lib.rs
+++ b/rust/foxglove_derive/src/lib.rs
@@ -588,9 +588,18 @@ fn derive_named_struct_impl(
         });
     }
 
-    let enum_defs = enum_defs.into_values().collect::<Vec<_>>();
-    let message_defs = message_defs.into_values().collect::<Vec<_>>();
-    let file_defs = file_defs.into_values().collect::<Vec<_>>();
+    // Emit definitions in a deterministic order: HashMap iteration order is
+    // randomized per process, so splicing `into_values()` directly makes the
+    // generated tokens (and the consuming crate's SVH) differ build-to-build.
+    let mut enum_defs: Vec<_> = enum_defs.into_iter().collect();
+    enum_defs.sort_by_key(|(t, _)| quote::quote!(#t).to_string());
+    let enum_defs = enum_defs.into_iter().map(|(_, d)| d).collect::<Vec<_>>();
+    let mut message_defs: Vec<_> = message_defs.into_iter().collect();
+    message_defs.sort_by_key(|(t, _)| quote::quote!(#t).to_string());
+    let message_defs = message_defs.into_iter().map(|(_, d)| d).collect::<Vec<_>>();
+    let mut file_defs: Vec<_> = file_defs.into_iter().collect();
+    file_defs.sort_by_key(|(t, _)| quote::quote!(#t).to_string());
+    let file_defs = file_defs.into_iter().map(|(_, d)| d).collect::<Vec<_>>();
     let has_generics = !input.generics.params.is_empty();
 
     // Extract computation bodies so we can conditionally wrap with OnceLock


gasmith:
This looks good. I'm going to suggest that we use a BTreeMap with type-token keys instead, so that we can just iterate over them in order, instead of bouncing through two vecs.

gasmith:
Thanks for the contribution! This will be carried forward in #1311.

PR #1301, opened 2026-06-09
remote_access: Publish VP8 instead of H.264 on macOS (FLE-579)

dave-foxglove:
### Changelog

Remote access video on macOS now publishes VP8 instead of H.264, so 1080p capture is delivered at full resolution and without the intermittent pauses seen on that path.

### Docs

None.

### Description

When a Mac runs the remote access gateway, video is published over H.264 via Apple's VideoToolbox encoder. While investigating FLE-579 we observed two things on that path, both from on-the-wire stats:

- **Resolution limited to 720p.** The negotiated H.264 parameters are `profile-level-id=42e01f` (Constrained Baseline, Level 3.1), and Level 3.1 allows up to 720p30. The stream stayed at 1280×720 even with ample headroom (held 720p at a ~6 Mbps estimate on a ~900 Mbps link).
- **Intermittent pauses.** Roughly every ~16 s, encoded output paused for a few seconds while the encoder adapted to a bitrate change (frames were still handed to `capture_frame`, but `frames_encoded` stopped advancing for the duration).

Publishing the same 1080p source over VP8 reached full **1920×1080 @ 30 fps without those pauses** at ~4.3 Mbps, so this PR selects VP8 on macOS.

Elsewhere (e.g. Linux devices) it keeps H.264 so the nvenc hardware encoder can be used; macOS has no nvenc to gain from H.264, and the software VP8 encode cost is acceptable there. Linux behavior is unchanged by this PR (H.264 was already the codec there). The codec choice is per-OS at compile time today; exposing a builder override is tracked in FLE-585.

This does not change 1080p on Linux/nvenc devices, which also negotiate Level 3.1 — raising the negotiated level there is tracked separately in FLE-584.

### Tested

- **Integration regression test** `livekit_video_1080p_resolution_not_capped` (committed code): on macOS the viewer decodes the full **1920×1080 over VP8**; in CI on Linux it asserts **H.264** is selected and video flows.
- **End-to-end on party** (macOS gateway → Foxglove app in Chrome): `chrome://webrtc-internals` inbound-rtp video showed `[codec]` VP8, `[framesDecoded/s]` ≈ 30, and `freezeCount` = 1 / `totalFreezesDuration` ≈ 0.28 s (one brief startup blip, stable — no recurring pauses).

Reviewers can reproduce: run `example_remote_access` at 1080p, connect a viewer, and in `chrome://webrtc-internals` (inbound-rtp video) check `[codec]` is VP8, `frameWidth`/`frameHeight` reach 1920×1080, `[framesDecoded/s]` ≈ 30, and `totalFreezesDuration` stays near 0.

Refs: [FLE-579](https://linear.app/foxglove/issue/FLE-579/investigate-1080p-encoder-initialization-and-frame-rate-cap)

### DIFF
--- a/rust/foxglove/src/remote_access/session.rs
+++ b/rust/foxglove/src/remote_access/session.rs
@@ -2182,12 +2182,12 @@ impl RemoteAccessSession {
                 // on Linux hosts that have nvenc available. VP8/VP9/AV1 paths
                 // are software-only in our builds, so H.264 is at worst parity elsewhere.
                 //
-                // Exception: on macOS the H.264 path uses Apple's VideoToolbox hardware
-                // encoder, which negotiates H.264 level 3.1 (capping resolution at 720p) and
-                // periodically freezes for several seconds while reconfiguring on WebRTC
-                // bitrate probes (FLE-579). macOS is only a dev machine here (real devices are
-                // Linux) with no nvenc to benefit from, so prefer software VP8 there — it has
-                // no level cap and no reconfigure freeze.
+                // Exception: on macOS we publish VP8 instead. On the macOS VideoToolbox
+                // H.264 path we observed (FLE-579) that the default negotiated H.264 level
+                // (Constrained Baseline 3.1) limits the stream to 720p, and that encoded
+                // output paused for several seconds at a time while the encoder adapted to
+                // bitrate changes; VP8 reached full 1080p without those pauses. The H.264
+                // level default is not macOS-specific and is tracked separately in FLE-584.
                 //
                 // Disable simulcast. We expect viewers will be mostly homogenous, and
                 // simulcast is a lot of work for the robot without much to gain.
--- a/rust/foxglove/src/remote_access/session.rs
+++ b/rust/foxglove/src/remote_access/session.rs
@@ -2181,12 +2181,25 @@ impl RemoteAccessSession {
                 // Prefer H.264 so that the libwebrtc nvenc encoder (H.264-only) can be used
                 // on Linux hosts that have nvenc available. VP8/VP9/AV1 paths
                 // are software-only in our builds, so H.264 is at worst parity elsewhere.
+                //
+                // Exception: on macOS the H.264 path uses Apple's VideoToolbox hardware
+                // encoder, which negotiates H.264 level 3.1 (capping resolution at 720p) and
+                // periodically freezes for several seconds while reconfiguring on WebRTC
+                // bitrate probes (FLE-579). macOS is only a dev machine here (real devices are
+                // Linux) with no nvenc to benefit from, so prefer software VP8 there — it has
+                // no level cap and no reconfigure freeze.
+                //
                 // Disable simulcast. We expect viewers will be mostly homogenous, and
                 // simulcast is a lot of work for the robot without much to gain.
                 // We observed that nvenc aggressively enforces the target bitrate,
                 // and combined with simulcast results in very low quality video with compression artifacts.
+                let video_codec = if cfg!(target_os = "macos") {
+                    VideoCodec::VP8
+                } else {
+                    VideoCodec::H264
+                };
                 let publish_options = TrackPublishOptions {
-                    video_codec: VideoCodec::H264,
+                    video_codec,
                     simulcast: false,
                     ..Default::default()
                 };
--- a/rust/remote_access_tests/src/test_helpers.rs
+++ b/rust/remote_access_tests/src/test_helpers.rs
@@ -602,6 +602,33 @@ impl ViewerConnection {
         }
     }
 
+    /// Waits for a `TrackSubscribed` room event for a video track and returns the
+    /// subscribed [`RemoteVideoTrack`], so callers can read receiver-side WebRTC stats
+    /// (e.g. the actually-received frame dimensions and frame rate).
+    ///
+    /// Non-matching `DataTrackPublished` events are buffered for later consumption.
+    pub async fn expect_video_track_subscribed(
+        &mut self,
+    ) -> Result<livekit::track::RemoteVideoTrack> {
+        let deadline = tokio::time::Instant::now() + EVENT_TIMEOUT;
+        loop {
+            let event = tokio::time::timeout_at(deadline, self.events.recv())
+                .await
+                .context("timeout waiting for TrackSubscribed event")?
+                .context("room events channel closed")?;
+            match event {
+                RoomEvent::TrackSubscribed {
+                    track: livekit::track::RemoteTrack::Video(track),
+                    ..
+                } => return Ok(track),
+                RoomEvent::DataTrackPublished(track) => {
+                    self.pending_data_tracks.push(track);
+                }
+                _ => continue,
+            }
+        }
+    }
+
     /// Waits for a `TrackUnsubscribed` room event and returns the track name.
     pub async fn expect_track_unsubscribed(&mut self) -> Result<String> {
         let deadline = tokio::time::Instant::now() + EVENT_TIMEOUT;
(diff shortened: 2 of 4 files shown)


rust/foxglove/src/remote_access/session.rs
```diff
@@ -2181,12 +2181,25 @@ impl RemoteAccessSession {
...
+                //
                 // Disable simulcast. We expect viewers will be mostly homogenous, and
                 // simulcast is a lot of work for the robot without much to gain.
                 // We observed that nvenc aggressively enforces the target bitrate,
                 // and combined with simulcast results in very low quality video with compression artifacts.
+                let video_codec = if cfg!(target_os = "macos") {
+                    VideoCodec::VP8
+                } else {
+                    VideoCodec::H264
+                };
```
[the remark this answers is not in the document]
dave-foxglove: [AI] [later] Good call on the public-SDK angle. The 'dev box' framing in the comment is reworded — it now describes the observed VideoToolbox behavior plus the level cap (tracked cross-platform in FLE-584), not an internal assumption. For configurability — a video_codec override on the Gateway builder defaulting to the per-OS choice — filed FLE-585 to keep this PR scoped to the fix.

dave-foxglove:
[AI] > Small, well-scoped fix, and the description is thorough (on-the-wire evidence, the Linux/nvenc caveat called out, manual-test checklist for reviewers). One inline question.
> 
> Main open question: this is the public SDK, and the codec choice is a non-configurable compile-time `cfg!(target_os = "macos")
> ...

[later] Configurability concern tracked in FLE-585 (video_codec builder override, defaulting to the per-OS choice); comment framing reworded. Keeping this PR scoped to the VP8-on-macOS fix.

dave-foxglove:
[AI] > Re-review of f8f17e0d (prior pass was 5a60115c). The only code change since then is the softened codec comment, which addresses the framing half of my inline thread — it now reads as observed VideoToolbox behavior + the level cap tracked in FLE-584, not "macOS is just our dev box." The two test file
> ...

[resolved] The open thread (macOS opt-out) is deferred to FLE-585. No other changes.

rust/remote_access_tests/tests/livekit_test.rs
```diff
@@ -925,6 +925,179 @@ async fn livekit_video_metadata_advertised_after_image_logged() -> Result<()> {
...
+            break;
+        }
+        tokio::time::sleep(Duration::from_millis(500)).await;
+        let stats = track.get_stats().await.context("get_stats")?;
+        let mut codec = String::new();
+        for s in &stats {
+            if let livekit::webrtc::stats::RtcStats::Codec(c) = s {
+                codec = c.codec.mime_type.clone();
+            }
+        }
```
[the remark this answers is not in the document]
dave-foxglove: [AI] [declined] This libwebrtc binding's InboundRtpStreamStats doesn't expose codec_id, so there's no field to correlate against here. And it's a single receive-only track, so the report carries exactly one negotiated video codec stat — last-wins resolves to the correct one. Leaving as-is given that plus the #[ignore]d/manual nature.

dave-foxglove:
[AI] > Re-review of 2c9cf178 (prior pass was f8f17e0d). The only change since then is the regression test going platform-aware: it now asserts the published codec on every platform (VP8 on macOS, H.264 elsewhere) and the 1080p resolution only on the VP8 path. Reads correctly. One new non-blocking note inli
> ...

[resolved] Inline codec-read note: InboundRtpStreamStats has no codec_id in this binding to correlate against, and it's a single receive-only track (one codec stat), so last-wins is correct — declined. The macOS opt-out concern is tracked in FLE-585.

rust/remote_access_tests/src/test_helpers.rs
```diff
@@ -602,6 +602,33 @@ impl ViewerConnection {
         }
     }
 
+    /// Waits for a `TrackSubscribed` room event for a video track and returns the
+    /// subscribed [`RemoteVideoTrack`], so callers can read receiver-side WebRTC stats
+    /// (e.g. the actually-received frame dimensions and frame rate).
+    ///
+    /// Non-matching `DataTrackPublished` events are buffered for later consumption.
+    pub async fn expect_video_track_subscribed(
```
eloff: This is almost identical to expect_track_subscribed above, the difference being the shape of the TrackSubscribed event it matches on.

I combined them here, and it seems to be an improvement, you can merge it if you agree: https://github.com/foxglove/foxglove-sdk/pull/1303

rust/remote_access_tests/tests/livekit_test.rs
```diff
@@ -925,6 +925,178 @@ async fn livekit_video_metadata_advertised_after_image_logged() -> Result<()> {
...
+    let channel_id = advertise.channels[0].id;
+
+    viewer
+        .subscribe_video_and_wait(&[channel_id], &video_channel)
+        .await?;
+    let track = viewer.expect_video_track_subscribed().await?;
+    info!("subscribed to video track {}", track.name());
+
+    // Pump 1080p frames at ~30fps from a background task until told to stop.
+    let frame = encode_raw_image_sized("camera_optical_frame", WIDTH, HEIGHT);
```
eloff: Nit: Every frame is identical, which is the holy grail for the video compression. Probably fine, but might undermine the test in some circumstances (or at least I can't be certain that it wouldn't.)

It's probably fine, but if you wanted to be robust against that you could add an integer counter argument to encode_raw_image_sized which varies the gradient, and then generate the frames in the loop instead.

rust/remote_access_tests/tests/livekit_test.rs
```diff
@@ -925,6 +925,178 @@ async fn livekit_video_metadata_advertised_after_image_logged() -> Result<()> {
...
+            last_codec.eq_ignore_ascii_case("video/VP8"),
+            "macOS should publish VP8, got {last_codec}"
+        );
+        assert_eq!(
+            (w, h),
+            (WIDTH, HEIGHT),
+            "VP8 on macOS should deliver the full {WIDTH}x{HEIGHT} (FLE-579); got {w}x{h}"
+        );
+    } else {
+        // Non-macOS keeps H.264 for nvenc; it is still level-3.1 capped (FLE-584), so we
```
eloff, line 1098: We should reach out to livekit if there is a knob we can turn to change this, or if they can make a knob for that. It'd be nice not to have this limit.

eloff (approved):
Some comments and suggestions, but all optional. Looks good.

dave-foxglove:
[AI] > Re-review of 47f5dabf (prior pass was 2c9cf178). The only changes since then are test-only: the receiver-stats polling collapsed into a single `match` pass, the gradient doc comment got more precise, and the frame-pump join now surfaces errors (`pump.await.expect("stopping frame pump")`) instead of 
> ...

[resolved] Acknowledged — thanks for the re-review.

rust/remote_access_tests/tests/livekit_test.rs
dave-foxglove: [AI] [later] Agreed — raising the negotiated H.264 level (and asking livekit about a knob for it) is tracked in FLE-584.

rust/remote_access_tests/src/test_helpers.rs
dave-foxglove: [AI] [resolved] Agreed — merged #1303; the helpers are now combined into expect_track_subscribed.

rust/remote_access_tests/tests/livekit_test.rs
dave-foxglove: [AI] [resolved] Good catch — the pump now cycles through distinct shifted-gradient frames so they're not identical (pre-encoded up front to avoid blocking the runtime in the loop).

PR #1305, opened 2026-06-10
remote_access: Add a FOXGLOVE_VIDEO_CODEC override for published video (FLE-587)

dave-foxglove:
### Changelog

Remote access: the `FOXGLOVE_VIDEO_CODEC` environment variable overrides the default codec for published video tracks. This is a developer aid, not a supported configuration surface. Platform defaults are unchanged (VP8 on macOS, H.264 elsewhere).

### Docs

One note on `Gateway::start()` in the Rust docs, where the variable is read — matching how `FOXGLOVE_API_URL` and `FOXGLOVE_API_TIMEOUT` are documented (Rust builder methods only, no C++/Python/crate-level mentions). The accepted values are documented on the internal parser.

### Description

FLE-587 set out to explore H.265 (HEVC) as the macOS default instead of VP8. The exploration found H.265 strong on the encode side — full 1080p with a VideoToolbox hardware encode path, versus VP8's software-only path (details in the issue):

| Metric (1080p test card) | VP8 (#1301) | H.265 |
|---|---|---|
| Resolution delivered | 1920×1080 | 1920×1080 |
| Frames decoded/s | ~30 | 31 |
| Freezes | 1 brief (~0.28 s) | 0 |
| Mac encode path | software | VideoToolbox hardware |

But the gateway publishes a single track with no codec fallback at the SFU, so a viewer that cannot decode the published codec gets no video — and browser HEVC decode support is not broad enough (no Firefox; Chrome only with a platform HEVC decoder) for a default whose goal is to "just work" in a browser. So **VP8 stays the macOS default**, and this PR ships the supporting work from the exploration:

1. **`FOXGLOVE_VIDEO_CODEC` escape hatch.** Read at `Gateway::start()` alongside `FOXGLOVE_DEVICE_TOKEN` / `FOXGLOVE_API_URL` / `FOXGLOVE_API_TIMEOUT`; an unrecognized value logs a warning and the default is used. The Rust core reads the variable, so Python and C++ gateways honor it with doc-only changes. H.265 remains reachable on macOS via `FOXGLOVE_VIDEO_CODEC=h265`.
2. **`DEFAULT_VIDEO_CODEC` constant.** The per-OS default moves out of `start_video_tracks` into a compile-time constant that carries the rationale, including the H.265 evaluation outcome.
3. **Codec in publish logs.** The publish success/failure log lines include the selected codec, so a codec that the host cannot encode is diagnosable from device logs.

Richer codec negotiation is tracked separately (FLE-585 builder override, FLE-593 viewer-driven preferences).

### Tested

- Unit tests for the env-value parser (all five codecs, case-insensitivity, invalid values) — pass locally.
- Integration test `livekit_video_codec_env_override`: with `FOXGLOVE_VIDEO_CODEC=vp9` (differs from the default on every platform), the negotiated inbound codec is `video/VP9`, exercising the override end-to-end. Runs in the remote-access-tests CI workflow; previously verified locally with the VP8 value.
- `cargo check` / clippy / fmt clean with the `remote-access` feature.
- H.265 exploration was verified end-to-end on party (macOS gateway → Foxglove app in Chrome) at 1080p: `chrome://webrtc-internals` inbound-rtp showed `[codec]` H265, 1920×1080 @ 31 fps, `freezeCount` 0.

Refs: [FLE-587](https://linear.app/foxglove/issue/FLE-587/explore-hevc-h265-instead-of-vp8-for-macos-remote-access-video)

### DIFF
--- a/cpp/foxglove/include/foxglove/remote_access.hpp
+++ b/cpp/foxglove/include/foxglove/remote_access.hpp
@@ -256,12 +256,9 @@ struct RemoteAccessGatewayOptions {
 /// The gateway connects to the Foxglove platform and allows remote clients to
 /// subscribe to channels and receive data.
 ///
-/// Published video tracks use a per-OS default codec (H.265 on macOS, H.264 elsewhere). The
-/// `FOXGLOVE_VIDEO_CODEC` environment variable overrides this default; it accepts `vp8`,
-/// `h264`, `vp9`, `av1`, or `h265` (case-insensitive; H.265 is the codec also known as
-/// HEVC). An unrecognized value logs a warning
-/// and the default is used. If the host platform cannot encode the selected codec, viewers
-/// receive no video; the device logs the selected codec when publishing each video track.
+/// The `FOXGLOVE_VIDEO_CODEC` environment variable overrides the default codec for published
+/// video tracks. This is a developer aid and an escape hatch, not a supported configuration
+/// surface; selecting a codec the host cannot encode leaves viewers without video.
 ///
 /// @note RemoteAccessGateway is fully thread-safe, but RemoteAccessGatewayCallbacks may be invoked
 /// concurrently from multiple threads, so you will need to use synchronization in your callbacks.
--- a/cpp/foxglove/include/foxglove/remote_access.hpp
+++ b/cpp/foxglove/include/foxglove/remote_access.hpp
@@ -256,6 +256,11 @@ struct RemoteAccessGatewayOptions {
 /// The gateway connects to the Foxglove platform and allows remote clients to
 /// subscribe to channels and receive data.
 ///
+/// Published video tracks use a per-OS default codec (H.265 on macOS, H.264 elsewhere). The
+/// `FOXGLOVE_VIDEO_CODEC` environment variable overrides this default; it accepts `vp8`,
+/// `h264`, `vp9`, `av1`, or `h265` (case-insensitive). An unrecognized value logs a warning
+/// and the default is used.
+///
 /// @note RemoteAccessGateway is fully thread-safe, but RemoteAccessGatewayCallbacks may be invoked
 /// concurrently from multiple threads, so you will need to use synchronization in your callbacks.
 class RemoteAccessGateway final {
--- a/cpp/foxglove/include/foxglove/remote_access.hpp
+++ b/cpp/foxglove/include/foxglove/remote_access.hpp
@@ -256,10 +256,6 @@ struct RemoteAccessGatewayOptions {
 /// The gateway connects to the Foxglove platform and allows remote clients to
 /// subscribe to channels and receive data.
 ///
-/// The `FOXGLOVE_VIDEO_CODEC` environment variable overrides the default codec for published
-/// video tracks. This is a developer aid and an escape hatch, not a supported configuration
-/// surface; selecting a codec the host cannot encode leaves viewers without video.
-///
 /// @note RemoteAccessGateway is fully thread-safe, but RemoteAccessGatewayCallbacks may be invoked
 /// concurrently from multiple threads, so you will need to use synchronization in your callbacks.
 class RemoteAccessGateway final {
--- a/cpp/foxglove/include/foxglove/remote_access.hpp
+++ b/cpp/foxglove/include/foxglove/remote_access.hpp
@@ -258,7 +258,8 @@ struct RemoteAccessGatewayOptions {
 ///
 /// Published video tracks use a per-OS default codec (H.265 on macOS, H.264 elsewhere). The
 /// `FOXGLOVE_VIDEO_CODEC` environment variable overrides this default; it accepts `vp8`,
-/// `h264`, `vp9`, `av1`, or `h265` (case-insensitive). An unrecognized value logs a warning
+/// `h264`, `vp9`, `av1`, or `h265` (case-insensitive; H.265 is the codec also known as
+/// HEVC). An unrecognized value logs a warning
 /// and the default is used. If the host platform cannot encode the selected codec, viewers
 /// receive no video; the device logs the selected codec when publishing each video track.
 ///
--- a/cpp/foxglove/include/foxglove/remote_access.hpp
+++ b/cpp/foxglove/include/foxglove/remote_access.hpp
@@ -259,7 +259,8 @@ struct RemoteAccessGatewayOptions {
 /// Published video tracks use a per-OS default codec (H.265 on macOS, H.264 elsewhere). The
 /// `FOXGLOVE_VIDEO_CODEC` environment variable overrides this default; it accepts `vp8`,
 /// `h264`, `vp9`, `av1`, or `h265` (case-insensitive). An unrecognized value logs a warning
-/// and the default is used.
+/// and the default is used. If the host platform cannot encode the selected codec, the video
+/// track fails to publish; the failure is logged on the device and viewers receive no video.
 ///
 /// @note RemoteAccessGateway is fully thread-safe, but RemoteAccessGatewayCallbacks may be invoked
 /// concurrently from multiple threads, so you will need to use synchronization in your callbacks.
--- a/cpp/foxglove/include/foxglove/remote_access.hpp
+++ b/cpp/foxglove/include/foxglove/remote_access.hpp
@@ -259,8 +259,8 @@ struct RemoteAccessGatewayOptions {
 /// Published video tracks use a per-OS default codec (H.265 on macOS, H.264 elsewhere). The
 /// `FOXGLOVE_VIDEO_CODEC` environment variable overrides this default; it accepts `vp8`,
 /// `h264`, `vp9`, `av1`, or `h265` (case-insensitive). An unrecognized value logs a warning
-/// and the default is used. If the host platform cannot encode the selected codec, the video
-/// track fails to publish; the failure is logged on the device and viewers receive no video.
+/// and the default is used. If the host platform cannot encode the selected codec, viewers
+/// receive no video; the device logs the selected codec when publishing each video track.
 ///
 /// @note RemoteAccessGateway is fully thread-safe, but RemoteAccessGatewayCallbacks may be invoked
 /// concurrently from multiple threads, so you will need to use synchronization in your callbacks.
--- a/rust/foxglove/src/remote_access/gateway.rs
+++ b/rust/foxglove/src/remote_access/gateway.rs
@@ -445,6 +445,8 @@ impl Gateway {
     /// The published video codec defaults to a per-OS choice, and can be overridden with the
     /// `FOXGLOVE_VIDEO_CODEC` environment variable (one of `vp8`, `h264`, `vp9`, `av1`,
     /// `h265`, case-insensitive). An unrecognized value logs a warning and the default is used.
+    /// If the host platform cannot encode the selected codec, the video track fails to
+    /// publish; the failure is logged on the device and viewers receive no video.
     pub fn start(mut self) -> Result<GatewayHandle, FoxgloveError> {
         crate::crypto::install_default_crypto_provider();
 
@@ -465,6 +467,8 @@ impl Gateway {
                 .and_then(|s| s.parse::<u64>().ok())
                 .map(Duration::from_secs)
         });
+        // A builder-level codec override is deferred to FLE-585; this environment variable
+        // is the configuration surface until then.
         let video_codec_override = std::env::var(FOXGLOVE_VIDEO_CODEC_ENV).ok().and_then(|s| {
             let codec = parse_video_codec(&s);
             if codec.is_none() {
--- a/rust/foxglove/src/remote_access/gateway.rs
+++ b/rust/foxglove/src/remote_access/gateway.rs
@@ -1,6 +1,7 @@
 use std::{collections::HashMap, fmt::Display, future::Future, sync::Arc, time::Duration};
 
 use indexmap::IndexSet;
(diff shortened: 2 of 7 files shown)


rust/foxglove/src/remote_access/session.rs
```diff
@@ -2182,19 +2182,24 @@ impl RemoteAccessSession {
...
+                // landed in PR #1301, to see whether the VideoToolbox HEVC encoder
+                // avoids the H.264 issues while keeping a hardware encode path.
                 //
                 // Disable simulcast. We expect viewers will be mostly homogenous, and
                 // simulcast is a lot of work for the robot without much to gain.
                 // We observed that nvenc aggressively enforces the target bitrate,
                 // and combined with simulcast results in very low quality video with compression artifacts.
                 let video_codec = if cfg!(target_os = "macos") {
-                    VideoCodec::VP8
+                    VideoCodec::H265
```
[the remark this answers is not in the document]
dave-foxglove: [AI] [later] Confirmed, and it holds at the SDK level too: the locked livekit 0.7.42 `TrackPublishOptions` has no backup-codec field, so the single H.265 track is all the SFU can forward — a viewer that can't decode HEVC gets no video, not degraded video. This is the stated land/no-land gate in the PR description and FLE-587: before this graduates from exploration, the non-HEVC viewer outcome (Firefox / Chrome-on-Linux against the party rig) needs to be confirmed, and routing goes through the FLE-585 builder codec override or a multi-codec approach. Tracked in FLE-587.

dave-foxglove:
[AI] > Exploration framing is clear, and the code change itself is small and does what it says. One inline finding on the actual blocker — viewer-side HEVC support / lack of fallback — which is your own open question, but I wanted to pin down that the code offers *no* fallback, so the failure mode is "no v
> ...

[resolved] Inline thread addressed (viewer-support gate tracked in FLE-587); the comment-hygiene notes were fixed in 7c32580e as your re-review confirmed.

dave-foxglove:
[AI] > Re-reviewed after the rebase + the "state the codec choice factually" commit. The diff is now just the two codec files (`session.rs` + `livekit_test.rs`); the comment hygiene I flagged last time is fixed — `EXPLORATION:` blocks and the `#1301` references are gone, and the remaining `FLE-*` refs matc
> ...

[resolved] The remaining open item (viewer-support gate) is tracked in FLE-587; inline thread resolved accordingly.

dave-foxglove:
[AI] > Re-reviewed at 4ea81f92. The only changes since my last pass (7c32580e) are two comment refinements, and they're both accurate:
> 
> - The test comment now honestly states H.265 *does* have level-based caps like H.264, but that level 6.0 (`level-id=180` → 180/30 = 6.0) is what's negotiated, which covers
> ...

[resolved] Acknowledged — no new findings.

rust/foxglove/src/remote_access/gateway.rs
```diff
@@ -444,6 +465,16 @@ impl Gateway {
                 .and_then(|s| s.parse::<u64>().ok())
                 .map(Duration::from_secs)
         });
+        let video_codec_override = std::env::var(FOXGLOVE_VIDEO_CODEC_ENV).ok().and_then(|s| {
```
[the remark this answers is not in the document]
dave-foxglove: [AI] [resolved] Addressed in the prior commit plus 3e9d04a9: the docs (Rust/C++/Python) now state the chosen codec must be encodable on the host and that viewers otherwise receive no video, and the publish success/failure log messages now include the selected codec, so a valid-but-unencodable pick is diagnosable from device logs. The docs state the observable outcome rather than the publish mechanics, since the exact failure path for an unencodable codec hasn't been verified on a non-macOS host. Declining the hevc alias: the accepted set deliberately matches livekit's canonical codec names exactly, and the warning for a rejected value lists the valid names, so a user who types hevc is pointed straight at h265.

dave-foxglove:
[AI] > Re-reviewed at 28501b30. This is the same change rebased onto current main — the two codec files are byte-for-byte identical to my last pass (4ea81f92), so there's no new code to review.
> 
> Still open, unchanged from prior:
> 
> - The land/no-land gate is still the viewer-support question in the open inli
> ...

[resolved] Recap of the pre-override state; the viewer-support gate is tracked in FLE-587 and the FOXGLOVE_VIDEO_CODEC escape hatch has since landed on the branch.

dave-foxglove:
[AI] > Reviewed the new `FOXGLOVE_VIDEO_CODEC` override commit (8d3a27a7) — the only new code since my last pass. The plumbing (gateway → connection → session → publish) is clean, the parse fn has a unit test, and the override has an end-to-end integration test. One new inline finding (override doesn't val
> ...

[resolved] All four items addressed: the PR title and description were rewritten after this review (they now cover the env override, the three-language docs, and the new integration test, with a real changelog entry); the H.265-on-macOS default is now the intended change of this PR rather than an open exploration question, so the docs stating it match the intent; a footnote under the metrics table now notes the bitrate row reflects mostly-static test-card content; and the inline thread on host encode support is resolved in 3e9d04a9 (docs + codec in publish logs).

cpp/foxglove/include/foxglove/remote_access.hpp
```diff
@@ -256,6 +256,13 @@ struct RemoteAccessGatewayOptions {
 /// The gateway connects to the Foxglove platform and allows remote clients to
 /// subscribe to channels and receive data.
 ///
+/// Published video tracks use a per-OS default codec (H.265 on macOS, H.264 elsewhere). The
+/// `FOXGLOVE_VIDEO_CODEC` environment variable overrides this default; it accepts `vp8`,
+/// `h264`, `vp9`, `av1`, or `h265` (case-insensitive; H.265 is the codec also known as
+/// HEVC). An unrecognized value logs a warning
+/// and the default is used. If the host platform cannot encode the selected codec, viewers
```
gasmith: Nit: Weird line break there. (And elsewhere in this diff.)

rust/foxglove/src/remote_access/session.rs
```diff
@@ -2182,22 +2187,30 @@ impl RemoteAccessSession {
...
-                    VideoCodec::H264
-                };
+                let video_codec =
+                    session
+                        .video_codec_override
+                        .unwrap_or(if cfg!(target_os = "macos") {
+                            VideoCodec::H265
+                        } else {
+                            VideoCodec::H264
+                        });
```
gasmith: Could we pull this out into a compile-time constant, and move the commentary about the choice of default there?

cpp/foxglove/include/foxglove/remote_access.hpp
```diff
@@ -256,6 +256,13 @@ struct RemoteAccessGatewayOptions {
 /// The gateway connects to the Foxglove platform and allows remote clients to
 /// subscribe to channels and receive data.
 ///
+/// Published video tracks use a per-OS default codec (H.265 on macOS, H.264 elsewhere). The
+/// `FOXGLOVE_VIDEO_CODEC` environment variable overrides this default; it accepts `vp8`,
+/// `h264`, `vp9`, `av1`, or `h265` (case-insensitive; H.265 is the codec also known as
+/// HEVC). An unrecognized value logs a warning
```
gasmith: Do we really need to explain here that H.265 is also known as HEVC?

cpp/foxglove/include/foxglove/remote_access.hpp
```diff
@@ -256,6 +256,13 @@ struct RemoteAccessGatewayOptions {
 /// The gateway connects to the Foxglove platform and allows remote clients to
 /// subscribe to channels and receive data.
 ///
+/// Published video tracks use a per-OS default codec (H.265 on macOS, H.264 elsewhere). The
+/// `FOXGLOVE_VIDEO_CODEC` environment variable overrides this default; it accepts `vp8`,
+/// `h264`, `vp9`, `av1`, or `h265` (case-insensitive; H.265 is the codec also known as
```
gasmith: I find it odd that these choices are not sorted alphabetically. I also wonder whether we want to advertise all of them. Do we want to claim support for all of them?

gasmith (commented):
Looks good, though I have a few comments.

I'm somewhat inclined to leave this as an environment variable for now, and tone down or even omit the documentation. I don't want to claim broad support for codecs that we might not want to continue supporting in the future. We should think of this as a developer-aid, and an escape hatch if we need it in the field.

Choosing [H265](https://caniuse.com/?search=h265) by default on MacOS gives me a little pause, because I'm not sure how much we can depend on browser support. Our main objective on MacOS is to make things "just work" when people are kicking the tires. If video doesn't show up when they try the SDK example, that's going to be a bad user experience. I think we should stick with VP8, but I'm open to hear other opinions.

cpp/foxglove/include/foxglove/remote_access.hpp
dave-foxglove: [AI] [resolved] Rewrapped — the docstring is now a short, properly wrapped paragraph (the awkward breaks came from an earlier edit that shortened lines without rewrapping).

rust/foxglove/src/remote_access/session.rs
dave-foxglove: [AI] [resolved] Pulled it out into a DEFAULT_VIDEO_CODEC constant and moved the rationale commentary there, including the outcome of the H.265 evaluation.

cpp/foxglove/include/foxglove/remote_access.hpp
dave-foxglove: [AI] [resolved] Dropped the HEVC explainer, along with the rest of the codec list in the public docs.

cpp/foxglove/include/foxglove/remote_access.hpp
dave-foxglove: [AI] [resolved] Removed the codec list from the public docstrings entirely, so we no longer advertise specific codecs. The accepted values are now documented only on the internal parser and the invalid-value warning, alphabetized.

dave-foxglove:
[AI] > Looks good, though I have a few comments.
> 
> I'm somewhat inclined to leave this as an environment variable for now, and tone down or even omit the documentation. I don't want to claim broad support for codecs that we might not want to continue supporting in the future. We should think of this as a 
> ...

[resolved] Agreed on both points. Reverted the macOS default to VP8 (the per-OS default now lives in a DEFAULT_VIDEO_CODEC constant whose commentary records the H.265 evaluation outcome), and trimmed the public docs to a brief escape-hatch mention with no codec list. The PR title and description now reflect the narrower scope; H.265 remains reachable via FOXGLOVE_VIDEO_CODEC=h265.

dave-foxglove:
[AI] > Re-reviewed at 3e9d04a9. Two new commits since my last pass (the override commit): the unsupported-codec doc note (Rust/C++/Python + gateway doc) and the codec-in-publish-logs change. Both are small and correct, and they close my prior gateway.rs finding — a valid-but-unencodable pick is now diagnos
> ...

[resolved] Stale recap of resolved inline threads — superseded by later reviews.

dave-foxglove:
[AI] > Re-reviewed at 29341bd4. This is the override + unsupported-codec doc note + publish-logs change rebased onto current main — the diff is byte-identical to my last pass (3e9d04a9), so no new code to review. Both prior inline threads are resolved: the gateway.rs host-encode-support finding (docs now s
> ...

[resolved] Stale recap — superseded by later reviews.

dave-foxglove:
[AI] > Re-reviewed at b4d40e8c. The only change since my last pass (29341bd4) is the "H.265 is also known as HEVC" clarification across the four docstrings and the invalid-value warning — accurate, no new code findings. Both prior inline threads (viewer-support gate, host-encode validation) are resolved.
> 
> ...

[resolved] The macOS default has since been reverted to VP8 per review, so the viewer-support concern about an H.265 default no longer applies.

dave-foxglove:
[AI] > Re-reviewed at 975292c9. The only change since my last pass (b4d40e8c) is dropping the `FLE-587:` prefix from the `livekit_video_codec_env_override` test doc comment — that's the right call per AGENTS.md (no issue refs in comments), no new code to review. Both prior inline threads (viewer-support ga
> ...

[resolved] The macOS default has since been reverted to VP8 per review, so the open H.265-default items no longer apply.

dave-foxglove:
[AI] > No new code findings — the codec diff is byte-identical to my last pass (the rebase only pulled in unrelated `main` changes), and both my prior inline threads (viewer-support gate, host-encode validation) are resolved. So this pass is about the thing that actually changed: this left draft, which tur
> ...

[resolved] The macOS default has since been reverted to VP8, so the default-flip decision this raised no longer applies.

dave-foxglove:
[AI] > Re-reviewed at aac7070d. The thing that's been the gate across every prior pass — the macOS default flipping VP8 → H.265 — is gone: this commit keeps VP8 as the macOS default, so the PR is now purely additive (env override + codec-in-publish-logs + the `DEFAULT_VIDEO_CODEC` refactor) rather than a d
> ...

[resolved] Acknowledged — no findings to address.

dave-foxglove:
[AI] > Looks good, though I have a few comments.
> 
> I'm somewhat inclined to leave this as an environment variable for now, and tone down or even omit the documentation. I don't want to claim broad support for codecs that we might not want to continue supporting in the future. We should think of this as a 
> ...

[resolved] Follow-up: trimmed further to match repo precedent — the env var is now documented only on Gateway::start in the Rust docs (like FOXGLOVE_API_URL / FOXGLOVE_API_TIMEOUT), with no C++/Python/crate-level mentions.

gasmith (approved):

PR #1308, opened 2026-06-11
Format library identifiers for consistency with mcap

amacneil:
### Changelog

Library identifiers and the Rust API client user agent now use user-agent-like tokens, including the Foxglove SDK version without a `v` prefix, and the MCAP Rust SDK version where appropriate.

### Docs

None.

### Description

This updates Foxglove SDK library identifiers from `foxglove-sdk-<language>/v<version>` to `foxglove-sdk-<language>/<version>`, for consistency with the MCAP libraries.

MCAP file headers append the MCAP Rust SDK token, producing `foxglove-sdk-<language>/<version> mcap-rust/<version>`. The MCAP token is sourced from `mcap::LIBRARY_IDENTIFIER` so the MCAP Rust SDK owns its identifier formatting.

WebSocket server, remote-access gateway `fg-library` metadata, and the Rust API client user agent advertise only the Foxglove SDK token.

### DIFF
--- a/rust/foxglove/src/library_version.rs
+++ b/rust/foxglove/src/library_version.rs
@@ -43,14 +43,16 @@ mod tests {
 
     #[test]
     fn library_version_uses_user_agent_like_tokens() {
+        let library_version = get_library_version();
+        let tokens = library_version.split(' ').collect::<Vec<_>>();
+
+        assert_eq!(tokens.len(), 2);
         assert_eq!(
-            get_library_version(),
-            format!(
-                "foxglove-sdk-{}/{} {}",
-                get_sdk_language(),
-                get_sdk_version(),
-                mcap::LIBRARY_IDENTIFIER
-            )
+            tokens[0],
+            concat!("foxglove-sdk-rust/", env!("CARGO_PKG_VERSION"))
         );
+        assert!(!tokens[0].contains("/v"));
+        assert_eq!(tokens[1], mcap::LIBRARY_IDENTIFIER);
+        assert!(tokens[1].starts_with("mcap-rust/"));
     }
 }
--- a/rust/foxglove/src/library_version.rs
+++ b/rust/foxglove/src/library_version.rs
@@ -29,7 +29,7 @@ pub(crate) fn get_sdk_version() -> &'static str {
 /// Returns a user-agent-like SDK identifier for this library, for use in log sinks
 /// and wire-visible metadata.
 /// Note that `set_sdk_language` must be called before this for it to have an effect.
-pub(crate) fn get_library_version() -> String {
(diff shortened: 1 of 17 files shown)


rust/foxglove/src/library_version.rs
```diff
@@ -29,5 +29,28 @@ pub(crate) fn get_sdk_version() -> &'static str {
 /// Returns an identifier for this library, for use in log sinks.
 /// Note that `set_sdk_language` must be called before this for it to have an effect.
 pub(crate) fn get_library_version() -> String {
-    format!("foxglove-sdk-{}/v{}", get_sdk_language(), get_sdk_version())
+    format!(
+        "foxglove-sdk-{}/{} {}",
+        get_sdk_language(),
+        get_sdk_version(),
+        mcap::LIBRARY_IDENTIFIER
+    )
```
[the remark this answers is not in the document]
amacneil: Yes this is intentional, it matches the mcap library. Nothing parses this today, it's purely informational.

rust/foxglove/src/library_version.rs
```diff
@@ -29,5 +29,28 @@ pub(crate) fn get_sdk_version() -> &'static str {
 /// Returns an identifier for this library, for use in log sinks.
 /// Note that `set_sdk_language` must be called before this for it to have an effect.
 pub(crate) fn get_library_version() -> String {
-    format!("foxglove-sdk-{}/v{}", get_sdk_language(), get_sdk_version())
+    format!(
+        "foxglove-sdk-{}/{} {}",
+        get_sdk_language(),
+        get_sdk_version(),
+        mcap::LIBRARY_IDENTIFIER
```
gasmith: Including this for mcap files makes sense, but not so much for the library metadata we advertise from the WebSocket server and remote access gateway, or the user agent we present from the api client.

I wonder whether it would've been better for the mcap interface for attaching library metadata to be additive, rather than an override.

amacneil: I thought about adding a new API to prepend versions, but its easy enough to do that using `mcap::LIBRARY_IDENTIFIER`, and matches what other mcap language libraries expose.

Will remove the mcap version from websocket server.

gasmith (approved):
Looks good. I was surprised to see that the API client (rust/foxglove/src/api_client/client.rs) is actually making its own user agent string, instead of using get_library_version(). Thoughts on updating that too, while we're here?

I just had claude take a look, and we are not parsing the user-agent string anywhere in the API server. We _do_ parse the user-agent string for data platform APIs authenticated by site-token, and for those we expect `<service> (<version>)`.

gasmith (approved):

PR #1311, opened 2026-06-11
rust: Emit Encode schema definitions in deterministic order

### LINKED ISSUE #1297: derive(Encode) is non-deterministic: schema definitions emitted in HashMap order (changes consuming crate SVH)

gasmith:
### Changelog
- rust: Emit Encode schema definitions in deterministic order

### Docs
None

### Description
This PR is carried forward from @pchikku's #1298.

The idea is to ensure that the Encode macro output is deterministic, by using a BTreeMap instead of a hashmap to collect the various definitions that it emits.

This change also fixes a pre-existing issue in which fields with container types (`Vec<T>`, `Option<T>`, `[T;N]`) were not being deduplicated properly. We now unwrap the container and dedupe based on element type.

### Links
Carried forward from: #1298 
Fixes: #1297

### DIFF
--- a/rust/foxglove_derive/tests/deterministic_order_test.rs
+++ b/rust/foxglove_derive/tests/deterministic_order_test.rs
@@ -1,0 +1,64 @@
+use ::foxglove::Encode;
+use prost::Message;
+
+// Ensure the macro properly references the foxglove crate
+mod foxglove {}
+
+#[derive(Encode)]
+struct Zulu {
+    v: u32,
+}
+
+#[derive(Encode)]
+struct Alpha {
+    v: u32,
+}
+
+#[derive(Encode)]
+struct Mike {
+    v: u32,
+}
+
+#[derive(Debug, Clone, Copy, Encode)]
+#[allow(dead_code)]
+enum Yankee {
+    A = 0,
+}
+
+#[derive(Debug, Clone, Copy, Encode)]
+#[allow(dead_code)]
+enum Bravo {
+    A = 0,
+}
+
+// Fields are declared in a deliberately non-alphabetical order. The generated
+// schema must emit the nested message and enum definitions sorted by type name,
+// regardless of declaration order, so that the encoded schema bytes are stable
+// across builds.
+#[derive(Encode)]
+struct Container {
+    zulu: Zulu,
+    yankee: Yankee,
+    alpha: Alpha,
+    bravo: Bravo,
+    mike: Mike,
+}
+
+#[test]
+fn nested_definitions_are_deterministically_ordered() {
+    let schema = Container::get_schema().expect("schema");
+    let fds = prost_types::FileDescriptorSet::decode(schema.data.as_ref()).expect("decode schema");
+
+    let container = fds
+        .file
+        .iter()
+        .flat_map(|f| f.message_type.iter())
+        .find(|m| m.name() == "Container")
+        .expect("Container descriptor");
+
+    let nested: Vec<&str> = container.nested_type.iter().map(|n| n.name()).collect();
+    assert_eq!(nested, ["Alpha", "Mike", "Zulu"]);
+
+    let enums: Vec<&str> = container.enum_type.iter().map(|e| e.name()).collect();
+    assert_eq!(enums, ["Bravo", "Yankee"]);
+}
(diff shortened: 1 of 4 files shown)


rust/foxglove_derive/tests/deterministic_order_test.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    let container = fds
+        .file
+        .iter()
+        .flat_map(|f| f.message_type.iter())
+        .find(|m| m.name() == "Container")
+        .expect("Container descriptor");
+
+    let nested: Vec<&str> = container.nested_type.iter().map(|n| n.name()).collect();
+    assert_eq!(nested, ["Alpha", "Mike", "Zulu"]);
```
dave-foxglove, line 60: IIUC, adding `alphas: Vec<Alpha>` would create a new entry which is sorted under "V" (for Vec). So this would then be `["Alpha", "Mike", "Alpha", "Zulu"]`. Doing `opt_alpha: Option<Alpha>` might also create another sorted by "O" (for Option). They would not be de-duped because their key name would be something like "Vec < Alpha >" or similar.

It looks like this was the case before this PR, so maybe it's out of scope (or maybe I'm wrong/missing something).

Regardless, this _is_ a deterministic order which is what the PR is about.

dave-foxglove (approved):

rust/foxglove_derive/tests/deterministic_order_test.rs
gasmith: Huh! That's a good catch. What we really want is the underlying type, not the container. That problem predates this PR, but it's probably worth fixing.

gasmith: Fixed.

PR #1312, opened 2026-06-11
ci: Add netem-perlink UDP path diagnostics (FLE-595)

dave-foxglove:
### Changelog

None

### Docs

None

### Description

**Do not merge as-is — CI diagnostics for FLE-595.**

The `netem-perlink-test` job intermittently fails (~60% of runs since June 11 00:00 UTC, across all branches including `main`) with `perlink_infra_link_a_has_more_packet_loss_than_link_b` observing ~98-99% UDP echo loss on both links, while the configured netem loss is 5% / 0%.

Evidence gathered so far points away from the tc setup under test and toward the runner environment policing forwarded UDP:

- TCP echoes to the same targets never fail in failing runs, and the flat `netem-test` job (UDP via host port mapping rather than the bridge FORWARD path) always passes.
- Only 1-3 of 130 datagrams survive per ~65 s measurement window, and link A (measured first) received >= link B in all 14 failing runs — consistent with a small-burst token-bucket limit, not random loss.
- Ruled out by direct comparison of passing vs failing runs: repo changes, Docker image digests, Alpine socat/iproute2 versions, runner image version, Azure region.

This PR adds a diagnostic script that localizes where the datagrams die — tc qdisc/class/filter counters, tcpdump at both ends of the echo path, NIC offload settings, host firewall rulesets with packet counters, and the eBPF program list — and runs it in the `netem-perlink-test` job before the tests and again on failure. The plan is to re-run this job a few times (failure odds ~60%), read the diagnostics from a failing run, then land the actual fix and drop or fold in these changes.

Relates to: [FLE-595](https://linear.app/foxglove/issue/FLE-595/ci-netem-perlink-test-flake-both-links-show-99percent-udp-loss-spiked)

### DIFF
--- a/.github/workflows/remote-access-tests.yml
+++ b/.github/workflows/remote-access-tests.yml
@@ -120,8 +120,15 @@ jobs:
           -f docker-compose.netem-perlink.yml
           up -d --wait
 
+      - name: Netem per-link diagnostics before tests (FLE-595)
+        run: ./scripts/netem_perlink_diag.sh
+
       - name: Run per-link netem integration tests
         run: cargo test -p remote_access_tests --features require-cuda -- --ignored perlink_
+
+      - name: Netem per-link diagnostics on failure (FLE-595)
+        if: failure()
+        run: ./scripts/netem_perlink_diag.sh
 
       - name: Capture LiveKit logs on failure
         if: failure()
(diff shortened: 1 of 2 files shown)


dave-foxglove:
[AI] Closing unmerged — this diagnostic instrumentation served its purpose. The failing run it captured (https://github.com/foxglove/foxglove-sdk/actions/runs/27372641137) pinpointed the root cause (the forking socat UDP echo servers wedging; see FLE-595), which is fixed in #1314.

PR #1314, opened 2026-06-11
netem: Replace forking UDP echo servers with a stateless loop (FLE-595)

dave-foxglove:
### Changelog

None (CI test infrastructure only).

### Docs

None.

### Description

Fixes the `netem-perlink-test` flake that has been failing ~60% of Remote Access Tests runs since June 11 (all branches, including `main`).

**Root cause** (confirmed by instrumenting a failing run via #1312): the tc/netem hierarchy under test was healthy — counters showed correct per-link classification and exactly the configured 5%/0% drops. The wedge was in the test's own UDP echo servers. The forking socat UDP echo shares its bound socket between the parent and forked children; the socat *client* sends a 0-length datagram at stdin EOF, and netem jitter can reorder it ahead of the payload. Under the right ordering a child lingers, steals datagrams sent by subsequent clients, and never replies — from then on nearly every echo times out. tcpdump on a failing run showed all 58 request datagrams arriving at the target while the target emitted only 3 replies, with a stale echo-server child visible in the process list. This also explains why both links always degraded together (same server design on both targets) and why TCP echoes kept working (per-connection sockets).

**Fix:** replace the forking UDP echo with `scripts/container/udp_echo.py`, a stateless single-socket recvfrom/sendto loop — no forking, no shared-socket state, 0-length datagrams ignored. Applied to both the per-link targets (port 7001) and the netem sidecar echo (port 9999), which used the same fragile construct. The targets switch from `alpine:3` to `python:3-alpine`, matching the sidecar. The TCP echo stays on socat, which is safe for TCP. The temporary CI skip from #1313 is reverted since the cause is fixed.

**Hardening (from review):** both containers' healthchecks now also verify the UDP echo server is listening (`netstat -uln | grep`), so a dead echo server fails `up --wait` loudly instead of presenting as ~100% packet loss. BusyBox `netstat` is built into `python:3-alpine`, so no extra packages are needed.

**Testing:**
- Unit-level: 40 fresh-source-port probes against `udp_echo.py` with empty datagrams interleaved before/after each payload (the exact poison sequence) — 40/40 echoed.
- Stack-level (local, Docker Desktop): 60-packet probe per link through the netem sidecar shows scattered ~5% loss on link A and 0% on link B — previously a hard cutoff after the first 3-4 packets.
- `cargo test -p remote_access_tests -- --ignored perlink_` (6/6) and `-- --ignored netem_` (6/6) against the rebuilt stacks.
- Healthchecks (final `netstat` form): stack reaches healthy under `up -d --wait`; killing `udp_echo.py` in a target flips it to unhealthy within ~25s, confirming busybox `netstat -uln` lists the bound UDP socket while the server is alive and stops listing it once dead.
- `cargo check`, `cargo clippy -D warnings`, `cargo fmt --check` all clean.

Open question tracked in FLE-595: why the race odds jumped on June 11 (0/11 on June 10 vs 14/23 on June 11 with identical images and package versions). The fix removes the race class entirely, so it holds regardless.

Fixes: [FLE-595](https://linear.app/foxglove/issue/FLE-595/ci-netem-perlink-test-flake-both-links-show-99percent-udp-loss-spiked)

### DIFF
--- a/docker-compose.netem-perlink.yml
+++ b/docker-compose.netem-perlink.yml
@@ -53,8 +53,15 @@ x-echo-target: &echo-target
   volumes:
     - ./scripts/container/udp_echo.py:/udp_echo.py:ro
   healthcheck:
-    # BusyBox nc doesn't support -z; use socat to probe the TCP port instead.
-    test: ["CMD-SHELL", "echo | socat - TCP:localhost:7000,connect-timeout=2"]
+    # BusyBox nc doesn't support -z; use socat to probe both echo servers.
+    # The UDP probe matters: a dead UDP echo server would otherwise present
+    # as near-total packet loss in the tests while the target reports healthy.
+    # ($$ escapes compose interpolation, producing a literal $ for the shell.)
+    test:
+      [
+        "CMD-SHELL",
+        "echo | socat - TCP:localhost:7000,connect-timeout=2 && [ -n \"$$(echo ping | socat -T2 - UDP-CONNECT:127.0.0.1:7001)\" ]",
+      ]
     interval: 2s
     timeout: 5s
     retries: 10
--- a/docker-compose.netem-perlink.yml
+++ b/docker-compose.netem-perlink.yml
@@ -35,17 +35,14 @@ networks:
       config:
         - subnet: 10.98.0.0/24
 
-# The targets use python:3-alpine (matching the netem sidecar) so the UDP echo
-# can be a stateless single-socket server; see scripts/container/udp_echo.py
-# for why a forking socat server wedges under netem jitter (FLE-595).
+# The targets use python:3-alpine to run the UDP echo server.
 x-echo-target: &echo-target
   image: python:3-alpine
   entrypoint: ["/bin/sh", "-c"]
   command:
     - |
       apk add --no-cache socat > /dev/null 2>&1
-      # TCP echo server on port 7000 for latency measurement. Forking is safe
-      # for TCP: each connection gets its own accepted socket.
+      # TCP echo server on port 7000 for latency measurement.
       socat TCP-LISTEN:7000,fork,reuseaddr EXEC:/bin/cat &
       # UDP echo server on port 7001 for packet loss measurement.
       python3 /udp_echo.py 7001 &
@@ -53,14 +50,12 @@ x-echo-target: &echo-target
   volumes:
     - ./scripts/container/udp_echo.py:/udp_echo.py:ro
   healthcheck:
-    # BusyBox nc doesn't support -z; use socat to probe both echo servers.
-    # The UDP probe matters: a dead UDP echo server would otherwise present
-    # as near-total packet loss in the tests while the target reports healthy.
-    # ($$ escapes compose interpolation, producing a literal $ for the shell.)
+    # BusyBox nc doesn't support -z; use socat to probe the TCP port, and
+    # netstat to confirm the UDP echo server is listening.
     test:
       [
         "CMD-SHELL",
-        "echo | socat - TCP:localhost:7000,connect-timeout=2 && [ -n \"$$(echo ping | socat -T2 - UDP-CONNECT:127.0.0.1:7001)\" ]",
+        "echo | socat - TCP:localhost:7000,connect-timeout=2 && netstat -uln | grep -q ':7001 '",
       ]
     interval: 2s
     timeout: 5s
--- a/docker-compose.netem-perlink.yml
+++ b/docker-compose.netem-perlink.yml
@@ -35,17 +35,23 @@ networks:
       config:
         - subnet: 10.98.0.0/24
 
+# The targets use python:3-alpine (matching the netem sidecar) so the UDP echo
+# can be a stateless single-socket server; see scripts/container/udp_echo.py
+# for why a forking socat server wedges under netem jitter (FLE-595).
 x-echo-target: &echo-target
-  image: alpine:3
+  image: python:3-alpine
   entrypoint: ["/bin/sh", "-c"]
   command:
     - |
       apk add --no-cache socat > /dev/null 2>&1
-      # TCP echo server on port 7000 for latency measurement.
+      # TCP echo server on port 7000 for latency measurement. Forking is safe
+      # for TCP: each connection gets its own accepted socket.
       socat TCP-LISTEN:7000,fork,reuseaddr EXEC:/bin/cat &
       # UDP echo server on port 7001 for packet loss measurement.
-      socat UDP-RECVFROM:7001,fork EXEC:/bin/cat &
+      python3 /udp_echo.py 7001 &
       exec sleep infinity
+  volumes:
+    - ./scripts/container/udp_echo.py:/udp_echo.py:ro
   healthcheck:
     # BusyBox nc doesn't support -z; use socat to probe the TCP port instead.
     test: ["CMD-SHELL", "echo | socat - TCP:localhost:7000,connect-timeout=2"]
--- a/docker-compose.netem-perlink.yml
+++ b/docker-compose.netem-perlink.yml
@@ -35,7 +35,6 @@ networks:
       config:
         - subnet: 10.98.0.0/24
 
-# The targets use python:3-alpine to run the UDP echo server.
 x-echo-target: &echo-target
   image: python:3-alpine
   entrypoint: ["/bin/sh", "-c"]
--- a/docker-compose.netem.yml
+++ b/docker-compose.netem.yml
@@ -76,7 +76,15 @@ services:
         # Keep the container alive so the qdisc rules persist.
         exec sleep infinity
     healthcheck:
-      test: ["CMD-SHELL", "tc qdisc show | grep -q netem"]
+      # Verify both the tc hierarchy and the UDP echo server: a dead echo
+      # server would otherwise present as near-total packet loss in the tests
+      # while the sidecar reports healthy.
+      # ($$ escapes compose interpolation, producing a literal $ for the shell.)
+      test:
+        [
+          "CMD-SHELL",
+          "tc qdisc show | grep -q netem && [ -n \"$$(echo ping | socat -T2 - UDP-CONNECT:127.0.0.1:9999)\" ]",
+        ]
       interval: 2s
       timeout: 5s
       retries: 10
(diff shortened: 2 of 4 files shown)


docker-compose.netem-perlink.yml
```diff
@@ -35,17 +35,23 @@ networks:
...
   command:
     - |
       apk add --no-cache socat > /dev/null 2>&1
-      # TCP echo server on port 7000 for latency measurement.
+      # TCP echo server on port 7000 for latency measurement. Forking is safe
+      # for TCP: each connection gets its own accepted socket.
       socat TCP-LISTEN:7000,fork,reuseaddr EXEC:/bin/cat &
       # UDP echo server on port 7001 for packet loss measurement.
-      socat UDP-RECVFROM:7001,fork EXEC:/bin/cat &
+      python3 /udp_echo.py 7001 &
```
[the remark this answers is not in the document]
dave-foxglove: [AI] [resolved] Good call — added UDP echo probes to both healthchecks (targets port 7001 and the netem sidecar port 9999). Verified the negative path locally: killing udp_echo in a target flips it to unhealthy within the retry window.

dave-foxglove:
[AI] > No blockers. I traced the bits that could break: `measure_udp_loss` sends `echo "pkt-$i"` (non-empty) and counts non-empty responses, so the empty-datagram skip in `udp_echo.py` is safe and won't drop real probes. `info!` is still used throughout the test, so dropping the CI skip doesn't orphan the 
> ...

[resolved] Healthcheck gap fixed in the follow-up commit (UDP probes on both the targets and the sidecar, negative path verified locally). On soaking: agreed — the plan is to re-run the netem-perlink-test job several times on this draft before un-drafting; attempt 1 is green. The ~0.13% false-fail floor on the lost_a > 0 assertion is pre-existing and acceptable; it can be revisited if it ever fires.

docker-compose.netem.yml
```diff
@@ -60,20 +60,31 @@ services:
...
     healthcheck:
-      test: ["CMD-SHELL", "tc qdisc show | grep -q netem"]
+      # Verify both the tc hierarchy and the UDP echo server: a dead echo
+      # server would otherwise present as near-total packet loss in the tests
+      # while the sidecar reports healthy.
+      # ($$ escapes compose interpolation, producing a literal $ for the shell.)
+      test:
+        [
+          "CMD-SHELL",
+          "tc qdisc show | grep -q netem && [ -n \"$$(echo ping | socat -T2 - UDP-CONNECT:127.0.0.1:9999)\" ]",
```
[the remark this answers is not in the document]
dave-foxglove: [AI] [resolved] Added the comment on the healthcheck noting the probe rides the shaped loopback and that retries absorb the default 2% loss, with the NETEM_ARGS-override caveat.

dave-foxglove:
[AI] > Re-reviewed the new commit (`f1b5b7db`, the healthcheck probes). The prior healthcheck-coverage thread is fixed and resolved — both the targets (7001) and the sidecar (9999) now fail unhealthy on a dead echo server instead of silently presenting as ~100% loss. 👍
> 
> One new non-blocking finding inline:
> ...

[resolved] Inline finding addressed with a documenting comment on the sidecar healthcheck.

docker-compose.netem-perlink.yml
```diff
@@ -35,20 +35,33 @@ networks:
       config:
         - subnet: 10.98.0.0/24
 
+# The targets use python:3-alpine (matching the netem sidecar) so the UDP echo
+# can be a stateless single-socket server; see scripts/container/udp_echo.py
```
gasmith: I don't see how the choice of image is related to the UDP echo implementation.

scripts/container/udp_echo.py
```diff
@@ -0,0 +1,43 @@
...
+Echoes every received datagram back to its sender from a single unconnected
+socket. Used by the netem test stacks (`docker-compose.netem.yml` port 9999,
+`docker-compose.netem-perlink.yml` port 7001 on the targets).
+
+A single-socket loop is used deliberately instead of a forking server such as
+`socat UDP-RECVFROM:<port>,fork`: forked children share the bound socket with
+the parent, and a child that lingers (e.g. after receiving the 0-length
+datagram a socat client emits at EOF, which netem jitter can reorder ahead of
+the payload) steals datagrams from new clients and never replies. That race
+intermittently blackholed the per-link loss measurements in CI (FLE-595).
```
gasmith: I think we can confine these explanations to the PR and linear ticket. They don't need to live on in the code.

docker-compose.netem-perlink.yml
```diff
@@ -35,20 +35,33 @@ networks:
...
+      python3 /udp_echo.py 7001 &
       exec sleep infinity
+  volumes:
+    - ./scripts/container/udp_echo.py:/udp_echo.py:ro
   healthcheck:
-    # BusyBox nc doesn't support -z; use socat to probe the TCP port instead.
-    test: ["CMD-SHELL", "echo | socat - TCP:localhost:7000,connect-timeout=2"]
+    # BusyBox nc doesn't support -z; use socat to probe both echo servers.
+    # The UDP probe matters: a dead UDP echo server would otherwise present
+    # as near-total packet loss in the tests while the target reports healthy.
```
gasmith: If it didn't matter, we wouldn't healthcheck it. Do we really need this extreme verbosity?

docker-compose.netem-perlink.yml
```diff
@@ -35,20 +35,33 @@ networks:
...
 x-echo-target: &echo-target
-  image: alpine:3
+  image: python:3-alpine
   entrypoint: ["/bin/sh", "-c"]
   command:
     - |
       apk add --no-cache socat > /dev/null 2>&1
-      # TCP echo server on port 7000 for latency measurement.
+      # TCP echo server on port 7000 for latency measurement. Forking is safe
+      # for TCP: each connection gets its own accepted socket.
```
gasmith: The "forking is safe for TCP" addition doesn't seem to add a lot of value, outside of the context of this PR.

docker-compose.netem.yml
```diff
@@ -60,20 +60,35 @@ services:
...
+      - ./scripts/container/udp_echo.py:/udp_echo.py:ro
     entrypoint: ["/bin/sh", "-c"]
     command:
       - |
         apk add --no-cache iproute2 socat > /dev/null 2>&1
         # Start a UDP echo server for packet loss testing. Each received
         # datagram is echoed back through the netem-shaped egress path.
-        socat UDP-RECVFROM:9999,fork EXEC:/bin/cat &
+        # A stateless single-socket server is required here; see
+        # scripts/container/udp_echo.py for why a forking server wedges.
```
gasmith: Too verbose.

docker-compose.netem.yml
```diff
@@ -60,20 +60,35 @@ services:
...
         exec sleep infinity
     healthcheck:
-      test: ["CMD-SHELL", "tc qdisc show | grep -q netem"]
+      # Verify both the tc hierarchy and the UDP echo server: a dead echo
+      # server would otherwise present as near-total packet loss in the tests
+      # while the sidecar reports healthy.
+      # The UDP probe rides the shaped loopback (netem applies to all
+      # interfaces in this namespace), so `retries` absorbs the default 2%
+      # simulated loss; an aggressive NETEM_ARGS loss override could make
+      # startup flaky.
```
gasmith: Instead of making startup flaky, maybe we could do `ss -uln | grep -q :9999` or so to make sure the program is listening on udp port 9999?

scripts/container/udp_echo.py
dave-foxglove: done

docker-compose.netem-perlink.yml
dave-foxglove: poor wording. will fix

docker-compose.netem-perlink.yml
dave-foxglove: Focused it.

docker-compose.netem.yml
dave-foxglove: done, with busybox netstat instead of ss so the targets don't need iproute2.

docker-compose.netem.yml
```diff
@@ -60,20 +60,26 @@ services:
...
         python3 /netem_setup.py || exit 1
         # Keep the container alive so the qdisc rules persist.
         exec sleep infinity
     healthcheck:
-      test: ["CMD-SHELL", "tc qdisc show | grep -q netem"]
+      # Verify the tc hierarchy is applied and the UDP echo server is listening.
+      test:
+        [
+          "CMD-SHELL",
+          "tc qdisc show | grep -q netem && netstat -uln | grep -q ':9999 '",
```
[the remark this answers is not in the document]
dave-foxglove: [AI] [resolved] Confirmed exercised: with the netstat form, the stack reaches healthy under up -d --wait, and killing udp_echo.py in a target flips it to unhealthy within ~25s — which also confirms busybox netstat -uln lists the bound (unconnected) UDP socket while the server is alive and stops listing it once dead. Updated the PR Testing notes to describe the netstat validation instead of the old socat round-trip probe.

dave-foxglove:
[AI] > No new blockers. The only change since my last pass (commit `40d79d8c`) is the doc comment on the sidecar healthcheck explaining the shaped-loopback probe — that closes my prior thread, and both inline threads are now resolved. Re-verified the two things that could have broken with the CI-skip remov
> ...

[resolved] Recap of resolved threads, no new findings.

dave-foxglove:
[AI] > No new blockers. Both of my prior inline threads (UDP healthcheck coverage, shaped-loopback caveat) are resolved. I re-verified the two things that could break with the CI-skip removal: `info!` is still used in `netem_perlink_test.rs` (lines 268, 285-287), so the import isn't orphaned; and `measure_
> ...

[resolved] Recap of resolved threads, no new findings.

dave-foxglove:
[AI] > Re-reviewed the only change since my last pass (`38299613`): both healthchecks now liveness-probe the UDP echo server via `netstat -uln | grep` instead of a round-trip socat echo, and the comments are trimmed. That swap removes the shaped-loopback dependency I flagged earlier (no more round-trip to 
> ...

[resolved] Inline netstat question answered and PR Testing notes updated.

docker-compose.netem-perlink.yml
```diff
@@ -35,20 +35,28 @@ networks:
       config:
         - subnet: 10.98.0.0/24
 
+# The targets use python:3-alpine to run the UDP echo server.
```
gasmith:
```suggestion
```

gasmith (approved):

PR #1315, opened 2026-06-12
Fix a leak when disconnecting clients that are still advertising

clalancette:
### Changelog

None.

### Docs

None.

### Description
    
When a websocket client disconnected without unadvertising its client channels, the server replayed `on_unsubscribe` for the client's subscriptions but silently dropped its advertised channels, never firing `on_client_unadvertise`. `ServerListener` implementations that allocate per-channel state on `on_client_advertise` (both ROS bridges create a ROS publisher there) leaked that state for every client that disconnected mid-advertisement.
    
Drain the client's advertised channels in `on_disconnect` and invoke `on_client_unadvertise` for each, mirroring the unsubscribe replay directly above and the remote-access gateway's existing behavior on participant removal. The handler runs after the advertised_channels lock is released, and before the server's `on_client_disconnect` notification, so listeners never observe events from a client they were already told is gone.

Tested as part of the ongoing work for the ROS 1 bridge, where it fixed a leak in subscribers.

Part of FLE-601

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,5 @@
 /.git/
+/.cargo/
 /.mypy_cache/
 /node_modules/
 /target/
--- a/ros/README.md
+++ b/ros/README.md
@@ -1,3 +1,25 @@
+# ROS packages
+
+The goal of the ROS packages here are to provide a simple way for ROS 2 users to connect their ROS middleware systems into foxglove, either through the foxglove websocket or through the remote access gateway.
+
+# Architecture
+
+The architecture consists of 2 packages:
+
+1. foxglove_msgs - Messages useful for publishing or subscribing from foxglove.  Not directly depended on by any of the other packages.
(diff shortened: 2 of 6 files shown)


dave-foxglove (dismissed):
There appears to be multiple topics in this PR. E.g. the README.md doesn't appear related to fixing a leak. Maybe the PR title/desc needs updating or maybe there's extra changes here?

clalancette:
> There appears to be multiple topics in this PR. E.g. the README.md doesn't appear related to fixing a leak. Maybe the PR title/desc needs updating or maybe there's extra changes here?

I just threw these extra changes in here because they are pretty small, and related, at least, to what this PR is doing.  But I can split them out if you really want a separate PR for them.

clalancette:
After discussion, we decide that operationally we should, in general, split these out.  But we're OK with having them in here for this PR, just so we don't churn.

gasmith (approved):

PR #1330, opened 2026-06-22
Add optional Ray Data datasource (`ray` feature)

defunctzombie:
## Summary

Adds an optional **`ray`** feature to the Python SDK, gated like the existing `notebook` feature. It exposes `foxglove.ray.FoxgloveDataset` — a [Ray Data](https://docs.ray.io/en/latest/data/data.html) datasource that downloads Foxglove recordings as **raw MCAP bytes**, in parallel across a Ray cluster:

```python
import ray
from foxglove.ray import FoxgloveDataset

ds = ray.data.read_datasource(
    FoxgloveDataset(token="fox_sk_...", dataset_id="ds_..."),
    parallelism=8,
)
```

Each row is `{"recording": {"id": str, "mcap": bytes}}`; decoding is left to a downstream Ray transform (e.g. `.flat_map(...)`), so the datasource stays domain-agnostic and reusable.

## What's included

- **`[ray]` extra** (`ray[data]`, `foxglove-client`, `pandas`)
- **`foxglove/ray/`**: gated `__init__` (friendly "install the extra" error) + `FoxgloveDataset`. Pickle-safe (stores config only; constructs a fresh client per worker); one MCAP-bytes block per recording so downloads stream.
- **Tests** (`test_ray.py`): gating, bucketing, one-block-per-recording (mocked client), pickle-safety. They skip when the extra isn't installed.
- **Docs** page + runnable **example** (`ray-datasource/`) that counts MCAP messages per recording.
- **CI**: a dedicated `test-sdk-ray` job / `test-python-ray` make target (see note below).

## Notes for reviewers

- **`dataset_id` is stubbed.** `foxglove-client` doesn't expose a dataset API yet, so `_resolve_recording_ids()` raises `NotImplementedError`. Everything downstream is built — the follow-up is a one-line swap to the real call once the client supports datasets.
- **Why a separate CI job?** The `ray` extra pulls in numpy 2.5, whose stubs use PEP-695 syntax that mypy rejects under our `python_version = "3.10"` target (and `follow_imports = skip` doesn't prevent the parse). So the ray deps are kept out of the lint/typecheck env and exercised in isolation.
- **Namespace check.** Both the SDK and `foxglove-client` install into the top-level `foxglove` package. Verified a normal wheel install resolves `from foxglove.client import Client` fine (foxglove-client is a PEP-420 namespace package). It only breaks under *editable* installs (dev/CI), where the tests mock `foxglove.client`, so nothing is affected.

## Verified locally
- `make test-python` → mypy clean + 97 passed, 14 skipped
- `make test-python-ray` → 5 passed
- `make docs-python` → builds with no warnings
- black / isort / flake8 clean; `uv lock --check` passes


defunctzombie:
https://github.com/foxglove/foxglove-sdk/pull/1442

PR #1338, opened 2026-06-24
bridge: Always initialize SDK logging

gasmith:
### Changelog
- Fixed a bug where the bridge would not initialize SDK logging unless debug:=true

### Docs
None

### Description
The bridge was only initializing SDK logging when debug:=true. It should
always initialize logging at the default level (info).

### Links
Fixes: FLE-625

### DIFF
--- a/ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
+++ b/ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
@@ -147,6 +147,8 @@ FoxgloveBridge::FoxgloveBridge(const rclcpp::NodeOptions& options)
   if (debug) {
     foxglove::setLogLevel(foxglove::LogLevel::Debug);
     this->get_logger().set_level(rclcpp::Logger::Level::Debug);
+  } else {
+    foxglove::setLogLevel(foxglove::LogLevel::Info);
   }
 
   _serverContext = foxglove::Context::create();


dave-foxglove (approved):

ros/src/foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -147,6 +147,8 @@ FoxgloveBridge::FoxgloveBridge(const rclcpp::NodeOptions& options)
   if (debug) {
     foxglove::setLogLevel(foxglove::LogLevel::Debug);
     this->get_logger().set_level(rclcpp::Logger::Level::Debug);
+  } else {
+    foxglove::setLogLevel(foxglove::LogLevel::Info);
   }
```
clalancette, line 152: I almost think that we should remove this parameter (and code block) entirely, and replace it with just:

```
foxglove::setLogLevel(foxglove::LogLevel::Info);
```

Reasoning: if you want to change the foxglove-sdk log level, you can do that by manipulating the environment (`FOXGLOVE_LOG_LEVEL=debug`).  If you want to change the ROS log-level, you can do that by passing in a common argument when starting the node (`--ros-args --log-level debug`).  The only thing that the extra `debug` parameter buys you is a way to set both at once.

If we want a way to set both at once, we could either make the parameter more generic so you could set any level, or we could read the log level value out of ROS 2 (with `rcutils_logging_get_logger_effective_level`) and use that to initialize `foxglove::setLogLevel`.

What do you think?

gasmith: We should definitely keep this code block around (see below), though we can separately debate whether we should have a `debug` parameter vs using `rcutils_logging_get_logger_effective_level`. If the latter is more conventional, we can make that switch (though we'll probably have to soft deprecate the parameter).

There's a subtle difference between `setLogLevel(Debug)` and `FOXGLOVE_LOG_LEVEL=debug`. The former turns on debug-level logging for the foxglove crate only, whereas the latter turns on debug-level logging for _all_ crates (including webrtc, which is extremely noisy). The `FOXGLOVE_LOG_LEVEL` offers fine-grained control over logging in the same way as `RUST_LOG` typically does. For example, you can do `FOXGLOVE_LOG_LEVEL=foxglove=debug,reqwest=debug,info` to enable debug logging for foxglove and reqwest, but info logging for other rust crates.

PR #1341, opened 2026-06-24
Teach ros/Makefile to clean deb build dirs before building debs

gasmith:
### Changelog
None

### Docs
None

### Description
When building debs, the ros/Makefile cleans the build dirs after the
build, but only on success. Which means it should also proactively clean
them before attempting to build, in case the previous build was a
failure. This change also updates the clean target to cover these
directories, for good measure.

### DIFF
--- a/ros/Makefile
+++ b/ros/Makefile
@@ -26,8 +26,18 @@ endif
 all: deps build
 
 .PHONY: clean
-clean:
+clean: clean-deb-bridge clean-deb-msgs
 	rm -rf build dist install log
+
+.PHONY: clean-deb-bridge
+clean-deb-bridge:
+	rm -rf src/foxglove_bridge/debian src/foxglove_bridge/.obj-*
+	rm -f src/ros-*-foxglove-bridge*
+
+.PHONY: clean-deb-msgs
+clean-deb-msgs:
+	rm -rf src/foxglove_msgs/debian src/foxglove_msgs/.obj-*
+	rm -f src/ros-*-foxglove-msgs*
 
 .PHONY: deps
 deps:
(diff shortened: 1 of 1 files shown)


ros/Makefile
```diff
@@ -56,22 +66,20 @@ BLOOM_BRIDGE_CMAKE_ARGS := \
...
-deb-bridge:
+deb-bridge: clean-deb-bridge
 	cd src/foxglove_bridge && bloom-generate rosdebian
 	sed -i 's|dh_auto_configure --|dh_auto_configure -- $(BLOOM_BRIDGE_CMAKE_ARGS)|' \
 		src/foxglove_bridge/debian/rules
 	cd src/foxglove_bridge && fakeroot debian/rules binary
 	mkdir -p dist && mv src/ros-*-foxglove-bridge*.deb dist/
-	rm -rf src/foxglove_bridge/debian src/foxglove_bridge/.obj-*
-	rm -f src/ros-*-foxglove-bridge*
+	$(MAKE) clean-deb-bridge
```
clalancette: In general, I'm not a huge fan of invoking `$(MAKE)` from within make.  It breaks Make's dependency processing; it can't tell what things are before what other things.  That also means that things like `make -j8` don't work properly.  Neither of those is a real concern here, but I'd prefer not to do it.

And it turns out there is a better way.  If we make the lines that remove the files a define, then we use that define both for `clean-deb-bridge` and for after, something like this:

```
define clean_deb_bridge =
rm -rf src/foxglove_bridge/debian src/foxglove_bridge/.obj-*
rm -f src/ros-*-foxglove-bridge*
endef

...

.PHONY: clean-deb-bridge
clean-deb-bridge:
      $(clean_deb_bridge)

.PHONY: deb-bridge
deb-bridge: clean-deb-bridge
      cd src/foxglove_bridge && bloom-generate rosdebian
      sed -i 's|dh_auto_configure --|dh_auto_configure -- $(BLOOM_BRIDGE_CMAKE_ARGS)|' \
              src/foxglove_bridge/debian/rules
      cd src/foxglove_bridge && fakeroot debian/rules binary
      mkdir -p dist && mv src/ros-*-foxglove-bridge*.deb dist/
      $(clean_deb_bridge)
```

clalancette (changes requested):

ros/Makefile
gasmith: Oh, good call. I totally forgot about makefile defines.

clalancette (approved):
Much better, thanks!

PR #1343, opened 2026-06-25
Expose preferred video encoder backend in SDK and bridge

clalancette:
### Changelog

Add ability to select the video encoder in use; defaults to `Auto`, which will automatically choose the best one.

### Docs

None.

### Description

LiveKit 0.7.46 exposes the preferred video encoder backend via TrackPublishOptions.video_encoder. Surface it as a gateway-wide setting across the whole stack: Rust SDK builder + FOXGLOVE_VIDEO_ENCODER env var, C API, C++ API, the Python SDK, and a ROS parameter on the bridge.

A non-Auto value is an explicit override that takes precedence over the env var; Auto (the default) means "no explicit preference" and defers to FOXGLOVE_VIDEO_ENCODER, then the SDK default. This rule is uniform across all layers. If the requested backend is unavailable, libwebrtc falls back to another compatible encoder.

- Rust: new VideoEncoderBackend enum + Gateway::video_encoder() builder and FOXGLOVE_VIDEO_ENCODER env var, plumbed gateway -> connection -> session -> TrackPublishOptions.video_encoder, with a debug log of the resolved backend at publish time.
- C API: FoxgloveVideoEncoderBackend enum + video_encoder option field; regenerated foxglove-c.h with the foxglove_video_encoder_backend rename.
- C++: VideoEncoderBackend enum + video_encoder field on RemoteAccessGatewayOptions, static_cast to the C enum in create().
- Python: VideoEncoderBackend PyO3 enum + video_encoder argument on start_gateway, with wrapper and .pyi stub updates.
- ROS bridge: read-only video_encoder string parameter parsed to foxglove::VideoEncoderBackend, documented in the README.
- example: --video-encoder flag exercising the Gateway builder option.

Fixes FLE-602


rust/foxglove/src/remote_access/session.rs
```diff
@@ -2210,8 +2215,14 @@ impl RemoteAccessSession {
...
                 // and combined with simulcast results in very low quality video with compression artifacts.
                 let video_codec = session.video_codec_override.unwrap_or(DEFAULT_VIDEO_CODEC);
+                let video_encoder_backend = session.video_encoder_override.unwrap_or_default();
+                debug!(
+                    "publishing video track for channel {channel_id:?} with codec {video_codec:?} \
+                     and preferred encoder backend {video_encoder_backend:?}"
+                );
                 let publish_options = TrackPublishOptions {
                     video_codec,
+                    video_encoder: video_encoder_backend.into(),
```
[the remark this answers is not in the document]
clalancette: This shouldn't be a problem; the default both before and after here is `Auto`, so this is a no-op.

rust/examples/remote_access/src/main.rs
```diff
@@ -81,6 +107,12 @@ struct Args {
     /// allocation at 8K.
     #[arg(long, default_value_t = 540, value_parser = RangedU64ValueParser::<usize>::new().range(540..=4320))]
     height: usize,
+
+    /// Preferred video encoder backend for published video tracks. If omitted, the SDK
+    /// chooses (honoring the `FOXGLOVE_VIDEO_ENCODER` environment variable). If the requested
+    /// backend is unavailable, the SDK falls back to another compatible encoder.
+    #[arg(long, value_enum)]
+    video_encoder: Option<VideoEncoderArg>,
```
gasmith: Nit: Could make this default to "auto"

rust/foxglove/src/remote_access/gateway.rs
```diff
@@ -193,6 +248,7 @@ pub struct Gateway {
     qos_classifier: Option<Arc<dyn QosClassifier>>,
     server_info: Option<HashMap<String, String>>,
     message_backlog_size: Option<usize>,
+    video_encoder_override: Option<VideoEncoderBackend>,
```
gasmith: Having both an "auto" variant and the Option wrapper seems redundant.

rust/foxglove/src/remote_access/gateway.rs
```diff
@@ -351,6 +409,19 @@ impl Gateway {
...
     }
 
+    /// Sets the preferred backend for encoding published video tracks.
+    ///
+    /// This preference applies to every video track the gateway publishes. If the requested
+    /// backend is unavailable on the host, libwebrtc falls back to another compatible encoder.
+    ///
+    /// If not set, or set to [`VideoEncoderBackend::Auto`], the backend is read from the
+    /// `FOXGLOVE_VIDEO_ENCODER` environment variable (one of: `auto`, `software`, `hardware`,
+    /// `nvenc`, `vaapi`, `videotoolbox`), ultimately falling back to [`VideoEncoderBackend::Auto`].
```
gasmith, line 440: I think this sounds reasonable.

gasmith (approved):
I think this looks good. My only suggestion is that we get rid of the Option wrapper, since it's redundant to the Auto variant.

rust/foxglove/src/remote_access/gateway.rs
```diff
@@ -158,6 +159,45 @@ const FOXGLOVE_DEVICE_TOKEN_ENV: &str = "FOXGLOVE_DEVICE_TOKEN";
...
+    /// Prefer a software encoder.
+    Software,
+    /// Prefer any available hardware encoder.
+    Hardware,
+    /// Prefer NVIDIA NVENC when available.
+    Nvenc,
+    /// Prefer VAAPI when available.
+    Vaapi,
+    /// Prefer VideoToolbox on Apple platforms when available.
+    VideoToolbox,
```
gasmith, line 188: We should consider putting this variant under a `#[cfg(target_os = "macos")]`. Same for the CLI args. I realize that there's no equivalent in C++, but the FFI layer could just collapse VideoToolbox->Auto and get the same behavior we have right now.

I don't feel strongly about this. There's no harm in leaving it as it is.

clalancette: I think I'd rather not do it, because:

1.  It has the potential to change the ordering of the items per-platform, especially if we add new ones in the future.  While this isn't the end of the world (it will be internally consistent), it makes it harder to debug.
2.  Some of the other options are *also* platform-specific, like vaapi (Linux) and nvenc (Linux/Windows).

So I'm going to leave this as-is for now.

rust/examples/remote_access/src/main.rs
clalancette: Yeah, good call.  Fixed in [6b7873b](https://github.com/foxglove/foxglove-sdk/pull/1343/commits/6b7873b9a8212885f902198b3bc8ebab085688bd)

rust/foxglove/src/remote_access/gateway.rs
clalancette: Fixed in [6b7873b](https://github.com/foxglove/foxglove-sdk/pull/1343/commits/6b7873b9a8212885f902198b3bc8ebab085688bd)

rust/foxglove/src/remote_access/session.rs
```diff
@@ -228,6 +228,10 @@ pub(super) struct RemoteAccessSession {
     video_codec_override: Option<VideoCodec>,
     /// Per-message size limit for lossy data-track channels; larger messages are dropped.
     max_data_track_message_size: usize,
+    /// The preferred encoder backend applied to published video tracks.
+    /// [`VideoEncoderBackend::Auto`](super::gateway::VideoEncoderBackend::Auto) leaves the
+    /// choice to libwebrtc.
+    video_encoder_override: super::gateway::VideoEncoderBackend,
```
gasmith: I don't think we need to call this an override anymore. Same comment applies elsewhere.

gasmith (approved):
One little internal naming nit, otherwise LGTM

rust/foxglove/src/remote_access/session.rs
clalancette: Good call, renamed to just "video_encoder" (matching the other APIs anyway).

clalancette:
Given that the last change was only a name change, I'm going to take the last approval as good and merge this in.  Thanks for the reviews!

PR #1345, opened 2026-06-26
remote_access: Drop oversized data-track messages (FLE-592)

dave-foxglove:
### Changelog

Remote access gateways now drop lossy data-track messages larger than a configurable limit (default 100 KiB), with a throttled warning, so one high-bandwidth channel can no longer starve the others; configure via `Gateway::max_data_track_message_size`.

### Docs

None — the new builder method is documented via rustdoc.

### Description

When streaming over remote access, subscribing a single high-bandwidth lossy channel (e.g. point clouds at ~1.25 MiB/message) could starve every other data channel while delivering nothing itself. All of a participant's data tracks multiplex onto one unreliable LiveKit `_data_track` channel fronted by a one-frame, drop-oldest send queue, so a stream of oversized frames monopolizes that single slot. The data-track stack is built for small, frequent messages; a multi-megabyte point cloud violates that assumption (and the same architecture is present in the latest LiveKit, so upgrading the dependency wouldn't help).

This change sheds the problem at the source: a configurable per-message size limit (default 100 KiB) on the lossy data-track publish path. Messages above the limit are dropped before publish, with a throttled warning that reports the running per-channel drop count. The limit is set via `Gateway::max_data_track_message_size` and must be at least 1200 bytes (one WebRTC data-channel packet); `start()` rejects smaller values. Reliable channels (which use the control bytestream) are unaffected.

This is the contained gateway-side fix. A viewer-facing drop signal (a versioned control-plane message so the app can surface lossiness, coordinated with FLE-603) is deferred to a follow-up.

Note: this is a default-behavior change — lossy data-track messages over 100 KiB are now dropped by default.

Manual testing — the gateway-side drop behavior is covered by unit and integration tests in this PR; the browser + netem end-to-end path was verified manually:

- [x] **End-to-end starvation check.** Ran `yarn stream-mcap` with `demo.mcap` (netem egress-shaped). `/velodyne_points` (~1.28 MB/msg) was dropped at the 100 KiB default and no longer held up the rest of the stream, with the gateway logging the throttled `dropping … exceeds data-track limit of 102400 bytes` warning. Verified on an earlier commit of this branch; the drop behavior is unchanged since (only the warning's wording was refined to report drops-since-last-warning), and the automated integration test re-ran green on the final code.

Fixes: [FLE-592](https://linear.app/foxglove/issue/FLE-592)

### DIFF
--- a/rust/foxglove/src/remote_access/gateway.rs
+++ b/rust/foxglove/src/remote_access/gateway.rs
@@ -364,9 +364,9 @@ impl Gateway {
     /// warning, rather than allowed to monopolize the shared data channel and
     /// starve other channels.
     ///
-    /// The limit must be at least one transport packet (16,000 bytes); a smaller
-    /// limit would drop messages that already fit in a single packet, and
-    /// [`start`](Self::start) rejects it. By default, the limit is 100 KiB.
+    /// The limit must be at least 1200 bytes (one WebRTC data-channel packet);
+    /// [`start`](Self::start) rejects anything smaller. By default, the limit is
+    /// 100 KiB.
     pub fn max_data_track_message_size(mut self, size: usize) -> Self {
         self.max_data_track_message_size = Some(size);
         self
@@ -547,7 +547,7 @@ impl Gateway {
             if size < MIN_DATA_TRACK_MESSAGE_SIZE {
                 return Err(FoxgloveError::ConfigurationError(format!(
                     "max_data_track_message_size ({size} bytes) is below the minimum of \
-                     {MIN_DATA_TRACK_MESSAGE_SIZE} bytes (one transport packet)."
+                     {MIN_DATA_TRACK_MESSAGE_SIZE} bytes (one data-channel packet)."
                 )));
             }
         }
--- a/rust/foxglove/src/remote_access/gateway.rs
+++ b/rust/foxglove/src/remote_access/gateway.rs
@@ -644,8 +644,8 @@ mod tests {
 
     #[test]
     fn max_data_track_message_size_below_minimum_rejected() {
-        // A limit smaller than one transport packet is a misconfiguration and
-        // must be rejected at startup.
+        // A limit below the minimum is a misconfiguration and must be rejected
+        // at startup.
         let result = Gateway::new()
             .device_token("test-token")
             .max_data_track_message_size(MIN_DATA_TRACK_MESSAGE_SIZE - 1)
--- a/rust/foxglove/src/remote_access/gateway.rs
+++ b/rust/foxglove/src/remote_access/gateway.rs
@@ -19,6 +19,7 @@ use crate::{
 use super::qos::{QosClassifier, QosClassifierFn, QosProfile};
 
 use super::connection::{ConnectionParams, ConnectionStatus, RemoteAccessConnection};
+use super::session::MIN_DATA_TRACK_MESSAGE_SIZE;
 use super::{Capability, Listener};
 use crate::remote_common::parameters::ParameterHandler;
 
@@ -130,6 +131,7 @@ impl GatewayHandle {
             qos_classifier: None,
             server_info: None,
             message_backlog_size: None,
+            max_data_track_message_size: None,
             video_codec_override: None,
             context: std::sync::Weak::new(),
         };
@@ -193,6 +195,7 @@ pub struct Gateway {
     qos_classifier: Option<Arc<dyn QosClassifier>>,
     server_info: Option<HashMap<String, String>>,
     message_backlog_size: Option<usize>,
+    max_data_track_message_size: Option<usize>,
     context: std::sync::Weak<Context>,
 }
 
@@ -214,6 +217,7 @@ impl Default for Gateway {
             qos_classifier: None,
             server_info: None,
             message_backlog_size: None,
+            max_data_track_message_size: None,
             context: Arc::downgrade(&Context::get_default()),
         }
     }
@@ -240,6 +244,10 @@ impl std::fmt::Debug for Gateway {
             .field("has_qos_classifier", &self.qos_classifier.is_some())
             .field("server_info", &self.server_info)
             .field("message_backlog_size", &self.message_backlog_size)
+            .field(
+                "max_data_track_message_size",
+                &self.max_data_track_message_size,
+            )
             .field("has_context", &(self.context.strong_count() > 0));
         dbg.finish()
     }
@@ -348,6 +356,19 @@ impl Gateway {
     /// By default, each participant gets a queue of 1024 messages.
     pub fn message_backlog_size(mut self, size: usize) -> Self {
         self.message_backlog_size = Some(size);
+        self
+    }
+
+    /// Sets the maximum size in bytes of a single message published to a lossy
+    /// channel's data track. Larger messages are dropped, with a throttled
+    /// warning, rather than allowed to monopolize the shared data channel and
+    /// starve other channels.
+    ///
+    /// The limit must be at least one transport packet (16,000 bytes); a smaller
+    /// limit would drop messages that already fit in a single packet, and
+    /// [`start`](Self::start) rejects it. By default, the limit is 100 KiB.
+    pub fn max_data_track_message_size(mut self, size: usize) -> Self {
+        self.max_data_track_message_size = Some(size);
         self
     }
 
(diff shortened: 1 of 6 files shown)


rust/foxglove/src/remote_access/session.rs
```diff
@@ -82,6 +82,16 @@ const ROOM_CLOSE_TIMEOUT: Duration = Duration::from_secs(5);
 
 pub(super) const DEFAULT_MESSAGE_BACKLOG_SIZE: usize = 1024;
 
+/// Default per-message size limit for lossy data-track channels, in bytes.
+/// Messages larger than this are dropped before publish so one oversized channel
+/// cannot monopolize the shared data channel and starve the others (see FLE-592).
+pub(super) const DEFAULT_MAX_DATA_TRACK_MESSAGE_SIZE: usize = 100 * 1024;
```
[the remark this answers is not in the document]
dave-foxglove: [AI] [resolved] The compressed-image case doesn't apply here: image channels (foxglove.CompressedImage / foxglove.RawImage and the ROS sensor_msgs image types) are published as LiveKit video tracks, not data tracks, so they never reach this gate — see detect_video_schema in session/video_track.rs. The 100 KiB limit only affects lossy *data-track* payloads (point clouds, marker arrays, etc.). 100 KiB was the value from the motivating discussion and it's configurable via Gateway::max_data_track_message_size; we're open to raising the default for mid-size payloads if reviewers prefer.

dave-foxglove:
[AI] > Carrying forward the open risk, untouched by the new test commit: this defaults on and silently sheds data while the viewer-facing drop signal (FLE-603) is deferred — a heavy lossy channel becomes an empty panel whose only trace is a gateway `warn!` the user never sees. Is shipping the default ahead
> ...

[resolved] Sequencing is intentional: the gateway-side shedding ships now (a strict improvement over the prior all-channels-stall), and the viewer-facing drop signal is tracked separately as FLE-603. In the interim the operator signal is the throttled gateway warn!, and the limit is configurable. On the 100 KiB default: see the inline reply — image frames are unaffected (they go over video tracks), so this only bounds lossy data-track payloads.

dave-foxglove:
[AI] > Newest commit just removes the now-unreachable `debug_assert` in `DataTrack::new()` — no concern there; the bound is enforced in `start()`, so nothing reaches the constructor below the minimum.
> 
> Still open from the last pass, and unchanged by this commit: this is a default-on change that silently sh
> ...

[resolved] Acknowledged on the debug_assert removal. The default-on / FLE-603 sequencing is the intended split — addressed in the reply on the latest review.

dave-foxglove:
[AI] > Logic is clean and the validation/default plumbing checks out. One rollout risk worth settling before this comes off draft: it's a default-on change that silently sheds data, and the viewer-facing drop signal (FLE-603) is deferred. So a user who subscribes a heavy lossy channel now gets an empty pan
> ...

[resolved] Intentional split — shedding ships now, viewer-facing signal tracked in FLE-603; addressed in the reply on the latest review.

rust/foxglove/src/remote_access/session/data_track.rs
```diff
@@ -99,12 +107,32 @@ impl DataTrack {
...
+    /// Drops the message (with a throttled warning) if it exceeds
+    /// [`max_message_size`](Self::max_message_size), or with a throttled debug
+    /// log if the track is not ready or full.
     pub fn log(&self, channel_id: ChannelId, msg: &[u8], metadata: &Metadata) {
+        if msg.len() > self.max_message_size {
+            let dropped = self.oversized_dropped.fetch_add(1, Ordering::Relaxed) + 1;
+            if self.oversized_throttler.lock().try_acquire() {
+                warn!(
+                    "dropping {}-byte message on channel {channel_id:?}: exceeds \
+                     data-track limit of {} bytes ({dropped} dropped on this channel so far)",
```
gasmith: This is the total number of dropped messages, not the number for a specific channel.

rust/foxglove/src/remote_access/session/data_track.rs
```diff
@@ -99,12 +107,32 @@ impl DataTrack {
...
     /// Build a sequenced frame and push it to the data track.
-    /// Drops with a throttled debug log if the track is not ready or full.
+    ///
+    /// Drops the message (with a throttled warning) if it exceeds
+    /// [`max_message_size`](Self::max_message_size), or with a throttled debug
+    /// log if the track is not ready or full.
     pub fn log(&self, channel_id: ChannelId, msg: &[u8], metadata: &Metadata) {
+        if msg.len() > self.max_message_size {
+            let dropped = self.oversized_dropped.fetch_add(1, Ordering::Relaxed) + 1;
+            if self.oversized_throttler.lock().try_acquire() {
```
gasmith: Might be a little more useful to log the number of messages dropped since the last time we logged, rather than the total number.

```rust
if self.oversized_throttler.lock().try_acquire() {
    let dropped = 1 + self.oversized_dropped.swap(0, Relaxed);
    warn!(...)
} else {
    self.oversized_dropped.fetch_add(1, Relaxed);
}
```

rust/foxglove/src/remote_access/gateway.rs
```diff
@@ -351,6 +359,19 @@ impl Gateway {
         self
     }
 
+    /// Sets the maximum size in bytes of a single message published to a lossy
+    /// channel's data track. Larger messages are dropped, with a throttled
+    /// warning, rather than allowed to monopolize the shared data channel and
+    /// starve other channels.
+    ///
+    /// The limit must be at least one transport packet (16,000 bytes); a smaller
```
gasmith: 16KiB is too generous. For lossy transport, there's no retransmission. If you lose a single UDP packet, you lose the whole message. LiveKit recommends a message size of no greater than 1200 to avoid segmentation (https://docs.livekit.io/transport/data/data-tracks/#choosing-a-frame-size).

rust/foxglove/src/remote_access/session.rs
```diff
@@ -82,6 +82,17 @@ const ROOM_CLOSE_TIMEOUT: Duration = Duration::from_secs(5);
...
+
+/// Minimum configurable data-track message limit, in bytes. Below one packet
+/// (livekit-datatrack's 16,000-byte max packet size, its `TRANSPORT_MTU`) the
```
gasmith: I suppose this must be application-level framing on top of UDP?

rust/foxglove/src/remote_access/gateway.rs
dave-foxglove: [AI] [resolved] Good catch. The 16 KiB came from misreading livekit-datatrack's TRANSPORT_MTU (16,000) as a wire packet — it's actually the application-level packetizer chunk. Lowered the minimum to 1200 bytes (one data-channel packet, per the frame-size guidance you linked). The default stays 100 KiB.

rust/foxglove/src/remote_access/session.rs
dave-foxglove: [AI] [resolved] Yes — 16,000 is livekit-datatrack's application-level packetizer chunk (its TRANSPORT_MTU); each chunk is segmented by SCTP/DTLS/UDP into ~1200-byte data-channel packets on the wire. Corrected the minimum and its comment accordingly.

rust/foxglove/src/remote_access/session/data_track.rs
dave-foxglove: [AI] [resolved] Done — now logs the number dropped since the last warning using the swap pattern you suggested.

rust/foxglove/src/remote_access/session/data_track.rs
dave-foxglove: I may be missing something, this counter reads as per-channel to me. Can you help me see where it ends up as a total.

- publish_data_tracks creates one DataTrack per channel and stores it keyed by ChannelId in data_tracks: HashMap<ChannelId, DataTrack> (publish_data_tracks, the map).
- oversized_dropped is a field on our foxglove per-channel DataTrack, and that's the channel's own track via get_subscribed_data_track(&channel_id).
- Channel IDs are not reused afaik, so a track maps to one channel for its lifetime.

It kinda looks like each channel's drops accumulate in its own counter. Thanks!

rust/remote_access_tests/tests/livekit_test.rs
```diff
@@ -138,6 +138,76 @@ async fn livekit_viewer_receives_message_after_subscribe() -> Result<()> {
...
+    let ctx = foxglove::Context::new();
+    let channel = ctx
+        .channel_builder("/test")
+        .message_encoding("json")
+        .build_raw()
+        .context("create channel")?;
+
+    // Use the minimum allowed limit (one transport packet). Messages larger than
+    // this are dropped before publish; smaller ones flow normally.
+    let limit = 16_000;
```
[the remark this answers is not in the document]
dave-foxglove: [AI] [resolved] Updated the comment to describe 16,000 as a representative limit, dropping the stale minimum / transport-packet framing.

dave-foxglove:
[AI] > The min→1200 change and the per-interval drop count look right, and the sequencing/100 KiB questions are settled — thanks. One leftover from the same change: the PR description still says the limit "must be at least one transport packet (16,000 bytes)", which now contradicts the code and rustdoc (12
> ...

[resolved] Updated the PR description to match the code: the limit must be at least 1200 bytes (one data-channel packet).

gasmith (approved):
Note that we'll also need to plumb this through the bridge as a parameter, though that can be a separate PR.

PR #1346, opened 2026-06-26
remote_access: Add transcoding support for flatbuffer, json, omgidl

gasmith:
### Changelog
- remote_access: Added support for transcoding images on topics that use
  foxglove schemas and flatbuffer, json, or omgidl encodings.

### Docs
None

### Description
We currently support transcoding ros1, ros2, and foxglove-protobuf image
topics. We were missing support for other supported encodings:
flatbuffer, json, and omgidl. This change adds support for these
encodings.

The flatbuffer implementation is hard-coded based on the present schema,
rather than pulling in a new codegen dependency on flatc.

### Testing
Tested an mcap file containing the same image stream, encoded to
multiple topics under the flatbuffer, json, omgidl, and protobuf
encodings, both raw and compressed (as png).

### Links
Fixes: FLE-634


rust/foxglove/src/remote_access/session/video_track.rs
```diff
@@ -304,6 +338,36 @@ fn decode_image_message<'a>(
...
+        #[cfg(feature = "img2yuv-flatbuffer")]
+        VideoInputSchema::FoxgloveFlatbufferCompressedImage => {
+            crate::img2yuv::flatbuffer::decode_compressed_image(data)
+                .map_err(|e| VideoEncodeError::Decode(e.to_string()))
+        }
+        #[cfg(feature = "img2yuv-flatbuffer")]
+        VideoInputSchema::FoxgloveFlatbufferRawImage => {
+            crate::img2yuv::flatbuffer::decode_raw_image(data)
+                .map_err(|e| VideoEncodeError::Decode(e.to_string()))
+        }
```
clalancette, line 358: This is a minor thing, but all of the rest of the decoders here do a decode, then call `ImageMessage::try_from(msg)` on the result.  I'm wondering if we should do the same for JSON and Flatbuffer here, mostly for consistency.  What do you think?

clalancette (commented):
I left one thing to thing about, otherwise this looks fine to me.

rust/foxglove/src/remote_access/session/video_track.rs
gasmith: I'm going to punt. There's a refactor here when we eventually get around to decoupling the foxglove message parser out of img2yuv and into its own crate. We'll align things when we get to that point.

clalancette (approved):

PR #1350, opened 2026-07-02
FLE-628: add a per-channel video-transcode opt-out for remote access

dave-foxglove:
### Changelog

Remote access producers can now opt a video-capable image channel out of video transcoding, so it's delivered as data instead, via the gateway builder (`Gateway::suppress_video_transcode` / `suppress_video_transcode_fn`). This is needed for image channels whose pixels must not pass through lossy video — compressed depth maps are the motivating case (their pixels encode depth). Rendering large depth frames end-to-end also requires raising `max_data_track_message_size` (see the delivery limitation below).

### Docs

Rustdoc: the `remote_access` module doc explains video transcoding and the data opt-out; the new public API (`Gateway::suppress_video_transcode`, `suppress_video_transcode_fn`, the `SuppressVideoTranscode` trait) is documented. User-facing docs.foxglove.dev coverage is tracked in [FLE-648](https://linear.app/foxglove/issue/FLE-648).

### Description

Over remote access the SDK classifies any video-capable image channel (`foxglove.CompressedImage`/`RawImage`, ROS `sensor_msgs` image types) as video and transcodes it to H.264/VP8. That corrupts compressed depth maps — their pixel values encode depth, so lossy video is meaningless for them and they don't render.

This adds an advertise-time opt-out expressed as a gateway-builder predicate, mirroring `qos_classifier`/`channel_filter`: a `SuppressVideoTranscode` trait plus the `suppress_video_transcode` (trait) and `suppress_video_transcode_fn` (closure) builder methods. When the predicate returns `true` for a channel, the channel is advertised without a video track and delivered as data. The SDK cannot tell such channels apart from a channel descriptor alone (the depth `format` lives in the message body, not the descriptor), so the producer identifies them here — e.g. by topic or schema. The trait is the seam the C/C++/Python bindings will implement. The decision lives on the gateway rather than in channel metadata: channel metadata is persisted to MCAP, so a metadata-based opt-out would leak remote-access-only config into recordings.

**Delivery limitation:** compressed-depth frames are large (a 640×480 frame is ~110–240 KiB) and exceed the default 100 KiB data-track message-size limit ([FLE-592](https://linear.app/foxglove/issue/FLE-592)), so they are dropped unless the operator raises `max_data_track_message_size` ([FLE-642](https://linear.app/foxglove/issue/FLE-642)). Delivering large data-as-data robustly is tracked in [FLE-656](https://linear.app/foxglove/issue/FLE-656).

**Follow-ups:** C/C++/Python `Gateway` bindings implementing the `SuppressVideoTranscode` trait, and the stacked ROS bridge PR wiring a topic matcher into it. (The MCAP example's channel-metadata forwarding moved to #1355.)

### Testing

Unit tests for the opt-out decision (`resolve_video_input_schema_honors_suppress_predicate`). The `remote_access_tests` LiveKit integration test (`livekit_suppress_video_transcode_opts_out_of_video_track`) configures the opt-out via the builder and asserts the opted-out channel is advertised without `foxglove.hasVideoTrack`; it is `#[ignore]` and needs a live gateway.

Manual testing:
- [x] Verified end-to-end with a local edit to the `remote_access` example (scratch only, not included in this PR): with `suppress_video_transcode_fn` opting a video-capable topic out, that topic is advertised without a video track and delivered as data (`chrome://webrtc-internals` shows no inbound video RTP), while a sibling topic still gets a video track. The same advertise-time behavior is asserted by the `livekit_suppress_video_transcode_opts_out_of_video_track` integration test above. (Full compressed-depth point-cloud render additionally depends on the ROS bridge #1351 and the delivery limit FLE-656.)

Part of: [FLE-628](https://linear.app/foxglove/issue/FLE-628)


dave-foxglove:
[AI] **Manual verification** (A/B on one protobuf compressed-depth recording, branch as the only variable):

- **This branch:** example logs `Delivering 8 compressed-depth channel(s) as data`; 3D panel "Depth map (point cloud)" renders a live point cloud; `chrome://webrtc-internals` shows no `kind=video` inbound-rtp for the depth topics — delivered over a data channel (data bytes climbing).
- **`main` (control):** no such log; a `kind=video` inbound-rtp appears with `framesDecoded` climbing; the app reconstructs an `rgba8` frame from the video and the depth map fails with *"Unsupported depth encoding 'rgba8'."*

Confirms the opt-out keeps compressed depth on a data channel and prevents the lossy video transcode.

rust/examples/remote_access_stream_mcap/src/main.rs
```diff
@@ -126,18 +137,77 @@ fn load_summary<R: Read + Seek>(file: &mut R) -> Result<Option<Summary>> {
...
+                        }
+                    }
+                    if pending.is_empty() {
+                        break;
+                    }
+                }
+            }
+        }
+    }
+    Ok(depth)
```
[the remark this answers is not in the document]
dave-foxglove: [AI] [resolved] Removed — this branch no longer modifies the example, so the up-front scan is gone entirely. The example's channel-metadata forwarding (which lets a recording declare the opt-out instead of the example classifying it by peeking at payloads) is handled separately in #1355, so there is no pre-playback scan to stall on.

dave-foxglove:
[AI] > Carrying forward one still-open question from prior passes (docs is now settled — rustdoc covers the SDK side and FLE-648 tracks the docs site 👍):
> 
> `foxglove.suppressVideoTranscode=true` is producer-set metadata, and `advertise.rs:194` clones producer metadata verbatim into the advertise message — s
> ...

[resolved] Fixed (082f7675). add_metadata_to_advertisement now strips foxglove.suppressVideoTranscode from the advertised channel metadata, so it stays a producer->SDK-only signal (consumed at advertise time by get_video_input_schema) and is no longer forwarded to subscribers as a reserved foxglove.* key. Added a unit test covering the strip.

rust/foxglove/src/remote_access/suppress_video_transcode.rs
```diff
@@ -0,0 +1,35 @@
+//! The video-transcoding opt-out predicate for remote access.
+
+use crate::channel::ChannelDescriptor;
+
+/// Decides, per channel, whether to opt out of video transcoding over remote access.
+///
+/// This callback is invoked when a channel is registered. Returning `true` advertises the channel
+/// without a video track, so its messages are delivered on the data plane unchanged. This is
+/// required for compressed depth maps, whose pixel values encode depth and would be corrupted by
+/// lossy video transcoding — [`is_compressed_depth_format`] classifies a compressed-depth `format`
```
gasmith: This doesn't work; `is_compressed_depth_format` operates on a `CompressedImage` message, not a channel descriptor. If we could detect compressed depth images from the channel descriptor, we would do that automatically. It is because we cannot that we must punt to the user to tell us which image channels cannot be represented as video.

gasmith (changes requested):
We should revert the changes to img2yuv, since there's no way for a user to effectively use that function to classify a channel based on its descriptor.

gasmith (commented):
Let's also rename this PR; it's about offering a mechanism to suppress video transcoding. Which _can be used_ for compressed depth images, but perhaps there are other uses as well.

rust/foxglove/src/remote_access/suppress_video_transcode.rs
dave-foxglove: [AI] [resolved] Good catch — you're right, is_compressed_depth_format reads a CompressedImage message's format field, which isn't available from a ChannelDescriptor, so it can't drive this predicate. Reverted the img2yuv helper; the docs now state that the SDK can't detect these channels from the descriptor, so the producer identifies them here (e.g. by topic or schema).

dave-foxglove:
[AI] > We should revert the changes to img2yuv, since there's no way for a user to effectively use that function to classify a channel based on its descriptor.

[resolved] Reverted the img2yuv changes (is_compressed_depth_format, its tests, and the re-exports).

dave-foxglove:
[AI] > Let's also rename this PR; it's about offering a mechanism to suppress video transcoding. Which _can be used_ for compressed depth images, but perhaps there are other uses as well.

[resolved] Renamed to 'add a per-channel video-transcode opt-out for remote access' and reframed the module/trait docs and PR description around the general mechanism, with compressed depth as the motivating example.

rust/foxglove/src/remote_access.rs
```diff
@@ -1,4 +1,17 @@
 //! Remote access implementation.
+//!
+//! # Video transcoding and the data opt-out
+//!
+//! For bandwidth-efficient live streaming, image channels are transcoded to a WebRTC video
+//! track rather than sent on the data plane: `foxglove.CompressedImage` and `foxglove.RawImage`
```
gasmith: Not just protobuf. We also support other encodings: json, flatbuffer, & omgidl.

gasmith (approved):
Minor comment on the docstring, otherwise LG

rust/foxglove/src/remote_access.rs
dave-foxglove: [AI] [resolved] Good catch — reworded the module doc to cover the foxglove image schemas in protobuf, JSON, FlatBuffer, and OMG IDL encodings (plus the ROS sensor_msgs types), not just protobuf (4d83a354).

dave-foxglove:
[AI] **Suggested release-note entry** (consolidates #1350, #1351, #1363 — the auto-generated notes would otherwise list three separate `FLE-628:` lines for one feature):

> **Remote access: deliver image channels as data instead of transcoding to video.** Transcoding corrupts compressed depth maps, whose pixel values encode distance. Opt a channel out per-channel via the SDK gateway's `suppress_video_transcode` callback (Rust/Python/C++), or the Foxglove Bridge's new `video_transcode_topic_denylist` launch parameter (defaults to `.*/compressedDepth`).

Docs for this land in app #16078.

PR #1354, opened 2026-07-07
remote_access: Warn viewers of dropped oversized messages (FLE-645)

Avan1ko:
### Changelog
Remote access viewers now see a warning in the app when the gateway drops oversized messages on a subscribed topic.

### Docs
None

### Description
FLE-592 made the remote access gateway drop lossy data-track messages that exceed the per-message size limit (default 100 KiB), but the only signal was a throttled warning in the gateway's logs. From the viewer's side, messages on an oversized topic just silently never arrive — and the person watching the stream usually can't see the device's logs, so the topic looks broken with no explanation.

This PR surfaces that same drop report to viewers using the existing `Status`/`RemoveStatus` control-plane messages, which the app already renders in the Problems panel:

- When the gateway's throttled drop warning fires, a `Status` warning is also sent to that channel's current data subscribers.
- The status uses a stable id (`channel-drops-{id}`), so there's only one such problem report per channel.
- The drop warning is latched so it can be replayed to new subscribers immediately. This is useful, because otherwise the 30s throttle period is a very long time to wait.
- The status is cleared via `RemoveStatus`:
  - For all participants, once a channel has had no drops for 60s.
  - For all participants when the channel is removed.
  - For a particular participant that unsubscribes from the channel.

Manual testing:

- [x] Verified end-to-end in the Foxglove app: streaming a topic with oversized messages over remote access shows the warning in the Problems panel (naming the topic, drop count, and size limit), repeated reports replace the entry rather than stacking, and the warning clears automatically ~60–75s after the oversized traffic stops (this manual check is the coverage for the sweeper's timed clear, which has no end-to-end test).

</td><td>

Screenshot 2026-07-07 at 12 26 55 PM

</td></tr></table>

Fixes: [FLE-645](https://linear.app/foxglove/issue/FLE-645)

### DIFF
--- a/rust/foxglove/src/remote_access/session.rs
+++ b/rust/foxglove/src/remote_access/session.rs
@@ -388,7 +388,7 @@ impl Sink for RemoteAccessSession {
         }
 
         if let Some((report, subscriber_sids)) = drop_report {
-            self.emit_drop_status(channel, channel_id, report, subscriber_sids);
+            self.emit_drop_status(channel, report, subscriber_sids);
         }
 
         Ok(())
@@ -621,10 +621,11 @@ impl RemoteAccessSession {
     fn emit_drop_status(
         &self,
         channel: &RawChannel,
-        channel_id: ChannelId,
         report: OversizedDropReport,
         subscriber_sids: SmallVec<[ParticipantSid; 4]>,
     ) {
+        let channel_id = channel.id();
+
         // The registry read lock is released before we get here, so a
         // concurrent `remove_channel` may have unadvertised the channel and run
         // its eager clear in the meantime. Skip emitting in that case so we
--- a/rust/foxglove/src/remote_access/session.rs
+++ b/rust/foxglove/src/remote_access/session.rs
@@ -606,8 +606,8 @@ impl RemoteAccessSession {
 
         // Don't take `subscription_lock` here: listeners can synchronously log
         // from subscribe/unsubscribe callbacks while that lock is held. The
-        // subscriber snapshot may be slightly stale, but unsubscribe/remove and
-        // the quiet-period sweeper all clear the stable status id.
+        // subscriber snapshot may be stale, but the status will be eventually
+        // cleared by the sweeper.
         if !self.channel_registry.read().has_channel(&channel_id) {
             return;
         }
--- a/rust/foxglove/src/remote_access/session.rs
+++ b/rust/foxglove/src/remote_access/session.rs
@@ -212,17 +212,27 @@ fn build_drop_status(channel_id: ChannelId, topic: &str, report: OversizedDropRe
     .with_id(drop_status_id(channel_id))
 }
 
+/// A channel's most recent oversized-drop report and when it was emitted.
+///
+/// `last_report` feeds the quiet-period sweeper; `report` lets the warning be
+/// rebuilt and replayed to viewers who subscribe while drops are ongoing.
+#[derive(Debug, Clone, Copy)]
+struct ActiveDropStatus {
+    last_report: std::time::Instant,
+    report: OversizedDropReport,
+}
+
 /// Drains and returns the channels whose most recent drop report is at least
 /// `quiet_period` old, leaving fresher entries in place. Pure and timer-free so
 /// the sweeper's expiry logic can be unit-tested with a synthetic `now`.
 fn drain_expired_drop_statuses(
-    statuses: &mut HashMap<ChannelId, std::time::Instant>,
+    statuses: &mut HashMap<ChannelId, ActiveDropStatus>,
     now: std::time::Instant,
     quiet_period: Duration,
 ) -> Vec<ChannelId> {
     let mut expired = Vec::new();
-    statuses.retain(|channel_id, last_report| {
-        if now.duration_since(*last_report) >= quiet_period {
+    statuses.retain(|channel_id, status| {
+        if now.duration_since(status.last_report) >= quiet_period {
             expired.push(*channel_id);
             false
         } else {
@@ -307,10 +317,11 @@ pub(super) struct RemoteAccessSession {
     video_codec_override: Option<VideoCodec>,
     /// Per-message size limit for lossy data-track channels; larger messages are dropped.
     max_data_track_message_size: usize,
-    /// Per-channel time of the most recent oversized-drop `Status` warning
+    /// Per-channel record of the most recent oversized-drop `Status` warning
     /// emitted to viewers. Feeds the quiet-period sweeper, which sends
-    /// `RemoveStatus` once a channel stops dropping. See [`DROP_STATUS_QUIET_PERIOD`].
-    active_drop_statuses: parking_lot::Mutex<HashMap<ChannelId, std::time::Instant>>,
+    /// `RemoveStatus` once a channel stops dropping, and lets the warning be
+    /// replayed to late subscribers. See [`DROP_STATUS_QUIET_PERIOD`].
+    active_drop_statuses: parking_lot::Mutex<HashMap<ChannelId, ActiveDropStatus>>,
     /// The preferred encoder backend applied to published video tracks.
     /// [`VideoEncoderBackend::Auto`](super::gateway::VideoEncoderBackend::Auto) leaves the
     /// choice to libwebrtc.
@@ -627,9 +638,13 @@ impl RemoteAccessSession {
             return;
         }
 
-        self.active_drop_statuses
-            .lock()
-            .insert(channel_id, std::time::Instant::now());
+        self.active_drop_statuses.lock().insert(
+            channel_id,
+            ActiveDropStatus {
+                last_report: std::time::Instant::now(),
+                report,
+            },
+        );
 
         let status = build_drop_status(channel_id, channel.topic(), report);
         let encoded = encode_json_message(&status);
@@ -980,6 +995,31 @@ impl RemoteAccessSession {
                 );
                 for descriptor in &subscribe_result.newly_subscribed_descriptors {
                     listener.on_subscribe(&client, descriptor);
+                }
+            }
+        }
+
+        // Replay any active oversized-drop warning to the new data subscribers,
+        // so a viewer subscribing mid-drop-storm sees the warning immediately
+        // instead of waiting up to a throttle window for the next report. Video
+        // subscribers are excluded to match the original warning's audience
+        // (they bypass the data plane). The sends are queued while holding the
+        // map lock: the sweeper drains entries under this lock strictly before
+        // broadcasting `RemoveStatus`, so a replayed `Status` can never arrive
+        // after the `RemoveStatus` that clears it. `subscription_lock` (held
+        // above) serializes this against `remove_channel`'s eager clear.
+        {
+            let statuses = self.active_drop_statuses.lock();
+            if !statuses.is_empty() {
+                for descriptor in &subscribe_result.newly_subscribed_descriptors {
+                    if !data_channel_ids.contains(&descriptor.id()) {
+                        continue;
+                    }
+                    if let Some(active) = statuses.get(&descriptor.id()) {
+                        let status =
+                            build_drop_status(descriptor.id(), descriptor.topic(), active.report);
+                        participant.send_control(encode_json_message(&status));
+                    }
                 }
             }
         }
@@ -3091,6 +3131,16 @@ mod tests {
         assert_eq!(format_byte_size(2 * 1024 * 1024), "2 MiB");
     }
 
+    fn active_drop_status(last_report: std::time::Instant) -> ActiveDropStatus {
+        ActiveDropStatus {
+            last_report,
+            report: OversizedDropReport {
+                dropped_since_last: 1,
+                size_limit: 1200,
+            },
+        }
+    }
+
     #[test]
     fn drain_expired_drop_statuses_removes_only_stale_entries() {
         let quiet = Duration::from_secs(60);
@@ -3100,9 +3150,9 @@ mod tests {
         let fresh = ChannelId::new(2);
         let mut statuses = HashMap::from([
             // Idle exactly the quiet period → expired (boundary is inclusive).
-            (stale, now - quiet),
+            (stale, active_drop_status(now - quiet)),
             // Idle less than the quiet period → retained.
-            (fresh, now - Duration::from_secs(30)),
+            (fresh, active_drop_status(now - Duration::from_secs(30))),
         ]);
 
         let expired = drain_expired_drop_statuses(&mut statuses, now, quiet);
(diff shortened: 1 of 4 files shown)


Avan1ko:
Both open items from the last review pass are addressed in 69875ab8:

**Late joiners** — active drop statuses are now replayed, scoped to *subscribe* time rather than participant add: the warning belongs to a channel's data subscribers, and a freshly joined participant hasn't subscribed to anything yet, so join-time replay would warn viewers who never look at the topic. The map now retains the last `OversizedDropReport` alongside its timestamp, and `handle_client_subscribe` rebuilds and sends the warning to the new data subscriber. The replay send is queued while holding the status-map lock, which orders it before any concurrent sweeper `RemoveStatus` — a replayed warning can never arrive after the message that clears it. Covered end-to-end: viewer-2 subscribes mid-storm and receives the status with the stable id, with no new drop report possible inside the throttle window.

**Clear paths untested** — the eager clear-on-remove is now exercised end-to-end: the integration test closes the channel and asserts every viewer receives `Unadvertise` followed by `RemoveStatus` with the channel's status id. The sweeper's *timed* clear remains without an end-to-end test on purpose: the 60s quiet period isn't configurable, so a real-time test would add ~75s of wall clock for one assertion. Its expiry decision logic is unit-tested with a synthetic clock, and the timed clear (including the ~60–75s latency) was verified manually in the app — noted in the PR description's Testing section.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -804,6 +996,31 @@ impl RemoteAccessSession {
...
+        {
+            let statuses = self.active_drop_statuses.lock();
+            if !statuses.is_empty() {
+                for descriptor in &subscribe_result.newly_subscribed_descriptors {
+                    if !data_channel_ids.contains(&descriptor.id()) {
+                        continue;
+                    }
+                    if let Some(active) = statuses.get(&descriptor.id()) {
+                        let status =
+                            build_drop_status(descriptor.id(), descriptor.topic(), active.report);
```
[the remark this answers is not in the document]
Avan1ko: I did it this way for consistency across users, since this is using the device-side gateway's throttled warns. The message drops tapering off/increasing are just a part of using a throttled warning. I think this is chill

rust/foxglove/src/remote_access/session.rs
```diff
@@ -286,6 +387,10 @@ impl Sink for RemoteAccessSession {
             }
         }
 
+        if let Some((report, subscriber_sids)) = drop_report {
+            self.emit_drop_status(channel, channel_id, report, subscriber_sids);
```
dave-foxglove: It _looks_ like we could pass just channel here and get the channel_id from that. Am I missing a reason to pass it separately?
OR
Since the call looks like it only uses channel.topic() currently, we could pass channel_topic as the parameter.

Avan1ko: Good catch, no specific reason. I think Claude just did it since channel_id would already be available, but I just changed it to pass channel alone.

rust/foxglove/src/remote_access/session/data_track.rs
```diff
@@ -12,6 +12,20 @@ use crate::{ChannelId, Metadata};
...
 const FRAME_HEADER_SIZE: usize = 8; // u16 LE flags + u16 LE data_offset + u32 LE sequence
 
+/// Details of a throttled oversized-message drop, surfaced to viewers as a
+/// `Status` warning.
+///
+/// Returned by [`DataTrack::log`] only at the moment the throttled gateway-side
+/// warning fires, so the viewer-facing signal reuses the same throttle window
+/// and aggregated count without a second throttle.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) struct OversizedDropReport {
```
dave-foxglove, line 25: Let's chat about whether this is over-focused on one topic. It feels odd to me to have the log function return an OversizedDropReport for example, compared to say, reporting metadata on the channel. Now, maybe the only metadata fields are the drops and size limit, but if there were to be further things added later, a more general name may make a nicer reporting mechanism. Hmm, I may just be suggesting changing the name here from  OversizedDropReport to something like ChannelReport (or open to other ideas - ChannelMetadata, ChannelStatus, ChannelStats, ...)

rust/foxglove/src/remote_access/session.rs
```diff
@@ -153,6 +162,85 @@ pub(super) fn encode_binary_message<'a>(message: &impl BinaryMessage<'a>) -> Byt
...
+    let formatted = format!("{value:.2}");
+    formatted
+        .trim_end_matches('0')
+        .trim_end_matches('.')
+        .to_string()
+}
+
+/// Builds the viewer-facing warning for a throttled oversized data-track drop.
+fn build_drop_status(channel_id: ChannelId, topic: &str, report: OversizedDropReport) -> Status {
+    let messages = if report.dropped_since_last == 1 {
```
dave-foxglove, line 197: Pluralizing text like this generally doesn't translate well. Though it also looks like this isn't being translated which could be an issue as well.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -153,6 +162,85 @@ pub(super) fn encode_binary_message<'a>(message: &impl BinaryMessage<'a>) -> Byt
...
+        "message"
+    } else {
+        "messages"
+    };
+    Status::warning(format!(
```
dave-foxglove, line 202: I like that this is being formatted nicely, but
- the presentation logic should be in the client (the app) rather than here (a kind of server)
- iiuc, this will not be translated on the client (while the rest of the app can/will be - so this will be just odd English text)
- the machine sending this is probably more cpu and and network limited (another reason to do it client side).

Can we send the pieces of data without formatting the string here

rust/foxglove/src/remote_access/session/data_track.rs
Avan1ko: I was thinking this log makes sense since its isolated onto a gateway-side restriction on file size, and is only triggered upon restriction. Other "metadata" or channel stats I think can mostly be derived from the messages itself? Lmk what you think

dave-foxglove: Hmm, I think something that is catching me is that we make a call to log a message and when that fails we return Some(...). The OversizedDropReport reads like an error and my first thought was to make it not read like an error. I'll make a note about using a Result below.

rust/foxglove/src/remote_access/session/data_track.rs
```diff
@@ -121,7 +135,17 @@ impl DataTrack {
...
+    /// Returns `Some(OversizedDropReport)` only on the throttled oversized-drop
+    /// path — the same moment the gateway-side warning fires — so the caller can
+    /// surface a matching viewer signal. Returns `None` on all other paths
+    /// (delivered, throttled, not-ready, queue-full).
+    pub fn log(
+        &self,
+        channel_id: ChannelId,
+        msg: &[u8],
+        metadata: &Metadata,
+    ) -> Option<OversizedDropReport> {
```
dave-foxglove: This feels like I'm getting something when something is going wrong. How about changing this to a Result of Ok(()) on success and  Err(OversizedDropReport).

rust/foxglove/src/remote_access/session/data_track.rs
```diff
@@ -12,6 +12,20 @@ use crate::{ChannelId, Metadata};
...
+/// Returned by [`DataTrack::log`] only at the moment the throttled gateway-side
+/// warning fires, so the viewer-facing signal reuses the same throttle window
+/// and aggregated count without a second throttle.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) struct OversizedDropReport {
+    /// Oversized messages dropped since the last report, including this one.
+    pub dropped_since_last: u64,
+    /// The configured per-message size limit, in bytes.
+    pub size_limit: usize,
+}
```
dave-foxglove, line 30: Since this is eventually sent to the client, should we consider sending start and end times on this data such that it can correlated with the presentation (of the data).

rust/foxglove/src/remote_access/session/data_track.rs
```diff
@@ -131,16 +155,19 @@ impl DataTrack {
                     msg.len(),
                     self.max_message_size
                 );
-            } else {
-                self.oversized_dropped.fetch_add(1, Ordering::Relaxed);
+                return Some(OversizedDropReport {
```
dave-foxglove: Rather than tracking and sending an out-of-band report of dropped messages, consider changing the in-band message to a message dropped message. I.e. we are still connected and able to send a message, this message was simply beyond a threshold were it's being dropped instead of being sent. We could still send a tiny message noting that the message was not sent. This provides accurate reporting of what and when messages were dropped without extra accounting or formatting GUI messages remotely (here).

dave-foxglove (dismissed):
Sorry, I made a few comments before realizing that I'd like to propose a pivot to how this is being reported. See https://github.com/foxglove/foxglove-sdk/pull/1354/changes#r3547910256

rust/foxglove/src/remote_access/session/data_track.rs
gasmith: That's not a bad idea, though it'd require a protocol change, to teach the viewer how to receive these tiny messages. It has the nice property of keeping presentation-layer concerns at the presentation layer.

On the other hand, the protocol overhead of lots of tiny little messages that the viewer is ultimately going to collapse into a summary is a bit of a bummer. We'd also be sending these messages over a lossy channel, for what it's worth, which means the viewer would only have partial information about tx errors.

If we do go down this path, we should consider the fact that the SDK may choose to drop messages for other reasons. For example, it might fail to decode an image for video transcoding. And in that case, we don't have a handy data track on which to send the tiny error message.

rust/foxglove/src/remote_access/session.rs
gasmith: I agree with the first two arguments, but I don't think formatting a string for a throttled log message is going to break the bank.

gasmith (commented):
On the whole, this is the shape I expected to see. There's a bit more machinery than I had anticipated (e.g., string formatting and the sweeper), but I suppose that's the price we pay for implementing this on the SDK side.

I like Dave's suggestions, but it's not something we're going to solve in this PR. We have other precedent of using Status messages to send lightweight notifications to the client, and while I strongly agree that I'd love for this to be sent as structured data, so that we can keep presentation layer concerns in the presentation layer, that's a larger undertaking, and it's going to take some time to design.

I'm inclined to suggest that we proceed with this approach for now, and file a ticket for future improvements. We urgently need _some_ user-visible indication that we're dropping over-sized messages.

Avan1ko:
Improvements to the current implementation require a whole protocol change, deferring to [FLE-663](https://linear.app/foxglove/issue/FLE-663/structured-error-reporting-protocol-for-the-remote-access-gateway), moving ahead with current implementation

rust/foxglove/src/remote_access/session/data_track.rs
Avan1ko: Notes added to FLE-663

dave-foxglove: Offline we agreed that a different solution is something we can consider in the future, but for now this solution is passable.

rust/foxglove/src/remote_access/session/data_track.rs
Avan1ko: Makes sense after our discussion, just committed this change.

rust/foxglove/src/remote_access/session/data_track.rs
Avan1ko: Noted on FLE-663

rust/foxglove/src/remote_access/session.rs
Avan1ko: Noted on FLE-663

rust/foxglove/src/remote_access/session.rs
Avan1ko: Noted on FLE-663

dave-foxglove (commented):
My concerns and requested changes are postponed by FLE-663. The remaining PR is what I think Greg was asking for. LG based on that.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -153,6 +163,91 @@ pub(super) fn encode_binary_message<'a>(message: &impl BinaryMessage<'a>) -> Byt
...
+
+/// Renders `value` with up to two decimal places, dropping any trailing zeros
+/// and a bare trailing decimal point (`100.00` → `100`, `1.50` → `1.5`).
+fn trim_trailing_zeros(value: f64) -> String {
+    let formatted = format!("{value:.2}");
+    formatted
+        .trim_end_matches('0')
+        .trim_end_matches('.')
+        .to_string()
+}
```
gasmith, line 193: Aw, c'mon, rust... I am flabbergasted and dismayed that there is not a format string for this.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -153,6 +163,91 @@ pub(super) fn encode_binary_message<'a>(message: &impl BinaryMessage<'a>) -> Byt
...
+/// count is a rolling per-window tally that updates in place, not a cumulative
+/// total — statuses carry no timestamp, so the message must say so itself.
+fn build_drop_status(channel_id: ChannelId, topic: &str, report: OversizedDropReport) -> Status {
+    let messages = if report.dropped_since_last == 1 {
+        "message"
+    } else {
+        "messages"
+    };
+    Status::warning(format!(
+        "Dropped {} {messages} on topic {topic} in the last {} seconds: \
```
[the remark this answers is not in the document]
gasmith: I'm kinda inclined towards towards just dropping the "in the last N seconds" entirely, and not replacing it with "recently". The fact that the warning is transient and removed after a relatively short period already speaks to its recency, and marginal inaccuracy in the number of messages dropped "recently" vs "closer to the previous warning" is not worth wringing our hands over.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -804,6 +1004,31 @@ impl RemoteAccessSession {
...
+                        continue;
+                    }
+                    if let Some(active) = statuses.get(&descriptor.id()) {
+                        let status =
+                            build_drop_status(descriptor.id(), descriptor.topic(), active.report);
+                        participant.send_control(encode_json_message(&status));
+                    }
+                }
+            }
+        }
```
gasmith, line 992: I think we can probably remove this bit. If we continue to drop messages, we'll eventually inform the client, which is good enough.

gasmith (approved):
Two minor suggestions and an observation of sadness. Otherwise LGTM.

rust/foxglove/src/remote_access/session.rs
gasmith: OK, after playing with this a bit, I see the value in promptly reporting drops after subscribe. This is particularly important if you unsubscribe and resubscribe rapidly, which a user might very well do e.g. in the 3d panel when a point cloud that isn't rendering.

PR #1361, opened 2026-07-09
fix(ros): resolve service-introspection event schemas to keep the graph poll alive

jbcpollak:
## Description

When ROS 2 [service introspection](https://design.ros2.org/articles/service_introspection.html) is enabled on a service client or server, rcl publishes a hidden `<service>/_service_event` topic whose type is the rosidl-generated `<pkg>/srv/<Srv>_Event` message. Running `foxglove_bridge` with `include_hidden:=true` (so these hidden topics are visible) permanently breaks topic discovery:

1. `<Srv>_Event` messages are generated by rosidl and have **no definition file on disk**, so `MessageDefinitionCache::get_full_text` throws `DefinitionNotFoundError`. `updateAdvertisedTopics` catches this and advertises the channel with a **null schema** (`schema = std::nullopt`), which is fine on its own.
2. On the **next** graph change, the channel-removal loop and the duplicate-advertisement check both call `channel.schema().value()` on that schema-less channel, throwing `std::bad_optional_access`. The exception propagates out of `updateAdvertisedTopics` and is only caught at the top of `rosgraphPollThread`:
   ```
   [ERROR] Exception thrown in rosgraphPollThread: bad optional access
   ```
   Because it unwinds before any new topic is advertised, and recurs on every subsequent poll, **no topic that appears after the first service-event topic is ever advertised** to connected clients. Topics advertised before it keep working, which makes this look intermittent.

This reproduces on any graph containing a service-event topic (plain services and the implicit action services alike) whenever the bridge can see hidden topics.

## Changes

- **`message_definition_cache.cpp`**: synthesize `<Srv>_Event` message definitions from the service's request/response definitions, mirroring the fields rosidl generates (`service_msgs/msg/ServiceEventInfo info` plus bounded `request`/`response` arrays). Also covers the implicit action services (`<Action>_SendGoal_Event`, `<Action>_GetResult_Event`). Service events now advertise with a real schema and decode in clients.
- **`ros2_foxglove_bridge.cpp`**: harden every `channel.schema().value()` access so a schema-less channel can never abort the graph poll again. Schema-less channels are matched by topic alone (avoiding advertise/remove churn), and their datatype is resolved from the ROS graph when a client subscribes.
- **Tests**:
  - `message_definition_cache_test.cpp` (new): service and action event schema synthesis, transitive dependency embedding, and the unknown-type error path.
  - `smoke_test.cpp`: regression test `serviceEventTopicDoesNotPreventFutureAdvertisements` — publishes a `std_srvs/srv/SetBool_Event`-typed topic, then a `std_msgs/String` topic, and asserts the second is still advertised. It fails against the current bridge with the exact `bad optional access` error and passes with this change.
- `package.xml` / `CMakeLists.txt`: add `service_msgs` (exec) and `test_msgs` (test) dependencies used by the synthesized schemas and the new test.

## Verification

- New regression test fails on the unpatched bridge (`Exception thrown in rosgraphPollThread: bad optional access`) and passes with the fix.
- Full package test suite green on ROS 2 Kilted: `colcon test` → 45/45 tests, 0 failures.


ros/src/foxglove_bridge/CMakeLists.txt
```diff
@@ -248,6 +248,13 @@ if(BUILD_TESTING)
   ament_add_gtest(utils_test tests/utils_test.cpp)
   target_link_libraries(utils_test foxglove_bridge_component)
   enable_strict_compiler_warnings(utils_test)
+
+  # message_definition_cache_test resolves schemas from installed interface packages
+  # (std_srvs for services, test_msgs for actions) via the ament index at runtime.
+  find_package(test_msgs REQUIRED)
```
clalancette, line 254: This isn't necessary; we are never reference test_msgs during CMake/build time, only during runtime.


```suggestion
```

ros/src/foxglove_bridge/package.xml
```diff
@@ -34,6 +35,9 @@
     <build_depend>ros_environment</build_depend>
 
     <depend>rcl_interfaces</depend>
+    <!-- Service event (<Srv>_Event) schemas synthesized by the message definition cache
+         reference service_msgs/msg/ServiceEventInfo, resolved via the ament index. -->
+    <exec_depend>service_msgs</exec_depend>
```
clalancette, line 40: ROS 2 Humble doesn't have service_msgs (that was introduced with Iron), so this needs to be:

```suggestion
    <exec_depend condition="$ROS_DISTRO != humble">service_msgs</exec_depend>
```

ros/src/foxglove_bridge/tests/smoke_test.cpp
```diff
@@ -846,6 +846,33 @@ TEST(FetchAssetTest, fetchNonExistingAsset) {
   EXPECT_FALSE(response.errorMessage.empty());
 }
 
+TEST(SmokeTest, serviceEventTopicDoesNotPreventFutureAdvertisements) {
```
clalancette, line 849: This test can't compile on Humble since it doesn't have `::Event`.  We can gate this with:

```
#if __has_include(<service_msgs/msg/service_event_info.hpp>)
```

ros/src/foxglove_bridge/tests/message_definition_cache_test.cpp
```diff
@@ -0,0 +1,64 @@
...
+  const auto [format, text] = cache.get_full_text("std_srvs/srv/SetBool_Request");
+  EXPECT_EQ(format, MessageDefinitionFormat::MSG);
+  EXPECT_NE(std::string::npos, text.find("bool data"));
+}
+
+TEST(MessageDefinitionCacheTest, ServiceEvent) {
+  // <Srv>_Event messages are generated by rosidl for service introspection and have no
+  // definition file of their own; the cache synthesizes them.
+  MessageDefinitionCache cache;
+  const auto [format, text] = cache.get_full_text("std_srvs/srv/SetBool_Event");
```
clalancette, line 21: Same problem on Humble here; these can compile there, but will never succeed since there is no `SetBool_Event`.  Gate this on the header existing.

ros/src/foxglove_bridge/tests/smoke_test.cpp
```diff
@@ -846,6 +846,33 @@ TEST(FetchAssetTest, fetchNonExistingAsset) {
...
+  // Regression test: topics typed with a rosidl-generated service event message (<Srv>_Event,
+  // published when service introspection is enabled) have no definition file on disk. They used
+  // to be advertised with a null schema, and every subsequent rosgraph poll then threw
+  // bad_optional_access (channel.schema().value()) before advertising anything, permanently
+  // stopping discovery of new topics.
```
clalancette, line 854: There's no real need to have history in comments; let's just remove this.

clalancette (changes requested):
First of all, thanks for the fix.  This is a bit of a specific corner case, so I'm glad you've uncovered it and made a patch for it.

Besides the stuff I've left inline, this needs to be rebased on the latest `main` to resolve a conflict.  Thanks!

clalancette:
Closing this in favor of #1407

PR #1372, opened 2026-07-14
FLE-671: standardize bridge terminology (whitelist → allowlist)

dave-foxglove:
### Changelog

Renamed the ROS 2 bridge's five topic/service/parameter list parameters from `*_whitelist` to `*_allowlist` (`topic_allowlist`, `service_allowlist`, `param_allowlist`, `client_topic_allowlist`, `best_effort_qos_topic_allowlist`); the old `*_whitelist` names keep working as deprecated aliases.

### Docs

Updated `ros/src/foxglove_bridge/README.md` (parameter names, a deprecation note, and how to disable a list). No separate docs PR needed.

### Description

The bridge's list parameters were inconsistently named: `asset_uri_allowlist` and `video_transcode_topic_denylist` already used allow/deny terminology, while five siblings still used `whitelist`. This standardizes the remaining five on `allowlist` (and fixes the `WhiteList`/`Whitelist` casing in the internal identifiers). No `blacklist` exists, so there is nothing to convert to `denylist`.

Because the parameter and launch-argument names are public API, the old names are kept working rather than removed:

- **Node parameters** (YAML params files, `--ros-args -p`, or a `<param>` in a user's own launch file): each deprecated `*_whitelist` parameter is still declared; a pure `resolveAliasedStringArray` prefers a non-empty deprecated value and `getStringArrayParamWithDeprecatedAlias` logs a one-time deprecation warning pointing at the new name.
- **Launch arguments** of the provided launch file: each deprecated `*_whitelist` arg defaults to its `*_allowlist` arg and feeds the `*_allowlist` parameter, so `ros2 launch ... topic_whitelist:=...` still overrides. (This path routes into the canonical parameter and therefore does not emit the node's deprecation warning — noted in a launch-file comment.)

An empty array can't serve as the "unset" sentinel via a launch argument because a ROS parameter override cannot be an empty string array — a bare `[]` is parsed as an untyped empty array and leaves the parameter uninitialized. That constraint (verified against ROS) is also why the shipped "match nothing" default is `["(?!)"]` rather than `[]`; a new `REGEX_MATCH_NOTHING` constant documents this where it's discoverable.

Historical `CHANGELOG.rst` entries are left intact with bracketed `[edit: now called ...]` annotations rather than rewritten.

Manual testing for reviewers:

- [ ] `ros2 launch foxglove_bridge foxglove_bridge_launch.xml topic_whitelist:="['/some_topic']"` still restricts topics (old launch arg).
- [ ] Setting `topic_whitelist` via a params file / `-p` logs a deprecation warning and still takes effect.
- [ ] `topic_allowlist` (new name) works and does not warn.

Verified in Docker (ROS rolling): bridge builds, all bridge tests pass (25 smoke + 18 utils, including the two new `resolveAliasedStringArray` cases), and the launch-arg alias routing was confirmed end-to-end.

Fixes: [FLE-671](https://linear.app/foxglove/issue/FLE-671)


clalancette:
We decided not to do this right now, so closing this PR out.

PR #1373, opened 2026-07-15
img2yuv: fix compressedDepth detection and align format parsing with frontend

gasmith:
### Changelog
- Remote access no longer attempts to transcode compressedDepth images as video.
- Remote access recognizes ROS-style format strings in foxglove.CompressedImage messages.

### Docs
None

### Description
`compressedDepth` payloads carry a transport-specific header before the
codec's data, so treating the inner codec name as a directly-decodable
Compression variant (as `try_from_ros_format` did for "png") produced
garbage.

This change reports `compressedDepth` as an unknown compression instead.

This change also updates the library to more closely mirror [frontend behavior](https://github.com/foxglove/app/blob/6f8d31885b6625ccd82e812aee8a3fc4159c0cfd/packages/viz/src/panels/ThreeDeeRender/renderables/Images/decodeImage.ts#L67-L93) for
parsing the format string. Previously, img2yuv only attempted to parse
ROS-style format strings for ROS1/ROS2 messages, but the frontend also
parses this style of format string for `foxglove.CompressedImage`
messages.

### Links
Fixes: FLE-673

### DIFF
--- a/rust/foxglove/src/img2yuv/compressed.rs
+++ b/rust/foxglove/src/img2yuv/compressed.rs
@@ -37,58 +37,15 @@ impl From<Compression> for image::ImageFormat {
 impl FromStr for Compression {
     type Err = UnknownCompressionError;
 
-    /// Parses a compressed image format string.
-    ///
-    /// This accepts a bare codec name (e.g. `"png"`), as used by the Foxglove `CompressedImage`
-    /// schema, as well as the documented [ROS 1][ros1] and [ROS 2][ros2]
-    /// `sensor_msgs/CompressedImage` format string, which additionally encodes the original pixel
-    /// format and a `compressed` marker:
-    ///
-    /// - `CODEC`
-    /// - `ORIG_PIXFMT; CODEC compressed [COMPRESSED_PIXFMT]`
-    ///
-    /// Other strings are rejected even if they contain a codec token.
-    ///
-    /// `ORIG_PIXFMT; compressedDepth CODEC` (the `compressed_depth_image_transport` format) is
-    /// always reported as an unknown compression: the payload has a transport-specific header
-    /// before the codec's data, so it can't be decoded as an ordinary compressed image of that
-    /// codec.
-    ///
-    /// [ros1]: https://docs.ros.org/en/noetic/api/sensor_msgs/html/msg/CompressedImage.html
-    /// [ros2]: https://docs.ros.org/en/rolling/p/sensor_msgs/msg/CompressedImage.html
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         let normalized = s.trim().to_ascii_lowercase();
-        if normalized.is_empty() {
-            return Err(UnknownCompressionError(s.to_string()));
-        }
-
-        if let Some(compression) = Self::from_codec(&normalized) {
-            return Ok(compression);
-        }
-
-        let Some((orig_pixfmt, rest)) = normalized.split_once(';') else {
-            return Err(UnknownCompressionError(s.to_string()));
-        };
-        if orig_pixfmt.split_whitespace().count() != 1 {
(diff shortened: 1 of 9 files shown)


rust/foxglove/src/img2yuv/compressed.rs
```diff
@@ -37,19 +37,70 @@ impl From<Compression> for image::ImageFormat {
...
+        if orig_pixfmt.split_whitespace().count() != 1 {
+            return Err(UnknownCompressionError(s.to_string()));
+        }
+
+        let mut tokens = rest.split_whitespace();
+        let Some(codec) = tokens.next() else {
+            return Err(UnknownCompressionError(s.to_string()));
+        };
+        if codec == "compresseddepth" || tokens.next() != Some("compressed") {
+            return Err(UnknownCompressionError(s.to_string()));
```
clalancette: I think we can do slightly better on this error message, saying something more like: `depth data through the video channel is not supported`.

rust/foxglove/src/img2yuv/compressed.rs
```diff
@@ -37,19 +37,70 @@ impl From<Compression> for image::ImageFormat {
...
+        }
+        let _compressed_pixfmt = tokens.next();
+        if tokens.next().is_some() {
+            return Err(UnknownCompressionError(s.to_string()));
+        }
+
+        Self::from_codec(codec).ok_or_else(|| UnknownCompressionError(s.to_string()))
```
clalancette: I'm wondering if this can be simplified with something like:

```
    let orig: Vec<&str> = orig.split_whitespace().collect();
    let rest: Vec<&str> = rest.split_whitespace().collect();

    match (orig.as_slice(), rest.as_slice()) {
        ([_], ["compresseddepth", ..]) => {
            Err(UnknownCompressionError::CompressedDepth(s.to_string()))
        }
        ([_], [codec, "compressed"] | [codec, "compressed", _]) => {
            Self::from_codec(codec).ok_or_else(unknown)
        }
        _ => Err(unknown()),
```

clalancette (commented):
Overall, this whole thing makes sense.  I left a couple of minor points for improvement.

rust/foxglove/src/img2yuv/compressed.rs
gasmith: I agree somewhat, but if we want to do that, we should do it at a higher layer. There's a possible future where this library _does_ support decompressing depth images, just not for the sake of transcoding to video via YUV.

Maybe the right answer is to have a separate error variant for compressed depth images. Or a method `UnknownCompressionError::is_compressed_depth()`? Maybe we'll cross that bridge when we get there.

Also note that there's a little bit of asymmetry here; we can only do this more nuanced error reporting for compressed depth images... for raw depth images, there's nothing to distinguish them from monochrome camera data.

rust/foxglove/src/img2yuv/compressed.rs
gasmith: Yeah, that reads a bit better. Will do.

gasmith (commented):
Thanks for the review! Apart from your suggestions, I decided that it was also better to keep the explicit `Compression::try_from_ros_format()` rather than hiding it under the `FromStr` trait (which is vestigial at this point).

rust/foxglove/src/img2yuv/compressed.rs
clalancette: Yeah, it did occur to me that if we want this to be a "generic" img2yuv library, we actually do want to support decompressing compressedDepth images.  We just don't want to do it for the video transcoding pipeline.  I'm OK with leaving this generic error message for now; I think we'll want to eventually make this library more broadly useful, but it doesn't have to be today.

clalancette (approved):

PR #1374, opened 2026-07-15
Recommend the FOXGLOVE_DEVICE_TOKEN env var over the device_token param

clalancette:
### Changelog

None.

### Docs

None.

### Description

A token passed as a parameter is readable by other nodes via the ROS 2 parameter services. Document that the environment variable is the preferred way to provide the token.

I verified that this works fine with the ROS 2 bridge locally.

Part of FLE-601


gasmith (approved):
🚢

PR #1388, opened 2026-07-20
Bump MSRV to 1.88

eloff:
### Changelog

None

### Docs

None

### Description

Raises the workspace MSRV (and the playground rustc / Emscripten / Pyodide pin chain) to 1.88 so the declared toolchain matches what the full feature set already requires — remote-access let-chains and locked dependency `rust-version` values.

**This is PR 1/9 in the series — foundation; review and merge first.**

**FLE-613 PR series (review in order where noted):**

| # | PR | Base | Depends on |
|---|----|------|------------|
| 1/9 | #1388 Bump MSRV to 1.88 | `main` | — |
| 2/9 | #1389 Fix PointCloud transform examples | `main` | — (independent) |
| 3/9 | #1390 Harden C++ websocket callback tests | `main` | — (independent; land before #1392) |
| 4/9 | #1391 Add RA Draco point cloud compression | #1388 | #1388 |
| 5/9 | #1392 Expose point cloud compression in C/C++ | #1391 | #1391 (+ includes #1390 until rebase) |
| 6/9 | #1393 Expose point cloud compression in Python | #1392 | #1392 |
| 7/9 | #1394 Add RA point cloud compression example | #1391 | #1391 |
| 8/9 | #1395 Add ROS2 PointCloud2 compression input | #1393 | #1393 |
| 9/9 | #1396 Add JSON/FlatBuffer PointCloud inputs | #1395 | #1395 |

Parent: [FLE-613](https://linear.app/foxglove/issue/FLE-613) · Original monolith: #1371

Fixes: [FLE-682](https://linear.app/foxglove/issue/FLE-682)

PR #1391, opened 2026-07-20
Add RA Draco point cloud compression

eloff:
### Changelog

Remote access now transparently uses Draco to lossily compress `foxglove.PointCloud` channels by default; compression can be configured or disabled on a per-topic basis.

### Docs

https://github.com/foxglove/app/pull/17110

### Description

Point cloud messages are often too large for lossy remote-access data tracks. This adds the `draco` crate to the `remote-access` feature with `compress_point_cloud` / `PointCloud::encode_draco`, and transparent remote-access transcoding: protobuf `foxglove.PointCloud` channels are advertised as `foxglove.CompressedPointCloud` and compressed off the logging hot path (queue depth 2).

Also included:
- Per-channel `point_cloud_compression` opt-out
- Reliable QoS channels skip compression automatically (lossy + drop-under-pressure would violate Reliable)
- 2-axis clouds are zero-filled to 3-axis for the encoder / app
- LiveKit integration tests

WebSocket sink is intentionally untouched.

### Testing

Tested by running the example in #1394 , which is built on top of this one.  With that running, I can visualize the compressed point cloud topic, but the uncompressed one shows the warning in the Problems tab.

**This is PR 4/9 in the series — core feature.**

**FLE-613 PR series (review in order where noted):**

| # | PR | Base | Depends on |
|---|----|------|------------|
| 1/9 | #1388 Bump MSRV to 1.88 | `main` | closed in favor of #1381  |
| 2/9 | #1389 Fix PointCloud transform examples | `main` | merged |
| 3/9 | #1390 Harden C++ websocket callback tests | `main` | merged |
| 4/9 | #1391 Add RA Draco point cloud compression | `main` | #1388 |
| 5/9 | #1392 Expose point cloud compression in C/C++ | #1391 | #1391 |
| 6/9 | #1393 Expose point cloud compression in Python | #1391 | #1391 |
| 7/9 | #1394 Add RA point cloud compression example | #1391 | #1391 |
| 8/9 | #1395 Add ROS2 PointCloud2 compression input | #1392 | #1392 |
| 9/9 | #1396 Add JSON/FlatBuffer PointCloud inputs | #1395 | #1395 |

Fixes: [FLE-684](https://linear.app/foxglove/issue/FLE-684)


rust/foxglove/src/remote_access/gateway.rs
```diff
@@ -279,6 +291,11 @@ impl Default for Gateway {
             message_backlog_size: None,
             max_data_track_message_size: None,
             video_encoder: VideoEncoderBackend::Auto,
+            // Transparent point-cloud compression is opt-out: enabled by default.
+            #[cfg(feature = "draco")]
+            point_cloud_compression: Some(crate::draco::CompressPointCloudOptions::default()),
```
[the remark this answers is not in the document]
clalancette: This is actually an improvement; today, all but the smallest pointclouds can't even be delivered (we have a 100KB cap on lossy channels).  But good callout about the Changelog, I'll add that.

clalancette:
During testing we found that the app cannot handle draco-compressed pointclouds in sequential mode with quantization > 0.  As I was debugging this, I also found that there are 3 situations where we can actually fail to compress at all:

1.  When quantization = 0, there is no work for the compression to do an compression is essentially skipped.
2.  When the pointcloud is empty, no compression happens.
3.  When the pointcloud has float64 data in it, draco cannot handle that and does not compress.

So the latest pushes here do the following:
a.  Remove sequential mode as a user-configurable option.  The infrastructure is left in place in case we want to add it in the future, once we fix the app.
b.  Don't allow quantization = 0.  Since no compression happens in this case, this is just a more confusing case of opting-out of compression.  The SDK now throws a hard error in this case.
c.  If a pointcloud has float64 data, it will print a warning to the console and send a warning to the app every 30 seconds so the user knows something is wrong.

rust/foxglove/Cargo.toml
```diff
@@ -11,6 +11,11 @@ license.workspace = true
 default = ["aws-lc-rs", "derive", "lz4", "websocket", "zstd"]
 chrono = ["dep:chrono"]
 derive = ["dep:foxglove_derive"]
+# Draco point-cloud compression: enables `foxglove::draco` (the
+# `compress_point_cloud` helper) and transparent transcoding of
+# `foxglove.PointCloud` channels in the remote-access sink
+# (enabled by default on that sink; opt out via its builder).
```
gasmith: Docs for features live in lib.rs.

```suggestion
```

rust/foxglove/Cargo.toml
```diff
@@ -25,6 +30,7 @@ websocket = ["_remote-common", "dep:tokio-tungstenite"]
 live_visualization = ["websocket"]
 remote-access = [
   "_remote-common",
+  "draco",
```
gasmith: Maybe we should let folks opt out of draco for remote access? Folks might have their own compression pipeline, or no point clouds at all.

If we don't allow folks to opt out, there's not much point in publicly documenting the feature, because it's also unused without remote access, and so no one would ever have a reason to opt in.

rust/remote_access_tests/src/test_helpers.rs
```diff
@@ -872,6 +872,10 @@ pub struct TestGatewayOptions {
     pub qos_classifier: Option<QosClassifierFn>,
     pub suppress_video_transcode: Option<SuppressVideoTranscodeFn>,
     pub max_data_track_message_size: Option<usize>,
+    /// When set, overrides the gateway's point-cloud compression config: `Some(None)`
+    /// disables compression, `Some(Some(opts))` customizes it. `None` leaves the gateway
+    /// default (compression enabled).
```
gasmith:
```suggestion
```

rust/foxglove/src/remote_access/session/point_cloud_track.rs
```diff
@@ -0,0 +1,129 @@
...
+                        // quantization), so it must not stay silent: lossless output
+                        // provides no size reduction over the raw cloud.
+                        if let Some(field) = &transcoded.lossless_fallback_field
+                            && lossless_throttler.try_acquire()
+                        {
+                            let message = format!(
+                                "point cloud compression on channel {channel_id:?}: field \
+                                 '{field}' is float64 and cannot be quantized; delivering \
+                                 lossless Draco (no size reduction). Use float32 or integer \
+                                 fields, or suppress compression for this channel."
```
gasmith: Oh, that's a sad story. Why doesn't draco support float64?

rust/foxglove/src/remote_access/session/point_cloud_track.rs
```diff
@@ -0,0 +1,129 @@
...
+                    Ok(Err(e)) => {
+                        if warn_throttler.try_acquire() {
+                            let message = format!(
+                                "point cloud compression error on channel {channel_id:?}: {e}"
+                            );
+                            warn!("{message}");
+                            if let Some(session) = session.upgrade() {
+                                session.publish_status(Status::warning(message));
+                            }
+                        }
```
gasmith: We can defer this, but I think we need a common mechanism for delivering a per-channel status message only to subscribers, which we can retract when the subscriber unsubscribes. Basically we need to generalize the work @Avan1ko did to add drop status reporting. There's a backlog ticket for reporting video encoding errors which asks for the same thing: https://linear.app/foxglove/issue/FLE-676/remote-access-video-encoding-failures-should-be-reported

rust/foxglove/src/remote_access/session/point_cloud_track.rs
```diff
@@ -0,0 +1,129 @@
...
+            // Throttles compression warnings so a stream of bad clouds doesn't flood the log.
+            let mut warn_throttler = Throttler::new(TRANSCODE_WARN_INTERVAL);
+            // Separate throttler for lossless-downgrade notices, so a burst of transcode
+            // errors can't mask them (or vice versa).
+            let mut lossless_throttler = Throttler::new(TRANSCODE_WARN_INTERVAL);
+            while let Ok((data, log_time)) = consumer_rx.recv_async().await {
+                let result = tokio::task::spawn_blocking(move || {
+                    transcode_point_cloud_message(&data, &options)
+                })
+                .await;
```
gasmith, line 54: I wonder whether we might just spawn this whole as a blocking task, and avoid the double-context-switch.

rust/foxglove/src/remote_access/gateway.rs
```diff
@@ -451,6 +473,74 @@ impl Gateway {
...
+    ///
+    /// Compression is enabled by default with [`CompressPointCloudOptions::default()`].
+    /// Note that the default settings are lossy: kd-tree encoding with positions quantized
+    /// to 12 bits.
+    /// [`DracoEncodeOptions::quantization_bits`](crate::draco::DracoEncodeOptions::quantization_bits)
+    /// must be between `1` and
+    /// [`MAX_QUANTIZATION_BITS`](crate::draco::MAX_QUANTIZATION_BITS) inclusive;
+    /// [`Self::start`] rejects values outside that range. In particular `0` (lossless) is
+    /// rejected because lossless Draco encoding provides no size reduction over the raw
+    /// point cloud — pass `None` instead to deliver point clouds unmodified.
```
gasmith: Is it worthwhile to have two methods of globally disabling point cloud compression? You alternatively provide a suppression callback that unconditionally returns true.

rust/foxglove/src/remote_access/gateway.rs
```diff
@@ -451,6 +473,74 @@ impl Gateway {
...
+    ///
+    /// Clouds containing float64 fields cannot be quantized and are delivered losslessly
+    /// (no size reduction); a throttled warning is emitted on the device and to viewers
+    /// when this happens.
+    ///
+    /// Channels classified as [`Reliability::Reliable`](crate::remote_access::Reliability::Reliable)
```
gasmith: Nit: I feel like this is a bit buried, and should probably be closer to the top of the doc.

rust/foxglove/src/remote_access/gateway.rs
```diff
@@ -671,6 +761,26 @@ impl Gateway {
...
+        if let Some(options) = &self.point_cloud_compression {
+            if let Err(e) = options.validate() {
+                return Err(FoxgloveError::ConfigurationError(format!(
+                    "compress_point_clouds: {e}"
+                )));
+            }
+            // Lossless Draco encoding provides no size reduction over the raw point
+            // cloud, so on the transparent path it is all overhead: reject it rather
+            // than transcode for nothing. (The direct compress_point_cloud API still
+            // accepts 0 for lossless encoding.)
```
gasmith: The parenthetical seems superfluous.

```suggestion
            // than transcode for nothing.
```

rust/foxglove/src/remote_access/gateway.rs
```diff
@@ -671,6 +761,26 @@ impl Gateway {
...
+            if options.draco_options().quantization_bits == 0 {
+                return Err(FoxgloveError::ConfigurationError(
+                    "compress_point_clouds: quantization_bits is 0 (lossless), which \
+                     provides no size reduction; pass None to deliver raw point clouds \
+                     instead"
```
gasmith: If you take my comment about two methods for globally disabling pointcloud compression, then this error message needs to be updated to refer to suppress_point_cloud_compression.

gasmith (commented):
Still working my way through, but I have a first batch of comments

rust/foxglove/src/remote_access/session.rs
```diff
@@ -635,6 +704,58 @@ impl RemoteAccessSession {
...
+                // advertised schema is the raw `foxglove.PointCloud`. Drop the message
+                // (panicking in debug builds).
+                debug_assert!(false, "point-cloud publisher exists for Reliable channel");
+                if self.point_cloud_qos_error_throttler.lock().try_acquire() {
+                    error!(
+                        "dropping transcoded point cloud for Reliable channel \
+                         {channel_id:?}: compression should never be enabled for \
+                         Reliable channels"
+                    );
+                }
```
gasmith: I don't think we need such ceremony; the comment, assert, and return is enough.

rust/foxglove/src/draco.rs
```diff
@@ -0,0 +1,1143 @@
...
+            let base = p * stride;
+            for (c, coord) in coords.iter().enumerate() {
+                let v = match *coord {
+                    Some((off, dt)) => read_as_f32(&cloud.data, base + off, dt),
+                    None => 0.0,
+                };
+                let o = (p * 3 + c) * 4;
+                dst[o..o + 4].copy_from_slice(&v.to_le_bytes());
+            }
+        }
```
gasmith, line 420: Might want to file a backlog ticket for SIMD-optimization here... but maybe we wait until we actually know that users are sending 2D point clouds over RA. Which raises the question of how we would find out about that.

rust/foxglove/src/draco.rs
```diff
@@ -0,0 +1,1143 @@
...
+    // - The kd-tree encoder doesn't support float64 attributes, and sequential encoding
+    //   with quantized positions is rejected by the reference Draco decoder (upstream
+    //   draco-core conformance bug; see `DracoMethod`), so float64-bearing clouds must
+    //   be fully lossless to stay decodable.
+    let has_f64 = fields.iter().any(|f| f.dtype == DataType::Float64);
+    let quantization_bits = if num_points == 0 || has_f64 {
+        0
+    } else {
+        options.quantization_bits
+    };
```
gasmith, line 363: I think we should drop this fallback, because it's extremely unlikely to buy us anything. The result is strictly larger than the original, and so it's unlikely to make it over the wire. We should just tell the user to suppress compression for float64 point clouds until draco supports them.

rust/foxglove/src/draco.rs
```diff
@@ -0,0 +1,1143 @@
...
+        let attr_id = draco_cloud.add_attribute(attr);
+
+        let name = &cloud.fields[idx].name;
+        let mut attr_metadata = Metadata::new();
+        // `set_string` rejects empty values; a nameless field simply gets no metadata.
+        if attr_metadata.set_string("name", name).is_ok() {
+            draco_cloud
+                .metadata_or_insert()
+                .set_attribute_metadata(attr_id as u32, attr_metadata);
+        }
```
gasmith, line 460: I'm still trying to figure out exactly what this is for. The comment could use some clarification. Separately, I wonder whether it should be optional since it involves iterating over all the points.

gasmith (commented):
A few more comments, but I think this looks really good.

rust/foxglove/Cargo.toml
clalancette: We already have runtime opt-out, so having an opt-out feature here mostly saves build time/disk size when building the SDK from source.  But webrtc dominates in both of those categories, so once you opt-in to remote-access at all, opting out of draco at build time doesn't really save much.  I'd just leave it as a required part of remote-access

rust/foxglove/src/remote_access/session/point_cloud_track.rs
clalancette: We talked about this one in person, but basically the problem is that the KD-Tree encoder doesn't support quantizing float64, and so you have to fall back to sequential (which we've already ruled out for other reasons).

I traced further, and it looks like the Foxglove app frontend *also* doesn't support float64, so even if we did encode it successfully somehow, we couldn't visualize it anyway.

rust/foxglove/src/remote_access/session/point_cloud_track.rs
clalancette: This is mirroring existing code in remote-access:

https://github.com/foxglove/foxglove-sdk/blob/35869165f1d17ab58156428ef21ff1f2e42c029f/rust/foxglove/src/remote_access/session/video_track.rs#L203-L237

According to Claude, the current design is slightly better than a blocking task in that it gracefully handles panics in `transcode_point_cloud_message` and continues.  While I don't think that will be a very common occurrence, combined with the above it does nudge me in the direction of keeping it like it is.

rust/foxglove/src/remote_access/gateway.rs
clalancette: I somewhat agree that having two mechanisms isn't great.  What if we flipped this on its head?  We get rid of the global options completely, and instead change the per-channel `suppress_point_cloud_compression_fn` to either return None (don't compress), or a `CompressPointCloudOptions` (compress with these options)?  That would get rid of the duplication, and allow per-topic tuning.

Thoughts?

rust/foxglove/src/remote_access/session.rs
clalancette: I went this way because the assert is in debug mode, so in the release binaries it wouldn't be there at all.  But this is generally unreachable, so yeah, I'll remove it.

rust/foxglove/src/draco.rs
clalancette: Reproducing our conversation from slack:

```
Basically the idea is that every point in a standard ROS/Foxglove pointcloud consists of a series of fields "x: 0.3", "y: 0.5", "z: 0.6", "intensity: 0.4", etc.  Each of those fields can have arbitrary names and types.  Draco doesn't support this level of configuration; you can basically only have x,y,z , plus GENERIC fields.  GENERIC fields only have a type and a value (no name).  So we need some way to give a name to those fields, and we do it through attributes.  That's what that loop is for; gathering the "other" properties of a point, adding them as generic, and attaching attributes.  I benchmarked it, and while that open-coded loop isn't the fastest thing in the world, it is drowned out by the other work that Draco does, so probably not worth optimizing.
```

As for the comment, I agree that it could be better, I'm rewriting it.

rust/foxglove/src/draco.rs
clalancette: Agreed.  Dropped in the latest push.

rust/foxglove/src/remote_access/gateway.rs
clalancette: Removed.

rust/foxglove/src/remote_access/gateway.rs
clalancette: Moved to the top.

rust/remote_access_tests/src/test_helpers.rs
clalancette: Removed.

rust/foxglove/Cargo.toml
clalancette: Removed.

rust/foxglove/src/remote_access/gateway.rs
gasmith: I definitely like the idea of per-channel configuration in principle. Sure, let's go with that.

clalancette: This is done now in the latest push.

rust/foxglove/src/remote_access/session.rs
```diff
@@ -635,6 +741,99 @@ impl RemoteAccessSession {
...
+        // Don't warn for a channel unadvertised out from under the transcoding task.
+        if !self.channel_registry.read().has_channel(&channel_id) {
+            return;
+        }
+        let now = std::time::Instant::now();
+        {
+            let mut warnings = self.active_compression_warnings.lock();
+            if warnings
+                .get(&channel_id)
+                .is_some_and(|w| now.duration_since(w.last_report) < COMPRESSION_WARN_INTERVAL)
```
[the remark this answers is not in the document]
clalancette: Adding additional tests here isn't really going to buy us much; we can't easily test the gap.  I'm going to decline doing anything here unless a human tells me I really need to.

clalancette:
@gasmith As I mentioned in our discussion yesterday, the fallout of two decisions we made was larger than expected.  Those two decisions are 1) dropping draco as a rust feature separate from remote-access, and 2) removing the gateway pointcloud compression policy, and only making it per-channel.  Combined with catering to the LLM gods, the big changes on this branch are:

* replace the global compression options with a per-channel policy, using a PointCloudCompression trait and point_cloud_compression_fn
* DracoEncodeOptions became valid-by-construction with a 30-bit cap (30-bits is the maximum that draco supports)
* float64 clouds are rejected rather than silently degraded
* viewer-facing compression warnings became a single session-owned, throttled, replay-on-subscribe, auto-clearing status per channel
* draco rust feature no longer exists as a separate flag

This is ready for review along with the C++, Python, example, and ROS 2 PRs that go along with it.

rust/foxglove/src/remote_access.rs
```diff
@@ -41,6 +43,8 @@ pub use client::Client;
 pub use connection::ConnectionStatus;
 pub use gateway::{Gateway, GatewayHandle, VideoEncoderBackend};
 pub use listener::Listener;
+#[cfg_attr(docsrs, doc(cfg(feature = "remote-access")))]
```
gasmith:
```suggestion
```

rust/foxglove/src/lib.rs
```diff
@@ -452,6 +456,10 @@ pub use remote_data_loader_backend as data_provider;
 #[allow(unused)]
 mod img2yuv;
 
+#[cfg(feature = "remote-access")]
+#[cfg_attr(docsrs, doc(cfg(feature = "remote-access")))]
```
gasmith: This shouldn't be necessary.
```suggestion
```

gasmith (approved):
> - viewer-facing compression warnings became a single session-owned, throttled, replay-on-subscribe, auto-clearing status per channel

Was not expecting that to land as part of this change, but I'm so glad it's here. I think there might be an opportunity to abstract this out a little bit, so that we can use it more broadly, but let's save that for another day.

rust/foxglove/src/draco.rs
clalancette: I've opened DB-1544 to track this.

clalancette:
During testing with real-world data, I found that float64 pointclouds are more prevalent than they seem; 2 out of the 3 in one of the datasets I looked at used float64.  Those would just fail to deliver anything.

Because this is compression, I think a loss of precision is OK; nobody is using compressed point clouds over a WebRTC connection to do high-fidelity comparisons.  And the advice of the draco developers is essentially to narrow float64 -> float32 before giving it to draco: https://github.com/google/draco/issues/360 .  The latest commit does that.

While doing that, I realized something else as well.  Many pointclouds have per-point metadata that makes less or no sense in a compressed-pointcloud world.  The biggest one is per-point timestamp, which spinning LIDARs make available to do motion compensation.  But again, that makes no sense over remote-access, and compression can change the order of the points so it is actively wrong.  point ids are in a similar situation.  And there are some pieces of data which are completely ignored by the app-side decoder, like `range` and `azimuth`.  Because of all of that, I decided to drop those fields in a compressed scenario; they won't provide value but will be taking up bandwidth.

With all of that in place, what happens on an incoming pointcloud is the following:
1.  drop_irrelevant_fields() - drop the timestamp and other data first
2.  reinterpret_packed_color_fields() - it is common for RGBD data to declare color fields as float32 when they are really uint32.  If the field is labeled as `rgb` or `rgba`, repack these as uint32.
3.  narrow_float64_fields() - narrow any float64 to float32
4. drop_non_finite_points() - pointclouds often have NaN in them, but draco errors on NaN in the data.  Drop these pieces of data.
5. compress_point_cloud() - use draco compression to compress the point cloud

With this in place, I was able to visualize both float32 and float64 pointclouds.

@gasmith for review, you should only have to re-review [53ae846](https://github.com/foxglove/foxglove-sdk/pull/1391/commits/53ae8466a9bde0533d0b4edf84a754d59ede250e); everything else stayed the same

rust/foxglove/src/remote_access/gateway.rs
```diff
@@ -479,30 +479,45 @@ impl Gateway {
...
-    /// removed before encoding: publishers commonly pad invalid returns with NaN, and
-    /// the Draco quantizer rejects non-finite input, which would otherwise fail the
-    /// whole cloud.
-    ///
-    /// Fields named `rgb` or `rgba` declared float32 — PCL's packed-color convention —
-    /// are reinterpreted as uint32 before encoding: quantizing the packed bits as a
-    /// float would destroy the colors, while integer attributes are copied losslessly.
-    ///
-    /// [`CompressPointCloudOptions::default()`]: crate::remote_access::CompressPointCloudOptions
+    /// Per-point fields that carry no value on a remote viewer are dropped before
```
gasmith, line 489: We should make sure that these caveats find their way into the foxglove docs as well.

gasmith (approved):

rust/foxglove/src/remote_access/gateway.rs
clalancette: Yes, good call.  I've folded these details into https://github.com/foxglove/app/pull/17110 now.

PR #1392, opened 2026-07-20
Expose point cloud compression in C/C++

eloff:
### Changelog

C and C++ remote access gateway options can configure or suppress transparent point cloud compression.

### Docs

None

### Description

Exposes the remote access gateway's transparent point-cloud compression in the C and C++ SDKs as a per-channel policy callback, mirroring the Rust PointCloudCompression trait: for each compressible Lossy foxglove.PointCloud channel, the callback returns Draco settings, the SDK default (a zero-initialized C struct, mode 0), or disabled to deliver that channel unmodified. Leaving the callback unset compresses every such channel with the default settings, matching a Rust gateway with no policy.

### Testing

I wrote a small C++ program locally that sets up two channels: `/cloud/wave` and `/cloud/raw`.  I then setup the options for the pointcloud like this:

```
  options.point_cloud_compression = [](const foxglove::ChannelDescriptor& channel) {
    if (channel.topic() == "/cloud/raw") {
      return foxglove::PointCloudCompression::disabled();
    }
    return foxglove::PointCloudCompression::withDraco({10});
  };
```

I was able to visualize the `/cloud/wave` topic in the app, but the `/cloud/raw` ones shows a warning in Problems.

**This is PR 5/9 in the series — C/C++ bindings.**

**Depends on:** #1391 (core). Base branch: `dan/fle-684-...`.

Fixes: [FLE-683](https://linear.app/foxglove/issue/FLE-683)


c/src/server.rs
```diff
@@ -92,6 +92,214 @@ impl From<FoxgloveServerCapability> for FoxgloveServerCapabilityBitFlags {
...
+                    "; 0 (lossless) is not supported here — use the Disabled compression \
+                     mode to deliver point clouds uncompressed"
+                } else {
+                    ""
+                };
+                foxglove::FoxgloveError::ConfigurationError(format!(
+                    "point_cloud_compression: quantization_bits ({}) must be between 1 and {}{hint}",
+                    options.quantization_bits,
+                    foxglove::draco::MAX_QUANTIZATION_BITS
+                ))
```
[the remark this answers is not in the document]
clalancette: The formatting is correct here, so there is nothing to do.  And yes, the log is a bonus.  So nothing to do here.

rust/foxglove/src/remote_access/point_cloud_transcode.rs
```diff
@@ -39,11 +39,40 @@ pub(crate) fn transcode_point_cloud_message(
...
+/// preserves the packed bits exactly. Only the declared type changes — both types are
+/// four bytes, so offsets, stride, and data are untouched — and the uint32 declaration is
+/// itself a common `PointCloud2` convention: the app already force-reads `rgb`/`rgba`
+/// color fields as uint32 whatever the declared type.
+///
+/// Name-matching is a heuristic, like the encoder's `x`/`y`/`z` handling: a field named
+/// `rgb` carrying a genuine continuous float would be retyped needlessly (still lossless,
+/// merely un-quantized), which is strictly safer than the certain mangling of packed
+/// colors today.
+fn reinterpret_packed_color_fields(cloud: &mut PointCloud) {
```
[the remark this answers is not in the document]
clalancette: This comment is just wrong; the commit in question is on #1391 as intended.

c/src/gateway.rs
```diff
@@ -112,6 +112,165 @@ impl foxglove::remote_access::SuppressVideoTranscode for SuppressVideoTranscode
...
+    /// `FOXGLOVE_POINT_CLOUD_COMPRESSION_MODE_DISABLED` instead.
+    pub quantization_bits: u8,
+}
+
+impl TryFrom<FoxgloveDracoEncodeOptions> for foxglove::draco::DracoEncodeOptions {
+    type Error = foxglove::FoxgloveError;
+
+    fn try_from(options: FoxgloveDracoEncodeOptions) -> Result<Self, Self::Error> {
+        // The SDK derives the Draco method (kd-tree for quantized encoding) internally, so
+        // only the quantization setting is forwarded.
```
gasmith: Nit: Feels like this isn't a concern worth documenting at this layer.
```suggestion
```

c/src/gateway.rs
```diff
@@ -112,6 +112,165 @@ impl foxglove::remote_access::SuppressVideoTranscode for SuppressVideoTranscode
...
+                     mode to deliver point clouds uncompressed"
+                } else {
+                    ""
+                };
+                foxglove::FoxgloveError::ConfigurationError(format!(
+                    "point_cloud_compression: quantization_bits ({}) must be between 1 and {}{hint}",
+                    options.quantization_bits,
+                    foxglove::draco::MAX_QUANTIZATION_BITS
+                ))
+            }
```
gasmith: Maybe we could just log a warning and do something reasonable. If 0, disable compression. Otherwise clamp to MAX_QUANTIZATION_BITS.

c/src/gateway.rs
```diff
@@ -112,6 +112,165 @@ impl foxglove::remote_access::SuppressVideoTranscode for SuppressVideoTranscode
...
+        channel: &foxglove::ChannelDescriptor,
+    ) -> Option<foxglove::remote_access::CompressPointCloudOptions> {
+        let c_channel_descriptor = FoxgloveChannelDescriptor(channel.clone());
+        let compression =
+            unsafe { (self.callback)(self.callback_context, &raw const c_channel_descriptor) };
+        match compression.to_compression_options() {
+            Ok(options) => options,
+            // The Rust policy trait makes invalid options unrepresentable, but a C callback
+            // can return them; the contract can no longer be enforced at startup, so report
+            // the error and deliver the channel unmodified.
```
gasmith: This comment feels a bit off. The fact that you can't return an invalid constructed `CompressPointCloudOptions` from the trait method doesn't mean that you can't experience that as a runtime error. I guess the difference is that in Rust, it's at least possible to construct validated options in advance, e.g. during startup? Maybe we should offer a validation function, or a C++ constructor that performs validation? Not a big deal, though. I'm not going to lose sleep over it.

cpp/foxglove/include/foxglove/point_cloud_compression.hpp
```diff
@@ -0,0 +1,64 @@
...
+  Default = 0,
+  /// Disable transparent point-cloud compression: point clouds are delivered unmodified.
+  Disabled = 1,
+  /// Draco compression with the settings in @ref PointCloudCompression::draco.
+  Draco = 2,
+};
+
+/// @brief Options for Draco point-cloud encoding.
+struct DracoEncodeOptions {
+  /// @brief Quantization bits for the position attribute; must be between 1 and 30
```
gasmith: Maybe expose a `MAX_QUANTIZATION_BITS` constant

cpp/foxglove/include/foxglove/point_cloud_compression.hpp
```diff
@@ -0,0 +1,64 @@
...
+  /// inclusive. Values outside that range are invalid, and a channel whose compression
+  /// callback returns them is delivered unmodified, with a logged warning. Values above 30
+  /// are rejected by the reference Draco decoder, and `0` (lossless) provides no size
+  /// reduction over the raw point cloud — use @ref PointCloudCompressionMode::Disabled
+  /// instead.
```
gasmith: I think this obscures the facts about how we actually handle invalid values in the C FFI layer.

c/src/gateway.rs
clalancette: Removed in the latest push.

c/src/gateway.rs
clalancette: Yeah, I switched this to a clamp on MAX_QUANTIZATION_BITS.

cpp/foxglove/include/foxglove/point_cloud_compression.hpp
clalancette: Done in the latest push.

cpp/foxglove/include/foxglove/point_cloud_compression.hpp
clalancette: With the clamping in place, I've updated this comment to describe it.  Let me know if the update is better.

c/src/gateway.rs
clalancette: I've removed this in the latest push.  I did not add a validation function right now, but let me know if you want one and I can add it.

gasmith (approved):

PR #1395, opened 2026-07-20
Add ROS2 PointCloud2 compression input

eloff:
### Changelog

The ROS 2 foxglove_bridge remote access gateway now automatically does lossy draco compression for topics of type `sensor_msgs/msg/PointCloud2`; the parameter `point_cloud_compression_topic_denylist` can be used to opt-out of this behavior on a topic-by-topic basis.

### Docs

https://github.com/foxglove/app/pull/17110

### Description

Extends transparent point-cloud compression to ROS 2 data: channels carrying CDR-encoded `sensor_msgs/msg/PointCloud2` messages are now recognized as compression input alongside protobuf `foxglove.PointCloud`. Compressible channels are re-advertised as protobuf `foxglove.CompressedPointCloud`, with the source schema recorded in channel metadata as `foxglove.originalSchemaName` so clients can surface the original type. The converter repacks row-padded and point-padded clouds to a dense layout and enforces the declared width/height on unpadded data, trimming excess and rejecting malformed clouds rather than compressing garbage.

On the bridge side, adds a `point_cloud_compression_topic_denylist` parameter mirroring `video_transcode_topic_denylist`: a list of topic-name regexes delivered with their original schema instead of being Draco-compressed — for topics whose exact values or point order must be preserved, since the default compression is lossy and reorders points. Defaults to empty (all compressible point-cloud topics are compressed).

### Testing

Tested out by building and running the foxglove_bridge on ROS 2 Jazzy, then using a small Python program to publish 2 topics via rclpy: `/cloud/compressed` and `/cloud/raw`.  I'm able to visualize the `/cloud/compressed` one in the app, but the `/cloud/raw` one shows a warning in Problems.

**This is PR 8/9 in the series — ROS 2 + bridge.**

**Depends on:** #1392 (C/C++ bindings). Base branch: `dan/fle-685-...`.

Fixes: [FLE-687](https://linear.app/foxglove/issue/FLE-687)


rust/foxglove/src/img2yuv/ros2.rs
gasmith, line 1: This'll make a it a little more difficult to backport changes to the copy of img2yuv that lives in the data platform repo. On the other hand, we should really separate message decoding from transcoding anyway. No need to do any of that as part of this change, of course.

rust/foxglove/src/remote_access/point_cloud_transcode.rs
```diff
@@ -142,16 +179,24 @@ fn drop_non_finite_points(cloud: &mut PointCloud) {
...
+/// channel metadata as `foxglove.originalSchemaName`, so clients can surface the source type
+/// of a transparently compressed channel.
 pub(crate) fn rewrite_advertisement(channel: &mut advertise::Channel<'_>) {
     let schema_data = protocol_schema::encode_schema_data(
         "protobuf",
         std::borrow::Cow::Borrowed(descriptors::COMPRESSED_POINT_CLOUD),
     )
     .expect("binary schema encoding is infallible");
+    channel.metadata.insert(
+        "foxglove.originalSchemaName".to_string(),
```
gasmith, line 397: Is there an app PR to update [docs/livekit.md](https://github.com/foxglove/app/blob/main/packages/viz/src/players/FoxgloveWebSocketPlayer/docs/livekit.md) with this key?

ros/src/foxglove_bridge/README.md
```diff
@@ -140,6 +140,7 @@ Parameters are provided to configure the behavior of the bridge. These parameter
...
+- **point_cloud_compression_topic_denylist**: List of regular expressions ([ECMAScript grammar](https://en.cppreference.com/w/cpp/regex/ecmascript)) of topic names delivered with their original schema over remote access instead of being transparently Draco-compressed. Use this for point cloud topics whose exact values or point order must be preserved, since the default compression is lossy (kd-tree encoding with positions quantized to 12 bits, which also reorders points). Defaults to `[]` (all compressible point cloud topics are compressed); the launch file defaults this argument to `['(?!)']`, a never-matching pattern, because an empty array cannot be passed as a string array through a launch file. Compression applies to `sensor_msgs/msg/PointCloud2` topics and to channels carrying protobuf-encoded `foxglove.PointCloud` messages. Channels classified as Reliable skip compression automatically and deliver the raw point cloud on the control bytestream.
```
gasmith, line 143: Should we also expose a global setting for quantization bits?

gasmith (approved):
You might also consider testing this against some real recordings with ros2 point clouds, with `example_remote_access_stream_mcap`.

rust/foxglove/src/img2yuv/ros2.rs
clalancette: Good point.  Do you want me to leave the duplication between `img2yuv` and `point_cloud_transcode` for now, and we can revisit it later?

ros/src/foxglove_bridge/README.md
clalancette: Good call, done in the latest push.  Note that I made it global as you suggested; making it per-topic is possible, but a bit painful in ROS 2 parameters.  My suggestion is that we leave it as global and then only implement per-topic if we get a feature request for it.

rust/foxglove/src/remote_access/point_cloud_transcode.rs
clalancette: Good call, done in https://github.com/foxglove/app/pull/17110

PR #1398, opened 2026-07-20
ci: Split msrv/rustdoc builds out of ci/rust

gasmith:
### Changelog
None

### Docs
None

### Description
Another change to try and speed up ci/rust a bit.

The MSRV & rustdoc builds are easy to separate out, because they use
different rust toolchains, and therefore don't benefit from running in
the same job as the rust checks we run against latest stable.

We follow the same pattern as the C/C++ and Python workflows, and
aggregate the rust build statuses as a single synthetic `rust-status`.
We'll need to manually update the repo settings to treat this as a
required status check after this PR merges.

I also noticed that we weren't building the full feature set against
MSRV, which was a mistake that I'm gonna fix while we're here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,8 +24,7 @@ jobs:
 
       - uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
-          # The last-specified toolchain is the default.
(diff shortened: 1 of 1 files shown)


Avan1ko (approved):
Actions run independently now, LG

PR #1415, opened 2026-07-29
Move python schemas makefile targets to top-level

gasmith:
### Changelog
None

### Docs
None

### Description
The python schema packages have their own Makefile under python/. This
change moves those targets up to the top-level Makefile, and updates
them to use the common toolchain container, so that they don't depend on
having the blessed version of flatc/protoc installed locally.

The new targets are:

- generate-python-schemas-{flatbuffer,protobuf}
- build-python-schemas
- test-python-schemas
- clean-python-schemas

This change also updates the python workflow to use protoc 29.6 for
building these packages, to match the version we use elsewhere in the
repo.

Incidentally, we've learned that this containerized toolchain only works
for linux/amd64. Should still work for mac users via emulation.


Container.mk
```diff
@@ -60,6 +60,61 @@ docs-python:
 clean-docs-python:
 	rm -rf python/foxglove-sdk/python/docs/_build
 
+.PHONY: generate-python-schemas-flatbuffer
```
clalancette, line 63: I appreciate wanting to be able to invoke these targets from the top-level, but I feel like this is going too far.  This file is already 175 lines without this change, and I expect it will grow in the future.  My suggestion is to keep `python/Makefile`, and then in this file add targets that forward to it (following the example of `clean-docs-python`, `build-cpp`, etc.).

Relatedly, but separately, I think we should probably move `build-cpp-dist` into cpp/Makefile for similar reasons.  If you agree, I can do a follow-up with that.

clalancette (changes requested):

Container.mk
gasmith: Oh, that's a good call.

clalancette (approved):
Much better!  Looks good to me, and I'll make the follow-up for the `build-cpp-dist` target.

PR #1434, opened 2026-08-20
bridge: Terminate strings for CDR serialization

### LINKED ISSUE #1428: [bug] JSON-encoded client-published messages deserialize as empty: rosx_introspection CDR conversion omits string null terminator
# [bug] JSON-encoded client-published messages deserialize as empty: rosx_introspection CDR conversion omits string null terminator

**Update to the original report** (title/body corrected after further investigation): the earlier claim of an rmw_cyclonedds-wide client-publish failure was **wrong** — the empty content observed with **CDR-encoded** client publishes was caused by my test client omitting the 4-byte rmw serialized-message header (`\x00\x01\x00\x00`). With the header, CDR client publish works correctly on rmw_cyclonedds (verified; matches what Foxglove Studio sends).

The **remaining genuine bug** is specific to **JSON-encoded** client publishes: the bridge's `rosx_introspection`-based json→CDR conversion produces CDR **without the string null terminator**, so typed subscribers deserialize the content as empty.

## Summary

When a WebSocket client publishes to a client-advertised topic using **`encoding: "json"`**, the bridge converts the JSON message to CDR via `rosx_introspection`, but **CDR string fields are emitted without the terminating null byte** (and the length field excludes it). Typed subscribers (`rclcpp` node, `ros2 topic echo`) then deserialize the message as empty (e.g. `std_msgs/String` arrives as `data: ''`).

**CDR-encoded** client publishes with the correct 4-byte rmw header work fine on the same setup — this issue is limited to the JSON conversion path.

## Environment

- ROS 2 Humble (Ubuntu 22.04, Docker, `--network host`)
- `ros-humble-foxglove-bridge` 3.4.2 (apt) / 3.4.3 (source, SDK 0.26.0) — both affected
- `ros-humble-rmw-cyclonedds-cpp` 1.3.4, cyclonedds 0.10.5 (also reproduced conceptually on any RMW — the malformed CDR is produced before publishing)
- Protocol: `foxglove.sdk.v1`

## Steps to reproduce

1. Run a trivial echo node: subscribes `std_msgs/String` on `/health/ping`, replies on `/health/pong` echoing the received `data`.
2. Start the bridge: `ros2 run foxglove_bridge foxglove_bridge --ros-args -p port:=8765`
3. From a WebSocket client (`@foxglove/ws-protocol`, subprotocol `foxglove.sdk.v1`):
   ```js
   const chId = client.advertise({
     topic: "/health/ping", encoding: "json",
     schemaName: "std_msgs/msg/String", schemaEncoding: "ros2msg",
     schema: "# This was originally provided as an example message.\n...\n\nstring data\n",
   });
   client.sendMessage(chId, new TextEncoder().encode(JSON.stringify({ data: "abcd" })));
   ```
4. Observe: the echo node receives the message but `data` is empty.

## Evidence

Wire bytes of the json-published message (`ros2 topic echo /health/ping --raw`):

```
b'\x00\x01\x00\x00\x04\x00\x00\x00abcd'
     ^-- rmw header  ^-- len=4     ^-- "abcd" — NO null terminator
```

Correct CDR for `std_msgs/String` requires `len=5` + `abcd\0` (ROS2 CDR string convention: length **includes** the terminating null). The bridge's json conversion emits `len=4` without the null → the subscriber's deserializer reads past the buffer → empty string.

**Control — CDR-encoded publish (client sends `[0x00 0x01 0x00 0x00][u32 len=5]["abcd\0"]`)**:
```
pong: {"node":"health_echo_node","seq":23,"echo":"header-cdr-0"}   ← correct content
```
Same setup, works — so the bug is in the JSON→CDR conversion, not in the publish path or the RMW.

## Expected behavior

`encoding: "json"` client publishes should produce valid CDR (strings null-terminated) so typed subscribers receive the full content, matching the `cdr`-encoded path.

## Impact

JSON encoding is the easiest for frontend/web clients. Currently a frontend publishing e.g. `{"data":"..."}` for teleop commands would have its content silently dropped, while the equivalent `cdr` encoding works. Either fix the conversion or document/validate the CDR encoding as required for client publish.

gasmith:
### Changelog
- bridge: Fixed an issue where JSON-encoded strings were not properly
  nul-terminated when reserialized as CDR.

### Links
Fixes #1428 

### Docs
None

### Description
The upstream rosx_introspection crate's CDR serializer does not properly
nul-terminate strings (facontidavide/rosx_introspection#40). This change
works around that issue by overriding ROS2_Serializer::serializeString()
to append a trailing '\0' byte.

A probe is used to statically determine whether rosx_introspection is
broken.

### DIFF
--- a/ros/src/foxglove_bridge/CMakeLists.txt
+++ b/ros/src/foxglove_bridge/CMakeLists.txt
@@ -249,6 +249,10 @@ if(BUILD_TESTING)
   target_link_libraries(utils_test foxglove_bridge_component)
   enable_strict_compiler_warnings(utils_test)
 
+  ament_add_gtest(cdr_serializer_test tests/cdr_serializer_test.cpp)
+  target_link_libraries(cdr_serializer_test foxglove_bridge_component)
+  enable_strict_compiler_warnings(cdr_serializer_test)
+
   ament_add_gtest(message_definition_cache_test tests/message_definition_cache_test.cpp)
   target_link_libraries(message_definition_cache_test foxglove_bridge_component)
   enable_strict_compiler_warnings(message_definition_cache_test)
--- a/ros/src/foxglove_bridge/include/foxglove_bridge/cdr_serializer.hpp
+++ b/ros/src/foxglove_bridge/include/foxglove_bridge/cdr_serializer.hpp
@@ -1,0 +1,43 @@
+#pragma once
+
+#include <string>
+
+#include <rosx_introspection/serializer.hpp>
+
+namespace foxglove_bridge {
+
+/// Reports whether the rosx_introspection ROS 2 serializer omits the null terminator when writing
+/// a string.
+///
+/// ROS 2 middlewares expect a string to be encoded as a length of `size() + 1`, followed by the
+/// characters and a trailing null byte. rosx_introspection writes the characters alone, prefixed
+/// by a length that excludes the terminator, which Cyclone DDS rejects outright
+/// (https://github.com/facontidavide/rosx_introspection/issues/40).
+///
+/// This is probed at runtime rather than assumed, so that a fixed rosx_introspection doesn't leave
+/// us writing two terminators. The encoded buffer holds the 4-byte encapsulation header and a
+/// 4-byte length, followed by the characters, so a one-character string occupies 9 bytes without a
+/// terminator and 10 bytes with one.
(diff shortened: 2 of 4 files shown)


clalancette (approved):
Nasty bug, I like the fix.

PR #1435, opened 2026-08-24
rust: Run cargo update

gasmith:
### Changelog
None

### Docs
None

### Description
Run cargo update to update dependencies, ahead of major crate upgrades.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1092,7 +1092,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


Cargo.lock
```diff
@@ -3985,14 +4014,14 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
```
clalancette: This isn't necessarily a problem, but a package downgrade does make my eyebrows raise a bit.  Just checking that this is intentional.

Cargo.lock
```diff
@@ -4624,10 +4654,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "32497e9a4c7b38532efcdebeef879707aa9f794296a4f0244f6f69e9bc8574bd"
 dependencies = [
  "fastrand",
- "getrandom 0.4.3",
+ "getrandom 0.3.4",
```
clalancette: Same here.

Cargo.lock
gasmith: Eagle eyes! This appears to have been damage caused by my attempt to rollback the livekit updates after they broke semver. I've restored these now.

clalancette (approved):

PR #1469, opened 2026-09-14
foxglove_bridge: broadcast server time unconditionally

MerinoSheep:
### Changelog

`foxglove_bridge` now advertises the WebSocket `Time` capability and broadcasts server time unconditionally, not just when `use_sim_time` is set.

### Docs

None.

### Description

`foxglove_bridge` only advertised the `Time` capability and called `broadcastTime(...)` when `use_sim_time` was true. On a real robot that flag is false, so the bridge never told the client what time it was. Per Foxglove Studio (v1.36.0): "Use system time for Foxglove WebSocket connections if the server does not publish time messages" — so the client fell back to its own wall clock.

Anything the 3D panel resolves at current time (coordinate frame axes, `frame_locked` scene entities) was then placed using a transform looked up at the *viewer's* clock instead of the robot's, while message-stamped geometry (which carries its own header stamp) still rendered correctly. The result: a scene internally inconsistent by exactly the client/server clock skew — any robot whose clock isn't synchronized with the client's (different timezone, unsynced NTP, stopped clock) would show axes and frame-locked entities offset by that skew.

Changes:
- Advertise the `Time` capability unconditionally, not gated on `_useSimTime`.
- Keep the existing sim-time path (`/clock` subscription → `broadcastTime`) unchanged.
- Add an `else` branch: when not using sim time, drive `broadcastTime(this->now())` from a 100ms wall timer, skipped when there are no connected clients.

This is deliberately a wall timer, not a `/clock` publisher + `use_sim_time: true`. Setting that flag would make `this->now()` read `/clock` instead of the system clock, and with nothing publishing `/clock` on a real robot the bridge's own clock would freeze at zero, taking every outgoing `log_time` with it. A `/clock`-publisher-based alternative would also make the bridge's notion of time depend on delivery through the same message bus it's serving — the failure mode this avoids.

Ran the full test suite via the repo's docker-based colcon build/test flow for ROS Jazzy. This surfaced a real regression in `smoke_test.cpp`'s `receiveMessagesOfMultipleTransientLocalPublishers`: it counted every binary WebSocket frame as a received publisher message, and the now-periodic Time broadcast frames (`ServerBinaryOpcode::TIME = 2`) were miscounted alongside `MESSAGE_DATA` frames (`= 1`). Fixed by filtering to `MESSAGE_DATA` frames only, matching the opcode-filtering pattern already used elsewhere in the test client (e.g. `waitForServiceResponse`); the same gap existed in the shared `waitForChannelMsg` test helper and was fixed there too.

After those fixes: `foxglove_bridge` and `foxglove_msgs` both build clean, all 3127 tests pass, 0 errors, 0 failures.

### DIFF
--- a/ros/src/foxglove_bridge/include/foxglove_bridge/ros2_foxglove_bridge.hpp
+++ b/ros/src/foxglove_bridge/include/foxglove_bridge/ros2_foxglove_bridge.hpp
@@ -164,6 +164,11 @@ private:
 
   void subscribeConnectionGraph(bool subscribe);
 
+  // Broadcasts the given timestamp (nanoseconds since epoch) to WebSocket clients via the Time
+  // capability. Shared by both the sim-time (/clock) and wall-clock (timer) sources in the
+  // constructor.
+  void broadcastTime(int64_t nanoseconds);
+
   void subscribe(ChannelId channelId, const foxglove::ClientMetadata& client);
 
   void unsubscribe(ChannelId channelId, const foxglove::ClientMetadata& client);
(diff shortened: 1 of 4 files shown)


MerinoSheep:
Let me know if you prefer this to be gated by some flag similar to use_sime_time.

gasmith:
Thanks for the contribution! Just wanted to let you know that we're discussing this internally. We're aware of this issue, but I don't think this is the right solution. The bridge is already stamping each outbound message with the device wallclock, so sending `Time` messages in addition is purely redundant.

MerinoSheep:
Hey @gasmith sounds good 😄

PR #1472, opened 2026-09-15
Expose Draco's sequential encoding method for point-cloud compression

clalancette:
### Changelog

* Expose draco point cloud sequential encoding as a non-default SDK option, which preserves point order at the expense of more size on the wire.

### Docs

None.

### Description

draco-core 2.0 (#1468) carries the upstream fix for sequential encoding with quantized positions: the reference Draco decoder rejected those bitstreams from draco-core 1.0.x, which is why the SDK only ever reached the sequential encoder as the lossless fallback. Verified against the app's vendored Draco 1.5.x wasm decoder that every sequential+quantized combination now decodes.

Rust: `DracoMethod` is public (KdTree default, Sequential), selected at construction with `DracoEncodeOptions::new(quantization_bits, method)` and reported by `method()`; `with_quantization_bits` is now shorthand for the kd-tree case. Under sequential encoding positions and float32 fields are quantized as under kd-tree, point order is preserved, and float64 fields are copied losslessly instead of being rejected; kd-tree still rejects float64 with `UnquantizableField`, whose message now points at the sequential method. Default output is unchanged. The remote_access_point_cloud example gains `--sequential`.

C: `foxglove_draco_method` (zero = kd-tree) as `foxglove_draco_encode_options.method`. C++: `DracoMethod` and `DracoEncodeOptions::method`, static-asserted against the C constants. Python: `DracoMethod` enum and `DracoEncodeOptions(method=...)`, with stubs, re-export, docs, and tests.

Tested locally by running the `remote_access_point_cloud` example with `--sequential --quantization-bits <num>`.  It correctly decodes with bits up to 17; beyond that, the remote access machinery rejects the packets as over 100 KB and thus too large to deliver.  That is exactly the behavior we want.


rust/foxglove/src/draco.rs
```diff
@@ -65,45 +67,63 @@ pub const MAX_QUANTIZATION_BITS: u8 = 30;
...
+            method,
         })
     }
 
+    /// Creates options that quantize positions to `bits` bits (lossy) with the default
+    /// kd-tree encoding; shorthand for [`DracoEncodeOptions::new`] with
+    /// [`DracoMethod::KdTree`].
+    pub fn with_quantization_bits(bits: u8) -> Result<Self, DracoEncodeError> {
+        Self::new(bits, DracoMethod::KdTree)
+    }
```
gasmith: This interface looks bad for compatibility. If we add new options in the future, we either have to break `new()` or add a new `with_...` builder that doesn't match the constructor-ish semantics of `with_quantization_bits()`. Let's rip the bandaid now. Make `new` private, and make `with_quantization_bits` a standard self-consuming builder. We can keep the `lossless` constructor.

gasmith (commented):
LG for the most part, but I think we should rework the DracoEncodeOptions interface in rust.

rust/foxglove/src/draco.rs
clalancette: Yeah, good point.  I should have done this on the earlier PRs, there was just too much to manage with it.

The latest commit fixes it, but in a different way than you suggested.  In particular I added a DracoEncodeOptions::builder() to use the Rust builder pattern, then hung everything off of that.  Besides using the builder pattern, this also leaves the current API intact (though deprecated), which is somewhat difficult to do with your suggestion.  Take a look and let me know what you think.

gasmith (approved):

rust/foxglove/src/draco.rs
gasmith: Yep, that's better than my suggestion. Nice!

